PR #23, opened 2019-11-26
demo(Arima): add doc on Arima ferioh wiufeh wf woi

Bugdup:
This is just a test pull request to test some elements with Pycharm
It should not be merged, only deleted


u8timeseries/models/arima.py
```diff
@@ -5,6 +5,9 @@
 
 
 class Arima(AutoRegressiveModel):
+    """
+    Implements ARIMA  model
```
Bugdup: This is a review on a change

PR #25, opened 2019-12-03
Feature/statistics

MounBen:
- Added tools for statistical analysis of seasonality.
- Updated the Air Passenger example to show these functionalities.


u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+    :param frequency: The seasonality period.
+    :param model: -the type of decomposition to use (additive or multipilcative)
+    :return: A tuple (ts, seasonality) of TimeSeries where ts is the adjusted original TimeSeries ad seasonality
+            is a TimeSeries containing the extracted seasonality.
+    """
+    decomp = seasonal_decompose(ts.values(), model=model, freq=frequency)
+    seasonality = TimeSeries.from_times_and_values(ts.time_index(), decomp.seasonal)
+
+    if (seasonality.values() < 1e-6).any():
+        print("WARNING seasonal indexes equal to zero, using non-seasonal Theta method ")
```
Bugdup: If no theta method will be used here, you might want to rephrase the warning. Else, all is fine

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+    for i in range(len(r)):
+        if i == m:
+            pass
+        else:
+            plt.plot((i, i), (0, r[i]), color='black', lw=.5)
+
+    if m is not None:
+        plt.plot((m, m), (0, r[m]), color='red', lw=.5)
+
+    band_up = (r[1:].mean() + norm.ppf(1 - alpha/200) * r[1:].var()) * stat
```
hrzn: You can also do: `norm.ppf(1. - alpha/2, scale=1.)`, maybe more readable.

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+
+def plot_acf(ts: 'TimeSeries', m: int = None, max_lag: int = 24, alpha: float = 5):
+    """
+    Plots the ACF of [ts], highlighting the seasonality at lag [m] and the significance interval.
+
+    :param ts: The TimeSeries whose ACF should be plotted.
+    :param m:  The seasonality period.
+    :param max_lag: The maximal lag order to consider.
+    :param alpha: The confidence interval.
+    :return: Shows the plot.
```
hrzn: I'd remove this line, as the function doesn't return anything

u8timeseries/models/statistics.py
hrzn: So here `band_up` is confidence interval for the `m`-th lag, correct?

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+            raise ValueError('m must be a positive integer.')
+
+    n_unique = np.unique(ts.values()).shape[0]
+
+    if n_unique > 1:  # Check for non-constant TimeSeries
+
+        r = acf(ts.values(), nlags=max_lag)  # In case user wants to check for seasonality higher than 24 steps.
+
+        grad = np.gradient(r)
+        signs_changes = np.diff(np.sign(grad))
```
hrzn: maybe rename to `gradient_sign_changes`

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if m is None:
+            # Tries to infer seasonality from AutoCorrelation Function if no value of m has been provided.
+            # We look for the first positive local maximum of the ACF by checking the sign changes in the gradient.
```
hrzn: thanks for the comment :)

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+
+            # Local maximum is indicated by signs_change == -2
+            if len(np.nonzero((signs_changes == -2))[0]) > 0:
+                m = np.nonzero((signs_changes == -2))[0][0] + 1
+            else:
+                return False, 0
+
+        # Check for local maximum when m is user defined
+        if np.nonzero((signs_changes == -2))[0][0] != m-1:
```
hrzn: So here you are testing only for the "shortest" period, right? There could be another, larger, period `m` if I'm correct?

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+        # Remove r[0], the auto-correlation at order 0, that introduces bias.
+        r = r[1:]
+
+        stat = bartlett_formula(r, m - 1, len(ts))
+
+        # The upper limit of the significance interval.
+        band_upper = r.mean() + norm.ppf(1 - alpha/200)*r.var()
+
+        # Only local maximum with positive values are candidates to be tested against significance level.
+        season = r[m-1] > stat * band_upper
```
hrzn: I would rename in `is_significant` or something showing it's a boolean value

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+
+def seasonal_adjustment(ts: 'TimeSeries', frequency, model='multiplicative') -> Tuple['TimeSeries', 'TimeSeries']:
+    """
+    Adjusts the TimeSeries [ts] for a seasonality of order [frequency] using the [model] decomposition.
+
+    :param ts: The TimeSeries to adjust.
+    :param frequency: The seasonality period.
+    :param model: -the type of decomposition to use (additive or multipilcative)
+    :return: A tuple (ts, seasonality) of TimeSeries where ts is the adjusted original TimeSeries ad seasonality
+            is a TimeSeries containing the extracted seasonality.
```
hrzn: Very cool. Maybe you could add a sentence explaining what technique is used in the `seasonal_decompose()` method.

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+    :return: A tuple (season, m), where s is a boolean indicating whether the TimeSeries has seasonality or not
+             and m is the seasonality period.
+    """
+    if m is not None:
+        if m < 0:
+            raise ValueError('m must be a positive integer.')
+
+    n_unique = np.unique(ts.values()).shape[0]
+
+    if n_unique > 1:  # Check for non-constant TimeSeries
```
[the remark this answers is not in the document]
hrzn: I had the same comment :)

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+import numpy as np
+from statsmodels.tsa.stattools import acf
+from scipy.stats import norm
+from statsmodels.tsa.seasonal import seasonal_decompose
+import matplotlib.pyplot as plt
+from typing import Tuple
+import math
+
+
+def check_seasonality(ts: 'TimeSeries', m: int = None, max_lag: int = 24, alpha: float = 5):
```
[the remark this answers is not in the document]
hrzn: Good point. Maybe specifying `alpha: float = 0.05`  would be better

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+            is a TimeSeries containing the extracted seasonality.
+    """
+    decomp = seasonal_decompose(ts.values(), model=model, freq=frequency)
+    seasonality = TimeSeries.from_times_and_values(ts.time_index(), decomp.seasonal)
+
+    if (seasonality.values() < 1e-6).any():
+        print("WARNING seasonal indexes equal to zero, using non-seasonal Theta method ")
+    else:
+        if model == 'multiplicative':
+            ts = ts / seasonality
```
hrzn: You should assign `ts / seasonality` to a new variable

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+    :param length: The size of the underlying sample to be used.
+    :return: The standard error of [r] with order [m].
+    """
+    if m == 1:
+        return math.sqrt(1/length)
+    else:
+        return math.sqrt((1 + 2 * sum(map(lambda x: x ** 2, r[:m-1]))) / length)
+
+
+def seasonal_adjustment(ts: 'TimeSeries', frequency, model='multiplicative') -> Tuple['TimeSeries', 'TimeSeries']:
```
hrzn: How about breaking this method in two:
```
def extract_seasonality_and_trend(ts: 'TimeSeries', frequency) -> Tuple['TimeSeries', 'TimeSeries']
```
and
```
def remove_seasonality(ts: 'TimeSeries', frequency, model='multiplicative') -> 'TimeSeries'
```
The second one would call the first, but first one might also be useful for users. In addition the second one can return only one time series (a new TimeSeries with seasonality removed).

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
Bugdup: Actually yours is better, less indentation

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+                return False, 0
+
+        # Check for local maximum when m is user defined
+        if np.nonzero((signs_changes == -2))[0][0] != m-1:
+            return False, m
+
+        # Remove r[0], the auto-correlation at order 0, that introduces bias.
+        r = r[1:]
+
+        stat = bartlett_formula(r, m - 1, len(ts))
```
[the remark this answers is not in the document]
Digma: One that comes to mind, is that there are some technique/algo that require stationary time-series, so removing the trend can sometimes be helpful in making these time-series stationary.

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
MounBen: Yes, both your ways make more sense! Fixed.

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
MounBen: True, it is more common to have alpha as a percent. Fixed.

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+
+    if m is not None:
+        plt.plot((m, m), (0, r[m]), color='red', lw=.5)
+
+    band_up = (r[1:].mean() + norm.ppf(1 - alpha/200) * r[1:].var()) * stat
+    plt.plot((0, len(r) - 1), (band_up, band_up), linestyle='--', color='blue')
+    plt.plot((0, len(r) - 1), (-band_up, -band_up), linestyle='--', color='blue')
+    plt.plot((0, len(r) - 1), (0, 0), color='black')
+
+    plt.show(fig)
```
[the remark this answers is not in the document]
MounBen: I didn't get any errors but if matplotlib systematically takes care of everything, I will let it do its thing.

u8timeseries/models/statistics.py
MounBen: Yes, left-over from the less general function. Thanks for pointing it out!

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+
+    r = acf(ts.values(), nlags=max_lag)
+
+    fig = plt.figure(figsize=(10, 5))
+
+    stat = bartlett_formula(r[1:], m - 1, len(ts))
+
+    for i in range(len(r)):
+        if i == m:
```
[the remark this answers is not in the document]
MounBen: Nice!

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,136 @@
...
+    If [m] is None, the seasonality period is inferred from the Auto-correlation Function (ACF).
+
+    :param ts: The TimeSeries to check for seasonality.
+    :param m: The seasonality period to check.
+    :param max_lag: The maximal lag allowed in the ACF.
+    :param alpha: The desired confidence level (default 5%).
+    :return: A tuple (season, m), where s is a boolean indicating whether the TimeSeries has seasonality or not
+             and m is the seasonality period.
+    """
+    if m is not None:
```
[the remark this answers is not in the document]
MounBen: Sure.

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
MounBen: Good point. I will change the function so that it also works in the case where true seasonality is not the first local maximum. Thanks!

u8timeseries/models/statistics.py
MounBen: Yes, for a seasonality of order m, you also get a seasonality of order k*m (k >=2, though they may not be significant).
This also somehow ties in with the comment above pointing out that the first local maximum may not be significant and we should check further.

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
hrzn: How about taking the global maximum of `r[1:]` ?

u8timeseries/models/statistics.py
Bugdup: Do we agree (my assumption) that if ACF at lag `m` is significant and that ACF at lag `2m` is also significant and even higher, we only consider `m` as seasonality right? 
In other words, we do not take the highest autocorrelation peak but the smallest significant lag as seasonality

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
hrzn: Or even
```
plt.plot((i, i), (0, r[i]), color=('red' if m is not None and i == m else 'black'), lw=.5)
```
🥇

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
Bugdup: I asked confirmation of an assumption here, not sure about it https://github.com/unit8co/u8timeseries/pull/25#discussion_r353212657

u8timeseries/models/statistics.py
hrzn: seasonalities need not be multiples of each others, right? So you could have a first peak at `m`, and then a larger peak at `1.5*m` for instance. In this case I would (intuitively) expect the method to return the largest peak.

MounBen: @Bugdup Yes! That is my understanding too. The seasonality order should explain all spikes, thus we consider the minimal lag that fits all the conditions.
You can think, as an example, of tides along a beach. There are two high tides and two low tides a day, and they sometimes are big tides ("grandes marées"). However any tide is relevant for fishermen :)

MounBen: @hrzn Let's say that we have two local maximums. Then I would expect their gcd to also provide a local significant local maximum.

Bugdup: @hrzn Yes, I just picked this specific example actually, did not see the specificity

u8timeseries/models/statistics.py
MounBen: Indeed.

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
Bugdup: So, this could make sense but perhaps some results exist stating the situation you described Julien is really not likely in reality (like first peak at `m` significant and then second peak at `1.5*m` significant with a higher value). Anyway if we go there, we should also check that any sub-frequency of the kept peak is not significant (or else we fall back to my case)

u8timeseries/models/statistics.py
hrzn: I guess this depends right? You could have time series with two seasonalities. Imagine for instance (in a not-far-fetch-at-all example) the time series describing the photovoltaic production on a planet with two suns that are out of lock :) So in general you'll have peaks at multiples of the first peak (e.g. `1m`, `2m`, etc), but you can also have other peaks that are not multiple of the first ones (`3.1421m`, 6.2842m`, etc). Right?
Then I think which peak we choose is a matter of convention. The first local maximum has the advantage of being simple & intuitive :)

MounBen: Ok. I think I (finally) have a satisfying answer :)
I modeled your example above using sine functions with different periods (3 and 5) and add them together, tried some stuff, and here is my conclusion:

- What has been implemented right now "won't work" if there are multiple seasonality in the sense that we have to make a choice. So it should be used only under the assumption that the seasonality is unique.

- There are two reasonable ways to extend the inferring part of the seasonality check that I can think of right now. If there are two distinct seasonality period (i.e. one is not a multiple of the other), then:
    -  either set the global seasonality to their LCM.
    -  or use Fourier Analysis to separate the two frequencies and treat them separately.

u8timeseries/models/statistics.py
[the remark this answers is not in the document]
MounBen: As far as I know Bartlett's formula is the most common one used for study of ACF, a few words could be said about that indeed.

Clearly it would be useful to also including de-trending as a tool in the library, @Digma's remark is very pertinent.

In terms of how effective it would be, in general, to first de-trend and then check seasonality, I believe it should not affect the result much since the underlying assumption with those two operations is that both trend and seasonality are deterministic. (Though @Bugdup, visually your example is very appealing!!)

u8timeseries/models/statistics.py
MounBen: It's clearer with name, done.

u8timeseries/models/statistics.py
MounBen: Broke it down into 2 pieces (extraction and seasonality removal), and added trend removal as a third piece.

u8timeseries/models/statistics.py
```diff
@@ -21,50 +24,59 @@ def check_seasonality(ts: 'TimeSeries', m: int = None, max_lag: int = 24, alpha:
...
-        return season, m
+    while not is_significant and i < len(candidate):
+        c = candidate[i] + 1
+        stat = _bartlett_formula(r, c - 1, len(ts))
 
-    return False, 0
+        is_significant = r[c-1] > stat * band_upper
+        i += 1
 
+    return is_significant, c
```
Bugdup: 👌

Bugdup: I will just check if we can avoid this additional `+ 1` and `- 1` later

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,198 @@
...
+
+    decomp = seasonal_decompose(ts.pd_series(), freq=freq, model=model)
+
+    season = TimeSeries.from_times_and_values(ts.time_index(), decomp.seasonal)
+    trend = TimeSeries.from_times_and_values(ts.time_index(), decomp.trend)
+
+    return trend, season
+
+
+def remove_from_series(ts: 'TimeSeries', other: 'TimeSeries', model: str) -> 'TimeSeries':
```
hrzn: How about renaming `model` to `method` ?

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,198 @@
...
+    stats = []
+    for i in range(1, max_lag+1):
+        stats.append(_bartlett_formula(r[1:], i, len(ts)))
+
+    plt.figure(figsize=fig_size)
+
+    for i in range(len(r)):
+        plt.plot((i, i), (0, r[i]), color=('red' if m is not None and i == m else 'black'), lw=.5)
+
+    acf_band = [(r[1:].mean() + norm.ppf(1 - alpha/2) * r[1:].var()) * stat for stat in stats]
```
Bugdup:
```suggestion
    upp_band = r[1:].mean() + norm.ppf(1 - alpha/2) * r[1:].var()
    acf_band = [upp_band * stat for stat in stats]
```
Small tweak to avoid unnecessary computation, especially if workin with big timeseries

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,198 @@
...
+    # Remove r[0], the auto-correlation at lag order 0, that introduces bias.
+    r = r[1:]
+
+    # The non-adjusted upper limit of the significance interval.
+    band_upper = r.mean() + norm.ppf(1 - alpha/2)*r.var()
+
+    # Significance test, stops at first admissible value.
+    is_significant = False
+    i = 0
+    c = 0
```
hrzn: To avoid managing `c` and `i` (which makes me sweat), you could have something like:
```
for i, candidate in enumerate(candidates):
    stat = _bartlett_formula(r, candidate, len(ts))
    if (r[candidate] > stat * band_upper):
        return True, candidate + 1
return False, 0
```
That's assuming `candidate` has been renamed to `candidates` (which would be a good idea IMO).

hrzn:
It'd be nice before merging to add one or a couple of unit tests, for instance to make sure the new method picks up the right seasonality on a simple sine wave.

u8timeseries/models/statistics.py
Bugdup: https://github.com/unit8co/u8timeseries/pull/25/files/cecfc132de401009cb6a5520db8c008ef3d939c5#r354174857

u8timeseries/models/statistics.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    Extracts trend and seasonality from the TimeSeries [ts] using statsmodels.seasonal_decompose.
+
+    :param ts: The TimeSeries to study.
+    :param freq: The seasonality period to use.
+    :param model: The type of decomposition to use ('additive' or 'multiplicative').
+    :return: A tuple (trend, seasonal) of TimeSeries containing the trend and seasonality respectively.
+    """
+
+    decomp = seasonal_decompose(ts.pd_series(), freq=freq, model=model)
```
Bugdup:
```suggestion
    decomp = seasonal_decompose(ts.pd_series(), freq=freq, model=model, extrapolate_trend='freq')
```
Avoid getting NaN values one beginning and end of trend

u8timeseries/models/statistics.py
MounBen: Not sure about that. The terminology comes from choosing either an additive or multiplicative decomposition of your time series  (the "model"). Also model is compatible with statstools.

Bugdup: Indeed, it is not the decomposition that use the addition or the multiplication it is dependent on the model we use for the timeseries (see notes at the end of the `seasonal_decompose` doc [here](https://www.statsmodels.org/dev/generated/statsmodels.tsa.seasonal.seasonal_decompose.html))

hrzn: Oh ok my bad

PR #28, opened 2020-01-22
Feature/missing values

MounBen:
Contains:
- revamp of docstrings.
- missing value methods.


u8timeseries.egg-info/PKG-INFO
```diff
@@ -0,0 +1,10 @@
+Metadata-Version: 1.0
```
hrzn: I guess this `egg-info` folder should be in git ignore

u8timeseries/utils/mising_values.py
```diff
@@ -0,0 +1,121 @@
...
+    return len(a[0]), list(a[0])
+
+
+def auto_fillna(ts: 'TimeSeries') -> 'TimeSeries':
+    """
+    This function automatically fills the missing value in the TimeSeries `ts`.
+
+    The rules for completion are given below.
+
+    Missing values at the beginning are set to 0.
```
hrzn: why not backward fill? That would seem more consistent with the way last and intermediates missing values are filled.

u8timeseries/utils/mising_values.py
```diff
@@ -0,0 +1,121 @@
...
+    :return: A new TimeSeries with all missing values filled according to the rules above.
+    """
+
+    # We compute the number of times entries of the TimeSeries go from mssing to numeric and vice-versa
+    arr = np.where(np.diff(np.isnan(ts.values())) == True)[0]
+
+    ts_temp = ts.pd_series()
+
+    if len(arr) == 0:
+        raise ValueError('Your TimeSeries has no missing value.')
```
hrzn: Not sure we should raise an error here. I think doing nothing would be appropriate.

u8timeseries/utils/mising_values.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+
+    a = np.where(np.diff(np.isnan(ts.values())) is True)
+
+    return len(a[0]), list(a[0])
+
+
+def auto_fillna(ts: 'TimeSeries') -> 'TimeSeries':
+    """
+    This function automatically fills the missing value in the TimeSeries `ts`.
```
hrzn: Perhaps a comment here saying that it is assumed that missing values are already np.nan in the TimeSeries. This is a good assumption most of the time, but there might be other "missing value" symbols (sometimes even numbers such as 0 or -inf are used as missing values symbols)... users should be aware I think.

u8timeseries/utils/mising_values.py
```diff
@@ -0,0 +1,121 @@
...
+    else:
+
+        nan_num = [arr[i] for i in range(len(arr)) if i % 2 == 1]
+        num_nan = [arr[i] for i in range(len(arr)) if i % 2 == 0]
+
+    # One has that len(nan_num) = len(num_nan) or len(num_nan) - 1
+
+    # As long as nan_num is not empty, the missing values are both preceded and followed by numeric values
+    # Thus we can use linear interpolation between the closest known values to fill in the gaps.
+    while len(nan_num) > 0:
```
hrzn: This logic looks neat, but it would be good to unit-test it before merging

u8timeseries/utils/mising_values.py
MounBen: I probably got biased towards zero because of the Kaggle Challenge. Maybe the simplest way would be to leave the user free to choose that value?

hrzn: Letting the user choose is a good idea, but we should still provide a default behavior. I would vote for backfilling first value here, instead of 0's, for two reasons:
* This is symmetric with what is done at the end of the time series
* 0 can be a very poor value in many time series. Think for instance about a time series representing the average heights of human beings over years - surely it was never 0

MounBen: Makes sense. I'll update it asap.

PR #30, opened 2020-02-18
correct issues and new features

Droxef:


u8timeseries/timeseries.py
```diff
@@ -271,6 +270,9 @@ def intersect(self, other: 'TimeSeries') -> 'TimeSeries':
         """
         Returns a slice containing the intersection of this TimeSeries and the one provided in argument.
 
+        TODO: This function does not really behave as described. Not really an intersect
+        TODO: In case of no intersection found (or too small), should it raise an error (like now), or return None?
```
hrzn: I would say raise an error

u8timeseries/metrics/metrics.py
```diff
@@ -28,23 +37,49 @@ def mape(true_series: TimeSeries, pred_series: TimeSeries) -> float:
...
+    warn("This function does not take into account the seasonality of the time series. "
+         "Please use mase_seasonal to be accurate", FutureWarning)
     y_true, y_pred = _get_values_or_raise(true_series, pred_series)
     errors = np.sum(np.abs(y_true - y_pred))
     t = y_true.size
     scale = t/(t-1) * np.sum(np.abs(np.diff(y_true)))
     return errors / scale
 
 
+def mase_seasonal(true_series: TimeSeries, pred_series: TimeSeries, m: int = 1) -> float:
```
hrzn: really cool! Since the seasonal version is equivalent to the non-seasonal one with `m=1`, how about renaming this method to `mase` ?

u8timeseries/timeseries.py
```diff
@@ -425,6 +441,84 @@ def append(self, other: 'TimeSeries') -> 'TimeSeries':
             conf_hi = self._confidence_hi.append(other.conf_hi_pd_series())
         return TimeSeries(series, conf_lo, conf_hi)
 
+    def add(self, values: np.ndarray, index: pd.DatetimeIndex = None,
```
hrzn: the name `add` can be a bit confusing. How about something different, like `append_values`?

u8timeseries/timeseries.py
Droxef: Ok, I will leave it like that then.
Another possibility would be to return the common index, instead of the resulting TimeSeries (an actual slice).
Thus, with a __getitem__ method, one could retrieve that slice for any object

u8timeseries/timeseries.py
Droxef: It seems more understandable that way. I will change it

u8timeseries/metrics/metrics.py
Droxef: I have no problem with it. It was just to show the difference with the previous method

PR #33, opened 2020-04-08
Dependency fix francesco

pennfranc:
Information was added to the README.md regarding macOS requirements and docker container usage. Furthermore, a dependency issue involving inconsistent versions of script and pmdarima was fixed. Finally, ipywidget was added to setup.py as dependency.


README.md
```diff
@@ -1,16 +1,37 @@
...
+## macOS install: 
 
 The install happens with `pip`. For `conda` users, do the following first:
 ```
 conda install gcc
 conda install -c conda-forge fbprophet
 ```
 
-And then:
+Also, for the installation to be successful, numpy has to be already installed.
```
TheMP, line 9: +1 for improving the comment quality

requirements-docker.txt
```diff
@@ -2,7 +2,7 @@ numpy
 pystan
 scikit-learn
 scipy<1.3  # >1.3 to use more recent pmdarima
-pmdarima  # <1.3 if scipy is kept this way
+pmdarima<1.3 # >1.3 if scipy>1.3 is chosen
```
TheMP, line 5: hope this doesn't break something :). One more reason to be careful with setup of unit tests

setup.py
```diff
@@ -9,15 +9,16 @@
       license='Apache License 2.0',
       packages=find_packages(),
       install_requires=[
+          'numpy>=1.15.4',
```
TheMP, line 12: Two files with same list of dependencies is risky. Let's try to make it better.

It would be nice to refer requirements.txt from setup.py like mentioned here https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py. 

Warning: you probably will have to also mount setup.py in the same way as docker requirements file in the Dockerfile.

PR #34, opened 2020-04-09
GitHub action ci

pennfranc:
- implemented GitHub Action to build docker container that runs unit tests
- added status badge of GitHub workflow to README
- fixed missing_value function that led to test failure
- implemented new smoke tests that check runnability of all forecasting and regression models


u8timeseries/tests/test_models.py
```diff
@@ -0,0 +1,54 @@
...
+    forecasting_horizon = 5
+    regression_window = 5
+
+    # dummy timeseries for autoregression forecasting and target for regression
+    times = pd.date_range('20000101', '20000130')
+    values = np.sin(range(len(times))) + np.array([2.0] * len(times))
+    ts: TimeSeries = TimeSeries.from_times_and_values(times, values)
+
+    # dummy feature timeseries for regression
+    feature_values = np.array(range(len(times))).astype(float)
```
TheMP, line 23: array has a dtype kwarg that could be used to specify type

u8timeseries/tests/test_models.py
```diff
@@ -0,0 +1,54 @@
...
+    feature_ts: TimeSeries = TimeSeries.from_times_and_values(times, feature_values)
+
+    def test_autoregressive_models_runnability(self):
```
TheMP, line 26: Good idea to split into regressive and autoregressive test cases - it can be even better to do completely two different python classes - AutoregressiveModelsTestCase and RegressiveModelsTestCase

run_test_docker.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run -it --entrypoint '/home/jovyan/run_tests.sh' eu.gcr.io/unit8-product/u8timeseries:latest
```
TheMP, line 3: Is this used somewhere in github action?

Dockerfile
```diff
@@ -23,3 +23,7 @@ WORKDIR /deps
 ADD . /home/jovyan/work
 WORKDIR /home/jovyan/work
 RUN pip install .
+
+# to run tests
+COPY run_tests.sh /home/jovyan/run_tests.sh
```
TheMP, line 28: Shouldn't this be covered already by line 23 `ADD . /home/jovyan/work`?

.github/actions/test_action/action.yml
```diff
@@ -0,0 +1,6 @@
+name: 'CI_action'
```
TheMP, line 1: would improve the "test_action" naming :)

run_test_docker.sh
pennfranc: No, I used this to test the same behaviour locally. Should I remove it?

Dockerfile
pennfranc: You're right!

.github/actions/test_action/action.yml
```diff
@@ -0,0 +1,6 @@
...
+description: 'Runs unit tests upon push.'
+runs:
+  using: 	'docker'
+  image: 	'./../../../Dockerfile'
+  entrypoint:	"/home/jovyan/run_tests.sh"
```
hrzn, line 6: who is this user?

u8timeseries/tests/test_models.py
```diff
@@ -0,0 +1,54 @@
+import unittest
+import pandas as pd
+import numpy as np
+
+from ..timeseries import TimeSeries
+from u8timeseries import Prophet, KthValueAgoBaseline, ExponentialSmoothing, TimeSeries, Arima, AutoArima
+from u8timeseries import StandardRegressiveModel
+from u8timeseries.models.theta import Theta
+
+class ModelsTestCase(unittest.TestCase):
```
hrzn, line 10: +1 for this. Thanks!

u8timeseries/tests/test_models.py
```diff
@@ -0,0 +1,54 @@
...
+    regression_window = 5
+
+    # dummy timeseries for autoregression forecasting and target for regression
+    times = pd.date_range('20000101', '20000130')
+    values = np.sin(range(len(times))) + np.array([2.0] * len(times))
+    ts: TimeSeries = TimeSeries.from_times_and_values(times, values)
+
+    # dummy feature timeseries for regression
+    feature_values = np.array(range(len(times))).astype(float)
+    feature_ts: TimeSeries = TimeSeries.from_times_and_values(times, feature_values)
```
hrzn, line 24: Not directly related, but I think it would be nice to have a collection of "utils" to easily create "dummy" TimeSeries which could be used as features. For example one could be the constant time series. Another could be a linearly increasing one, or one which 1's during public holidays and 0's else, etc.

.github/actions/test_action/action.yml
pennfranc: jovyan is the user that is created in the jupyter/base-notebook Dockerfile. jovyan/work is the WORKDIR where our Dockerfile installs u8timeseries.

u8timeseries/tests/test_models.py
pennfranc: That makes sense! I will add this to Jira.

PR #35, opened 2020-04-14
Timeseries generation

pennfranc:
- Created 6 different functions for the creation of TimeSeries objects.
- Included test cases for every function.
- Removed unnecessary lines of code from all test cases.

### DIFF
--- a/u8timeseries/utils/timeseries_generation.py
+++ b/u8timeseries/utils/timeseries_generation.py
@@ -83,7 +83,7 @@ def white_noise_timeseries(length: int = 10, offset : str = 'D', mean: float = 0
     :param length: The length of the returned TimeSeries.
     :param offset: The time differene between two adjacent entries in the returned TimeSeries. A DateOffset alias is expected;
     :param start_date: The time index of the first entry in the returned TimeSeries.
-    :return: A random walk TimeSeries created as indicated above.
+    :return: A white noise TimeSeries created as indicated above.
     """
 
     times = pd.date_range(periods=length, freq=offset, start=start_date)
@@ -104,7 +104,7 @@ def random_walk_timeseries(length: int = 10, offset : str = 'D', mean: float = 0
     :param length: The length of the returned TimeSeries.
     :param offset: The time differene between two adjacent entries in the returned TimeSeries. A DateOffset alias is expected;
     :param start_date: The time index of the first entry in the returned TimeSeries.
-    :return: A random walk TimeSeries created as indicated above..
+    :return: A random walk TimeSeries created as indicated above.
     """
 
     times = pd.date_range(periods=length, freq=offset, start=start_date)
--- a/u8timeseries/utils/timeseries_generation.py
+++ b/u8timeseries/utils/timeseries_generation.py
@@ -9,8 +9,8 @@ from u8timeseries.timeseries import TimeSeries
 
 
 
-def constant_timeseries(value: float = 0, length: int = 10, offset : str = 'D',
-                        start_date: pd.Timestamp = pd.Timestamp('20000101')) -> 'TimeSeries':
+def constant_timeseries(value: float = 0, length: int = 10, offset: str = 'D',
+                        start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
     """
     Creates a constant TimeSeries with the given value, length, start date and offset.
 
@@ -28,8 +28,8 @@ def constant_timeseries(value: float = 0, length: int = 10, offset : str = 'D',
     return TimeSeries.from_times_and_values(times, values)
 
 
-def linear_timeseries(start_value: float = 0, value_delta: float = 1, length: int = 10, offset : str = 'D',
-                      start_date: pd.Timestamp = pd.Timestamp('20000101')) -> 'TimeSeries':
+def linear_timeseries(start_value: float = 0, value_delta: float = 1, length: int = 10, offset: str = 'D',
+                      start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
     """
     Creates a TimeSeries with a starting value of 'start_value' that increases by 'value_delta' at each step.
     The last entry of the time series will be equal to 'start_value' + ('length' - 1) * 'value_delta'.
@@ -37,7 +37,7 @@ def linear_timeseries(start_value: float = 0, value_delta: float = 1, length: in
     :param start_value: The value of the first entry in the TimeSeries.
     :param value_delta: The difference in value between to adjacent entries in the TimeSeries.
     :param length: The length of the returned TimeSeries.
-    :param offset: The time differene between two adjacent entries in the returned TimeSeries.
+    :param offset: The time differene between two adjacent entries in the returned TimeSeries. A DateOffset alias is expected.
     :param start_date: The time index of the first entry in the returned TimeSeries.
     :return: A linear TimeSeries with gradient 'value_delta'.
     """
@@ -49,8 +49,8 @@ def linear_timeseries(start_value: float = 0, value_delta: float = 1, length: in
 
 
 def periodic_timeseries(frequency: float = 0.1, amplitude: float = 1, phase: float = 0, y_offset: float = 0,
-                        length: int = 10, offset : str = 'D',
-                        start_date: pd.Timestamp = pd.Timestamp('20000101')) -> 'TimeSeries':
+                        length: int = 10, offset: str = 'D',
+                        start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
     """
     Creates a TimeSeries with a sinusoidal value progression with a given frequency, amplitude, phase and y offset.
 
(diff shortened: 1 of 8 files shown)


u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+import math
+from pandas.tseries.offsets import DateOffset
+from pandas.tseries.frequencies import to_offset
+from pandas.tseries.holiday import USFederalHolidayCalendar
+
+from u8timeseries.timeseries import TimeSeries
+
+
+
+def constant_timeseries(value: float = 0, length: int = 10, offset : str = 'D',
```
TheMP, line 12: +1 for type hints and docstrings. 

small nitpick: in offset there is a redundant space before : ;).

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+                        start_date: pd.Timestamp = pd.Timestamp('20000101')) -> 'TimeSeries':
```
TheMP, line 13: In case someone tries to plot/use this would be nicer to use meaningful timestamp here.

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    times = pd.date_range(periods=length, freq=offset, start=start_date)
+    values = np.linspace(start_value, start_value + (length - 1) * value_delta, length)
+
+    return TimeSeries.from_times_and_values(times, values)
+
+
+def periodic_timeseries(frequency: float = 0.1, amplitude: float = 1, phase: float = 0, y_offset: float = 0,
+                        length: int = 10, offset : str = 'D',
+                        start_date: pd.Timestamp = pd.Timestamp('20000101')) -> 'TimeSeries':
```
TheMP, line 53: Same comment for timestamp

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Creates a TimeSeries with a sinusoidal value progression with a given frequency, amplitude, phase and y offset.
+
+    :param frequency: The number of periods that take place within one time unit given in 'offset'.
+    :param amplitude: The maximum  difference between any value of the returned TimeSeries and 'y_offset'.
+    :param phase: The relative position within one period of the first value of the returned TimeSeries (in radians).
+    :param y_offset: The shift of the sine function along the y axis.
+    :param length: The length of the returned TimeSeries.
+    :param offset: The time differene between two adjacent entries in the returned TimeSeries.
```
TheMP, line 62: Would put a comment that according to pandas doc we expect str or DateOffset

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+import math
+from pandas.tseries.offsets import DateOffset
+from pandas.tseries.frequencies import to_offset
+from pandas.tseries.holiday import USFederalHolidayCalendar
+
+from u8timeseries.timeseries import TimeSeries
+
+
+
+def constant_timeseries(value: float = 0, length: int = 10, offset: str = 'D',
```
hrzn, line 12: I would have recommended a non-zero default `value` here, so that it can be used out-of-the box as an external regressor in additive models (e.g. `value: float = 1.` would do I think).
Also, I would rename `offset` in `freq`, and maybe `start_date` in `start_ts` to reflect it needs not be an actual date?

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+    :return: A constant TimeSeries with value 'value'.
+    """
+
+    times = pd.date_range(periods=length, freq=offset, start=start_date)
+    values = np.full(length, value)
+
+    return TimeSeries.from_times_and_values(times, values)
+
+
+def linear_timeseries(start_value: float = 0, value_delta: float = 1, length: int = 10, offset: str = 'D',
```
hrzn, line 31: how about specifying `end_value` instead of `value_delta` ?
`value_delta` has the potential to be slightly misleading if one doesn't read the doc carefully, as it could be either the delta from one point to the next, or the delta between the first and last point of the series.

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+                        start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
+    """
+    Creates a TimeSeries with a sinusoidal value progression with a given frequency, amplitude, phase and y offset.
+
+    :param frequency: The number of periods that take place within one time unit given in 'offset'.
+    :param amplitude: The maximum  difference between any value of the returned TimeSeries and 'y_offset'.
+    :param phase: The relative position within one period of the first value of the returned TimeSeries (in radians).
+    :param y_offset: The shift of the sine function along the y axis.
+    :param length: The length of the returned TimeSeries.
+    :param offset: The time differene between two adjacent entries in the returned TimeSeries. A DateOffset alias is expected.
```
hrzn, line 62: `differene` -> `difference`
Again, I would rename in `freq` everywhere :)

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+    values = f(values)
+
+    return TimeSeries.from_times_and_values(times, values)
+
+
+def white_noise_timeseries(length: int = 10, offset: str = 'D', mean: float = 0, std: float = 1,
+                           start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
+    """
+    Creates a white noise TimeSeries by sampling a gaussian distribution with mean 'mean' and 
+    standard deviation 'std'. Each value represents an indipendent sample of the distribution.
```
hrzn, line 79: indipendent -> independent

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    times = pd.date_range(periods=length, freq=offset, start=start_date)
+    values = np.array(range(length), dtype=float)
+    f = np.vectorize(lambda x: amplitude * math.sin(2 * math.pi * frequency * x + phase) + y_offset)
+    values = f(values)
+
+    return TimeSeries.from_times_and_values(times, values)
+
+
+def white_noise_timeseries(length: int = 10, offset: str = 'D', mean: float = 0, std: float = 1,
```
hrzn, line 75: How about calling this one `gaussian_timeseries` ? As when `mean` is not 0 this wouldn't (technically) be white noise.

hrzn: You could even optionally support the case where `mean` is a vector and `std` is a matrix, and the resulting distribution is a multivariate Gaussian. If `mean` and `std` are scalars you could revert to of i.i.d samples as you have now.

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+    :param mean: The mean of the gaussian distribution that is sampled at each step.
+    :param std: The standard deviation of the gaussian distribution that is sampled at each step.
+    :param length: The length of the returned TimeSeries.
+    :param offset: The time differene between two adjacent entries in the returned TimeSeries. A DateOffset alias is expected.
+    :param start_date: The time index of the first entry in the returned TimeSeries.
+    :return: A random walk TimeSeries created as indicated above.
+    """
+
+    times = pd.date_range(periods=length, freq=offset, start=start_date)
+    values = [np.random.normal(mean, std)]
```
hrzn, line 111: I think here there are probably ways to generate the random vector of steps once, and then take its cumulative sum in one go using numpy (which would likely be more efficient)

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+    :return: A white noise TimeSeries created as indicated above.
+    """
+
+    times = pd.date_range(periods=length, freq=offset, start=start_date)
+    values = np.random.normal(mean, std, size=length)
+
+    return TimeSeries.from_times_and_values(times, values)
+
+
+def random_walk_timeseries(length: int = 10, offset: str = 'D', mean: float = 0, std: float = 1,
```
hrzn, line 95: cool idea 👍

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+from pandas.tseries.holiday import USFederalHolidayCalendar
+
+from u8timeseries.timeseries import TimeSeries
+
+
+
+def constant_timeseries(value: float = 0, length: int = 10, offset: str = 'D',
+                        start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
+    """
+    Creates a constant TimeSeries with the given value, length, start date and offset.
```
hrzn, line 15: +1 for the very good comments :)

u8timeseries/utils/timeseries_generation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    times = pd.date_range(periods=length, freq=offset, start=start_date)
+    values = [np.random.normal(mean, std)]
+    while (len(values) < length):
+        values.append(values[-1] + np.random.normal(mean, std))
+
+    return TimeSeries.from_times_and_values(times, values)
+
+
+def us_holiday_timeseries(length: int = 10, start_date: pd.Timestamp = pd.Timestamp('2000-01-01')) -> 'TimeSeries':
```
hrzn, line 118: would be nice to have few other regions as well

PR #39, opened 2020-04-20
Feature/logging

pennfranc:
- Logged assert statements
- Logged manually raised errors
- Logged print statements
- Added function decorator that logs runtime

### DIFF
--- a/u8timeseries/backtesting/backtesting_udf.py
+++ b/u8timeseries/backtesting/backtesting_udf.py
@@ -1,7 +1,10 @@
 import pandas as pd
 from u8timeseries.timeseries import TimeSeries
 from u8timeseries.models.autoregressive_model import AutoRegressiveModel
+from ..custom_logging import assert_log, get_logger
 from typing import Tuple, List, Callable, Any
+
+logger = get_logger(__name__)
 
 """ This file contains some backtesting utilities, to evaluate user-defined functions (such as error functions)
     on some validation sets, sliding over time.
@@ -25,8 +28,8 @@ def get_train_val_series(series: TimeSeries, start: pd.Timestamp, nr_points_val:
     :return: a list of (training_set, validation_set) pairs
     """
 
-    assert start in series, 'The provided start timestamp is not in the time series.'
-    assert start != series.end_time(), 'The provided start timestamp is the last timestamp of the time series'
+    assert_log(start in series, 'The provided start timestamp is not in the time series.', logger)
+    assert_log(start != series.end_time(), 'The provided start timestamp is the last timestamp of the time series', logger)
     # TODO: maybe also check that valset_duration >= series frequency
 
     curr_val_start: pd.Timestamp = start
--- a/u8timeseries/backtesting/backtesting_udf.py
+++ b/u8timeseries/backtesting/backtesting_udf.py
@@ -1,7 +1,7 @@
 import pandas as pd
 from u8timeseries.timeseries import TimeSeries
 from u8timeseries.models.autoregressive_model import AutoRegressiveModel
-from ..custom_logging import assert_log, get_logger
+from ..custom_logging import check_value_log, get_logger
 from typing import Tuple, List, Callable, Any
 
 logger = get_logger(__name__)
@@ -28,8 +28,8 @@ def get_train_val_series(series: TimeSeries, start: pd.Timestamp, nr_points_val:
     :return: a list of (training_set, validation_set) pairs
     """
 
-    assert_log(start in series, 'The provided start timestamp is not in the time series.', logger)
-    assert_log(start != series.end_time(), 'The provided start timestamp is the last timestamp of the time series', logger)
+    check_value_log(start in series, 'The provided start timestamp is not in the time series.', logger)
+    check_value_log(start != series.end_time(), 'The provided start timestamp is the last timestamp of the time series', logger)
     # TODO: maybe also check that valset_duration >= series frequency
 
     curr_val_start: pd.Timestamp = start
--- a/u8timeseries/backtesting/backtesting_udf.py
+++ b/u8timeseries/backtesting/backtesting_udf.py
@@ -1,7 +1,7 @@
 import pandas as pd
 from u8timeseries.timeseries import TimeSeries
 from u8timeseries.models.autoregressive_model import AutoRegressiveModel
-from ..custom_logging import check_value_log, get_logger
+from ..custom_logging import raise_if_not, get_logger
 from typing import Tuple, List, Callable, Any
 
 logger = get_logger(__name__)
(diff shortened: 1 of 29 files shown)


u8timeseries/custom_logging.py
```diff
@@ -0,0 +1,74 @@
...
+    :param exception: The exception instance to be raised.
+    :param logger: The logger instance to log the exception type and message.
+    """
+    exception_type = str(type(exception)).split("'")[1]
+    message = str(exception)
+    logger.error(exception_type + ": " + message)
+
+    raise exception
+
+def time_log(logger: logging.Logger = get_logger('main_logger')):
```
hrzn, line 55: This is pretty cool :)

u8timeseries/custom_logging.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    A decorator function that logs the runtime of the function it is decorating
+    to the logger object that is taken as an argument.
+
+    :param logger: The logger instance to log the runtime of the function.
+    """
+    def time_log_helper(method):
+        def timed(*args):
+            start_time = time.time()
+            result = method(*args)
```
hrzn, line 65: shouldn't we also add the kwargs here?

u8timeseries/custom_logging.py
```diff
@@ -0,0 +1,74 @@
...
+            end_time = time.time()
+            duration = int((end_time - start_time) * 1000)
+
+            logger.info(method.__name__ + " function ran for {} milliseconds".format(duration))
```
hrzn, line 69: Perhaps you can use a better looking format like {0:.2f} or something, to avoid too many places

u8timeseries/backtesting/backtesting_udf.py
```diff
@@ -25,8 +28,8 @@ def get_train_val_series(series: TimeSeries, start: pd.Timestamp, nr_points_val:
     :return: a list of (training_set, validation_set) pairs
     """
 
-    assert start in series, 'The provided start timestamp is not in the time series.'
-    assert start != series.end_time(), 'The provided start timestamp is the last timestamp of the time series'
+    assert_log(start in series, 'The provided start timestamp is not in the time series.', logger)
```
hrzn, line 31: This is a good idea, however I think here, and in most places where we check user-provided arguments, it would be more appropriate to use `ValueError` exceptions (most of the time).
Until now we were using assert's, but this was a (bad) shortcut, as the exception system is made for this, and will be overall more informative than asserts.

u8timeseries/custom_logging.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.INFO)
+    stderr_handler = logging.StreamHandler()
+    formatter = logging.Formatter('[%(asctime)s] %(levelname)s | %(name)s | %(message)s')
+    stderr_handler.setFormatter(formatter)
+    logger.addHandler(stderr_handler)
+    return logger
+
+def assert_log(boolean_value: bool, message: str = "", logger: logging.Logger = get_logger('main_logger')):
```
hrzn, line 25: not sure we need to keep this one

u8timeseries/models/autoregressive_model.py
```diff
@@ -31,7 +33,7 @@ def predict(self, n: int) -> TimeSeries:
         """
         :return: A TimeSeries containing the `n` next points, starting after the end of the training time series.
         """
-        assert self._fit_called, 'fit() must be called before predict()'
+        assert_log(self._fit_called, 'fit() must be called before predict()', logger)
```
hrzn, line 36: Here `ValueError` is probably not a good fit, maybe there is a better exception type? (Other than AssertError or ValueError)

u8timeseries/models/statistics.py
```diff
@@ -124,7 +126,7 @@ def remove_from_series(ts: 'TimeSeries', other: 'TimeSeries', model: str) -> 'Ti
     :return: A TimeSeries `new_ts`, defined by removing `other` from `ts`.
     """
     if (other.values() < 1e-6).any() and model == 'multiplicative':
-        print("WARNING indexes equal to zero, cannot remove for a multiplicative model.")
+        log.warning("Indexes equal to zero, cannot remove for a multiplicative model.")
```
hrzn, line 129: `log` --> `logger` ?

u8timeseries/custom_logging.py
pennfranc: Do you mean decimal places? Right now I am casting the number to an int, so I'm getting rid of decimal places altogether. Should I include 2 decimal places?

hrzn: Oh right sorry, I missed this. Then it's fine :)

u8timeseries/custom_logging.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.INFO)
+    stderr_handler = logging.StreamHandler()
+    formatter = logging.Formatter('[%(asctime)s] %(levelname)s | %(name)s | %(message)s')
+    stderr_handler.setFormatter(formatter)
+    logger.addHandler(stderr_handler)
+    return logger
+
+def check_value_log(boolean_value: bool, message: str = "", logger: logging.Logger = get_logger('main_logger')):
```
hrzn, line 25: I'm nitpicking a bit, but this is used very often. I think `check_value_log()` doesn't really reflect the fact it's checking a boolean predicate and possible.

How about naming it `raise_if_not()`or something like this? @TheMP any idea?

u8timeseries/custom_logging.py
pennfranc: done

u8timeseries/backtesting/backtesting_udf.py
pennfranc: done

u8timeseries/custom_logging.py
pennfranc: done

u8timeseries/models/autoregressive_model.py
pennfranc: used generic Exception, done

u8timeseries/models/statistics.py
pennfranc: done

u8timeseries/custom_logging.py
```diff
@@ -0,0 +1,71 @@
...
+    stderr_handler.setFormatter(formatter)
+    logger.addHandler(stderr_handler)
+    return logger
+
+def raise_if_not(boolean_value: bool, message: str = "", logger: logging.Logger = get_logger('main_logger')):
+    """
+    Checks provided boolean expression and raises a ValueError if it evaluates to False.
+    It logs the error to the provided logger before raising it.
+
+    :param boolean_value: The boolean expression (or value) to be checked.
```
hrzn, line 30: I'm nitpicking here, but instead of "boolean expression (or value)" you can use "predicate", as this is the term for it :)

u8timeseries/custom_logging.py
hrzn: I still see `result = method(*args)`

u8timeseries/custom_logging.py
pennfranc: Should I also change the name of the variable 'boolean_value'?

hrzn: In fact I just googled it and a predicate is a function returning a boolean. Here we do have a boolean value everywhere in fact, so maybe you could just rename it "condition" everywhere? :D

pennfranc: Ok will do!

pennfranc: done

PR #40, opened 2020-04-21
Feature/versioning

pennfranc:
First PR into the newly created debug branch:
- Implemented centralised version information in u8timeseries/VERSION.
- Set initial version of develop branch to 0.0.2-SNAPSHOT.
- Added versioning support to docker bash scripts.
- Added (unrelated) reference in custom_logging.py.


build_docker.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
 cd "$( dirname "${BASH_SOURCE[0]}" )"
-docker build -t eu.gcr.io/unit8-product/u8timeseries:latest .
+version=$(head -n 1 u8timeseries/VERSION)
+docker build -t eu.gcr.io/unit8-product/u8timeseries:${version} .
```
TheMP, line 4: Would also leave the latest tag, since we can have multiple:

docker build -t eu.gcr.io/unit8-product/u8timeseries:${version} -t eu.gcr.io/unit8-product/u8timeseries:latest .

setup.py
```diff
@@ -24,5 +25,6 @@
           'testfixtures'
       ],
       zip_safe=False,
-      python_requires='>=3.6'
+      python_requires='>=3.6',
+      package_data={'u8timeseries': ['VERSION']}
```
TheMP, line 29: What is this line adding?

pennfranc: setuptools doesn't include all files in the installed package by default (see https://packaging.python.org/guides/using-manifest-in/). Without this line, the VERSION file would not be available for access after installation. I chose this approach in favour of including a MANIFEST.in because it doesn't require an extra file.

build_docker.sh
pennfranc: done

setup.py
TheMP: sounds good :)

PR #41, opened 2020-04-23
Version bump script

pennfranc:
- Created a python script that updates the current VERSION file of u8timeseries according to the settings given by the user.


README.md
```diff
@@ -46,3 +46,20 @@ pd.plotting.register_matplotlib_converters()
 
 ## unit test status
 ![ci_workflow](https://github.com/unit8co/u8timeseries/workflows/ci_workflow/badge.svg)
+
```
hrzn, line 49: really cool, thanks for the documentation. We can keep it in the README for now, however at some point I think this should move elsewhere, as README will be meant for general public mostly.

version_bump.py
```diff
@@ -0,0 +1,43 @@
...
+    suffix = "-SNAPSHOT"
+
+# update version
+if (bump_type != 0):
+    relevant_part = int(version_parsed[bump_type - 1])
+    version_parsed[bump_type - 1] = str(relevant_part + 1)
+new_version_string = ".".join(version_parsed) + suffix
+
+# write new version to VERSION file
+version_file = open("u8timeseries/VERSION", "w")
```
hrzn, line 40: I think it'd be slightly better to do this as:
```
with open('u8timeseries/VERSION', 'w') as f:
  f.write('new_version_string)
```
It should ensure that the file pointer is always closed in the (unlikely) case something fishy happens before it reaches the close() command.
```

File: version_bump.py [hunk 3]
```diff
@@ -0,0 +1,43 @@
...
+bump_type = args.bump
+while (not bump_type in ["0", "1", "2", "3"]):
+    print("Enter the type of version bump: 0 (No increment), 1 (Major), 2 (Minor), 3 (Patch)")
+    bump_type = input()
+bump_type = int(bump_type)
+
+# get information on whether SNAPSHOT should be added (if no valid argument has been passed)
+answer = args.release
+while (not answer in ["y", "n"]):
+    print("Is this an official release? (y/n)")
```
hrzn, line 26: perhaps in the prompt you could add `Is this an official release? I will add a SNAPSHOT tag if not. (y/n)`
so the user knows what this is doing.

version_bump.py
```diff
@@ -0,0 +1,43 @@
...
+# set up arguments
+parser = argparse.ArgumentParser(prog="version bump")
+parser.add_argument('-b', '--bump', default="", help="BUMP should be an integer between 0 and 3" \
+                                                     " indicating the type of version bump.")
+parser.add_argument('-r', '--release', default="", help="RELEASE is expected to be 'y' or 'n', indicating" \
+                                                        " whether or not the SNAPSHOT suffix should be added.")
+args = parser.parse_args()
+
+# retrieve current version
+version_string = open("u8timeseries/VERSION", "r").read()
```
TheMP, line 12: for the sake of reusing this script in other projects, could you move u8timeseries to an argument (eg. VERSION_FILE)?

version_bump.py
TheMP, line 40: Same as previous comment, this could be read from a variable

README.md
TheMP: Could be actually moved to the top of version bumping script

version_bump.py
```diff
@@ -0,0 +1,43 @@
+import argparse
+
+# set up arguments
+parser = argparse.ArgumentParser(prog="version bump")
+parser.add_argument('-b', '--bump', default="", help="BUMP should be an integer between 0 and 3" \
```
TheMP, line 5: no need to specify default if it is optional argument

version_bump.py
pennfranc: Do you mean I should define this as a constant somewhere (at the top of the script), or should I pass it as an argument like -b and -r?

TheMP: the second one :)

hrzn: We could maybe have both? An argument, and a default value (specified at the top of the file in a variable), so that it keeps the ease of use of the script.

version_bump.py
```diff
@@ -0,0 +1,67 @@
...
+import argparse
+
+VERSION_FILE = "u8timeseries/VERSION"
+
+# set up arguments
+parser = argparse.ArgumentParser(prog="version bump")
+parser.add_argument('-b', '--bump', help="BUMP should be an integer between 0 and 3" \
+                                         " indicating the type of version bump.")
+parser.add_argument('-r', '--release', help="RELEASE is expected to be 'y' or 'n', indicating" \
+                                            " whether or not the SNAPSHOT suffix should be added.")
```
hrzn, line 30: I'm nitpicking again :) but here it could be read as if `y` = adding `SNAPSHOT`.
I'd rather write: `RELEASE is expected to be 'y' or 'n'. If set to 'n' the SNAPSHOT suffix will be added.`

setup.py
```diff
@@ -1,7 +1,8 @@
 from setuptools import setup, find_packages
+import os
 
 setup(name='u8timeseries',
-      version='0.0.2',
+      version=open("u8timeseries/VERSION", "r").read(),
```
endrjuskr, line 5: Is it possible to reuse tags from github? Then we don't need to add an additional file.

version_bump.py
```diff
@@ -0,0 +1,67 @@
+### INSTRUCTIONS ###
```
endrjuskr, line 1: IMO there is already some lib to do it, so maybe we can reuse it

PR #42, opened 2020-04-23
Add documentation

endrjuskr:
* use sphinx
* use autodoc
* build and update gh-pages on commit only to master


docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
TheMP, line 1: Probably we dont need a .bat file

docs_requirements.txt
```diff
@@ -0,0 +1,94 @@
+alabaster==0.7.12
```
TheMP, line 1: could be probably more minimal :)

PR #43, opened 2020-04-23
Fix/timeseries.plot

pennfranc:
Restored functionality to plot TimeSeries on same x axis.

### DIFF
--- a/u8timeseries/timeseries.py
+++ b/u8timeseries/timeseries.py
@@ -402,15 +402,8 @@ class TimeSeries:
         """
         Currently this is just a wrapper around pd.Series.plot()
         """
-        # temporary work-around for the pandas.plot issue
-        # errors = self._combine_or_none(self._confidence_lo, self._confidence_hi,
-        #                                lambda x, y: np.vstack([x.values, y.values]))
-        # self._series.plot(yerr=errors, *args, **kwargs)
-        if new_plot:
-            fig, ax = plt.subplots()
-        else:
-            fig, ax = plt.gcf(), plt.gca()
-        self._series.plot()
+        fig = (plt.figure() if new_plot else plt.gcf())
+        self._series.plot(figure=fig)
         x_label = self.time_index().name
         if x_label is not None and len(x_label) > 0:
             plt.xlabel(x_label)
--- a/u8timeseries/timeseries.py
+++ b/u8timeseries/timeseries.py
@@ -8,6 +8,9 @@ from .custom_logging import raise_log, raise_if_not, get_logger
 
 logger = get_logger(__name__)
 
+fig = None
+ax = None
+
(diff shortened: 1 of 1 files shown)


u8timeseries/timeseries.py
```diff
@@ -398,15 +401,17 @@ def from_times_and_values(times: pd.DatetimeIndex,
...
+    def plot(self, *args, plot_ci=True, new_plot=True, **kwargs):
         """
         Currently this is just a wrapper around pd.Series.plot()
         """
         # temporary work-around for the pandas.plot issue
         # errors = self._combine_or_none(self._confidence_lo, self._confidence_hi,
         #                                lambda x, y: np.vstack([x.values, y.values]))
         # self._series.plot(yerr=errors, *args, **kwargs)
-        fig, ax = plt.subplots()
+        global fig, ax
```
TheMP, line 412: Not a fan of keeping those as global, can backfire in other places. Would new_plot be enough?

pennfranc: Fair enough, another possibility would be to return ax, and provide it as an optional argument to the next invocation of plot()

pennfranc: As far as I can see, we need a way to hold on to the ax object to be able to plot on the same axis.

hrzn: You probably don't need to keep these globals yourself. You should be able to get the current Figure or Axes currently active in pyplot. I would keep the `new_plot` flag, default it to `False`, and then use such tricks (havent' tested):

```
if new_plot:
   fig, ax = plt.subplots()
else:
   fig, ax = plt.gcf(), plt.gca()
self._series.plot()
fig.autofmt_xdate()
```

pennfranc: This fixes it, thanks!

u8timeseries/timeseries.py
```diff
@@ -398,17 +398,19 @@ def from_times_and_values(times: pd.DatetimeIndex,
...
         #                                lambda x, y: np.vstack([x.values, y.values]))
         # self._series.plot(yerr=errors, *args, **kwargs)
-        fig, ax = plt.subplots()
-        ax.plot_date(self.time_index(), self.values(), marker='', linestyle='-', *args, **kwargs)
-        fig.autofmt_xdate()
+        if new_plot:
+            fig, ax = plt.subplots()
+        else:
+            fig, ax = plt.gcf(), plt.gca()
+        self._series.plot()
```
hrzn, line 413: So now that you do it this way, you never use the `fig` and `ax` variables, so I don't think the `new_plot` flag will work when you set it to True. What you need is probably something like:
```
fig = (plt.figure() if new_plot else plt.gcf())
self._series.plot(figure=fig)
```

PR #45, opened 2020-04-24
use action cache

endrjuskr:


.github/workflows/update_docs.yml
```diff
@@ -13,11 +13,22 @@ jobs:
...
         python-version: 3.6
-        
-    - name: Install dependencies
+
+    - name: Prepare machine
       run: |
         sudo apt-get install -y pandoc
         python -m pip install --upgrade pip
+
+    - uses: actions/cache@v1
```
TheMP, line 22: I hope it won't kick us as badly as in gitlab :)

PR #46, opened 2020-04-24
Improve docs

endrjuskr:
* Add readme
* Clean up package and module sites.
* Different way of handling dependencies


requirements/main.txt
```diff
@@ -0,0 +1,13 @@
+numpy>=1.18.2
```
TheMP, line 1: main is a requirement file for which part?

PR #47, opened 2020-04-26
Feature/rnns improvements

hrzn:
A version of RNN implementations, which (I think) should be ready for merging. It is based on @Droxef original work. I have (quite drastically) simplified it.

Notable changes:
* There is no workaround for multi-variate time series. This version assumes the TimeSeries class is univariate only. We will extend the class to multi-variate in due time.
* I have changed quite a lot the arguments and signatures of some of the methods of the RNNModel class. The main goal was to simplify, and avoid methods with external effects, or modifying class variables from everywhere. It should now be much simpler to understand the state of this class.
* I have also changed the checkpointing. Before it required having an instance of the class with the exact same & correct state before loading a checkpointed model. Now checkpoints contain the whole state, and there is a static method which allows to retrieve a (full) checkpointed model.
* There is no obligation to have called "set_optimizer()" etc before fitting the model. It's all taken care of with the constructor.
* The DataLoader does not anymore take care of transforming / normalizing the time series (which was introducing quite some complexity). Now there is a separate kind of model called `Transformer` which can be applied on any `TimeSeries` for normalization based on the usual scikit-learn Scalers.

I recommend going though the notebook `examples/RNN_examples.ipynb` to have a glimpse of how RNNs can be used.


u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+        :return: A TimeSeries containing the `n` next points, starting after the end of the training time series.
+        """
+
+        super().predict(n)
+
+        scaled_series = self.training_series.values()[-self.seq_len:]
+        pred_in = torch.from_numpy(scaled_series).float().view(1, -1, 1).to(self.device)
+        test_out = []
+        self.model.eval()
+        for i in range(n):
```
pennfranc, line 296: Apologies in advance if I'm just misunderstanding the code. But it seems to me that we're only ever using the first element of the predictions by the RNN. If that's the case, when does it make sense to set output_length to a value larger than 1?

hrzn: That is a good question. It's not used for predictions, but for training. At training time, we can compute the loss over a predicted sequence (see line 329). Note that this means that when output_length > 1, we do not train for the same task as what is done at prediction time. We might want to have this behaviour captured by a hyper-parameter at some point.

examples/Daily-temperature-example.ipynb
```diff
@@ -0,0 +1,1020 @@
...
...RK0AdIfPEnZENhHFE9Eq+IZw05jZimwr1/w2gMcBlFqQZ0cuA/iTiJYT0QgHZZ8OIAvAOCJaSURjiaimA3IDuR4+N4ZVTMlm5t0A3vBv2wsgh5n/lC3Xv39fImpARDUAXBxwjCjZqwFcTkQJRHQafG6TFvAZXxkBx2fAgEEmQK4t3KjA7wBwLxEth881UejfngKgBEBTAKcBeISITlctl4hqAfgRPn+dVWvRtGxmLmHmbgCaA0ghoi6y5RLRMACZzLzcykValev/7lxm7gFgqP/Yvg7JToBvmPwBM3eHT/GMckAuAICIqsCnEH6wINOSbCI6BcDl/m1N4bOWb5Itl5nTALwKn+U6BT6lV2xBbjjZn8GnnFPhM0YW+mVQkHNYCdEzK9cWrquJycwbAAwGACJqB99wHfD5zaYwcxGATCJaAN8QZJsquUSUCJ/y/pqZf3JSdsCxh4loNoAh8L39ZcrtDuAyIroYvonUOkT0FTOberitXC8z7/Efm0lEP8OnBOZWOrl42XMBZASMcCbCggK38RsPBbCCmfeblWlDNgPYzsxZ/mN+AtAHwFeS5W5j5k/hd1cR0UsobxXblu13zTxUth8RLQSwGUA2fMZQGc0RwW0kSK4tXGeBE1Ej//9xAP4F4EP/VzsBDCAfNeFz/G9QJZeICL6OlsbMbzosO4mI6vmPqQ5gECzcC7NymXk0Mzdn5lbwDetnmlXeFq+3JhHV9h9TE74HxNTLysY17wOwi4ja+/cbCGC9bLkBh94Ae+4TK7J3AuhNRDX8/XwgfHM8suUGHtMSwFWweO2hZPuvqab/84UAipl5vd8HfZSIevuv+RYAk2TLtXJt5TA76ynyH3w/zl4ARfC9aYfDF0a0yf/vFZxcbFQLvmHkOvgeoMfCnUe2XADnwWeprMHJUK+LnbhmAF3hC+NbA58i+49T9zrgfP1gLApFxPWeDt9wuixs8imH+1c3+Ia+awD8D8ApDsmtAeAggLoKnqln4VOqawF8CaCqQ3Ln+betBjBQwjW3ArARvhfSdARE9sA3ClgLYCuA98qOcUDua/7jS/3/P2P099YrMTUajcajuM6FotFoNBpjaAWu0Wg0HkUrcI1Go/EoWoFrNBqNR9EKXKPRaDyKVuAajUbjUbQC12g0Go+iFbhGo9F4lP8HiDIGj23JcSIAAAAASUVORK5CYII=\n",
```
TheMP, line 74: I hope the pics won't take too much space on github :)

requirements-docker.txt
```diff
@@ -0,0 +1,16 @@
+numpy
```
TheMP, line 1: I think this can be dropped as @endrjuskr split int into requirements/main.txt and extenstions. @endrjuskr can you confirm?

requirements/main.txt
```diff
@@ -9,5 +9,7 @@ fbprophet>=0.5
 scikit-learn>=0.21.2
 pandas>=0.23.1
 tqdm>=4.32.1
-ipywidgets
-testfixtures
\ No newline at end of file
+torch>=1.4.0
+tensorboard>=2.2.1
+ipywidgets>=7.5.1
+testfixtures
```
TheMP, line 15: will prevent future issues if we freeze the version... which probably applies to the whole file.

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+            last = feature
+        self.fc = nn.Sequential(*feats)
+
+    def forward(self, x):
+        # data is of size (batch_size, input_length, input_size)
+        batch_size = x.size(0)
+
+        out, hidden = self.rnn(x)
+
+        # Reshaping the outputs such that it can be fit into the fully connected layer
```
TheMP, line 91: Can it be dropped?

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+        # predictions is of size (batch_size, output_length, 1)
+        return predictions
+
+
+class RNNModel(AutoRegressiveModel):
+    def __init__(self,
+                 model: Union[str, nn.Module] = 'RNN',
+                 output_length: int = 1,
+                 input_length: int = 12,
+                 hidden_size: int = 25,
```
TheMP, line 117: what if there is more than 1 hidden layer? How should the params look like?

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+
+        if model in ['RNN', 'LSTM', 'GRU']:
+            hidden_fc_size = [] if hidden_fc_size is None else hidden_fc_size
+            self.model = RNNModule(name=model, input_size=self.input_size, hidden_dim=hidden_size,
+                                   num_layers=n_rnn_layers, output_length=output_length,
+                                   num_layers_out_fc=hidden_fc_size, dropout=dropout)
+        elif isinstance(model, nn.Module):
+            self.model = model
+        else:
+            raise ValueError('{} is not a valid RNN model.\n Please specify' \
```
TheMP, line 189: @pennfranc introduced log_error function that could be useful here

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+        """
+        :param series: The training time series
+        :param val_series: Optionally, a validation time series that will
+                           be used to compute validation loss throughout training
+        """
+
+        super().fit(series)
+
+        if self.from_scratch:
+            shutil.rmtree(_get_checkpoint_folder(self.work_dir, self.model_name), ignore_errors=True)
```
TheMP, line 243: can be dangerous if variables are not set correctly.

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+        if self.batch_size is None:
+            self.batch_size = len(series) // 10
+            print('No batch size set. Using: {}'.format(self.batch_size))
+
+        # Prepare training data:
+        dataset = TimeSeriesDataset1D(series, self.seq_len, self.output_length)
+        train_loader = DataLoader(dataset, batch_size=self.batch_size, shuffle=True,
+                                  num_workers=0, pin_memory=True, drop_last=True)
+        if len(train_loader) == 0:
+            raise ValueError('The provided training time series is too short for obtaining even one training point.')
```
TheMP, line 254: same comment as one about log_error from @pennfranc

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+
+        scaled_series = self.training_series.values()[-self.seq_len:]
+        pred_in = torch.from_numpy(scaled_series).float().view(1, -1, 1).to(self.device)
+        test_out = []
+        self.model.eval()
+        for i in range(n):
+            out = self.model(pred_in)
+            pred_in = pred_in.roll(-1, 1)
+            pred_in[:, -1, :] = out[:, 0]
+            test_out.append(out.cpu().detach().numpy()[0, 0])
```
TheMP, line 300: ATM we can only use CPU?

u8timeseries/models/RNN_model.py
```diff
@@ -0,0 +1,469 @@
...
+        filename = 'checkpoint_{0}.pth.tar'.format(epoch)
+        os.makedirs(folder, exist_ok=True)
+        filename = os.path.join(folder, filename)
+
+        with open(filename, 'wb') as f:
+            pickle.dump(self, f)
+
+        if len(checklist) >= 5:
+            # remove older files
+            for chkpt in checklist[:-4]:
```
TheMP, line 414: isn't it basically checklist[0]?

examples/Daily-temperature-example.ipynb
hrzn: Yeah... I wonder how we should treat notebooks in general, it's kind of a pain to have them tracked by git (as usually executing them changes them).

u8timeseries/models/RNN_model.py
hrzn: The RNNs would be stacked, with same hidden layer sizes. Here, we follow PyTorch, see e.g.: https://pytorch.org/docs/stable/nn.html?highlight=rnn#torch.nn.RNN

u8timeseries/models/RNN_model.py
hrzn: We should be able to train and run inference on GPU (although I haven't tested), but here, the method `predict()` returns a `TimeSeries` to the user. And thus the returned data is not on GPU memory, but backed by a numpy array on regular memory.

u8timeseries/models/RNN_model.py
hrzn: That's `checklist[0]` only when `len(checklist) == 5` but not when `len(checklist) > 5`

u8timeseries/models/RNN_model.py
hrzn: True... I don't like it too much. what would be a better way to manage our checkpoints folders?

u8timeseries/models/RNN_model.py
hrzn: Yes, done

u8timeseries/timeseries.py
```diff
@@ -312,10 +301,9 @@ def rescale_with_value(self, value_at_first_step: float) -> 'TimeSeries':
         Numerical imprecisions appear with [value_at_first_step] > 1e+24
 
         :param value_at_first_step: The new value for the first entry of the TimeSeries.
-        :return: A new TimeSeries, whose first value was changed to [value_at_first_step] and whose others values
-                have been scaled accordingly.
+        :return: A new TimeSeries, where the first value is [value_at_first_step] and other values
```
MounBen, line 304: The brackets should be changed with quotation marks (so their appear in italic in the documentation :) )

hrzn: Good one, thanks. I plan to anyway rework all the comments in the codebase to follow Numpy style.

PR #48, opened 2020-04-27
Merge workflows.

endrjuskr:


endrjuskr:
Removed docker from tests in github actions and you can see output in logs. There are Errors but it is expected as we test logging, still it would be nice to not output in to console as it is confusing, but this is other PR.

PR #54, opened 2020-04-28
Feature/fft [WIP]

pennfranc:
Implemented FFT-based forecasting model that works by cropping the training set to whole periods, detrending the data, and filtering low-amplitude frequencies in the frequency domain.
Please see https://github.com/unit8co/u8timeseries/blob/feature/FFT/examples/FFT-example.ipynb for a demonstration.


u8timeseries/models/fft.py
```diff
@@ -0,0 +1,125 @@
...
+        if compare_seasonality(pred_ts, curr_ts, required_matches):
+            new_series = series.drop_before(curr_ts)
+            return new_series
+    
+    logger.warning("No matching timestamp could be found, returning original TimeSeries.")
+    return series
+
+class FFT(AutoRegressiveModel):
+
+    def __init__(self, filter_first_n: int = -1, required_matches: set = {}, detrend: bool = False, detrend_poly_degree: int = 3):
```
hrzn, line 57: How about renaming `filter_first_n` into `nr_freqs_to_keep` ?
Also concerning the `required_matches`, I think this will likely be quite complex to understand & use (especially because it requires knowing the pd.Timestamps attributes). How about an "auto" version, which, when predict() is called, first tries to figure out the correct phase shift?

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        Returns the most significant frequencies that were chosen when training the model.
+        """
+        if (not self._fit_called):
+            raise_log(Exception('fit() must be called before chosen_frequencies()'), logger)
+        frequencies = np.fft.fftfreq(len(self.fft_values))
+        chosen_frequencies = frequencies[self.filtered_indices]
+        return chosen_frequencies
+
+    @time_log(logger=logger)
```
hrzn, line 90: I wonder if it's a good idea to always log/display the time it takes. In notebooks it produces such outputs:
image

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,125 @@
...
+    def fit(self, series: 'TimeSeries'):
+        super().fit(series)
+
+        # determine trend
+        if (self.detrend):
+            trend_coefficients = np.polyfit(range(len(self.training_series)), series.values(), 3)
```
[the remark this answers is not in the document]
hrzn: Also, it might be a good idea to optionally fit an exponential trend (rather than logarithmic), as it can also be a pretty common one.

u8timeseries/models/fft.py
hrzn: Also: I would rename `detrend` to `trend` and `detrend_poly_degree` to `trend_poly_degree`.
Finally, instead of default values of -1, it's probably better to use a type hint Optional[int] to make it clear this is not required (and set it to None by default).

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Compares two timestamps according two a given set of attributes (such as minute, hour, day, etc.).
+    It returns true if and only if the two timestamps are matching in all given attributes.
+
+    :param ts_1: First timestamp that will be compared.
+    :param ts_2: Second timestamp that will be compared.
+    :required_matches: A set of pd.Timestamp attributes which ts_1 and ts_2 will be checked on.
+    :return: True if and only if 'ts_1' and 'ts_2' match in all attributes given in 'required_matches'.
+    """
+    is_matching = True
```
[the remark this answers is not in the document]
pennfranc: Very nice, thank you

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,125 @@
...
+        # subtract trend
+        detrended_values = self.cropped_series.values() - self.trend(range(len(series) - len(self.cropped_series), len(series)))
+
+        # perform dft
+        self.fft_values = np.fft.fft(detrended_values)
+
+        # filter out low-amplitude frequencies
+        first_n = self.filter_first_n
+        if (first_n < 0 or first_n > len(self.fft_values)): first_n = len(self.fft_values)
+        self.filtered_indices = np.argpartition(abs(self.fft_values), -first_n)[-first_n:]
```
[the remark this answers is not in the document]
pennfranc: Yeah I believe this would work just as well. But if I understand correctly, it might be a bit less efficient since we sort the whole array, whereas np.argpartition can be seen as one single iteration (partition) of quicksort.

u8timeseries/models/fft.py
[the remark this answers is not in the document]
pennfranc: done

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,125 @@
...
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+
+logger = get_logger(__name__)
+
+
+### helper functions that might have to be relocated ###
+
+def compare_seasonality(ts_1: pd.Timestamp, ts_2: pd.Timestamp, required_matches: set):
```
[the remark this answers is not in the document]
pennfranc: done

u8timeseries/models/fft.py
pennfranc: done

u8timeseries/models/fft.py
pennfranc: kind of done, I suspended info logging for historical_forecasting

u8timeseries/models/fft.py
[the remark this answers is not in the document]
hrzn: OK, that was also my understanding. But it might be worth to add this sentence of yours as a comment in the code, to make it easier for our future selves to understand it again :)

u8timeseries/models/fft.py
```diff
@@ -54,30 +48,30 @@ def crop_to_match_seasons(series: 'TimeSeries', required_matches: set):
 
 class FFT(AutoRegressiveModel):
 
-    def __init__(self, filter_first_n: int = -1, required_matches: set = {}, detrend: bool = False, detrend_poly_degree: int = 3):
+    def __init__(self, nr_freqs_to_keep: Optional[int] = None, required_matches: set = {}, trend: bool = False, trend_poly_degree: int = 3):
```
hrzn, line 51: about the `{}` default value: you shouldn't have mutable defaults. It can lead to nasty behaviors (see e.g. this: https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)

u8timeseries/models/fft.py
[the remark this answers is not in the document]
pennfranc: Ah I see, that makes sense!

u8timeseries/models/fft.py
pennfranc: Wow, I would not have expected that behavior of default arguments at all. Thanks for making me aware of this!

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,201 @@
...
+from ..custom_logging import raise_if_not, time_log, get_logger
+from ..models.statistics import check_seasonality
+import numpy as np
+import pandas as pd
+from statsmodels.tsa.stattools import acf
+from typing import List, Optional
+
+logger = get_logger(__name__)
+
+def check_approximate_seasonality(series: TimeSeries, seasonality_period: int,
```
hrzn, line 12: I guess this method (and the following ones below) are not meant to be called from outside this file, so I would prepend them with `_`.

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,201 @@
...
+class FFT(AutoRegressiveModel):
+
+    def __init__(self, nr_freqs_to_keep: Optional[int] = None, required_matches: Optional[set] = None, 
+                 trend: bool = None, trend_poly_degree: int = 3, automatic_matching=False):
+        """
+        Forecasting based on a discrete fourier transform using FFT of the (cropped and detrended) training sequence
+        with subsequent selection of the most significant frequencies to remove noise from the prediction.
+
+        :param nr_freqs_to_keep: The total number of frequencies that will be used for forecasting.
+        :param required_matches: The attributes of pd.Timestamp that will be used to create a training
```
hrzn, line 128: I think this description could be made a bit more clear, and would also benefit from a couple of examples of the correct parameters for a couple of time series.

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,201 @@
...
+                                 sequence that is cropped at the beginning such that the first timestamp 
+                                 of the training sequence and the first prediction point have matching 'phases'.
+                                 If the series has a yearly seasonality, include 'month', if it has a monthly 
+                                 seasonality, include 'day', etc.
+        :param trend: Boolean value indicating whether or not detrending will be applied before performing DFT.
+        :param trend_poly_degree: The degree of the polynomial that will be used for detrending.
+        :param automatic_matching: If set to True, the model tries to find the pd.Timestamp attributes that
```
hrzn, line 135: I think you don't need this one. You can just do the automatic mode when `required_matches` is None.

u8timeseries/models/fft.py
```diff
@@ -0,0 +1,201 @@
...
+        if (check_approximate_seasonality(series, 12, 1, 0)):
+            relevant_attributes.add('month')
+    elif (type(series.freq()) == pd.tseries.offsets.Day):
+        # check for yearly seasonality
+        if (check_approximate_seasonality(series, 365, 5, 20)):
+            relevant_attributes.update({'month', 'day'})
+        # check for monthly seasonality
+        elif (check_approximate_seasonality(series, 30, 2, 2)):
+            relevant_attributes.add('day')
+    # TODO: add other seasonality cases
```
hrzn, line 72: Yes, could to add more combinations :)

u8timeseries/models/fft.py
pennfranc: done

u8timeseries/models/fft.py
pennfranc: done (I hope it got better)

u8timeseries/models/fft.py
pennfranc: done

u8timeseries/models/fft.py
pennfranc: done

PR #55, opened 2020-04-28
Add conda

radujica:
* add conda recipe for publishing
* add conda environment for dev setup and instructions
* clean/separate requirements

Looks like we can't do pure conda for now :(
https://github.com/alkaline-ml/pmdarima/issues/326

For publishing/CD (both pypi and conda) we could depend on github releases and setup github action, as long as we sync setup.py version with a git tag (e.g. bump2version?). Actual release creation can trigger it, so it's still manual, but no individual developer setup would be required.

Other notes:
`conda-build conda_recipe --python 3.x`
`conda convert --platform all <path-to-bz2>`
`conda-verify <path-to-bz2>`


conda_recipe/u8timeseries/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+build:
+  number: 0
+  script: '{{ PYTHON }} -m pip install . --no-deps -vv'
+
+requirements:
+  host:
+    - python >=3.6
+  run:
+  # only essential requirements to use this library ~ setup.py install_requires ~ requirements/main.txt
```
hrzn, line 19: Isn't there a way to tie this to `requirements/main.txt` ?

.github/workflows/workflow.yml
```diff
@@ -22,13 +22,14 @@ jobs:
         id: cache
         with:
           path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('requirements/main.txt') }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements/main.txt', 'requirements/tests.txt') }}
```
TheMP, line 25: Pls align with https://github.com/unit8co/u8timeseries/pull/49

requirements/publish.txt
```diff
@@ -0,0 +1 @@
+twine
```
TheMP, line 1: pls freeze the version

conda_recipe/u8timeseries/meta.yaml
radujica: Made it work :)

I can now do jinja 'coding', lolz

requirements/tests.txt
```diff
@@ -0,0 +1 @@
+testfixtures
```
TheMP, line 1: I feel safer when the versions are frozen :)

setup.py
```diff
@@ -1,24 +1,23 @@
...
+      version=open('u8timeseries/VERSION', 'r').read(),
       description='A collection of easy-to-use timeseries forecasting models',
       url='http://github.com/unit8co/u8timeseries',
       author='Unit8 SA',
       author_email='info@unit8.co',
       license='Apache License 2.0',
       packages=find_packages(),
-      install_requires=requirements["main"],
-      extras_require={"docs": requirements["docs"]},
+      install_requires=read_requirements('requirements/main.txt'),
```
TheMP, line 19: @endrjuskr will this work without extras_require?

.github/workflows/workflow.yml
radujica: I split the test requirements so the tests requirement also need to be cached (and installed)

setup.py
radujica: just pointing out that the github actions do work with this change

.github/workflows/workflow.yml
```diff
@@ -54,14 +55,15 @@ jobs:
...
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements/main.txt', 'requirements/docs.txt', 'requirements/docs.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
 
     - name: Install dependencies
       run: |
         pip install .
-        pip install -e .[docs]
+        pip install -r requirements/docs.txt
+        pip install -r requirements/tests.txt
```
radujica, line 66: docs might be also building docs for the tests; if that's the case, could exclude the tests and also not install the deps here

PR #57, opened 2020-04-29
Put plotting dependencies as extras

radujica:
Plotting does not seem as core functionality of u8timeseries atm (and it's just bloating our builds and dependencies) so mark them as optional/extra requirements, i.e. to make use also of plot stuff:

    pip install .[plots]

   instead of 

    pip install .

(and once published, pip install u8timeseries[plots] should work)

But everything else (so >95%) will still just work without them.

Notes
- like ~20s faster CI
- IPython and tqdm also seem optional?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@
 ### pure pip
 This approach most likely requires other non-Python dependencies.
 
+    # if plotting tweaks provided by u8timeseries are required
+    pip install .[plots]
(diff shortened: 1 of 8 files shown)


hrzn:
I'm not quite sure about this one. I think plotting is likely to be used very often. Perhaps could be worth having a discussion? @TheMP

PR #59, opened 2020-04-29
fix(logging): removed all custom_logging messages from unittest output

pennfranc:
Removed all custom_logging messages from unittest output by
1. Suppressing logging in all unittest files where log messages are printed.
2. Redirect log messages to NullHandler() for test_logging.py (logging cannot be suppressed here since we test for the right logging messages).

### DIFF
--- a/u8timeseries/custom_logging.py
+++ b/u8timeseries/custom_logging.py
@@ -4,7 +4,7 @@ from datetime import datetime
 import pathlib
 import time
 
-def get_logger(name: str):
+def get_logger(name: str, handler=logging.StreamHandler()):
     """
     Internally calls the logging.getLogger function with the 'name' argument to create or 
     retrieve a logger object. It is recommended to pass __name__ as argument when calling 
@@ -12,11 +12,11 @@ def get_logger(name: str):
     the messages appropriately.
 
     :param name: The name that gets passed to the logger.getLogger function.
+    :param handler: The handler instance to be added to the logger.
     :return: A logger instance with the given name.
     """
     logger = logging.getLogger(name)
     logger.setLevel(logging.INFO)
-    handler = logging.StreamHandler()
     formatter = logging.Formatter('[%(asctime)s] %(levelname)s | %(name)s | %(message)s')
     handler.setFormatter(formatter)
     logger.addHandler(handler)
--- a/u8timeseries/custom_logging.py
+++ b/u8timeseries/custom_logging.py
@@ -4,7 +4,7 @@ from datetime import datetime
 import pathlib
 import time
 
-def get_logger(name):
+def get_logger(name: str):
     """
     Internally calls the logging.getLogger function with the 'name' argument to create or 
     retrieve a logger object. It is recommended to pass __name__ as argument when calling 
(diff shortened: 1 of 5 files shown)


u8timeseries/custom_logging.py
```diff
@@ -16,10 +16,10 @@ def get_logger(name):
     """
     logger = logging.getLogger(name)
     logger.setLevel(logging.INFO)
-    stderr_handler = logging.StreamHandler()
+    handler = logging.StreamHandler()
```
endrjuskr, line 19: How about putting handler as parameter?

u8timeseries/tests/test_metrics.py
```diff
@@ -17,6 +18,10 @@ class MetricsTestCase(unittest.TestCase):
     series2: TimeSeries = TimeSeries(pd_series2)
     series3: TimeSeries = TimeSeries(pd_series3)
 
+    @classmethod
+    def setUpClass(cls):
+        logging.disable(logging.CRITICAL)
```
TheMP, line 23: what is this doing?

u8timeseries/custom_logging.py
endrjuskr, line 19: Is there a way to check what was logged, but not print it? For instance, keep every log in memory and then check if some message is in memory?

u8timeseries/tests/test_logging.py
```diff
@@ -12,10 +13,15 @@
...
 
+    @classmethod
+    def setUpClass(cls):
+        logging.disable(logging.NOTSET)
+
     def test_raise_log(self):
         exception_was_raised = False
         with LogCapture() as lc:
             logger = get_logger(__name__)
+            logger.handlers = [logging.NullHandler()]
```
TheMP, line 24: what is this line doing?

u8timeseries/tests/test_metrics.py
pennfranc: This suppresses the logging of all log messages below the level of CRITICAL. In this case, this happens before the execution of all test cases for timeseries. I chose this approach here because for test_timeseries.py, we do not need any log messages.

u8timeseries/custom_logging.py
pennfranc: I think the behaviour you're describing is achieved by the LogCapture class from testfixtures. This class can catch logs without printing them and lets us check for the presence of a specific log. But this only works for the root logger. That means that if we create our own logger with an additional handler, the logs from that handler do not get prevented from printing (note that every logger by default propagates its logs to the root logger as well). This is why, before testing log messages using LogCapture, I replace the handler with the NullHandler (although I could have just removed the handler as well).
I hope all of this makes sense.

pennfranc: And set logging.StreamHandler() as its default value?

u8timeseries/tests/test_logging.py
pennfranc: It replaces our custom handler (which was created in the get_logger() function from custom_logging.py) by a handler that does not produce any output. This is done because although LocCapture prevents the root logger from printing the log message, it does not capture the log message of our custom handler. I could have also just removed our custom handler, which might be a better option. I will do this as part of the next push.

TheMP: ok, makes sense :)

PR #60, opened 2020-04-30
Move plotting dependencies as extras

radujica:
Plotting does not seem as a core functionality of u8timeseries (and it's just bloating our builds and dependencies) so mark them as optional/extra requirements, i.e. to make use also of plot stuff:

    pip install .[plots]

   instead of 

    pip install .

(and once published, `pip install u8timeseries[plots]`)

But everything else (so >95%) will still just work without them.

Notes
- like ~20s faster CI
- IPython and tqdm also seem optional? if that's the case, could add those as optional in the same extras bundle, e.g. `u8timeseries[visuals]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,12 @@ and combine the predictions of several models and external regressors.
 ### pure pip
 This approach most likely requires other non-Python dependencies.
 
-```
+    # if plotting tweaks provided by u8timeseries are required
+    pip install .[plots]
+    # otherwise just
     pip install .
     # install any additional dev requirements, e.g.:
     pip install -r requirements/docs.txt
-```
 
 **Running the examples only without installing:**
 
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -2,9 +2,6 @@ numpy>=1.18.1
 scipy>=1.4.1
 statsmodels>=0.11.1
 pmdarima>=1.5.3
-matplotlib>=3.2.1
 fbprophet>=0.5
 pandas>=0.23.1
 tqdm>=4.32.1
-ipywidgets
-plotly>=4.6.0
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -5,3 +5,4 @@ pmdarima>=1.5.3
 fbprophet>=0.5
 pandas>=0.23.1
 tqdm>=4.32.1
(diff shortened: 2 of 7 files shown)


hrzn:
I'm not too sure... I think plotting will likely be very often useful, and it's a bummer to have to launch a special pip command to have it. The goal of the library should be simplicity of use (including install), and "getting everything out of the box". I think that should have precedence over CI tests time.

endrjuskr:
IMO if library does not work 100% without plots then we should not remove it.
Library is intended to compare timeseries, which does not say we need to visualise them, so I would recommend remove all visual related things from core library and allow use to chose ploting lib. We can provide ways to integrate library with matplotlib, but this is something on top of core lib.

hrzn:
That is one way of seeing it. However I think 80% of the time when you work with time series, you want to visualise them at some point. We could see this like the functionality on Pandas Series or Dataframe objects: although Pandas is not a visualisation tool per-se, you can still call my_series.plot(), which is extremely useful, and IMO should come out-of-the-box with u8timeseries. 

Perhaps we could have the reverse? `pip install u8timeseries[core]` to install only the core (for instance for server use)?

radujica:
pandas also doesn't include matplotlib by default. Users are required to install it separately.

By making it an extra, we would actually do more (compared to no extra at all and just core lib) by providing recommended and possibly tested visualization libs (matplotlib, plotly, etc.)

    >>> df.plot()
    Traceback (most recent call last):
    File "<stdin>", line 1, in <module>
    File "/Users/radujica/anaconda3/envs/pandas-plot/lib/python3.8/site-packages/pandas/plotting/_core.py", line 767, in __call__
    plot_backend = _get_plot_backend(kwargs.pop("backend", None))
    File "/Users/radujica/anaconda3/envs/pandas-plot/lib/python3.8/site-packages/pandas/plotting/_core.py", line 1666, in _get_plot_backend
      raise ImportError(
    ImportError: matplotlib is required for plotting when the default backend "matplotlib" is selected.

hrzn:
OK I see your point @radujica. I'm hesitating.. because on the other hand, why removing plotting from core, if anyway most users will plot (at least I think)? @TheMP @pennfranc @endrjuskr any opinion?

PR #61, opened 2020-04-30
Fix/rework_imports

pennfranc:
Removed all imports from __init__ files, changed all imports within the library to relative imports, adjusted imports in notebooks.


u8timeseries/backtesting/__init__.py
```diff
@@ -1,2 +0,0 @@
-from .backtesting_udf import get_train_val_series, backtest_autoregressive_model
-from .forecasting_simulation import simulate_forecast_ar, simulate_forecast_regr
```
hrzn, line 2: Can we still do `from u8timeseries.backtesting import simulate_forecast_ar` without these?

u8timeseries/metrics/__init__.py
```diff
@@ -1 +1 @@
-from .metrics import mape, mase, overall_percentage_error, marre, r2_score, coefficient_variation
```
endrjuskr, line 1: Please keep comments in headers for documentation purposes. They will appear when you merge develop back to your branch

u8timeseries/backtesting/__init__.py
pennfranc: No, this does not work anymore. Now it would have to be imported like this: `from u8timeseries.backtesting.forecasting_simulation import simulate_forecast_ar`. Unless there's another way I don't know about.

u8timeseries/metrics/__init__.py
pennfranc: You mean keep 'Removed all imports...' in the header of the PR?

u8timeseries/backtesting/__init__.py
hrzn: Well I believe that if you don't remove these lines in the sub-modules `__init__.py` files it should still work? In any case, I don't think the users should be exposed to the names of the files we are using - i.e., users should write `from u8timeseries.backtesting import simulate_forecast_ar` but not `from u8timeseries.backtesting.forecasting_simulation import simulate_forecast_ar`

pennfranc: Ok yes, I think I misinterpreted your reply on Slack. I will reintroduce the submodule `__init__.py` file imports.

examples/Air-passengers-example.ipynb
```diff
@@ -6,26 +6,25 @@
    "metadata": {
```
TheMP, line 6: Will be easier to keep output here as this renders on the github page for people that browse the repo. Also much less diff to store for git

pennfranc: Does this mean it's important to run the notebook before pushing it to GitHub (assuming it has changed)?

u8timeseries/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+Tests
+-----
+"""
```
TheMP, line 4: might need a newline :)

u8timeseries/utils/statistics.py
```diff
@@ -0,0 +1,204 @@
+from ..timeseries import TimeSeries
```
TheMP, line 1: Is this a brand new file or was it moved from somewhere?

pennfranc: This was moved from the the models module. Julien suggested that the models module should only contain forecasting (and probably regression) models. @hrzn please correct me if I'm wrong here.

u8timeseries/tests/__init__.py
pennfranc: done

u8timeseries/metrics/__init__.py
pennfranc: done

examples/Air-passengers-example.ipynb
hrzn: @pennfranc I think it would be nice yes. However notebooks are kind of a pain to maintain in the repo overall... it'd be nice to find a better setup for them somehow.

hrzn:
@pennfranc Re-reading this PR, I think it might actually be better to first merge my branch `feature/refactor` (PR coming soon), and then this one. I think it'll be a bit less messy this way, I hope you don't mind waiting a little (sorry for this).

PR #62, opened 2020-04-30
Add flake8 linter

radujica:
Apart from the first 3 files, all changes are just making me .. er.. the linter happy

Can look at second-to-last 9a63328 commit for the kind of things it caught.

Could also add e.g. pylint but that one is much more picky.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -35,6 +35,10 @@ jobs:
         run: |
           python -m unittest
 
+      - name: Run linter
+        run: |
+          flake8 --config=setup.cfg u8timeseries
+
   docs:
     runs-on: ubuntu-latest
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ and combine the predictions of several models and external regressors.
 `pmdarima` is currently not supported on conda which means we need to reinstall it through pip
 
     conda env create -f conda_recipe/environment.yml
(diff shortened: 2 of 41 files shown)


.github/workflows/workflow.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         run: |
           python -m unittest
 
+      - name: Run linter
+        run: |
+          flake8 --config=setup.cfg u8timeseries
```
hrzn, line 40: does this automatically apply the changes on the code?

endrjuskr: @radujica is it failing on some warnings or errors or correct them? I believe former but double-checking :)

endrjuskr: Can we run it before tests or even as separate step so we can run it simultaneously? Unless it is < 2 secs then ignore it

radujica: It's like 1s so easiest to just keep it here.

It doesn't correct anything by itself, just fails the github action if an issue is found (after printing what failed, ofc).

IMO we shouldn't let some external tool edit our code and commit it, e.g. through black lib. It also subtly pushes for a proper dev environment if developing many things ;)

radujica:
> Can we also somehow run it from inside PyCharm?

Guess so: https://gist.github.com/tossmilestone/23139d870841a3d5cba2aea28da1a895

PR #65, opened 2020-04-30
Feature/more unittests

pennfranc:
Added unit tests that test the performance of the autoregression and regression models on real data. In the case of autoregression, all non-baseline models are expected to perform at least as well as the best baseline model (currently there is only 1).


u8timeseries/tests/test_autoregression_models.py
```diff
@@ -3,31 +3,67 @@
...
+        # iterate through all baseline models and save best scores
+        best_mape = 100
+        for baseline in self.baseline_models:
+            baseline.fit(self.ts_pass_train)
+            prediction = baseline.predict(len(self.ts_pass_val))
+            best_mape = min(mape(prediction, self.ts_pass_val), best_mape)
+
+        # iterate through all models and check if they are at least as good as the baselines
+        for model in self.models:
+            model.fit(self.ts_pass_train)
```
TheMP, line 64: out of curiosity, how long do tests execute?

pennfranc: Good question, I think it's best to measure once RNN is included too. I will merge it with develop later and measure the total time.

u8timeseries/models/arima.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, start_p=1, max_p=12, start_q=0, max_q=12, max_P=2, max_Q=2, s
     def __str__(self):
         return 'auto-ARIMA'
 
+    @time_log(logger=logger)
```
hrzn, line 75: I vote for removing these time logging in models. Printing progress can be useful (when possible), but logging time should be up to the user I think.

u8timeseries/tests/test_autoregression_models.py
```diff
@@ -3,31 +3,67 @@
...
-    # forecasting horizon used in predictions
+    # forecasting horizon used in runnability tests
     forecasting_horizon = 5
 
-    # dummy timeseries for autoregression forecasting
-    times = pd.date_range('20000101', '20000130')
-    values = np.sin(range(len(times))) + np.array([2.0] * len(times))
-    ts: TimeSeries = TimeSeries.from_times_and_values(times, values)
+    # dummy timeseries for runnability tests
+    np.random.seed(0)
```
hrzn, line 17: `0` stands for random seed right? I think you can use another value, for reproducibility and making the tests deterministic.

u8timeseries/tests/test_autoregression_models.py
```diff
@@ -3,31 +3,67 @@
...
+    ]
+    baseline_models = [
+        KthValueAgoBaseline()
+    ]
 
-        for model in models:
-            model.fit(self.ts)
+    # maximum error values for baselines
+    max_mape_baseline = 40
+
```
hrzn, line 39: It is quite cool to test both the models and the baselines. However here you introduce some coupling between them, which may render failures a bit less intuitive to interpret (did the baseline get better or did the model get worse?)

I think you could simply consider baselines as any other model, and test a list of models. Since this unit test is on a fixed dataset, we already know what MAPE to expect of each model, so we could use this as an upper bound (with some small tolerance). This way, you can have per-model thresholds, and test "more tightly" the models performing better. It would ensure that there is no performance regression if we change the models. For examples, if ExponentialSmoothing goes from 10% MAPE to 20% MAPE, we would catch it even if stays better than the baseline. You can also try building a few different versions of each model, with different parameters.

u8timeseries/tests/test_regression_models.py
```diff
@@ -3,28 +3,62 @@
...
+        for model in self.models:
             # training and predicting on same features, since only runnability is tested
-            model.fit([self.feature_ts], self.target_ts)
-            prediction = model.predict([self.feature_ts])
-            self.assertTrue(len(prediction) == len(self.feature_ts))
+            model.fit([self.ts_periodic, self.ts_gaussian], self.ts_sum)
+            prediction = model.predict([self.ts_periodic, self.ts_gaussian])
+            self.assertTrue(len(prediction) == len(self.ts_periodic))
+
+    def test_models_denoising(self):
```
hrzn, line 55: Very cool, thanks

u8timeseries/models/arima.py
pennfranc: Should I do it as part of this branch?

u8timeseries/tests/test_autoregression_models.py
pennfranc: done

u8timeseries/tests/test_autoregression_models.py
pennfranc: done

u8timeseries/models/arima.py
hrzn: No not needed

u8timeseries/tests/test_autoregression_models.py
```diff
@@ -4,34 +4,56 @@
...
+    np.random.seed(1)
+    ts_gaussian = tg.gaussian_timeseries(length=100, mean=50)
+
+    # real timeseries for functionality tests
+    df = pd.read_csv('examples/AirPassengers.csv', delimiter=",")
+    ts_passengers = TimeSeries.from_dataframe(df, 'Month', '#Passengers')
+    ts_pass_train, ts_pass_val = ts_passengers.split_after(pd.Timestamp('19570101'))
+
+    # autoregressive models - maximum error tuples
+    models = [
```
hrzn, line 27: cool!

u8timeseries/tests/test_autoregression_models.py
pennfranc: @TheMP Currently all tests run in 25 seconds. I decided not to include RNN in the test_autoregression_models test cases, since separate test cases already exist, and RNN takes quite long compared to other models.

PR #66, opened 2020-05-01
improved README

hrzn:
Made the README more public-facing. I moved all the notes about dev setups using Conda & pip to a private Quip page: https://unit8io.quip.com/EHR0A7vQSZWL/u8timeseries-dev-setup

Let's keep the README as simple and user-friendly as possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,38 +8,24 @@ The models can all be used in the same way, using `fit()` and `predict()` functi
 similar to scikit-learn. The library also makes it easy to backtest models,
 and combine the predictions of several models and external regressors.
 
-## install
+# Install
 
-### using conda
-`pmdarima` is currently not supported on conda which means we need to reinstall it through pip
-
-    conda env create -f conda_recipe/environment.yml
-    conda install -c conda-forge -c alkaline-ml --name u8timeseries-dev --file requirements/main.txt
-    # here install any additional dev requirements from the other requirements/*.txt files
-    source activate u8timeseries-dev
-    pip install "pmdarima>=1.5.3"
-    pip install --no-deps -e .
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,38 +8,24 @@ The models can all be used in the same way, using `fit()` and `predict()` functi
 similar to scikit-learn. The library also makes it easy to backtest models,
 and combine the predictions of several models and external regressors.
 
-## install
-
-### using conda
```
TheMP, line 13: Should we also keep conda steps somewhere?

README.md
```diff
@@ -69,17 +55,18 @@ plt.xlabel('Year')
...
+# Features
 Currently, the library contains the following features: 
 
 **Forecasting Models:** 
 * Exponential smoothing, 
 * ARIMA & auto-ARIMA,
 * Facebook Prophet,
-* Theta method, 
+* Theta method,
+* FFT (Fast Fourier Transform)
```
TheMP, line 69: This one is not actually released yet :)

README.md
radujica: I see conda setup needs to be checked/rewritten with the addition of pytorch, so maybe a different PR?

issues include:
- it's torch on pip, pytorch on conda -.-
- pytorch actually has some very specific install steps depending on the GPU and on whether using pip vs conda

hrzn: If we have a general way to `conda install u8timeseries` then yes we should document it here. But for a future PR I think.

README.md
hrzn: I know it will be very soon though, right @pennfranc ? :D

PR #69, opened 2020-05-06
Clean & document dependencies

radujica:
* easy install steps for all envs
* need fbprophet to 0.6 because they fix an easter 'egg' (https://github.com/facebook/prophet/issues/1300)
* make sure CI uses non-cuda torch

Ideally, we would just add per-OS requirements in requirements.txt, e.g.

    torch==1.6.0+cpu -f https://download.pytorch.org/whl/torch_stable.html; platform_system == "Windows"

but -f does not seem to work inside requirements.txt atm (https://github.com/pypa/pip/issues/8103)


.github/workflows/develop.yml
```diff
@@ -33,6 +33,8 @@ jobs:
 
       - name: "5. Install main and dev dependencies"
         run: |
+          # this avoids an unnecessary cuda install on Linux
```
endrjuskr, line 36: Do you think we can put it in some requirements?

README.md
```diff
@@ -9,11 +9,29 @@ similar to scikit-learn. The library also makes it easy to backtest models,
...
-```
-pip install u8timeseries
-```
+The following steps assume running inside a conda environment. 
+If that's not possible, first follow the official instructions to install 
+[fbprophet](https://facebook.github.io/prophet/docs/installation.html#python)
+and [torch](https://pytorch.org/get-started/locally/), then skip to 
+[Install u8timeseries](#install-u8timeseries)
+
+To create a conda environment for Python 3.7
```
endrjuskr, line 22: I would put 3.6 as it is our min requirement.

.github/workflows/develop.yml
radujica: nope, see PR:

> * easy install steps for all envs
> * need fbprophet to 0.6 because they fix an easter 'egg' ([facebook/prophet#1300](https://github.com/facebook/prophet/issues/1300))
> * make sure CI uses non-cuda torch
> 
> Ideally, we would just add per-OS requirements in requirements.txt, e.g.
> 
> ```
> torch==1.6.0+cpu -f https://download.pytorch.org/whl/torch_stable.html; platform_system == "Windows"
> ```
> 
> but -f does not seem to work inside requirements.txt atm ([pypa/pip#8103](https://github.com/pypa/pip/issues/8103))

README.md
radujica: sure, though it's just an example for users that never used conda

README.md
```diff
@@ -9,11 +9,29 @@ similar to scikit-learn. The library also makes it easy to backtest models,
...
+    conda create --name <choose-your-name> python=3.7
+
+#### MAC
+    conda install -c conda-forge -c pytorch pip fbprophet pytorch
+
+#### Linux and Windows
+    conda install -c conda-forge -c pytorch pip fbprophet pytorch cpuonly
+    
+## Install u8timeseries
+    pip install u8timeseries
```
hrzn, line 34: Running only this one out-of-the-box in a non-conda virtualenv doesn't work?

radujica: Nope, because of the non-Python platform-specific stuff, in this case fbprophet (and indirectly pystan) and possibly more.

Essentially everything in the Dockerfile is needed (plus whatever is in the FROM image) and Conda handles it.

hrzn: I see, bummer :/

PR #70, opened 2020-05-06
add badges

endrjuskr:
Screenshot 2020-05-06 at 11 25 19

* Build is failing because it is on master and we haven't merged it yet
* Last release is not present because we haven't released it yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![u8timeseries](static/images/logo-72dpi.png "u8timeseries")
 
-![[PyPI version](https://badge.fury.io/py/u8timeseries.svg)](https://badge.fury.io/py/u8timeseries)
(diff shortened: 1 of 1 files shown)


endrjuskr:
We can add docker hub badge as well

TheMP:
What would show on those badges if PR is merged? Especially the release date - is it taken from github release?
Are we able co confirm that verison above python 3.6 would run?

endrjuskr:
Release is coming from Github

Tests are passing on 3.6 version.

When we merge to develop they will show the same. We need to merge it to master to change values.

hrzn:
Very cool :) 
About downloads / month: maybe not worth showing until we reach some significant numbers?

PR #72, opened 2020-05-06
Fix/suppress prophet logs

pennfranc:
- Added a function that conditionally suppresses all stdout and stderr messages depending on the level of the logger and a given severity threshold.
- Used this function to suppress fbprophet logs created by the fit function.

### DIFF
--- a/u8timeseries/models/prophet.py
+++ b/u8timeseries/models/prophet.py
@@ -11,7 +11,7 @@ from ..custom_logging import time_log, get_logger, execute_and_suppress_output
 import logging
 
 logger = get_logger(__name__)
-logger.level = logging.WARNING # set to warning to suppress prophet logs
+logger.level = logging.WARNING  # set to warning to suppress prophet logs
 
 
 class Prophet(AutoRegressiveModel):
@@ -79,7 +79,6 @@ class Prophet(AutoRegressiveModel):
             self.model.add_country_holidays(self.country_holidays)
 
         execute_and_suppress_output(self.model.fit, logger, logging.WARNING, in_df)
-
 
     def predict(self, n):
         super().predict(n)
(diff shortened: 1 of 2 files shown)


u8timeseries/models/prophet.py
```diff
@@ -7,9 +7,11 @@
 import pandas as pd
 
 from u8timeseries.models.autoregressive_model import AutoRegressiveModel
-from ..custom_logging import time_log, get_logger
+from ..custom_logging import time_log, get_logger, execute_and_suppress_output
+import logging
 
 logger = get_logger(__name__)
+logger.level = logging.WARNING  # set to warning to suppress prophet logs
```
radujica, line 14: Will this warning level apply only to this module, or to the entire script using the module? If the latter, it would mean logging level switches to warning whenever u8timeseries fbprophet is used.

Reading https://github.com/facebook/prophet/issues/223 might be more accurate to use

    logging.getLogger('fbprophet').setLevel(logging.WARNING)

pennfranc: It will apply to that specific logger instance, which is used exclusively in prophet.py. So yes, the entire prophet.py script will be on that level you could say. But every other part of the project uses a different logger instance.
I'm doing something very similar to the code you're quoting, except I'm naming the logger 'u8timeseries.models.prophet' instead of 'fbprophet'. This is consistent with the naming I used in all scripts/modules.
I could replace `logger.level = ` with `logger.setLevel()`, that might be nicer.

radujica: Ah was thinking we could 'tap into' fbprophet's logger only.

Think this works fine though, as long as we don't have any info logs in our prophet.py which we would never see, if I understood correctly (unless we change the level in the code).

pennfranc: Yeah tapping into the fbprophet's logger would definitely be more elegant, but I couldn't make it work. Apparently the output originates from compiled code in C or Fortran according to this post: https://stackoverflow.com/questions/45551000/how-to-control-output-from-fbprophet. 

So right now we would have to change the level of the logger back to logging.INFO for the logs to be visible again, otherwise it will not appear.

PR #73, opened 2020-05-06
Add more info about lib

endrjuskr:
Based on https://github.com/alkaline-ml/pmdarima/blob/master/setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,16 +6,51 @@ def read_requirements(path):
     return list(Path(path).read_text().splitlines())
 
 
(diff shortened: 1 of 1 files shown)


endrjuskr:
@hrzn probably we need to update pip page as well after release

radujica:
> @hrzn probably we need to update pip page as well after release

could publish to https://test.pypi.org/ and make sure it looks ok before release. Uploads can be deleted afterwards, so no worries about versions.

PR #76, opened 2020-05-07
Feature/refactor

hrzn:
Big refactoring (sorry for the big PR):

* Added Numpy-style docstrings everywhere, and improved the documentation
  * (we still have to add the docstrings in utils/ and preprocessing/scaler_wrapper.py to the documentation, not quite sure how to do it)
* Renamed "AutoRegressiveModel" --> "ForecastingModel"
* Renamed "RegressiveModel" --> "RegressionModel"
* Added more baseline models, and renamed some
* Corrected quite a few bugs here and there
* Removed a few deprecated/useless things (notably `backtesting_udf`)
* Changed the way we wrap around some of the models; like AutoARIMA


TheMP:
Where should we focus the most?

requirements/dev.txt
```diff
@@ -1,5 +1,15 @@
+# docs
+sphinx
```
TheMP: Will be more stable with versions frozen

docs/source/conf.py
```diff
@@ -19,8 +19,8 @@
 # -- Project information -----------------------------------------------------
 
 project = 'u8timeseries'
-copyright = '2020, Unit8'
-author = 'Unit8'
+copyright = '2020, Unit8 SA'
```
TheMP, line 22: Nice touch ;]

u8timeseries/tests/test_logging.py
```diff
@@ -79,21 +76,25 @@ def test_timeseries_split_error_log(self):
...
         lc.check(
             ('u8timeseries.timeseries', 'ERROR',
              'ValueError: Timestamp must be between 2000-01-01 00:00:00 and 2000-01-03 00:00:00')
         )
 
     def test_time_log(self):
-        # test time log decorator log message when running theta model
-        ts = constant_timeseries(length=100)
-        model = Theta()
+        logger = get_logger(__name__)
```
pennfranc, line 88: Thanks, this is definitely a better test. But if we don't want the error log to show up when running unittest we should also add `logger.handlers = []`.

u8timeseries/models/rnn_model.py
```diff
@@ -399,7 +490,7 @@ def _save_model(self,
         """
 
         checklist = glob(os.path.join(folder, "checkpoint_*"))
-        checklist = sorted(checklist, key=lambda x: float(re.findall('(\\d+)', x)[-1]))
+        checklist = sorted(checklist, key=lambda x: float(re.findall('(\d+)', x)[-1]))
```
radujica, line 493: think the regex string should be `r'(\d+)'`

u8timeseries/utils/statistics.py
```diff
@@ -124,89 +156,122 @@ def extract_trend_and_seasonality(ts: 'TimeSeries', freq: int = None,
...
-            new_ts = ts / other
-        else:
-            new_ts = ts - other
 
+    if model == 'multiplicative':
+        new_ts = ts / other
+    elif model == 'additive':
+        new_ts = ts - other
+    else:
+        raise_log(ValueError('Invalid parameter; must be either "additive" or "multiplicative". Was: {}'.format(model)))
```
pennfranc, line 186: Might be useful to add logger as argument to raise_log.

radujica:
LGTM though linter will require a bunch of changes 🏳️

u8timeseries/models/standard_regression_model.py
```diff
@@ -0,0 +1,72 @@
...
+        # Get (at most) the last [train_n_points] of each series
+        last_train_ts = train_features[0].end_time()
+        last_n_points_features = [s.slice_n_points_before(last_train_ts, train_n_points) for s in train_features]
+        last_n_points_target = train_target.slice_n_points_before(last_train_ts, train_n_points)
+
+        super().fit(last_n_points_features, last_n_points_target)
+
+        self.model.fit(self._get_features_matrix_from_series(last_n_points_features), last_n_points_target.values())
+
+    def predict(self, features: List[TimeSeries]):
```
pennfranc, line 69: Question: Is there rule as to when we use a new line for every function argument? Because here we stay on the same line, while in the 'fit' function we use a new line for every parameter.

pennfranc:
> LGTM though linter will require a bunch of changes 🏳️

u8timeseries/models/standard_regression_model.py
hrzn: No rule so far, but whenever it's a bit too long, I try to split in several lines (didn't do it everywhere though)

u8timeseries/tests/test_logging.py
hrzn: thank you!

u8timeseries/models/rnn_model.py
```diff
@@ -30,8 +34,55 @@ def _get_runs_folder(work_dir, model_name):
     return os.path.join(work_dir, RUNS_FOLDER, model_name)
 
 
+class _TimeSeriesDataset1D(Dataset):
```
pennfranc, line 37: Any specific reason this was moved here? I guess if this is our only torch-based model (which it currently is) it's a good place. But it might be useful to move this into utils for future torch-based models to have access to. What do you think?

hrzn:
@TheMP I think you can focus mostly on the docstrings, see if the documentation style makes sense for you.

u8timeseries/models/rnn_model.py
hrzn: The main reason is that I think from now on we should consider the submodule `utils` as something "public-facing". The users of u8timeseries are unlikely to care about the Pytorch dataset, as it is mosty meant only to be used internally by the RNN model. I think if we later have more Pytorch models requiring such Datasets, we should maybe factor this one out, but not in the public-facing `utils` submodue. Makes sense for you?

pennfranc: Got it! Maybe we can add it in a separate file in the models module but not include it in `__init__.py`? I guess we still have time to think about it.

hrzn: Yes that could be a good option, however I think we should leave it for a next PR

PR #78, opened 2020-05-11
Add missing dates

endrjuskr:


u8timeseries/timeseries.py
```diff
@@ -44,7 +45,13 @@ def __init__(self,
         raise_if_not(isinstance(series.index, pd.DatetimeIndex), 'Series must be indexed with a DatetimeIndex.', logger)
         raise_if_not(np.issubdtype(series.dtype, np.number), 'Series must contain numerical values.', logger)
 
-        self._series: pd.Series = series.sort_index()  # Sort by time
+        series = series.sort_index()  # Sort by time
+        if fill_missing_dates:
```
endrjuskr, line 49: Maybe user wants to fails if such cases happens so adding a flag

u8timeseries/timeseries.py
```diff
@@ -804,6 +811,13 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
         conf_hi = self._combine_or_none(self._confidence_hi, other.conf_hi_pd_series(), combine_fn)
         return TimeSeries(series, conf_lo, conf_hi)
 
+    def _fill_missing_dates(self, series: pd.Series) -> pd.Series:
+        date_axis = series.axes[0]
```
endrjuskr, line 815: Is it always true for timeseries coming from external source?

u8timeseries/timeseries.py
```diff
@@ -804,6 +811,13 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+        available_frequencies = np.unique(np.diff(date_axis.to_numpy()).astype(int))
```
endrjuskr, line 816: We need use gcd as we have for instance 2 and 3 days frequency.

u8timeseries/timeseries.py
```diff
@@ -804,6 +811,13 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+        gcd_frequency = np.gcd.reduce(available_frequencies)
+        desired_delta = pd.Timedelta(gcd_frequency)
+        return series.resample(desired_delta).mean()
```
endrjuskr, line 819: This is not super intuitive but found such suggestion on SO.

u8timeseries/timeseries.py
```diff
@@ -804,6 +810,17 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
         return TimeSeries(series, conf_lo, conf_hi)
 
+    @staticmethod
+    def _fill_missing_dates(series: pd.Series) -> pd.Series:
+        date_axis = series.index
+        size = 3
+        available_frequencies = [date_axis[x:x + size].inferred_freq for x in range(len(date_axis) - size + 1)]
+        available_frequencies = set(filter(None.__ne__, available_frequencies))
+
+        raise_if_not(len(available_frequencies) == 1, 'Could not infer explicit frequency.', logger)
```
hrzn, line 820: Should this be `len(available_frequencies) >= 1` ?

u8timeseries/timeseries.py
```diff
@@ -804,6 +810,17 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+        suggested_frequency = available_frequencies.pop()
+        return series.resample(suggested_frequency).sum()
```
hrzn, line 822: I played a bit with the current version and it seems that the interpolation is not really what one would expect (with some 0 values inserted). Perhaps here you could follow a 2 steps approach. 1st: fill in the missing dates with `np.nan`, and 2nd: interpolate the missing values, using for instance https://github.com/unit8co/u8timeseries/blob/2139a0f40d032b936c76158b01f6f48656567a0c/u8timeseries/utils/missing_values.py#L41 or `pd.Series.interpolate()` directly.
image

hrzn:
This is a very much needed feature!

u8timeseries/timeseries.py
endrjuskr: If there is more than one for instance 2D and 3D then we cannot decide, I would suggest to fail in such case

u8timeseries/timeseries.py
endrjuskr: I was thinking that we need to fill blanks and lib will automatically interpolate it or leave it to user

pennfranc: I think that's a good idea, maybe leave it to the user if an additional flag is passed? But I believe it would still be good to fill the missing timestamps with NaNs, since otherwise the TimeSeries instance cannot be created in the first place, and the user cannot call the missing_value functions either.

u8timeseries/timeseries.py
```diff
@@ -37,19 +38,26 @@ def __init__(self,
...
 
         # cannot create a timeseries with n<3 -> can add less than 3 elements with add function
         raise_if_not(len(series) >= 3, 'Series must have at least three values.', logger)
         raise_if_not(isinstance(series.index, pd.DatetimeIndex), 'Series must be indexed with a DatetimeIndex.', logger)
         raise_if_not(np.issubdtype(series.dtype, np.number), 'Series must contain numerical values.', logger)
 
-        self._series: pd.Series = series.sort_index()  # Sort by time
-        self._freq: str = self._series.index.inferred_freq  # Infer frequency
+        series = series.sort_index()  # Sort by time
+        if series.index.inferred_freq is None:
```
TheMP, line 51: `if not series.index.inferred_freq and fill_missing_dates`

TheMP, line 51: if not series.index.inferred_freq

u8timeseries/timeseries.py
```diff
@@ -804,10 +815,41 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+            date_axis[x:x + samples_size].inferred_freq
+            for x
+            in range(len(date_axis) - samples_size + 1)]
+
+        observed_frequencies = set(filter(None.__ne__, observed_frequencies))
+
+        raise_if_not(len(observed_frequencies) == 1, 'Could not infer explicit frequency. '
+                                                     'Is Series too short (n=2)?', logger)
+        inferred_frequency = observed_frequencies.pop()
+        return series.resample(inferred_frequency).asfreq(fill_value=None)
```
TheMP, line 846: how is this working, does it set the series.index.inferred_freq?

u8timeseries/timeseries.py
```diff
@@ -804,10 +815,41 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+        series
+            The actual time series, as a pandas Series with a proper time index.
+
+        Returns
+        -------
+        pandas.Series
+            A new Pandas Series without missing dates.
+        """
+        date_axis = series.index
+        samples_size = 3
```
TheMP, line 835: do we have some sort of global variable for that? Would be easier to use it from one place until we start working on https://unit8.atlassian.net/browse/U8TS-66

u8timeseries/timeseries.py
```diff
@@ -804,10 +815,41 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+        observed_frequencies = [
+            date_axis[x:x + samples_size].inferred_freq
```
TheMP, line 837: How does this inferred_freq work internally?

We take a chunk of 3 samples and try to infer frequency, then take another 3 from x+1. Could you make sure this is not O(N^2)?

u8timeseries/timeseries.py
```diff
@@ -804,10 +815,41 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+            for x
+            in range(len(date_axis) - samples_size + 1)]
+
+        observed_frequencies = set(filter(None.__ne__, observed_frequencies))
+
+        raise_if_not(len(observed_frequencies) == 1, 'Could not infer explicit frequency. '
```
TheMP, line 843: Can add a number of detected freq's to log output

u8timeseries/timeseries.py
radujica: Whichever way, would be best to avoid default/automatic behavior that the user is probably unaware of (despite it possibly being written in the docs); if allowing it, log it.

u8timeseries/timeseries.py
pennfranc: @TheMP Actually I'm not sure whether solving the issue you mentioned (which I did as part of the residuals pull request) is going to change much in that regard. As far as I can tell 3 is just the minimum series length to infer the frequency. Since Series instances shorter than 3 will be allowed, I will be sure to throw an exception if the frequency is not given explicitly. 
Sorry if this sounds confusing, once I'm done with my PR it should become clear when looking at the TimeSeries constructor.

u8timeseries/timeseries.py
endrjuskr: it sets `freq` and it for each date according to new frequency it takes one from original series or NaN if it is not present

u8timeseries/timeseries.py
endrjuskr: sure

u8timeseries/timeseries.py
endrjuskr: Looks like linear, because it works on deltas between points, so at the end it is linear.

PR #79, opened 2020-05-11
Fix notebooks

radujica:
Status:
- Air pass <- needs `get_train_val_series and backtest_autoregressive_model` which were deleted
- Daily <- `daily-max-temp.csv` file doesn't exist
- FFT <- works
- RNN <- works
- Intro <- Timeseries does not have method `intersect`

Observations:
- do we really need all notebooks?
  - they seem overlapping
  - air passengers dataset is used in at least 3 notebooks
  - FFT uses energy, temps, and air pass. datasets
- there are a _lot_ of warnings; looks bad

Notes:
- I would really manually look if the examples make sense
- notebooks currently have no output
- if installed papermill as in feature/examples, can run notebook (which will save output to notebook) like:

        papermill x.ipynb x.ipynb


hrzn:
Thanks @radujica for taking a look at this. I need to make a pass over all of them, especially after the last refactoring. Some comments:

* I Agree we probably can keep fewer notebooks. Probably a general "intro" one with few models and how to backtest and combine them would do, plus the FFT and RNNs.
* Let's also remove the warnings in the corresponding sections of the code.
* "Air pass <- needs get_train_val_series and backtest_autoregressive_model which were deleted": this corresponds to the "old way" of doing backtesting, and this section of the notebook can be removed (I will commit a change sometime soon)

PR #80, opened 2020-05-12
Feature/residuals

pennfranc:
- Implemented function computing residuals given a ForecastingModel and a TimeSeries instance.
- Removed the constraint that a TimeSeries instance needs at least three elements. Instead a new get_min_train_series_length method was introduced for ForecastingModel, which can be overridden in the (and already is for some) concrete models.
- Fixed FFT problems that arose when calling fit multiple times, updated docstrings to new format.


u8timeseries/models/forecasting_model.py
```diff
@@ -63,6 +66,9 @@ def predict(self, n: int) -> TimeSeries:
         if (not self._fit_called):
             raise_log(Exception('fit() must be called before predict()'), logger)
 
+    def get_min_train_series_length(self) -> int:
```
radujica, line 69: How about using a `@property`? `base_class.min_series_length` and `sub_class.min_series_length` seems a better approach. It makes it clearer what it represents, i.e. a property of ForecastingModel's

radujica: And if it needs to be specific to each subclass, could raise NotImplementedError in base

pennfranc: I like the property decorator idea, I will implement that. In regards to your second comment, I thought it might be useful to have a default value, so that models that do not have a specific requirement can just inherit the default.

u8timeseries/backtesting/backtesting.py
```diff
@@ -11,6 +11,10 @@
 
 from u8timeseries.utils import _build_tqdm_iterator
 from ..logging import raise_if_not, get_logger
+from ..utils.statistics import plot_acf
+import matplotlib.pyplot as plt
```
endrjuskr, line 15: Let's move external libs to the top

endrjuskr: In general how about python lib then external lib and at the end our modules

u8timeseries/backtesting/backtesting.py
```diff
@@ -166,3 +170,91 @@ def backtest_regression(feature_series: Iterable[TimeSeries],
...
+    p = backtest_forecasting(series, model, first_index, fcast_horizon_n, True, verbose=verbose)
+
+    # compute residuals
+    series_trimmed = series.slice_intersect(p)
+    residuals = series_trimmed - p
+
+    return residuals
+
+
+def analyze_residuals(residuals: TimeSeries, num_bins: int = 20):
```
endrjuskr, line 220: maybe mention `plot` in the name of method

u8timeseries/models/arima.py
```diff
@@ -54,6 +54,10 @@ def predict(self, n):
         forecast = self.model.forecast(steps=n)[0]
         return self._build_forecast_series(forecast)
 
+    @property
+    def min_train_series_length(self) -> int:
+        return 30
```
endrjuskr, line 59: Why is that? Can we put docs here?

u8timeseries/models/fft.py
```diff
@@ -1,29 +1,39 @@
 from .forecasting_model import ForecastingModel
 from ..timeseries import TimeSeries
-from ..logging import time_log, get_logger, raise_log
+from ..logging import get_logger
 import numpy as np
```
endrjuskr, line 4: Please rearrange

u8timeseries/models/fft.py
```diff
@@ -49,14 +59,22 @@ def _check_approximate_seasonality(series: TimeSeries, seasonality_period: int,
     return order <= max_seasonality_order
 
 
-def _find_relevant_timestamp_attributes(series: TimeSeries):
-    """
+def _find_relevant_timestamp_attributes(series: TimeSeries) -> set:
+    """ Finds pd.Timestamp attributes relevant for seasonality.
```
endrjuskr, line 63: Should it be in new line?

u8timeseries/backtesting/__init__.py
```diff
@@ -3,4 +3,4 @@
 -----------
 """
 
-from .backtesting import backtest_forecasting, backtest_regression
+from .backtesting import backtest_forecasting, backtest_regression, forecasting_residuals, analyze_residuals
```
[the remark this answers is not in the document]
pennfranc: If we did that couldn't it be confused with the models themselves?

u8timeseries/models/fft.py
pennfranc: I probably misunderstood, but my thinking was that this first line (before any line break) will be used as title in the docs. So I assume that's wrong?

u8timeseries/backtesting/backtesting.py
pennfranc: done

u8timeseries/backtesting/backtesting.py
pennfranc: done

u8timeseries/models/arima.py
pennfranc: done

u8timeseries/models/fft.py
pennfranc: done

u8timeseries/models/arima.py
TheMP: should it be 30 or 3?

pennfranc: I made it 30 on purpose since ARIMA requires a significantly longer minimum length in most cases. The number is still arbitrary though.

u8timeseries/backtesting/backtesting.py
```diff
@@ -64,7 +68,7 @@ def backtest_forecasting(series: TimeSeries,
     raise_if_not(start != series.end_time(), 'The provided start timestamp is the last timestamp of the time series',
                  logger)
 
-    last_pred_time = series.time_index()[-fcast_horizon_n - 2] if trim_to_series else series.time_index()[-2]
+    last_pred_time = series.time_index()[-fcast_horizon_n - 1] if trim_to_series else series.time_index()[-1]
```
pennfranc, line 71: @hrzn This change (from -2 to -1) should be ok, right?

hrzn: Yes, I think that makes sense, I don't remember why I put -2. 🤔

hrzn: Maybe you can add a `raise_if_not` check that `fcast_horizon_n >= 1`

u8timeseries/backtesting/backtesting.py
```diff
@@ -166,3 +170,91 @@ def backtest_regression(feature_series: Iterable[TimeSeries],
...
 
     return TimeSeries.from_times_and_values(pd.DatetimeIndex(times), np.array(values))
+
+
+def forecasting_residuals(model: ForecastingModel, series: TimeSeries, fcast_horizon_n: int = 1,
+                          verbose: bool = True) -> TimeSeries:
+    """ A function for computing the residuals produced by a given model and time series.
+
+    This function computes the difference between the actual observations from 'series'
+    and the fitted values vector p obtained by training 'model' on 'series'.
```
hrzn, line 180: I think to get good results on Sphinx you should use these ticks: `, instead of '

u8timeseries/backtesting/backtesting.py
pennfranc: done

u8timeseries/backtesting/backtesting.py
pennfranc: Done for that file. I created a jira taks to go over all files to update all ticks.

u8timeseries/models/forecasting_model.py
```diff
@@ -63,6 +66,14 @@ def predict(self, n: int) -> TimeSeries:
         if (not self._fit_called):
             raise_log(Exception('fit() must be called before predict()'), logger)
 
+    @property
+    def min_train_series_length(self) -> int:
```
hrzn, line 70: very cool

u8timeseries/timeseries.py
```diff
@@ -38,26 +38,33 @@ def __init__(self,
...
-        self._freq: str = self._series.index.inferred_freq  # Infer frequency
+        if (len(series) < 3):
+            self._freq: str = freq
+        else:
+            if not series.index.inferred_freq:
+                if fill_missing_dates:
+                    self._series = self._fill_missing_dates(self._series)
+                else:
+                    raise_if_not(False, 'Could not infer frequency. Are some dates missing? '
+                                        'Is Series too short (n=2)?', logger)
```
hrzn, line 64: You can remove the `Is Series too short` part from this message. You can maybe tell the user something like:
```
Could not infer frequency. Are some dates missing? Try specifying a `freq` when building the `TimeSeries`, or specify `fill_missing_dates=True`.
```

u8timeseries/timeseries.py
```diff
@@ -504,7 +511,8 @@ def from_dataframe(df: pd.DataFrame,
     def from_times_and_values(times: pd.DatetimeIndex,
                               values: np.ndarray,
                               confidence_lo: np.ndarray = None,
-                              confidence_hi: np.ndarray = None) -> 'TimeSeries':
+                              confidence_hi: np.ndarray = None,
+                              freq: Optional[str] = None) -> 'TimeSeries':
```
hrzn, line 515: How about the `fill_missing_dates` argument?

u8timeseries/timeseries.py
pennfranc: How about just `Could not infer frequency. Are some dates missing? Try specifying `fill_missing_dates=True`.`? Because the way it is implemented right now, if a series has a length of at least 3, the frequency will always be set by calling infer_freq.

u8timeseries/timeseries.py

PR #84, opened 2020-05-14
Upsample timeseries

endrjuskr:


u8timeseries/utils/timeseries_generation.py
```diff
@@ -241,3 +242,44 @@ def holiday_timeseries(country_code: str,
     values = times.isin(scoped_country_holidays).astype(int)
 
     return TimeSeries.from_times_and_values(times, values)
+
+
+def upsample_timeseries(ts: TimeSeries, freq: str, method: str = 'pad') -> TimeSeries:
```
pennfranc, line 247: Cool function! But maybe we could place it somewhere else? Because all other functions in timeseries_generation.py generate a new time series from scratch, so I'm not sure this fits. How about having this function as a method of TimeSeries (possibly static)?

endrjuskr: Good point

u8timeseries/timeseries.py
```diff
@@ -743,6 +743,48 @@ def update(self,
                 new_hi.update(conf_hi)
             return TimeSeries(new_series, new_lo, new_hi)
 
+    def upsample(self, freq: str, method: str = 'pad') -> 'TimeSeries':
```
hrzn, line 746: what happens if this is called with a lower frequency than the original series? Will this downsample?

endrjuskr: It will throw an exception

u8timeseries/timeseries.py
```diff
@@ -743,6 +743,48 @@ def update(self,
...
+        """
+
+        new_offset = to_offset(freq)
+        old_offset = self.freq()
+
+        raise_if_not(
+            new_offset < old_offset,
+            "Frequency needs to higher than current one - "
+            + self.freq_str() + ". "
+            + "Did you try to downsample?")
```
TheMP, line 775: Maybe we actually also need downsampling ;]?

u8timeseries/timeseries.py
hrzn: oh yes sorry missed it :) but how about also supporting downsampling, and renaming this one to `resample()` ?

u8timeseries/tests/test_timeseries.py
```diff
@@ -428,3 +428,26 @@ def test_fill_missing_dates(self):
...
         self.assertTrue(math.isnan(series_test.pd_series().get('20130105')))
+
+    def test_upsample_timeseries(self):
+        times = pd.date_range('20130101', '20130110')
+        pd_series = pd.Series(range(10), index=times)
+        timeseries = TimeSeries(pd_series)
+
+        resampled_timeseries = timeseries.upsample('H')
+        self.assertEqual(resampled_timeseries.freq_str(), 'H')
+        self.assertEqual(resampled_timeseries.pd_series().at[pd.Timestamp('20130101020000')], 0)
```
hrzn, line 439: You should be able to index directly into the `TimeSeries`. Something like `resampled_timeseries[pd.Timestamp('20130101020000')]`

u8timeseries/tests/test_timeseries.py
```diff
@@ -428,3 +428,26 @@ def test_fill_missing_dates(self):
...
+        self.assertEqual(resampled_timeseries.pd_series().at[pd.Timestamp('20130109090000')], 8)
+
+        with self.assertRaises(ValueError):
+            # try to downsample
+            timeseries.upsample('2D')
+
+        confidence_hi_series = pd.Series(range(5, 15), index=times)
+        timeseries = TimeSeries(pd_series, confidence_hi=confidence_hi_series)
+        resampled_timeseries = timeseries.upsample('H')
+        self.assertEqual(resampled_timeseries.conf_hi_pd_series().index.inferred_freq, 'H')
```
hrzn, line 450: you should be able to call `resampled_timeseres.freq_str()`

PR #85, opened 2020-05-15
Feature/TCN

pennfranc:
- Created base class for torch-based forecasting models called 'TorchForecastingModel'.
- RNNModel now builds on top of base class.
- New forecasting model: TCNModel implemented, also inherits TorchForecastingModel
- Preliminary example notebook for TCN: https://github.com/unit8co/u8timeseries/blob/feature/TCN/examples/TCN-examples.ipynb


u8timeseries/models/torch_forecasting_model.py
```diff
@@ -0,0 +1,505 @@
+"""
+Recurrent Neural Networks
```
hrzn, line 2: Torch Forecasting Model

hrzn:
Could you add unit tests for TCN?

u8timeseries/models/tcn_model.py
```diff
@@ -0,0 +1,254 @@
...
+
+        super().__init__(**kwargs)
+
+    def create_dataset(self, series):
+        return _TimeSeriesDataset1DShifted(series, self.input_length, self.output_length)
+
+    def predict(self, n: int) -> TimeSeries:
+        super().predict(n)
+
+        scaled_series = self.training_series.values()[-self.seq_len:]
```
hrzn, line 243: The name `scaled_series` was there in RNN for historical reasons (as the model was previously also scaling the series). We should use a better name like `input_sequence` or something

u8timeseries/models/torch_forecasting_model.py
```diff
@@ -0,0 +1,505 @@
...
+
+        if torch_device_str is None:
+            self.device = self._get_best_torch_device()
+        else:
+            self.device = torch.device(torch_device_str)
+
+        self.input_size = 1  # We support only univariate time series currently
+        self.input_length = input_length
+        self.output_length = output_length
+        self.seq_len = input_length
```
hrzn, line 208: Could we remove this one and keep only `self.input_length` ?

u8timeseries/models/tcn_model.py
```diff
@@ -0,0 +1,254 @@
...
+                                kernel_size=kernel_size, num_filters=num_filters,
+                                num_layers=num_layers, dilation_base=dilation_base,
+                                output_length=output_length, dropout=dropout, weight_norm=weight_norm)
+
+        super().__init__(**kwargs)
+
+    def create_dataset(self, series):
+        return _TimeSeriesDataset1DShifted(series, self.input_length, self.output_length)
+
+    def predict(self, n: int) -> TimeSeries:
```
hrzn, line 240: Do you need this implementation? It looks like `TorchForecastingModel` is already doing this.

u8timeseries/models/tcn_model.py
```diff
@@ -0,0 +1,254 @@
...
+        self.input_length = input_length
+        self.n_filters = num_filters
+        self.kernel_size = kernel_size
+        self.out_len = output_length
+        self.dilation_base = dilation_base
+        self.dropout = nn.Dropout(p=dropout)
+
+        # If num_layers is not passed, compute number of layers needed for full history coverage
+        if (num_layers is None and dilation_base > 1):
+            num_layers = math.ceil(math.log((input_length - 1) / (kernel_size - 1), dilation_base))
```
hrzn, line 158: that's cool :)

u8timeseries/models/tcn_model.py
```diff
@@ -0,0 +1,254 @@
...
+        self.input_size = input_size
+        self.input_length = input_length
+        self.n_filters = num_filters
+        self.kernel_size = kernel_size
+        self.out_len = output_length
+        self.dilation_base = dilation_base
+        self.dropout = nn.Dropout(p=dropout)
+
+        # If num_layers is not passed, compute number of layers needed for full history coverage
+        if (num_layers is None and dilation_base > 1):
```
hrzn, line 157: parentheses not needed (here and below)

u8timeseries/models/tcn_model.py
```diff
@@ -0,0 +1,254 @@
...
+        elif (num_layers is None):
+            num_layers = (input_length - kernel_size) / 2 + 1
+
+        # Building TCN module
+        self.res_blocks_list = []
+        for i in range(num_layers):
+            res_block = _ResidualBlock(num_filters, kernel_size, dilation_base,
+                                       self.dropout, weight_norm, i, num_layers)
+            self.res_blocks_list.append(res_block)
+        self.res_blocks = nn.ModuleList(self.res_blocks_list)
```
hrzn, line 168: Is this needed?

u8timeseries/models/tcn_model.py
```diff
@@ -0,0 +1,254 @@
...
+            The number of filters in a convolutional layer of the TCN.
+        kernel_size
+            The size of every kernel in a convolutional layer.
+        dilation_base
+            The base of the exponent that will determine the dilation on every level.
+        dropout
+            The dropout rate for every convolutional layer.
+        weight_norm
+            Boolean value indicating whether to use weight normalization.
+        i
```
hrzn, line 37: could we find maybe a better name? For instance `nr_blocks_below`

u8timeseries/models/tcn_model.py
pennfranc: I changed it slightly when overriding it, since by default the TorchForecastingModel implementation uses the first point of the predicted series as the next point in time. This would not work for the TCN because it predicts a shifted sequence.

u8timeseries/models/tcn_model.py
pennfranc: If I don't add this line the optimizer will not find the parameters, I get a 'optimizer got an empty parameter list' error.

u8timeseries/models/torch_forecasting_model.py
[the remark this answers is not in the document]
pennfranc: done

u8timeseries/models/torch_forecasting_model.py
```diff
@@ -0,0 +1,505 @@
...
+class _TimeSeriesDataset1DSequential(Dataset):
+
+    def __init__(self,
+                 series: TimeSeries,
+                 data_length: int = 1,
+                 target_length: int = 1):
+        """
+        A PyTorch Dataset from a univariate TimeSeries.
+        The Dataset iterates a moving window over the time series. The resulting slices contain `(data, target)`,
+        where `data` is a 1-D sub-sequence of length [data_length] and target is the 1-D sub-sequence of length
```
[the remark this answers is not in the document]
pennfranc: done

u8timeseries/models/torch_forecasting_model.py
```diff
@@ -0,0 +1,505 @@
...
+        data = self.series_values[idx:idx + self.data_length]
+        target = self.series_values[idx + self.data_length:idx + self.data_length + self.target_length]
+        return torch.from_numpy(data).float().unsqueeze(1), torch.from_numpy(target).float().unsqueeze(1)
+
+
+class _TimeSeriesDataset1DShifted(torch.utils.data.Dataset):
+
+    def __init__(self,
+                 series: TimeSeries,
+                 data_length: int = 3,
```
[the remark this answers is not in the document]
pennfranc: done

u8timeseries/models/tcn_model.py
pennfranc: done

u8timeseries/models/torch_forecasting_model.py
pennfranc: done

u8timeseries/models/tcn_model.py
pennfranc: done

u8timeseries/models/torch_forecasting_model.py
```diff
@@ -0,0 +1,505 @@
...
+                 loss_fn: nn.modules.loss._Loss = nn.MSELoss(),
+                 model_name: str = "torch_model_run",  # TODO uid
+                 work_dir: str = os.getcwd(),
+                 log_tensorboard: bool = False,
+                 nr_epochs_val_period: int = 10,
+                 torch_device_str: Optional[str] = None):
+        """ Pytorch-based Forecasting Model.
+
+        This class is meant to be inherited to create a new pytorch-based forecasting module.
+        When subclassing this class, please make sure to set the self.model attribute
```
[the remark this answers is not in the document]
pennfranc: The reason for why I opted to create the pytorch module in the __init__ function instead of an abstract method is because the model-specific parameters would all have to be passed to that method. So for instance, for the TCN the variable `dilation`would be an argument to the init function, which would then have to be forwarded in some way to the invocation of `model`, which would probably occur somewhere in the TorchForecastingModel superclass. I feel like this would get messy, no?

u8timeseries/models/tcn_model.py
pennfranc: done

u8timeseries/models/torch_forecasting_model.py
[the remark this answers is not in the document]
hrzn: I think you could maybe keep the creation of the pytorch module in init (as of now), but just implement an abstract method in all the sub-classes:
```
get_model():
    return self.model
```
By making it abstract you force the children classes to implement it (and you don't rely on the next person reading the comment to do it right). The when using the model you can call `self.get_model()` instead of `self.model`.
Does it make sense?

u8timeseries/models/tcn_model.py
hrzn: Oh! I think there should be a neater way to specify the parameters to the optimizer directly with pytorch

pennfranc: Ok I'll look into it!

u8timeseries/models/tcn_model.py
hrzn: Is there a way to slightly generalize the `TorchForecastingModel` implementation have it only once?

u8timeseries/models/torch_forecasting_model.py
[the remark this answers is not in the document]
pennfranc: I see how this would be cleaner in terms of subclass requirements, but the problem with this that I see is that the model hyperparameters passed to the `__init__`function of TorchForecastingModel will be required by the `get_model()` function. And since different subclasses can have different hyperparameters, wouldn't the `get_model()`function signature be inconsistent?

u8timeseries/models/tcn_model.py
pennfranc: I added a new integer property to `TorchForecastingModel`called `first_prediction_index`. Predict now accesses this property, so it suffices to override `first_prediction_index`.

u8timeseries/models/torch_forecasting_model.py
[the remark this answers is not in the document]
hrzn: OK, fair enough, let's leave it like that then.

PR #86, opened 2020-05-18
Feature/gridsearch

pennfranc:
- Implemented a function called 'backtest_gridsearch' that takes a subclass of ForecastingModel, a TimeSeries instance, a dictionary containing lists of possible hyperparameter values and returns the model that performs the best according to the 'metric' argument. The performance is evaluated by using 'backtest_forecasting' as a subroutine.
- Included a basic unit test for backtest_gridsearch

### DIFF
--- a/u8timeseries/backtesting/backtesting.py
+++ b/u8timeseries/backtesting/backtesting.py
@@ -5,9 +5,11 @@ Backtesting Functions
 
 import pandas as pd
 import numpy as np
+import matplotlib.pyplot as plt
 from ..timeseries import TimeSeries
 from ..models.forecasting_model import ForecastingModel
 from ..models.regression_model import RegressionModel
+from ..models import AutoARIMA, ExponentialSmoothing, FFT, Prophet, Theta
 from .. import metrics
 from ..utils import _build_tqdm_iterator
 from ..logging import raise_if_not, get_logger
@@ -246,3 +248,38 @@ def backtest_gridsearch(model_class: type, parameters: dict,
             best_param_combination = param_combination_dict
     logger.info('Chosen parameters: ' + str(best_param_combination))
     return model_class(**best_param_combination)
+
+
+def explore_models(train_series: TimeSeries, val_series: TimeSeries):
+
+    model_parameter_tuples = [
+        (AutoARIMA, {}),
+        (FFT, {
+            'nr_freqs_to_keep': [5, 10, 25, 50, 100],
+            'trend': [None, 'poly', 'exp']
+        }),
+        (Theta, {
+            'theta': list(range(3, 10))
+        }),
+        (Prophet, {})
+    ]
+
+    for model_class, params in model_parameter_tuples:
+
+        if (len(params.keys()) > 0):
+            model = backtest_gridsearch(model_class, params, train_series, val_series)
+        else:
+            model = model_class()
+
+        # fit and predict
+        model.fit(train_series)
+        predictions = model.predict(len(val_series))
+
+        # visualize
+        print(str(model), ", MAPE:", metrics.mape(predictions, val_series))
+        train_series.plot(label='train')
+        val_series.plot(label='val')
+        predictions.plot(label='pred')
+        plt.legend()
+
+        plt.show()
--- a/u8timeseries/backtesting/backtesting.py
+++ b/u8timeseries/backtesting/backtesting.py
@@ -265,8 +265,8 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
 def backtest_gridsearch(model_class: type,
                         parameters: dict,
                         train_series: TimeSeries,
+                        fcast_horizon_n: Optional[int] = None,
                         val_series: Optional[TimeSeries] = None,
-                        fcast_horizon_n: int = 5,
                         num_predictions: int = 10,
                         metric: Callable[[TimeSeries, TimeSeries], float] = metrics.mape,
                         verbose=False):
@@ -275,15 +275,15 @@ def backtest_gridsearch(model_class: type,
     This function has 2 modes of operation: Expanding window mode and split mode.
     Both modes of operation evaluate every possible combination of hyperparameter values
     provided in the `parameters` dictionary by instantiating the `model_class` subclass
-    of ForecastingModel with each combination and returning the best-performing model with regards
+    of ForecastingModel with each combination, and returning the best-performing model with regards
     to the `metric` function. The `metric` function is expected to return an error value,
     thus the model resulting in the smallest `metric` output will be chosen.
     The relationship of the training data and test data depends on the mode of operation.
 
-    Expanding window mode (default):
-    For every model, `num_predictions` predictions with horizon `fcast_horizon_n` are computed
-    by using the `backtest_forecast` function as a subroutine on `train_series`. This means that
-    the model is both trained and evaluated on `train_series`.
+    Expanding window mode (activated when `fcast_horizon_n` is passed):
+    For every hyperparameter combination, the model is repeatedly trained and evaluated on different
+    splits of `train_series`. The number of splits is equal to `num_predictions`, and the
+    forecasting horizon used when making a prediction is `fcast_horizon_n`.
     Note that the model is retrained for every single prediction, thus this mode is slower.
 
     Split window mode (activated when `val_series` is passed):
@@ -315,10 +315,14 @@ def backtest_gridsearch(model_class: type,
     Returns
     -------
     ForecastingModel
-        A 'model_class' instance with the best-performing hyperparameters from the given selection.
+        An untrained 'model_class' instance with the best-performing hyperparameters from the given selection.
     """
 
-    backtest_start_time = train_series.end_time() - (num_predictions + fcast_horizon_n) * train_series.freq()
+    raise_if_not((fcast_horizon_n is None) ^ (val_series is None),
+                 "Please pass exactly one of the arguments 'forecast_horizon_n' or 'val_series'.", logger)
+
+    if val_series is None:
+        backtest_start_time = train_series.end_time() - (num_predictions + fcast_horizon_n) * train_series.freq()
     min_error = float('inf')
     best_param_combination = {}
 
@@ -330,13 +334,13 @@ def backtest_gridsearch(model_class: type,
     for param_combination in iterator:
         param_combination_dict = dict(list(zip(parameters.keys(), param_combination)))
         model = model_class(**param_combination_dict)
-        if (val_series is None):  # expanding window mode
+        if val_series is None:  # expanding window mode
             backtest_forecast = backtest_forecasting(train_series, model, backtest_start_time, fcast_horizon_n)
             error = metric(backtest_forecast, train_series)
         else:  # split mode
             model.fit(train_series)
             error = metric(model.predict(len(val_series)), val_series)
-        if (error < min_error):
+        if error < min_error:
             min_error = error
             best_param_combination = param_combination_dict
     logger.info('Chosen parameters: ' + str(best_param_combination))
@@ -346,8 +350,9 @@ def backtest_gridsearch(model_class: type,
 def explore_models(train_series: TimeSeries,
                    val_series: TimeSeries,
                    test_series: TimeSeries,
-                   width: int = 3,
                    metric: Callable[[TimeSeries, TimeSeries], float] = metrics.mape,
+                   model_parameter_tuples: Optional[list] = None,
+                   plot_width: int = 3,
                    verbose: bool = True):
     """ A function for exploring the suitability of multiple models on a given train/validation/test split.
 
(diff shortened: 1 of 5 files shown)


u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +252,117 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+            model = backtest_gridsearch(model_class, params, train_series, val_series)
+        else:
+            model = model_class()
+
+        # fit and predict
+        model.fit(train_series)
+        predictions = model.predict(len(val_series))
+
+        # visualize
+        print(str(model), ", MAPE:", metrics.mape(predictions, val_series))
```
hrzn, line 362: I think the metric function could be an (optional) parameter of the function `explore_models()`

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +252,117 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+            'nr_freqs_to_keep': [5, 10, 25, 50, 100],
+            'trend': [None, 'poly', 'exp']
+        }),
+        (Theta, {
+            'theta': list(range(3, 10))
+        }),
+        (Prophet, {})
+    ]
+
+    for model_class, params in model_parameter_tuples:
```
hrzn, line 350: how about keeping track of the best performing model (according to user-defined metric) here, and printing a summary at the end, sorting the models from best to worse?

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +252,117 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
     ax3.set_ylabel('ACF value')
     ax3.set_xlabel('lag')
     ax3.set_title('ACF')
+
+
+def backtest_gridsearch(model_class: type, parameters: dict,
```
hrzn, line 257: `type` --> `ForecastingModel` ? and new line after before `parameters`

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +252,117 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
     ax3.set_ylabel('ACF value')
     ax3.set_xlabel('lag')
     ax3.set_title('ACF')
+
+
+def backtest_gridsearch(model_class: type, parameters: dict, 
+                        train_series: TimeSeries, 
+                        val_series: Optional[TimeSeries] = None,  # only used by split mode
+                        fcast_horizon_n: int = 5, num_predictions: int = 10,  # only used by expanding window mode
```
hrzn, line 260: I think all the parameters should be on a separate line

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +252,117 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+                        metric='mape', verbose=False):
```
hrzn, line 261: for `metric` you could consider directly passing a function (Callable) that takes 2 TimeSeries in argument. This way it can also be a user-made one

pennfranc: done

u8timeseries/backtesting/backtesting.py
pennfranc: done

u8timeseries/backtesting/backtesting.py
pennfranc: done

u8timeseries/backtesting/backtesting.py
pennfranc: I saw this comment a bit late, but I included something similar, namely a bar chart comparing the metrics of the different model types. Do you think that works too?

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+    provided in the `parameters` dictionary by instantiating the `model_class` subclass
+    of ForecastingModel with each combination and returning the best-performing model with regards
+    to the `metric` function. The `metric` function is expected to return an error value,
+    thus the model resulting in the smallest `metric` output will be chosen.
+    The relationship of the training data and test data depends on the mode of operation.
+
+    Expanding window mode (default):
+    For every model, `num_predictions` predictions with horizon `fcast_horizon_n` are computed
+    by using the `backtest_forecast` function as a subroutine on `train_series`. This means that
+    the model is both trained and evaluated on `train_series`.
```
hrzn, line 286: I understand what you mean, but would phrase it differently:
```
This means that the model will be repeatedly trained and evaluated on different splits of `train_series`.
```

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+        The number of train/prediction cycles performed in one iteration of expanding window mode.
+    metric:
+        A function that takes two TimeSeries instances as inputs and returns a float error value.
+    verbose:
+        Whether to print progress.
+
+    Returns
+    -------
+    ForecastingModel
+        A 'model_class' instance with the best-performing hyperparameters from the given selection.
```
hrzn, line 318: Maybe mention that this instance is not fit (only created)

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+
+    # compute all hyperparameter combinations from selection
+    params_cross_product = list(product(*parameters.values()))
+
+    # iterate through all combinations of the provided parameters and choose the best one
+    iterator = _build_tqdm_iterator(params_cross_product, verbose)
+    for param_combination in iterator:
+        param_combination_dict = dict(list(zip(parameters.keys(), param_combination)))
+        model = model_class(**param_combination_dict)
+        if (val_series is None):  # expanding window mode
```
hrzn, line 333: no need for the parentheses

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
     ax3.set_ylabel('ACF value')
     ax3.set_xlabel('lag')
     ax3.set_title('ACF')
+
+
+def backtest_gridsearch(model_class: type,
+                        parameters: dict,
+                        train_series: TimeSeries,
+                        val_series: Optional[TimeSeries] = None,
+                        fcast_horizon_n: int = 5,
```
hrzn, line 269: I think we should maybe force the user to select a forecast horizon (i.e. not specify a default value)

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+    for param_combination in iterator:
+        param_combination_dict = dict(list(zip(parameters.keys(), param_combination)))
+        model = model_class(**param_combination_dict)
+        if (val_series is None):  # expanding window mode
+            backtest_forecast = backtest_forecasting(train_series, model, backtest_start_time, fcast_horizon_n)
+            error = metric(backtest_forecast, train_series)
+        else:  # split mode
+            model.fit(train_series)
+            error = metric(model.predict(len(val_series)), val_series)
+        if (error < min_error):
```
hrzn, line 339: no need for parentheses

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+    metric:
+        A function that takes two TimeSeries instances as inputs and returns a float error value.
+    verbose:
+        Whether to print progress.
+    """
+
+    raise_if_not(width > 1, "Please choose an integer 'width' value larger than 1", logger)
+
+    # list of tuples containing model classes and hyperparameter selection, if required
+    model_parameter_tuples = [
```
hrzn, line 384: How about having this as a default, but let users specify their own `model_parameters_tuples` if they want to?

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+            min_error = error
+            best_param_combination = param_combination_dict
+    logger.info('Chosen parameters: ' + str(best_param_combination))
+    return model_class(**best_param_combination)
+
+
+def explore_models(train_series: TimeSeries,
+                   val_series: TimeSeries,
+                   test_series: TimeSeries,
+                   width: int = 3,
```
hrzn, line 349: I would move this parameter after `metric`, since it is more accessory and `metric` is really core to the logic. Also, I would maybe rename it `plot_width` or something like this, to make it clear it's only something for the plots.

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+    model_parameter_tuples = [
+        (NaiveSeasonal, {
+            'K': list(range(1, 31))
+        }),
+        (FFT, {
+            'nr_freqs_to_keep': [5, 10, 25, 50, 100, 150, 200],
+            'trend': [None, 'poly', 'exp']
+        }),
+        (Theta, {
+            'theta': [0, 1] + list(range(3, 50))
```
hrzn, line 393: maybe make it go from -10 to 10 in 50 steps?

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+        (AutoARIMA, {})
+    ]
+
+    # set up plot grid
+    height = math.ceil(len(model_parameter_tuples) / width) + 1
+    fig = plt.figure(constrained_layout=True, figsize=(width * 4, height * 4))
+    gs = fig.add_gridspec(height, width)
+
+    # arrays to store data for bar charts
+    mape_vals = []
```
hrzn, line 405: `metric_vals` ?

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+        end_time = time.time()
+        predictions = model.predict(len(test_series))
+
+        # plot predictions against observations
+        y = int(i / width)
+        x = i - y * width
+        ax = fig.add_subplot(gs[y, x])
+        mape_val = metrics.mape(predictions, test_series)
+        ax.set_title(str(model) + "\n" + metric.__name__ + ": " + str(round(mape_val, 2)))
+        train_and_val_series[-len(test_series):].plot(label='train and val', ax=ax)
```
hrzn, line 432: Is it correct to take this indexing `[-len(test_series):]` ?

u8timeseries/backtesting/backtesting.py
```diff
@@ -259,3 +260,193 @@ def plot_residuals_analysis(residuals: TimeSeries, num_bins: int = 20):
...
+        # record data for bar charts
+        mape_vals.append(mape_val)
+        times.append(end_time - start_time)
+        index.append(str(model)[:3] + "..")
+
+    # plot bar charts
+    ax_mapes = fig.add_subplot(gs[-1, 0])
+    ax_mapes.set_title("Performance")
+    ax_mapes.set_ylabel(metric.__name__ + ' value')
+    pd.Series(mape_vals, index=index).plot.bar(ax=ax_mapes, rot=0)
```
hrzn, line 446: does this way of plotting print the actual numeric values? I think it'd be nice for users to be able to read the actual numbers (not just see them) when comparing metrics and execution time.

u8timeseries/backtesting/backtesting.py
pennfranc: I did this so that half of the graph (in terms of the x axis) would be reserved for the prediction, so that it is clearly visible. Or do you mean to ask whether this form of indexing even works on TimeSeries objects? If that's the case then yes, it seems to be working perfectly.

hrzn: No I meant to ask if it was really your intent to plot only the `len(test_series)` last points. I think it might be weird, especially for tiny test series. I would find it more intuitive to plot the whole thing here (and let the users zoom in if needed).

u8timeseries/backtesting/backtesting.py
pennfranc: There does not seem to be a built-in function like that for bar plots. Currently I am printing the values of the metric and the runtime in the plot title of the respective model. Do you think that's ok?

u8timeseries/backtesting/backtesting.py
pennfranc: done

PR #91, opened 2020-06-03
fix a typo in the doc

guillaumeraille:


hrzn:
Nice :) could you re-open the PR into develop though?

PR #92, opened 2020-06-05
Feature/multivariate - step 1

pennfranc:
- Changed TimeSeries underlying representation from pandas.Series to pandas.DataFrame
- Adapted TimeSeries methods to work with DataFrame
- Added new TimeSeries getter function to interface with models that do not support the multivariate case
- Changed models/utilities to call the above mentioned univariate getter functions so that they are guaranteed to use the same univariate data as before
- Removed confidence interval functionality from TimeSeries class
- Removed unit tests which involved testing of confidence intervals
- Adapted other unit tests where necessary


u8timeseries/timeseries.py
```diff
@@ -20,35 +19,34 @@
 
 class TimeSeries:
     def __init__(self,
-                 series: pd.Series,
-                 confidence_lo: Optional[pd.Series] = None,
-                 confidence_hi: Optional[pd.Series] = None,
+                 series: Union[pd.Series, pd.DataFrame],
```
TheMP, line 22: Do we want to allow for both or maybe it would be easier just to assume DataFrame, and Series would be just a dataframe for univariate case (ie "1 row")?

u8timeseries/timeseries.py
```diff
@@ -20,35 +19,34 @@
...
+            The actual time series, as a pandas Series or DataFrame with a proper time index.
         freq
             Optionally, a string representing the frequency of the Pandas Series. When creating a TimeSeries
             instance with a length smaller than 3, this argument must be passed.
         fill_missing_dates
             Optionally, a boolean value indicating whether to fill missing dates with NaN values
             in case the frequency of `series` cannot be inferred.
         """
+        # create dataframes
+        if (isinstance(series, pd.Series)):
```
TheMP, line 42: doing just DataFrame representation would help to avoid those considerations ;]

u8timeseries/timeseries.py
```diff
@@ -678,92 +621,59 @@ def append_values(self,
...
 
-        raise_if_not(not (values is None and conf_lo is None and conf_hi is None),
-                     "At least one parameter must be filled other than index", logger)
+        if isinstance(values, (list, range)):
+            values = np.array(values)
+
+        if values is not None and len(values.shape) == 1:
+            values = values.reshape((len(values), 1))
+
+        raise_if_not(not (values is None), "'values' parameter should not be None.", logger)
```
TheMP, line 656: maybe time for raise_if that would be basically implemented as not (raise_if_not)

u8timeseries/timeseries.py
pennfranc: Yeah I thought about that too. I kept supporting it for compatibility with existing code. Thoughts on this @hrzn?

hrzn: I think we could keep only `DataFrame` here, but perhaps have a static method `from_series` that would allow to build a univariate one from a `Series`.

u8timeseries/timeseries.py
```diff
@@ -20,35 +19,34 @@
...
     def __init__(self,
-                 series: pd.Series,
-                 confidence_lo: Optional[pd.Series] = None,
-                 confidence_hi: Optional[pd.Series] = None,
+                 series: Union[pd.Series, pd.DataFrame],
                  freq: Optional[str] = None,
                  fill_missing_dates: Optional[bool] = True):
         """
-        A TimeSeries is an object representing a univariate time series, and optional confidence intervals.
+        A TimeSeries is an object representing a time series.
```
hrzn, line 26: `representing a univariate or multivariate time series` ?

u8timeseries/timeseries.py
```diff
@@ -20,35 +19,34 @@
...
             Optionally, a boolean value indicating whether to fill missing dates with NaN values
             in case the frequency of `series` cannot be inferred.
         """
+        # create dataframes
+        if (isinstance(series, pd.Series)):
+            series = pd.DataFrame(series)
+        series.columns = range(series.shape[1])
 
-        raise_if_not(len(series) > 0, 'Series must not be empty.', logger)
+        raise_if_not(len(series) > 0 and series.shape[1] > 0, 'Series must not be empty.', logger)
```
hrzn, line 46: `Dataframe must not be empty`

u8timeseries/timeseries.py
```diff
@@ -20,35 +19,34 @@
...
         raise_if_not(isinstance(series.index, pd.DatetimeIndex), 'Series must be indexed with a DatetimeIndex.', logger)
-        raise_if_not(np.issubdtype(series.dtype, np.number), 'Series must contain numerical values.', logger)
+        raise_if_not(series.dtypes.apply(lambda x: np.issubdtype(x, np.number)).all(), 'Series must'
```
hrzn, line 48: `Dataframe`

u8timeseries/timeseries.py
```diff
@@ -73,48 +71,30 @@ def __init__(self,
...
+    def _assert_univariate(self):
         """
-        Returns
-        -------
-        pandas.Series
-            A copy of the Pandas Series underlying this time series
+        Raises an error if the current TimeSeries instance is not univariate.
         """
-        return self._series.copy()
+        raise_if_not(self._series.shape[1] == 1, 'Only univariate TimeSeries instances support this method', logger)
```
hrzn, line 78: This one will raise a `ValueError`. Depending on the context (e.g. when called from `pd_series()`), it might not be the best error to raise, as we are not in a case where the provided function arguments are bad. Could you raise better errors that reflect the context? Maybe `AssertionError` is already better.

u8timeseries/timeseries.py
```diff
@@ -73,48 +71,30 @@ def __init__(self,
...
 
-    def conf_lo_pd_series(self) -> Optional[pd.Series]:
+    def pd_series(self) -> pd.Series:
```
hrzn, line 80: I think this could be changed to something like `get_pd_series(self, column=None)`. If column is `None` the series must be univariate, otherwise return the specified column as a series.

u8timeseries/timeseries.py
```diff
@@ -20,35 +19,34 @@
...
+        series.columns = range(series.shape[1])
 
-        raise_if_not(len(series) > 0, 'Series must not be empty.', logger)
+        raise_if_not(len(series) > 0 and series.shape[1] > 0, 'Series must not be empty.', logger)
         raise_if_not(isinstance(series.index, pd.DatetimeIndex), 'Series must be indexed with a DatetimeIndex.', logger)
-        raise_if_not(np.issubdtype(series.dtype, np.number), 'Series must contain numerical values.', logger)
+        raise_if_not(series.dtypes.apply(lambda x: np.issubdtype(x, np.number)).all(), 'Series must'
+                     ' contain only numerical values.', logger)
         raise_if_not(len(series) >= 3 or freq is not None, 'Series must have at least 3 values if the "freq" argument'
                      'is not passed', logger)
```
hrzn, line 52: Just below this line, could we rename `self._series` in `self._df` ?

hrzn: I'd also rename the __init__ argument `series` into `df`

u8timeseries/timeseries.py
```diff
@@ -141,7 +121,8 @@ def first_value(self) -> float:
         float
             The first value of this series
         """
-        return self._values[0]
+        self._assert_univariate()
```
hrzn, line 124: Why do you need to enforce univariate here? All the dimensions in the multivariate case have the same time index, so it would be consistent to just return the first timestamp of this index.

u8timeseries/timeseries.py
```diff
@@ -150,7 +131,8 @@ def last_value(self) -> float:
         float
             The last value of this series
         """
-        return self._values[-1]
+        self._assert_univariate()
```
hrzn, line 134: same comment here

u8timeseries/timeseries.py
```diff
@@ -141,7 +121,8 @@ def first_value(self) -> float:
         float
             The first value of this series
         """
-        return self._values[0]
+        self._assert_univariate()
+        return self._values[0][0]
```
hrzn, line 125: how about defining `first_values()` and `last_values()` counterpart functions?

u8timeseries/timeseries.py
```diff
@@ -489,10 +466,6 @@ def from_dataframe(df: pd.DataFrame,
             The time column name (mandatory). If set to `None`, the DataFrame index will be used.
         value_col
             The value column name (mandatory).
-        conf_lo_col
-            The lower confidence interval column name (optional).
-        conf_hi_col
-            The upper confidence interval column name (optional).
         freq
             Optionally, a string representing the frequency of the Pandas Series.
```
hrzn, line 470: Change `Series` into time series

u8timeseries/timeseries.py
```diff
@@ -574,13 +533,10 @@ def plot(self,
 
         fig = (plt.figure() if new_plot else (kwargs['figure'] if 'figure' in kwargs else plt.gcf()))
         kwargs['figure'] = fig
-        self._series.plot(*args, **kwargs)
+        self.pd_series().plot(*args, **kwargs)
```
hrzn, line 562: How about plotting all the dimensions w.r.t the time index, and maybe printing an error if there are too many dimensions (e.g. 15 or more), and letting the users optionally choose which dimensions to plot?

u8timeseries/backtesting/backtesting.py
```diff
@@ -91,9 +91,8 @@ def backtest_forecasting(series: TimeSeries,
         train = series.drop_after(pred_time)  # build the training series
         model.fit(train)
         pred = model.predict(fcast_horizon_n)
-        values.append(pred.values()[-1])  # store the N-th point
+        values.append(pred.univariate_values()[-1])  # store the N-th point
```
hrzn, line 94: we will eventually need to extend model forecasts (and backtesting) to multivariate forecasts.

u8timeseries/models/baselines.py
```diff
@@ -28,7 +28,7 @@ def __str__(self):
 
     def fit(self, series: TimeSeries):
         super().fit(series)
-        self.mean_val = np.mean(series.values())
+        self.mean_val = np.mean(series.univariate_values())
```
hrzn, line 31: For univariate models, it would be nice if they print a user-friendly message when they are called with a multi-variate timeseries.
Perhaps one way to go would be to have 2 super-classes: `UnivariateForecastingModdel` and `MultivariateForecastingModel`, and the univariate one would do a check in its `fit()` method (which is called by all children), to ensure that the series it's called on is univariate, or alternatively that the user specified a column in the series. WDYT?

u8timeseries/models/standard_regression_model.py
```diff
@@ -46,7 +46,7 @@ def __init__(self,
 
     @staticmethod
     def _get_features_matrix_from_series(features: List[TimeSeries]):
-        return np.array([s.values() for s in features]).T  # (n_samples x n_features)
+        return np.array([s.univariate_values() for s in features]).T  # (n_samples x n_features)
```
hrzn, line 49: We'll also need to change this. Perhaps now we could have just one multivariate time series containing all the features.

u8timeseries/timeseries.py
```diff
@@ -458,22 +447,36 @@ def shift(self, n: int) -> 'TimeSeries':
...
-        new_conf_hi = None
-        if self._confidence_lo is not None:
-            new_conf_lo = self._confidence_lo.copy()
-            new_conf_lo.index = new_time_index
-        if self._confidence_hi is not None:
-            new_conf_hi = self._confidence_hi.copy()
-            new_conf_hi.index = new_time_index
-        return TimeSeries(new_series, new_conf_lo, new_conf_hi)
+        return TimeSeries(new_series)
+
```
hrzn, line 451: How about a `stack` function to stack several (multi or uni- variate) series into one?

u8timeseries/timeseries.py

u8timeseries/timeseries.py

u8timeseries/timeseries.py
pennfranc: My thinking was that although we do not use a pandas series as underlying data type anymore, we are still talking about time series. One (maybe minor) issue that I can see happening if we replace 'Series' with 'Dataframe' in this error message is that if this error is thrown by the user calling TimeSeries.from_series(pd_series) where pd_series is empty, the error message will be 'Dataframe must not be empty' which might confuse the user.

u8timeseries/timeseries.py
pennfranc: This is addressed in the step 2 branch by a method called `univariate_component` which accepts an integer index and returns a new univariate TimeSeries instance. So the functionality you suggest could be achieved by `series.univariate_component(column).pd_series()`. Do you think this is ok or should this be possible more directly, such as with the function you suggest?

u8timeseries/timeseries.py
hrzn: Good point. How about `Time series must not be empty` ?

u8timeseries/timeseries.py
hrzn: Your solution looks good :) thanks

u8timeseries/timeseries.py
pennfranc: That's definitely a good idea. I will do this as part of step 2, since I have an appropriate test file for new multivariate functions.

u8timeseries/timeseries.py
pennfranc: done

u8timeseries/timeseries.py
pennfranc: done

u8timeseries/timeseries.py
pennfranc: Yeah that sounds like a good solution

u8timeseries/timeseries.py
pennfranc: I changed it to `Time series`, like above.

u8timeseries/timeseries.py

u8timeseries/timeseries.py

u8timeseries/timeseries.py

u8timeseries/timeseries.py

u8timeseries/timeseries.py
pennfranc: done in step 2

u8timeseries/timeseries.py
pennfranc: Sounds good, I will do this in step 2 since I think it fits better there

u8timeseries/backtesting/backtesting.py
pennfranc: I will add this to my todo list for step 3

u8timeseries/models/standard_regression_model.py
pennfranc: I will add this to my todo list for step 3. Although requiring just one big multivariate TimeSeries instance means that the user has to stack all his predictions manually beforehand.

hrzn: Perhaps we can support an arbitrary number of (possibly multivariate) series and stack them anyway.

PR #98, opened 2020-06-16
fix(examples/notebooks): fix issue with renaming u8timeseries to darts

Droxef:
Correct the issue in the notebooks where the library name was not changed everywhere


examples/FFT-examples.ipynb
```diff
@@ -183,12 +183,12 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "MAPE: 521378672076109.06\n"
+      "MAPE: nan\n"
```
TheMP, line 186: In what state will those notebooks be in github? 

Are the output cleared now? Can we somehow do it with a 'minimum' amount of changes in the notebooks to keep the diffs small?

Droxef: They are rerun.
I can indeed clear those notebooks, so that there should be only necessary change.

TheMP: Ah sorry for misleasing comment, I don't think that cleaning is necessary. In fact, output of notebooks is very nice and allows us to see how the notebooks should look like before executing them: https://github.com/unit8co/darts/blob/develop/examples/RNN-examples.ipynb. 

What I was trying to say is do we really need such a big delta in PR or is it possible to just change the library imports and leave the res as-is - but it seems that indeed reexecuting all notebooks is the best approach here.

Droxef: Sorry, I needed to rerun them to see if they did work. Thus losing their previous state. Even opening them lead to unwanted changes

PR #99, opened 2020-06-17
fix/short-timeseries-support

pennfranc:
added support for operations on short (length < 3) TimeSeries objects

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -56,6 +56,8 @@ class TimeSeries:
 
         if (len(series) < 3):
             self._freq: str = freq
+            logger.info('A TimeSeries with length below 3 is being created. Please note that this can lead to'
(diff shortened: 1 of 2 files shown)


darts/timeseries.py
```diff
@@ -212,9 +212,9 @@ def copy(self, deep: bool = True) -> 'TimeSeries':
             A copy of this time series.
         """
         if deep:
-            return TimeSeries(self.pd_series(), self.conf_lo_pd_series(), self.conf_hi_pd_series())
+            return TimeSeries(self.pd_series(), self.conf_lo_pd_series(), self.conf_hi_pd_series(), self.freq())
```
TheMP, line 215: Maybe some warning in the constructor that the series is small and issues can occur would be useful

pennfranc: done

PR #102, opened 2020-06-18
Fix/release workflow fix

hrzn:
Fix the `pip install torch` command that was failing in tests actions.

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -80,7 +80,7 @@ jobs:
       - name: "5. Install main and release dependencies"
         run: |
           # this avoids an unnecessary cuda install on Linux
-          pip install -q torch==1.5.0
+          pip install -q torch==1.5.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
           pip install -q .
           pip install -q -r requirements/release.txt
 
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -38,7 +38,7 @@ jobs:
       - name: "5. Install main and dev dependencies"
         run: |
           # this avoids an unnecessary cuda install on Linux
-          pip install -q torch
+          pip install -q torch==1.5.0
           pip install -q .
           pip install -q -r requirements/dev.txt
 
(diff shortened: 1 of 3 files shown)


.github/workflows/develop.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: "5. Install main and dev dependencies"
         run: |
           # this avoids an unnecessary cuda install on Linux
-          pip install -q torch==1.5.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+          pip install -q torch
```
TheMP, line 41: @endrjuskr why was this version crashing?

@hrzn will out code be compatible with future releases of torch?

TheMP: Seems like the version '1.5.0' exists on pip -> maybe it's better to fix it on that one.

From the error message:

Run # this avoids an unnecessary cuda install on Linux
ERROR: Could not find a version that satisfies the requirement torch==1.5.0+cpu (from versions: 0.1.2, 0.1.2.post1, 0.1.2.post2, 0.1.7.post2, 0.1.8.post1, 0.1.9.post1, 0.1.9.post2, 0.1.10.post1, 0.1.11.post4, 0.1.11.post5, 0.1.12.post1, 0.1.12.post2, 0.2.0.post1, 0.2.0.post2, 0.2.0.post3, 0.3.0, 0.3.0.post2, 0.3.0.post3, 0.3.0.post4, 0.3.1, 0.4.0, 0.4.1, 1.0.0, 1.0.1, 1.0.1.post2, 1.1.0, 1.1.0.post2, 1.2.0, 1.3.0, 1.3.0.post2, 1.3.1, 1.4.0, 1.5.0)
ERROR: No matching distribution found for torch==1.5.0+cpu
##[error]Process completed with exit code 1.

endrjuskr: So on macos we should install 1.5.0, but for ubuntu we can install cpu only

hrzn: Right, will do that

hrzn: @endrjuskr we can install cpu only, but I think we don't loose anything installing the cuda version (it shouldn't change anything, and maybe simplifies the install command altogether)

TheMP: +1, optimising at CI level is not that important =]

PR #103, opened 2020-06-18
Remove `.darts` directory after tests.

Kostiiii:
Fix minor issue with dangling `.darts` directory with checkpoints after running tests.

### Summary
TCNModel was creating `.darts` directory without removing it after tests.


darts/tests/test_TCN.py
```diff
@@ -12,6 +13,10 @@ class TCNModelTestCase(unittest.TestCase):
     def setUpClass(cls):
         logging.disable(logging.CRITICAL)
 
+    @classmethod
```
TheMP, line 16: Why remove this directory only in TCN test case?

Kostiiii: I checked other test cases and the ones that use model that creates this directory already cleanup after themselves.

pennfranc: Yeah I forgot to add this, thanks @Kostiiii!

TheMP: Approved

PR #123, opened 2020-07-03
Feature/four theta

Droxef:
Implement 4Theta method from M4 competition

### Summary

Naive implementation of the 4Theta method from the organizers of M4 competition.
Modify backtesting gridsearch to let an automatic search of the best hyper-parameters


darts/models/theta.py
```diff
@@ -126,3 +127,217 @@ def predict(self, n: int) -> 'TimeSeries':
...
+            forecast = forecast**self.wses * drift**self.wdrift
+        else:
+            raise_log(ValueError("model_mode cannot be {}".format(self.model_mode)))
+
+        # Re-apply the seasonal trend of the TimeSeries
+        if self.is_seasonal:
+
+            replicated_seasonality = np.tile(self.seasonality.pd_series()[-self.season_period:],
+                                             math.ceil(n / self.season_period))[:n]
+            if self.season_mode in ['multiplicative', 'mul']:
```
TheMP, line 287: for simplicity I would also pick one of them, and also an ENUM that we pass here might be helpful to avoid typos https://docs.python.org/3/library/enum.html

darts/models/theta.py
```diff
@@ -88,13 +89,13 @@ def fit(self, series: TimeSeries, component_index: Optional[int] = None):
...
         # SES part of the decomposition.
-        self.model = hw.SimpleExpSmoothing(new_ts.values()).fit()
+        self.model = hw.SimpleExpSmoothing(new_ts.values()).fit(initial_level=0.2)
 
         # Linear Regression part of the decomposition. We select the degree one coefficient.
         b_theta = np.polyfit(np.array([i for i in range(0, self.length)]), (1.0 - self.theta) * new_ts.values(), 1)[0]
 
         # Normalization of the coefficient b_theta.
-        self.coef = b_theta / (2.0 - self.theta)
+        self.coef = b_theta / (2.0 - self.theta)  # change to b_theta / (-self.theta) if classical theta
```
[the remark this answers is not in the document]
Droxef: Mathematically, I'm still not quite sure about the rest, but I am sure it is correct for theta = 0 and 1.
And the definition of the theta lines are not quite the same, thus the confusing theta.

But comparing with the classical theta (4theta with default parameter), we have the exact same results if we change the normalization by -1/theta. (because there is a symmetry: 2-theta_1 = theta_2)

darts/models/theta.py
```diff
@@ -126,3 +127,217 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        self.drift = np.poly1d(np.polyfit(np.arange(self.length), linreg, 1))
+        theta0_in = self.drift(np.arange(self.length))
+        if self.trend_mode == 'exponential':
+            theta0_in = np.exp(theta0_in)
+
+        if self.model_mode == 'additive':
+            theta_t = self.theta * new_ts.values() + (1 - self.theta) * theta0_in
+        elif self.model_mode == 'multiplicative' and (theta0_in > 0).all():
+            theta_t = (new_ts.values() ** self.theta) * (theta0_in ** (1 - self.theta))
+        else:
```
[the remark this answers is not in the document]
Droxef: For a first draft, I blindly replicated the step of the original alorithm. So I will change it. 
But we still need to check if all values are positive if 'multiplicative' models are chosen or else it will fail. 
I can either raise an error, or simply keep the fallback as it is.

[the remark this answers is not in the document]
Droxef: There should be a more optimized implementation, like our theta. But we need to do the math to find it

darts/models/theta.py
Droxef: To be sure I understand correctly, the different model arguments will accept an Enum value? 
I can change the theta method as well in the same manner. Thus Enum classes will be declared exterior to the class

TheMP: Yes this comment actually applies to both theta methods - instead of retyping "multiplicative"/"additive" string we can have an enum that would have 2 values - MULTIPLICATIVE or ADDITIVE and if you make a typo the IDE will tell you something is not matching (not the case for raw strings)

darts/backtesting/backtesting.py
```diff
@@ -313,15 +313,16 @@ def backtest_gridsearch(model_class: type,
 
     Parameters
     ----------
-    model
+    model_class
```
pennfranc, line 316: I let this slip through, thanks!

darts/backtesting/backtesting.py
```diff
@@ -360,6 +361,10 @@ def backtest_gridsearch(model_class: type,
         if val_series is None:  # expanding window mode
             backtest_forecast = backtest_forecasting(train_series, model, backtest_start_time, fcast_horizon_n)
             error = metric(backtest_forecast, train_series)
+        elif val_series == 'train':
+            model.fit(train_series)
+            # Use ndarray because casting to TimeSeries takes too much time
+            error = metric(model.fitted_values, train_series.univariate_values())
```
pennfranc, line 367: As far as I can tell, FourTheta is the only model that possesses the attribute `fitted_values`. So it would be good to have a check here to test whether the given model supports this functionality, right?

darts/models/theta.py
```diff
@@ -88,13 +89,13 @@ def fit(self, series: TimeSeries, component_index: Optional[int] = None):
             new_ts = remove_seasonality(ts, self.season_period, model=self.mode)
 
         # SES part of the decomposition.
-        self.model = hw.SimpleExpSmoothing(new_ts.values()).fit()
+        self.model = hw.SimpleExpSmoothing(new_ts.values()).fit(initial_level=0.2)
```
pennfranc, line 92: What is the effect of this change?

darts/backtesting/backtesting.py
Droxef: At least all ExponentialSmoothing have a fitted attribute that can be retrieved (So the current theta model could have one too, I think). If you approve this functionality, it might be interesting to add this attribute in other models.
But yes, a check is necessary.

darts/models/theta.py
Droxef: The problem with SES in statsmodels is that alpha is not constrained between [0.1, 0.99] as it should, thus giving NaN values when alpha is 0.
Setting the initial_level to 0.2 (or anything else) avoided all encountered cases where the optimization gave alpha=0.
It is more a hotfix rather than an actual solution before statsmodels corrects the problem.

Droxef: Seems like it leads to different results. I will add it only in the case alpha is 0, and it will recompute it

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        # - if theta = 1, then the theta method restricts to a simple exponential smoothing (SES)
+        # - if theta = 0, then the theta method restricts to a simple `trend_mode` regression.
+
+    def fit(self, ts, component_index: Optional[int] = None):
+        super().fit(ts, component_index)
+
+        self.length = len(ts)
+        # normalization of data
+        self.mean = ts.mean().mean()
+        if self.mean == 0:
```
hrzn, line 205: That seems weird, do they do the same thing in the original implementation?

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+            logger.warn("Time series has negative values. Fallback to additive and linear model")
+
+        # Check for statistical significance of user-defined season period
+        # or infers season_period from the TimeSeries itself.
+        if self.season_mode is Season.NONE:
+            self.season_period = 1
+        if self.season_period is None:
+            max_lag = len(ts) // 2
+            self.is_seasonal, self.season_period = check_seasonality(ts, self.season_period, max_lag=max_lag)
+            logger.info('Theta model inferred seasonality of training series: {}'.format(self.season_period))
```
hrzn, line 222: `FourTheta model`

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+
+        # Store and remove seasonality effect if there is any.
+        if self.is_seasonal:
+            _, self.seasonality = extract_trend_and_seasonality(new_ts, self.season_period,
+                                                                model=self.season_mode.value)
+            new_ts = remove_seasonality(new_ts, self.season_period, model=self.season_mode.value)
+
+        # Drift part of the decomposition
+        if self.trend_mode is Trend.LINEAR:
+            linreg = new_ts.univariate_values()
```
hrzn, line 235: You could start extracting the univariate values earlier in the method, in order to do it only once, and fail early if it's not univariate.

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+                self.model = hw.SimpleExpSmoothing(theta_t).fit()
+                theta2_in = self.model.fittedvalues
+            self.fitted_values = self.wses * theta2_in + self.wdrift * theta0_in
+        if self.is_seasonal:
+            if self.season_mode is Season.ADDITIVE:
+                self.fitted_values += self.seasonality.univariate_values()
+            elif self.season_mode is Season.MULTIPLICATIVE:
+                self.fitted_values *= self.seasonality.univariate_values()
+        self.fitted_values *= self.mean
+        # takes too much time to create a time series for fitted_values
```
hrzn, line 271: what's the overhead?

hrzn: Also, could you add a comment explaining what `fitted_values` represent?

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+                forecast *= replicated_seasonality
+            else:
+                forecast += replicated_seasonality
+
+        forecast *= self.mean
+
+        return self._build_forecast_series(forecast)
+
+    @staticmethod
+    def select_best_model(ts: TimeSeries, thetas: List[int] = None, m: Optional[int] = None):
```
hrzn, line 305: Could you add type annotation for the return type?

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        ----------
+        ts
+            The TimeSeries on which the model will be tested.
+        thetas
+            A list of thetas to loop over.
+        m
+            Optionally, the season used to decompose the time series.
+        Returns
+        -------
+        theta
```
hrzn, line 319: Here you should have a type

darts/models/theta.py
hrzn: Also, since you plan for the case `thetas is None` later, update `thetas: Optional[List[int]]`

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+    def select_best_model(ts: TimeSeries, thetas: List[int] = None, m: Optional[int] = None):
+        """
+        Performs a grid search over all hyper parameters to select the best model.
+
+        Parameters
+        ----------
+        ts
+            The TimeSeries on which the model will be tested.
+        thetas
+            A list of thetas to loop over.
```
hrzn, line 314: Perhaps mention the default values here.

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        -------
+        theta
+            The best performing model on the time series.
+        """
+        # Only import if needed
+        from ..backtesting.backtesting import backtest_gridsearch
+        from sklearn.metrics import mean_absolute_error as mae
+        if thetas is None:
+            thetas = [1, 2, 3]
+        elif isinstance(thetas, int):
```
hrzn, line 327: I would remove this case, that's abusing the types, better let it crash IMO.

darts/models/theta.py
```diff
@@ -115,14 +123,228 @@ def predict(self, n: int) -> 'TimeSeries':
...
+            thetas = [thetas]
+        if (ts.values() <= 0).any():
```
hrzn, line 329: Perhaps add a log message if this happens?

darts/models/theta.py
hrzn, line 305: We will have to think where to put this method.

darts/backtesting/backtesting.py
```diff
@@ -399,8 +400,13 @@ def backtest_gridsearch(model_class: type,
     """
 
     if (val_series is not None):
-        raise_if_not(train_series.width == val_series.width, "Training and validation series require the same"
-                     " number of components.", logger)
+        if val_series == 'train':
```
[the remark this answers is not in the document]
Droxef: I can change val_series to take the train TimeSeries, and perform a check `train_series == val_series` to activate this option so there is no abuse

darts/backtesting/backtesting.py
```diff
@@ -365,15 +365,16 @@ def backtest_gridsearch(model_class: type,
...
         The ForecastingModel subclass to be tuned for 'series'.
     parameters
         A dictionary containing as keys hyperparameter names, and as values lists of values for the
         respective hyperparameter.
     train_series
         The univariate TimeSeries instance used for training (and also validation in split mode).
-    test_series
+    val_series
         The univariate TimeSeries instance used for validation in split mode.
+        Use `train` to compare with model.fitted_values
```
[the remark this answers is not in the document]
Droxef: `model.fitted_values` are the values obtain when fitting a model (ExponentialSmoothing in our case) on a time series, the result of the fit (a TimeSeries with same index as the train time series). It is a quick way to assess the model (we don't have to compute the predictions each time), but I understand that we cannot see if the fitted model underfit or overfit and will give bad predictions.
It is a trade off between computation time and efficiency. Thus, it can be dropped if you think it is not a good thing to have.
I mainly used this version for the M4 benchmark because the original script used this method to select the best model.

darts/models/theta.py
Droxef: Sorry, I forgot about that. I will change it

Droxef: > I think it will depend on the refactoring of backtesting. If gridsearch become an attribute of ForcastingModel, it becomes possible to overload it here.

darts/models/theta.py
Droxef: The TimeSeries in M4 are all stricly positive, so no.
I will raise an error in this case, but the model can run well without normalization in this case.

Droxef: Additionally, I can let the choice to use the normalization to the user with a parameter.

darts/backtesting/backtesting.py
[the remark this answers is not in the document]
hrzn: If we do that the type will be OK, but I think it might be misleading because the behavior will be different, as we wouldn't in this case evaluate `model.predict(len(train_series))`. I almost prefer taking the "train" string in this case. @pennfranc @guillaumeraille any good idea?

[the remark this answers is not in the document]
guillaumeraille: I think the easiest and clearest would be to add a separate boolean argument. On top of what Julien mentionned `val_series` ends with series which makes it very confusing to be a string here in my opinion :)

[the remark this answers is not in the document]
hrzn: Maybe a boolean argument `use_fitted_values` could do indeed. It would override the `val_series` when set.

darts/models/theta.py
```diff
@@ -41,8 +42,8 @@ def __init__(self,
         seasonality_period
             User-defined seasonality period. If not set, will be tentatively inferred from the training series upon
             calling `fit()`.
-        mode
-            Type of seasonality. Either "additive" or "multiplicative".
+        season_mode
+            Type of seasonality. Must be a Season Enum member. Defaults to `Season.MULTIPLICATIVE`.
```
hrzn, line 46: I think for all the enums we should either link to the possible values, or list them (including `Season.NONE`).

darts/models/theta.py
```diff
@@ -94,9 +96,12 @@ def fit(self, series: TimeSeries, component_index: Optional[int] = None):
         b_theta = np.polyfit(np.array([i for i in range(0, self.length)]), (1.0 - self.theta) * new_ts.values(), 1)[0]
 
         # Normalization of the coefficient b_theta.
-        self.coef = b_theta / (2.0 - self.theta)
+        self.coef = b_theta / (2.0 - self.theta)  # change to b_theta / (-self.theta) if classical theta
 
         self.alpha = self.model.params["smoothing_level"]
+        if self.alpha == 0.:
+            self.model = hw.SimpleExpSmoothing(new_ts.values()).fit(initial_level=0.2)
```
hrzn, line 103: I don't feel too comfortable with this magic number. Could we at the very least put it as a constant at the top of the file?
How did you pick 0.2?

darts/models/theta.py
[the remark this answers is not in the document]
hrzn: Could we unify the meaning of theta? E.g. by changing to `b_theta / (-self.theta)` and adapting the default value to `theta=2` ?

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        """
+        An implementation of the 4Theta method with configurable `theta` parameter.
+
+        See M4 competition `solution <https://github.com/Mcompetitions/M4-methods/blob/master/4Theta%20method.R>`_.
+
+        The training time series is de-seasonalized according to `seasonality_period`,
+        or an inferred seasonality period.
+
+        When called with `theta = 2 - X`, `model_mode = Model.ADDITIVE` and `trend_mode = Trend.LINEAR`,
+        this model is equivalent to calling `Theta(theta=X)`.
```
hrzn, line 151: Do we still need the other Theta model implementation? Wouldn't it be enough to just call this one?

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        ----------
+        theta
+            Value of the theta parameter. Defaults to 2.
+            If theta = 1, then the fourtheta method restricts to a simple exponential smoothing (SES).
+            If theta = 0, then the fourtheta method restricts to a simple `trend_mode` regression.
+        seasonality_period
+            User-defined seasonality period. If not set, will be tentatively inferred from the training series upon
+            calling `fit()`.
+        model_mode
+            Type of model combining the Theta lines. Must be a Model Enum member. Defaults to `Model.ADDITIVE`.
```
hrzn, line 163: Same comment as above, let's link to a place where users can see all possible values for the Enums

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+
+    def fit(self, ts, component_index: Optional[int] = None):
+        super().fit(ts, component_index)
+
+        self.length = len(ts)
+        # normalization of data
+        if self.normalization:
+            self.mean = ts.mean().mean()
+            raise_if_not(not np.isclose(self.mean, 0),
+                         "The mean value of the TimeSeries must not be 0 when using normalization", logger)
```
hrzn, line 205: Perhaps you can write "The mean value of the provided series is too close to zero to perform normalization"

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+            new_ts = ts / self.mean
+        else:
+            new_ts = ts.copy()
```
hrzn, line 208: do you need this copy? Normally everything should be immutable :)

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+            max_lag = len(ts) // 2
+            self.is_seasonal, self.season_period = check_seasonality(ts, self.season_period, max_lag=max_lag)
+            logger.info('FourTheta model inferred seasonality of training series: {}'.format(self.season_period))
+        else:
+            # force the user-defined seasonality to be considered as a true seasonal period.
+            self.is_seasonal = self.season_period > 1
+
+        # Store and remove seasonality effect if there is any.
+        if self.is_seasonal:
+            _, self.seasonality = extract_trend_and_seasonality(new_ts, self.season_period,
```
hrzn, line 224: could you change `extract_trend_and_seasonality()` in order to accept Enums instead of strings? And also everywhere else it's needed.

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+                                                                model=self.season_mode.value)
+            new_ts = remove_seasonality(new_ts, self.season_period, model=self.season_mode.value)
+
+        ts_values = new_ts.univariate_values()
```
hrzn, line 228: You can do this earlier in the method, so that it fails upfront if users call it with a multivariate series

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+                self.fitted_values += self.seasonality.univariate_values()
+            elif self.season_mode is Season.MULTIPLICATIVE:
+                self.fitted_values *= self.seasonality.univariate_values()
+        # Fitted values are the results of the fit of the model on the train series. A good fit of the model
+        # will lead to fitted_values similar to ts. But one cannot see if it overfits.
+        if self.normalization:
+            self.fitted_values *= self.mean
+        # Takes too much time to create a TimeSeries
+        # Overhead: 30% ± 10 (2-10 ms in average)
+        self.fitted_values = TimeSeries.from_times_and_values(ts.time_index(), self.fitted_values)
```
hrzn, line 277: You could build the `TimeSeries` only in the backtesting method, this way the time overhead is paid only for backtesting and not for simply fitting the model.

darts/tests/test_autoregression_models.py
```diff
@@ -37,6 +36,8 @@ class AutoregressionModelsTestCase(unittest.TestCase):
         (Theta(), 11.3),
         (Theta(1), 20.2),
         (Theta(3), 9.8),
+        (FourTheta(1), 20.2),
+        (FourTheta(-1), 9.8),
```
hrzn, line 40: could you add testing for a few more modes?

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        if self.normalization:
+            forecast *= self.mean
+
+        return self._build_forecast_series(forecast)
+
+    @staticmethod
+    def select_best_model(ts: TimeSeries, thetas: Optional[List[int]] = None,
+                          m: Optional[int] = None, normalization: bool = True) -> 'FourTheta':
+        """
+        Performs a grid search over all hyper parameters to select the best model.
```
hrzn, line 314: Perhaps mention that it is using the fitted values on the training series here.

darts/tests/test_4theta.py
```diff
@@ -0,0 +1,69 @@
...
+        series = sine_series + linear_series
+        train_series, val_series = series.split_before(series.time_index()[-10])
+        thetas = np.linspace(-3, 3, 30)
+        best_model = FourTheta.select_best_model(train_series, thetas)
+        model = FourTheta(random.choice(thetas), model_mode=random.choice(list(Model)),
+                          trend_mode=random.choice(list(Trend)), season_mode=random.choice(list(Season)))
+        model.fit(train_series)
+        best_model.fit(train_series)
+        forecast_random = model.predict(10)
+        forecast_best = model.predict(10)
```
hrzn, line 68:
```suggestion
        forecast_best = best_model.predict(10)
```

darts/models/theta.py
```diff
@@ -115,14 +120,241 @@ def predict(self, n: int) -> 'TimeSeries':
...
+            else:
+                forecast += replicated_seasonality
+
+        if self.normalization:
+            forecast *= self.mean
+
+        return self._build_forecast_series(forecast)
+
+    @staticmethod
+    def select_best_model(ts: TimeSeries, thetas: Optional[List[int]] = None,
```
hrzn, line 311: We can keep it temporarily, but then eventually I think we should have each model (if possible) given good reasonable default hyper parameters sets, and then let the user call `backtest_gridsearch(params='default', ...)` or something like this.
Ping @grll

darts/models/theta.py
Droxef: I cannot do this earlier, or I will need to modify this value (if normalization and seasonality) and rewrite the functions for ndarray, or recall the function.
But I will add an assert_univariate at the beginning

hrzn: Oh right, OK

darts/models/theta.py
Droxef: It is not a perfect solution, but a temporary one. alpha should be bounded in ]0, 1[, and is now [0,1] in statsmodels.
0.2 is just good practice to reach good results, and gave similar results as the R function.

darts/models/theta.py
```diff
@@ -33,16 +34,21 @@ def __init__(self,
...
+            Value of the theta parameter. Defaults to 2. Cannot be set to 0.
             If theta = 1, then the theta method restricts to a simple exponential smoothing (SES)
         seasonality_period
             User-defined seasonality period. If not set, will be tentatively inferred from the training series upon
             calling `fit()`.
-        mode
-            Type of seasonality. Either "additive" or "multiplicative".
+        season_mode
+            Type of seasonality.
+            Either SeasonalityMode.MULTIPLICATIVE, SeasonalityMode.ADDITIVE or SeasonalityMode.NONE.
```
hrzn, line 50: Thanks :)

darts/models/theta.py
hrzn: Solved after our discussion - The original Theta implementation is faster.

darts/models/theta.py
```diff
@@ -115,14 +125,258 @@ def predict(self, n: int) -> 'TimeSeries':
...
+        if self.is_seasonal:
+            if self.season_mode is SeasonalityMode.ADDITIVE:
+                self.fitted_values += self.seasonality.univariate_values()
+            elif self.season_mode is SeasonalityMode.MULTIPLICATIVE:
+                self.fitted_values *= self.seasonality.univariate_values()
+        # Fitted values are the results of the fit of the model on the train series. A good fit of the model
+        # will lead to fitted_values similar to ts. But one cannot see if it overfits.
+        if self.normalization:
+            self.fitted_values *= self.mean
+        # Takes too much time to create a TimeSeries
```
hrzn, line 294: I think you can now remove this comment (or move to the corresponding backtesting function)

PR #125, opened 2020-07-06
Refactor/backtesting

guillaumeraille:
Fixes #DARTS-133.

### Summary

Contain a proposal to refactor backtesting.

Add reviews and comments directly to `PROPOSAL.md`


hrzn:
That looks very good @guillaumeraille 
+1 for having `model.backtest()` methods and +1 for the submodule `darts.model_selection`
Some very minor comments that I think of now: I think we could find better names for `fcast_horizon_n` and `explore_models()`.

PROPOSAL.md
```diff
@@ -0,0 +1,100 @@
...
+```
+Parameters of backtesting will always involve a `model` being backtested and a `TimeSeries` that act as ground truth.
+Since backtesting is usually refering to a model I propose to add backtest as a method of the `ForecastingModel` and 
+`RegressionModel`class.
+
+This would lead to the following API for the end user:
+
+```python
+model = ExponentialSmoothing()
+historical_forecast = model.backtest(series, pd.Timestamp('19550101'), fcast_horizon_n=3, verbose=True))
```
TheMP, line 52: What would be returned here? A pandas series or a new "Backtesting" object with some properties?

guillaumeraille: A time series containing the forecast values for `series`, when successively applying the specified model with the specified forecast horizon

guillaumeraille: But my favorite is the next use case example where it also directly return the residuals so that we can remove the forecast_residual method

PROPOSAL.md
```diff
@@ -0,0 +1,100 @@
...
+forcasting anyway:
+
+```python
+historical_forecast, residuals = model.backtest(series, pd.Timestamp('19550101'), fcast_horizon_n=3, verbose=True))
+```
+
+We could also add the gridSearch logic directly on the model class as a `classmethod`:
+
+```python
+best_model = ExponentialSmoothing.gridsearch(series, fcast_horizon...)
```
Droxef, line 65: I like this idea, as for some models, we must perform this search to find the best model (theta and ETS are good example, and I had to do one for 4Theta)
One other option to do the gridsearch would be to compare the fittedvalues when it exists, instead of the results of a backtest (faster and should assess the performance of the model as well)

guillaumeraille: in that case we could create a dedicated method for that on the model as well (cant be a classmethod if it needs some fitted value) what do you think ?

Droxef: Yes it seems more appropriate. Moreover that the fittedvalues call is not necessarily consistent between models

pennfranc: I like it, definitely more elegant. My only concern is whether this is an intuitive way to get to the residuals from the perspective of a new user.

PROPOSAL.md
```diff
@@ -0,0 +1,100 @@
...
+```python
+model = ExponentialSmoothing()
+historical_forecast = model.backtest(series, pd.Timestamp('19550101'), fcast_horizon_n=3, verbose=True))
+```
+
+We could also add residuals computation for free and return it as computation seems unexpansive and require a backtest
+forcasting anyway:
+
+```python
+historical_forecast, residuals = model.backtest(series, pd.Timestamp('19550101'), fcast_horizon_n=3, verbose=True))
```
pennfranc, line 59: I like it, definitely more elegant. My only concern is whether this is an intuitive way to get to the residuals from the perspective of a new user.

PROPOSAL.md
```diff
@@ -0,0 +1,100 @@
...
+Reference: #DARTS-133
+
+---
+
+## Summary
+
+The main idea behind this proposal is to refactor `backtesting.py` by implementing it directly in the model 
+class. Here would be a quick list of changes proposed:
+
+- `backtest_forecasting`, `backtest_regression`, `backtest_gridsearch` -> moved in corresping model methods
```
pennfranc, line 12: Do you think `backtest_gridsearch` could be made to a static method of `ForecastingModel`? (Since it takes a class, and not an instance as argument)

PROPOSAL.md
guillaumeraille: an other possibility would be to store residuals on the model and retrieve it on request but I believe residuals shouldn't belong to the model

PROPOSAL.md
hrzn: How about having an abstract method `ForecastingModel.hyperparams()` that returns the params dict that `backtest_gridsearch()` would iterate over? This way model implementers would just have to implement this one method for the model to be grid-searchable. @guillaumeraille @pennfranc wdyt?

PROPOSAL.md
hrzn: I guess that's a possible solution to https://github.com/unit8co/darts/pull/123/files#r453315099

PROPOSAL.md
guillaumeraille: Not a bad idea but I think the params dict is more dataset / application dependent hence unkown at model implementation time dont you think ?

hrzn: Right that's true. What I had in mind was more for this method to return decent default hyper-params, but we would still need to accept user-defined parameters.

PROPOSAL.md
```diff
@@ -0,0 +1,106 @@
+# Proposal for Refactoring Backtesting
+
+Reference: #DARTS-133
+
+---
+## Progress
+
+- [x] backtest_forcasting moved in model methods
```
TheMP, line 8: +1

darts/models/forecasting_model.py
```diff
@@ -95,6 +96,194 @@ def _build_forecast_series(self,
...
+        Raises
+        ------
+        ValueError
+            if an input doesn't pass sanity checks.
+        """
+        raise_if(not retrain, "Only 'TorchForecastingModel' instances support the option 'retrain=False'.", logger)
+
+    def _backtest_build_fit_and_predict_kwargs(self,
+                                               target_indices: Optional[List[int]],
+                                               component_index: Optional[int],
```
hrzn, line 138: this doesn't seem to be used

darts/models/forecasting_model.py
```diff
@@ -95,6 +96,194 @@ def _build_forecast_series(self,
...
+        """
+        fit_kwargs = {}
+        predict_kwargs = {}
+        fit_kwargs['target_indices'] = target_indices
+        return fit_kwargs, predict_kwargs
+
+    def backtest(self,
+                 series: TimeSeries,
+                 start: pd.Timestamp,
+                 forcast_horizon: int,
```
hrzn, line 169: `forecast_horizon`

darts/models/torch_forecasting_model.py
```diff
@@ -620,3 +620,50 @@ def _get_best_torch_device(self):
     def _get_learning_rate(self):
         for p in self.optimizer.param_groups:
             return p['lr']
+
+    def _backtest_model_specfic_sanity_checks(self, retrain: bool):
```
hrzn, line 624: `specific`

darts/models/forecasting_model.py
guillaumeraille: You mean because it is overriden? It is used as default and not rewritten in the multivariate case so used

hrzn: As we discussed, I was talking about `component_index` that is apparently not used in this method.

darts/models/forecasting_model.py
```diff
@@ -95,61 +97,241 @@ def _build_forecast_series(self,
...
+            Whether to print progress
+
+        Returns
+        -------
+        forecast
+            A time series containing the forecast values for `target_series`, when successively applying the specified
+            model with the specified forecast horizon.
+        residuals
+            Difference between the `forecast` and the actual `target_series` provided.
+        """
```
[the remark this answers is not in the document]
pennfranc: Yeah this is my fault. I'm changing the function signatures somewhat, but I didn't adapt the docstring.

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+            The first prediction time, at which a prediction is computed for a future time
+        forecast_horizon
+            The forecast horizon for the point prediction
+        stride
+            The number of time steps (the unit being the frequency of `series`) between two consecutive predictions.
+        retrain
+            Whether to retrain the model for every prediction or not. Currently only `TorchForecastingModel`
+            instances as `self` argument support setting `retrain` to `False`.
+        use_full_output_length
+            Optionally, if `self` is an instance of `TorchForecastingModel`, this argument will be passed along
```
hrzn, line 199: I'd replace `self` by "the model" everywhere

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+        # check start parameter
+        if hasattr(n, 'start'):
+            if isinstance(n.start, float):
+                raise_if_not(n.start >= 0.0 and n.start < 1.0, '`start` should be between 0.0 and 1.0.', logger)
+            elif isinstance(n.start, pd.Timestamp):
+                raise_if_not(n.start in training_series, '`start` timestamp is not in the `series`.', logger)
+                raise_if(n.start == training_series.end_time(), '`start` timestamp is the last timestamp of `series`',
+                         logger)
+            else:
+                raise_if(training_series[n.start].time_index()[0] == training_series.end_time(), '`start` timestamp '
```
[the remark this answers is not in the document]
hrzn: We could also add an additional check that `start + forecast_horizon` is in the series if `trim_to_series` is False.

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+        the model will only be trained only on the initial training window (up to `start` time stamp), and only if it
+        has not been trained before. Then, at every iteration, the newly expanded 'training sequence' will be fed to the
+        model to produce the new output.
+
+        Parameters
+        ----------
+        training_series
+            The training time series on which to backtest
+        target_series
+            The target time series on which to backtest
```
hrzn, line 188: Maybe mention for what kinds of models this one applies.

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+                   parameters: dict,
+                   training_series: TimeSeries,
+                   target_series: Optional[TimeSeries] = None,
+                   forecast_horizon: Optional[int] = None,
+                   use_full_output_length: Optional[bool] = None,
+                   val_target_series: Optional[TimeSeries] = None,
+                   num_predictions: int = 10,
+                   use_fitted_values: bool = False,
+                   metric: Callable[[TimeSeries, TimeSeries], float] = metrics.mape,
+                   verbose=False):
```
hrzn, line 280: Would be nice to annotate the return type

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+        forecast = TimeSeries.from_times_and_values(pd.DatetimeIndex(times), np.array(values))
+
+        return forecast
+
+    @classmethod
+    def gridsearch(model_class,
+                   parameters: dict,
+                   training_series: TimeSeries,
+                   target_series: Optional[TimeSeries] = None,
+                   forecast_horizon: Optional[int] = None,
```
hrzn, line 274: Could we support the same formats as for backtesting here?

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+        provided in the `parameters` dictionary by instantiating the `model_class` subclass
+        of ForecastingModel with each combination, and returning the best-performing model with regards
+        to the `metric` function. The `metric` function is expected to return an error value,
+        thus the model resulting in the smallest `metric` output will be chosen.
+        The relationship of the training data and test data depends on the mode of operation.
+
+        Expanding window mode (activated when `forecast_horizon` is passed):
+        For every hyperparameter combination, the model is repeatedly trained and evaluated on different
+        splits of `training_series` and `target_series`. The number of splits is equal to `num_predictions`, and the
+        forecasting horizon used when making a prediction is `forecast_horizon`.
```
hrzn, line 294: Perhaps add a sentence "The `num_predictions` point predictions are obtained by predicting the last `num_predictions` points of the training series with the specified `forecast_horizon`"

Or maybe you'll find a better formulation :D

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+        Note that the model is retrained for every single prediction, thus this mode is slower.
+
+        Split window mode (activated when `val_series` is passed):
+        This mode will be used when the `val_series` argument is passed.
+        For every hyperparameter combination, the model is trained on `training_series` + `target_series` and
```
hrzn, line 299: Just a thought: This naming of `training_series` and `target_series` is slightly confusing because usually in ML parlance the training set contains the targets as well. Could we maybe change to something better, like `train_feature_series` and `train_target_series`? Or even perhaps `X_train` and `y_train`. We would likely have to change in many places though.

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+        The three modes of operation evaluate every possible combination of hyperparameter values
+        provided in the `parameters` dictionary by instantiating the `model_class` subclass
+        of ForecastingModel with each combination, and returning the best-performing model with regards
+        to the `metric` function. The `metric` function is expected to return an error value,
+        thus the model resulting in the smallest `metric` output will be chosen.
+        The relationship of the training data and test data depends on the mode of operation.
+
+        Expanding window mode (activated when `forecast_horizon` is passed):
+        For every hyperparameter combination, the model is repeatedly trained and evaluated on different
+        splits of `training_series` and `target_series`. The number of splits is equal to `num_predictions`, and the
```
hrzn, line 293: You can add a mention that this uses the `backtest()` function.

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+                min_error = error
+                best_param_combination = param_combination_dict
+        logger.info('Chosen parameters: ' + str(best_param_combination))
+        return model_class(**best_param_combination)
+
+    def residuals(self,
+                  series: TimeSeries,
+                  forecast_horizon: int = 1,
+                  verbose: bool = False) -> TimeSeries:
+        """ A function for computing the residuals produced by the current model and a univariate time series.
```
hrzn, line 406: "and" --> "on" ?

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+
+        This function computes the difference between the actual observations from `series`
+        and the fitted values vector p obtained by training `self` on `series`.
```
hrzn, line 409: Replace `self` by "the model" in the whole docstring

darts/models/regression_model.py
```diff
@@ -79,6 +84,84 @@ def predict(self, features: List[TimeSeries]) -> TimeSeries:
...
+                 feature_series: Iterable[TimeSeries],
+                 target_series: TimeSeries,
+                 start: pd.Timestamp,
+                 forecast_horizon: int,
+                 trim_to_series: bool = True,
+                 verbose=False) -> TimeSeries:
+        """ A function for backtesting `RegressionModel`'s.
+
+        This function computes the time series of historical predictions
+        that would have been obtained, if the current model had been used to predict `series`
```
hrzn, line 97: `series` -> `target_series`

darts/models/regression_model.py
```diff
@@ -79,6 +84,84 @@ def predict(self, features: List[TimeSeries]) -> TimeSeries:
...
+        using the `feature_series`, with a certain time horizon.
+
+        To this end, it repeatedly builds a training set composed of both features and targets,
+        from `feature_series` and `target_series`, respectively.
+        It trains the current model on the training set, emits a (point) prediction for a fixed
+        forecast horizon, and then moves the end of the training set forward by one
```
hrzn, line 103: It'd be nice to introduce strides here too.

darts/models/theta.py
```diff
@@ -337,7 +337,7 @@ def select_best_model(ts: TimeSeries, thetas: Optional[List[int]] = None,
         using the fitted values on the training series `ts`.
 
 
-        Uses 'backtesting.backtest_gridsearch' with 'use_fitted_values=True' and 'metric=metrics.mae`.
+        Uses 'ForecastingModel.gridsearch' with 'use_fitted_values=True' and 'metric=metrics.mae`.
```
hrzn, line 340: This function `select_best_model` is basically doing a gridsearch, but it's misplaced IMO, and I think we should remove it if we can.
Instead one option could be to have each model implement a method `default_hyperparams` returning some good ranges of hyper parameters to be iterated on when grid-searching. This way afterwards in `ForecastingModel` we could have only one method `auto_select_model` doing the gridsearch on reasonable hyper-params.

darts/timeseries.py
```diff
@@ -1007,7 +1007,7 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
             series_df = pd_serie.to_frame()
         # multivariate case
         else:
-            raise_if(len(set(df_a.columns.to_list() + df_b.columns.columns.to_list())) != len(df_a.columns),
```
hrzn, line 1010: good catch ;)

darts/models/forecasting_model.py
pennfranc: Yes, I think the naming here is tricky. Originally Guillaume used the names `covariate_series` and `target_series`, which I also thought could be misleading. How about `input_series` and `target_series`? None of the words are included in the set of {`train`, `validation`, `test`}, and, as supposed to using X and y, they preserve the notion of time series objects instead of any set of features and labels. I think X and y could also be somewhat confusing since we are not dealing with the conventional concept of labels in the sense that our 'labels' might also be included in the 'features' (which is also the case by default when training on univariate series). What do you think?

darts/models/theta.py
pennfranc: I agree, some standardization may be good here. But should we leave this for a separate PR?

darts/models/forecasting_model.py
```diff
@@ -101,6 +105,343 @@ def _build_forecast_series(self,
...
+            elif isinstance(n.start, pd.Timestamp):
+                raise_if_not(n.start in training_series, '`start` timestamp is not in the `series`.', logger)
+                raise_if(n.start == training_series.end_time(), '`start` timestamp is the last timestamp of `series`',
+                         logger)
+            else:
+                raise_if(training_series[n.start].time_index()[0] == training_series.end_time(), '`start` timestamp '
+                         'is the last timestamp of `series`', logger)
+
+    def _backtest_model_specific_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
+        """Method to be overriden in subclass for model specific sanity checks"""
```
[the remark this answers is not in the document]
pennfranc: But then we need to provide an implementation for every forecasting model, even though currently most models don't have one. Do you think we should enforce it and add one to every model?

darts/models/forecasting_model.py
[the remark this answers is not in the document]
pennfranc: > Instead of doing `.time_index()[0]` you should be able to call `.start_time()`.
> Also, shouldn't we replace the `==` by `>=`, and add a preliminary check that (when it's an int), it is >= 0.

done

darts/models/forecasting_model.py
pennfranc: Are you referring to the data types of the function parameters? If yes, do you think for all parameters that are shared they should be the same? Because the reason for why `forecast_horizon` is optional in this case is that specifying it will lead to a expanding window evaluation, whereas leaving it set as `None` will require either the validation series arguments or the `use_fitted_values` parameter to be passed. That's why we decided not to set it to a default value of 1 here. But maybe you're referring to something else completely?

darts/models/forecasting_model.py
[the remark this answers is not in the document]
pennfranc: > We could also add an additional check that `start + forecast_horizon` is in the series if `trim_to_series` is False.

I assume what you meant to say is if `trim_to_series` is **True**', since otherwise we can have points in our predicted series that go beyond the input series time index, right?

darts/models/forecasting_model.py
pennfranc: > Maybe mention for what kinds of models this one applies.

Done. Could you check if the new parameter documentation for `target_series` looks good to you?

darts/models/forecasting_model.py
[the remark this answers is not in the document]
hrzn: Ok, let's leave it as is for now.

darts/models/forecasting_model.py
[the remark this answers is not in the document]
hrzn: Right, correct

darts/models/forecasting_model.py
hrzn: lgtm, thanks!

darts/models/forecasting_model.py
hrzn: Sorry, my comment was not super clear :)
I only meant that for the `forecast_horizon` parameter, it would be nice to support `float`, `int` and `pd.Timestamp` types (when it's not `None`), so that this parameter is treated the same as in the `backtest` method.

darts/models/forecasting_model.py
hrzn: I agree, I like your suggestion. Do you think we can do this change (in a consistent way, wherever needed across the codebase) in this PR, or do you prefer to wait for a next PR?

hrzn (approved):
Nice improvement :)

darts/models/forecasting_model.py
pennfranc: I think you're talking about the `start` parameter of the `backtest` method, since `forecast_horizon` is also just expected to be an integer in `backtest`. The thing is, `gridsearch` does not take `start` as a parameter right now, instead it takes `num_predictions`, from which we derive start. Do you suggest we replace `num_predictions` by `start` and treat it like we do in `backtest`?

darts/models/forecasting_model.py
pennfranc: Let's do it in a separate one!

darts/models/forecasting_model.py
hrzn: Right, sorry! You are right I mixed up the two parameters, my bad. It would be nice though to find a way to make the two signatures more compatible, and replace `num_predictions` by `start` like you suggest.

pennfranc: Done. We could make the `RegressionModel.backtest` function conform more as well by introducing the same `start` data types as for `ForecastingModel`. But this requires a bit more work since we need to change the checks, and we will probably want to reuse code from `ForecastingModel` to avoid duplication.

PR #111, opened 2020-06-22
feature/multivariate - step 4

pennfranc:
### Summary
- Adds multivariate functionality to backtesting functions.
- Further improves backtest_regression by introducing new parameters.

PR #126, opened 2020-07-07
Fix build failure

endrjuskr:
Fixes build failure.

### Summary

Notebook was failing due to incorrect path. It assumed that we are in root folder.


endrjuskr:
Already solved with #111 :)

PR #127, opened 2020-07-07
Fix/gradle

guillaumeraille:
- change workingdir for notebook in CI to darts root
- update every notebook to use that path
- update CI coverage run to not include `tests` and `__init__.py`

 ## Summary


examples/RNN-examples.ipynb
```diff
@@ -2,7 +2,16 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
```
endrjuskr, line 5: Please do not save it, because it will create a lot of diffs in the future. It is not idempotent

Kostiiii:
In docker after you go to examples directory and open notebook the notebook current directory is `examples`. So I feel like the checks and workdir of it should stay in `examples`.

examples/RNN-examples.ipynb
guillaumeraille: Yes I know it's a lot of diffs, do you know how to avoid that ?

guillaumeraille:
> In docker after you go to examples directory and open notebook the notebook current directory is `examples`. So I feel like the checks and workdir of it should stay in `examples`.

I am not sure to follow, you mean when you use the docker image and run `jupyter notebook` from within the `examples` folder within the docker container ?

examples/RNN-examples.ipynb
endrjuskr: If I needed to fix some issues in a notebook, I run it manually, then I committed and after that check using papermill if it works. Worth to check https://papermill.readthedocs.io/en/latest/usage-cli.html. Particularly `--no-request-save-on-cell-execute`. IMO we use papermill for nonstandard case :)

Kostiiii:
> > In docker after you go to examples directory and open notebook the notebook current directory is `examples`. So I feel like the checks and workdir of it should stay in `examples`.
> 
> I am not sure to follow, you mean when you use the docker image and run `jupyter notebook` from within the `examples` folder within the docker container ?

I run `./gradlew docker && ./gradlew dockerRun` then open link to jupyter, then from root folder I go to examples folder and open some notebook. Running `%pwd` inside it shows that workdir is set to `examples`.

guillaumeraille:
> > > In docker after you go to examples directory and open notebook the notebook current directory is `examples`. So I feel like the checks and workdir of it should stay in `examples`.
> > 
> > 
> > I am not sure to follow, you mean when you use the docker image and run `jupyter notebook` from within the `examples` folder within the docker container ?
> 
> I run `./gradlew docker && ./gradlew dockerRun` then open link to jupyter, then from root folder I go to examples folder and open some notebook. Running `%pwd` inside it shows that workdir is set to `examples`.

I was trying to find a fix for it without success.
I also tried to change the path to use the filename directory in VSCode but the problem is that then it can't find darts in the path anymore when the path is set to `examples`. I believe living the working dir in the root folder is better as we can change the code and instantly inspect changes in cells calling the corresponding code

guillaumeraille:
to add some clarification here is my problem @Kostiiii:
image

because of this path used inside the jupyter notebook in docker, when you actually make a change you have to build the package again to see some changes reflected in the notebook

Kostiiii:
> to add some clarification here is my problem @Kostiiii:
> image
> 
> because of this path used inside the jupyter notebook in docker, when you actually make a change you have to build the package again to see some changes reflected in the notebook

Okey, now I understand what is your concern. Hmm I always though that notebooks are supposed to use installed version of darts not local code.

guillaumeraille:
> > to add some clarification here is my problem @Kostiiii:
> > image
> > because of this path used inside the jupyter notebook in docker, when you actually make a change you have to build the package again to see some changes reflected in the notebook
> 
> Okey, now I understand what is your concern. Hmm I always though that notebooks are supposed to use installed version of darts not local code.

if that's what we want then we can discard this changes and I will create my own notebooks locally :)

TheMP:
> > > to add some clarification here is my problem @Kostiiii:
> > > image
> > > because of this path used inside the jupyter notebook in docker, when you actually make a change you have to build the package again to see some changes reflected in the notebook
> > 
> > 
> > Okey, now I understand what is your concern. Hmm I always though that notebooks are supposed to use installed version of darts not local code.
> 
> if that's what we want then we can discard this changes and I will create my own notebooks locally :)

@guillaumeraille Bit late to the party but would simply adding darts to the path in notebook help? https://www.johnny-lin.com/cdat_tips/tips_pylang/path.html?

guillaumeraille:
> > > > to add some clarification here is my problem @Kostiiii:
> > > > image
> > > > because of this path used inside the jupyter notebook in docker, when you actually make a change you have to build the package again to see some changes reflected in the notebook
> > > 
> > > 
> > > Okey, now I understand what is your concern. Hmm I always though that notebooks are supposed to use installed version of darts not local code.
> > 
> > 
> > if that's what we want then we can discard this changes and I will create my own notebooks locally :)
> 
> @guillaumeraille Bit late to the party but would simply adding darts to the path in notebook help? https://www.johnny-lin.com/cdat_tips/tips_pylang/path.html?

it would but then it would mean adding a code cell that should be run only locally. Actually, I think even adding something to the path that is not used would also work so running it in docker would be useless but fine. I can make the changes and try thanks

examples/FFT-examples.ipynb
```diff
@@ -8,6 +8,21 @@
...
+   "outputs": [],
+   "source": [
+    "# fix pythonpath when working locally\n",
+    "from os.path import dirname, basename\n",
+    "import sys\n",
+    "\n",
+    "cwd = os.getcwd()\n",
+    "if basename(cwd) == 'examples':\n",
+    "    sys.path.insert(0, dirname(cwd))"
+   ]
```
Kostiiii, line 24: Maybe it could be put as a separate function common to all notebooks (in separate file maybe). Plus setting this path could be configurable by env variable for examples 'local-develop' which you could set locally. What do you think about it?

PR #130, opened 2020-07-08
Fix/cleaner code for TorchForecastingModel

pennfranc:
### Summary

- Created some subroutines for the biggest methods of `TorchForecastingModel` to make the code more structured.
- Added an additional test case to `test_RNN.py`
- Added missing check to `backtesting`.


guillaumeraille:
The PR might need to be rebased ? It is a lot of commits (https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request)

darts/models/torch_forecasting_model.py
```diff
@@ -416,6 +379,9 @@ def first_prediction_index(self) -> int:
         """
         return 0
 
+    def create_dataset(self, series):
+        return _TimeSeriesSequentialDataset(series, self.input_length, self.output_length, self.target_indices)
```
guillaumeraille, line 383: should we add docstring to the method ?

darts/models/torch_forecasting_model.py
```diff
@@ -537,8 +503,67 @@ def _save_model(self,
                 for chkpt in checklist[:-1]:
                     os.remove(chkpt)
 
-    def create_dataset(self, series):
-        return _TimeSeriesSequentialDataset(series, self.input_length, self.output_length, self.target_indices)
+    def _prepare_validation_data(self, val_series):
+        if val_series is not None:
```
guillaumeraille, line 507: should we add docstring to private method as well ?

darts/tests/test_RNN.py
```diff
@@ -59,6 +59,13 @@ def test_fit(self):
         pred4 = model3.predict(n=1)
         self.assertEqual(len(pred4), 1)
 
+        # test validation series input
+        model3.fit(self.series[:60], val_series=self.series[60:])
+        pred4 = model3.predict(n=6)
+        self.assertEqual(len(pred4), 6)
```
guillaumeraille, line 65: I believe test should be short and contain as little assertion as possible (usually 1) this allows to quickly see directly from the test runner what is failing and act accordingly. We haven't been following this practice in the past but it seems that this test could be split up and the method name could be more explicit, what do you think ?

darts/models/torch_forecasting_model.py
pennfranc: I asked @TheMP about this as well. I think in general it's not necessary for private methods, especially in this case where the name is quite descriptive.

darts/tests/test_RNN.py
pennfranc: That's probably a good idea, but I think this should be done in a separate PR, as it would require a complete refactoring of the current test cases.

pennfranc:
> The PR might need to be rebased ? It is a lot of commits (https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request)

@guillaumeraille That's true, but does it still matter when we squash merge?

darts/models/torch_forecasting_model.py
pennfranc: Here I did not include it since we're overriding a method from the parent class. Do you think I should still include it?

guillaumeraille:
> > The PR might need to be rebased ? It is a lot of commits (https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request)
> 
> @guillaumeraille That's true, but does it still matter when we squash merge?

I guess you are right, it doesn't really mater since it will be squashed but in that case we should also be careful of the name of the description of the commit as it will automatically add a bullet list with 190 elements which might be hard to track should we revert something in the futur

darts/models/torch_forecasting_model.py
guillaumeraille: from (PEP-257)[https://www.python.org/dev/peps/pep-0257/]:

hrzn: are you talking about `create_dataset()`? Should it be made private instead? I'm also thinking that the whole `TorchForecastingModel` should be made private.

pennfranc: create_dataset is private now

PR #132, opened 2020-07-09
Proposal for preprocessing pipeline

Kostiiii:
Follow up on our discussion on preprocessing.

BaseTransformer stay the same as was shown - but of course please raise again any concerns/improvement.
transform in Timeseries also - again please raise any concerns/improvement.

New class Pipeline for managing transforms and validations. Pipeline is intended only for one object of timeseries - this way when we call inverse we are sure that state of transformer is for correct series. Pipeline can be traversed forward and maybe backward (sometimes transform may not have inverse we may have to be extra careful about it)

Interface of pipeline:
```
add_transformer - function that add transformer to pipeline, order of transformers are determined as an order that transformers were added,
add_validator - function that adds validator for each step
resolve - function that resolves pipeline (traverse path of transformations to the end either forward or backward
resolve_n_steps - same as resolve but traverse n step
set_data/get_data -setter/getter for data - getter would return data that we have at our current step of transformations
```

I left out implementations details as I feel like they are more appropriate to do after proposal in some form will be accepted.

Ping @TheMP @hrzn @endrjuskr @pennfranc @guillaumeraille @radujica @Droxef. I hope I included all interested.

### DIFF
--- a/darts/preprocessing/base_transformer.py
+++ b/darts/preprocessing/base_transformer.py
@@ -1,0 +1,32 @@
+
+from typing import TypeVar, Generic, Callable
+from darts.logging import raise_if_not, get_logger
+
+logger = get_logger(__name__)
+T = TypeVar('T')
+
+
+class BaseTransformer(Generic[T]):
+    def __init__(self):
+        self.validators = [
+            lambda x: True
+        ]
+
+    def add_validator(self, validator: Callable[[T], bool]):
+        self.validators.append(validator)
+        return self
+
+    def validate(self, data: T) -> bool:
+        for validator in self.validators:
+            raise_if_not(validator(data), "Validation failed", logger)
+        return True
+
+    def transform(self, data: T, *args, **kwargs) -> T:
+        return data
+
+    def inverse_transform(self, data: T, *args, **kwargs) -> T:
+        return data
+
+    def __call__(self, data: T, *args, **kwargs):
+        self.validate(data)
+        return self.transform(data, *args, **kwargs)
--- a/darts/preprocessing/pipeline.py
+++ b/darts/preprocessing/pipeline.py
@@ -1,0 +1,30 @@
+
+from typing import TypeVar, Generic
+from darts.preprocessing.base_transformer import BaseTransformer
+
+logger = get_logger(__name__)
+T = TypeVar('T')
+
+
+class Pipeline(Generic[T]):
+    def __init__(self):
+        pass
+
+    def add_transformer(self, transformer: BaseTransformer[T]) -> 'Pipeline':
+        pass
+
+    # ? validator as separate class or as just callable ?
+    def add_validator(self, validator) -> 'Pipeline':
+        pass
+
+    def resolve(self, inverse: bool = False) -> 'Pipeline':
+        pass
+
+    def resolve_n_steps(self, n: int) -> 'Pipeline':
+        pass
+
+    def set_data(data: T) -> 'Pipeline':
+        pass
+
+    def get_data(self) -> T:
+        pass
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -967,6 +967,9 @@ class TimeSeries:
         inferred_frequency = observed_frequencies.pop()
         return series.resample(inferred_frequency).asfreq(fill_value=None)
 
+    def transform(self, f: Callable[['TimeSeries'], 'TimeSeries'], *args, **kwargs) -> 'TimeSeries':
+        return f(self, *args, **kwargs)
+
     """
     Definition of some useful statistical methods.
     These methods rely on the Pandas implementation.
--- a/preprocess.md
+++ b/preprocess.md
@@ -110,9 +110,11 @@ pipeline = (Pipeline()
 
 pipeline = pipeline.resolve()
 train, val = pipeline.get_data()
-original_ts = pipeline.resolve(reverse=True).get_data()
+original_ts = pipeline.resolve(reverse=True).get_data()[0]
 ```
 
 ## Potential problems
-What about transformers without inverse - maybe storing input and returning it would work but then we have to say that trasformer works only for one list of input.
-Maybe overcomplicated
+* What about transformers without inverse - maybe storing input and returning it would work but then we have to say that trasformer works only for one list of input.
+* Maybe overcomplicated
+
+* In approach with lists of timeseries it really isn't (mathematically) `transform` anymore.
--- a/preprocess.md
+++ b/preprocess.md
@@ -1,0 +1,118 @@
+## Pipeline
+
+```python
+class Pipeline(Generic[T]):
+    def __init__(self):
+        pass
+    
+    # change from last time - passing args and kwargs to transformer
+    def add_transformer(self, transformer: BaseTransformer[T], *transformer_args, **transformer_kwargs) -> 'Pipeline':
+        pass
+
+    # ? validator as separate class or as just callable ?
+    def add_validator(self, validator) -> 'Pipeline':
+        pass
+
+    def resolve(self, inverse: bool = False) -> 'Pipeline':
+        pass
+
+    def resolve_n_steps(self, n: int) -> 'Pipeline':
+        pass
+
+    def set_data(data: T) -> 'Pipeline':
+        pass
+
+    def get_data(self) -> T:
+        pass
+```
+
+```python
+ts # timeseries we are working with
+transformers = [transformer_1, ..., transformer_n] # objects deriving from BaseTransformer
+common_validators = [all_positive, ...]
+```
+## init pipeline either like that
+```python
+pipeline = Pipeline(transformers, common_validators)
+```
+## or like that
+```python
+pipeline = Pipeline()
+for t in transformes:
+    pipeline = pipeline.add_transformer(t)
+for v in common_validators:
+    pipeline = pipeline.add_validator(v)
+```
+
+## use like that
+
+```python
+pipeline.set_data(ts)
+pipeline = pipeline.resolve_n_steps(1) # pipeline traversed using first trnasformation
+ts_transformed = pipeline.get_data() # ts after 1st transform
+pipeline = pipeline.resolve() # pipeline traversed to end (all transformations applied)
+pipeline = pipeline.resolve(inverse=True) # pipeline reverted to begining
+```
+
+## But
+Common pattern is first to split data, train/fit transformer on train part of data - apply all to train, test and val data.
+
+So maybe instead of transformers and pipeline working on
+`timeseries -> timeseries` making it `List[timeseries] -> List[timeseries]` would work better.
+
+
+```Python
+
+class BaseTransformer(Generic[T]):
+    def __init__(self, validators):
+        pass
+
+    def add_validator(self, validator: Callable[[T], bool]):
+        pass
+
+    def validate(self, data: List[T]) -> bool:
+        pass
+
+    def transform(self, data: T, *args, **kwargs) -> List[T]:
+        pass
+
+    def inverse_transform(self, data: List[T], *args, **kwargs) -> List[T]:
+        pass
+
+    def fit(self, data: T):
+        pass
+
+    # inverse whether to apply inverse transformation
+    # fit_on: index of data to fit on, -1 to skip fitting
+    def __call__(self, data: List[T], *args, inverse: bool = False, fit_on: int = -1, **kwargs) -> List[T]:
+        pass
+```
+
+```python
+# Changes to Pipeline
+class Pipeline(Generic[t])
+    def set_data(data: List[T]) -> 'Pipeline':
+        pass
+
+    def get_data(self) -> List[T]:
+        pass
+```
+
+then we could pass timeseries splitted in, next transformer could be fitted on train data, applied to all.
+Example
+```python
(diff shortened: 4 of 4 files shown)


darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def transform(self, data: T, *args, **kwargs) -> T:
+        return data
+
+    def inverse_transform(self, data: T, *args, **kwargs) -> T:
+        return data
+
+    def __call__(self, data: T, *args, **kwargs):
+        self.validate(data)
+        return self.transform(data, *args, **kwargs)
```
guillaumeraille, line 32: In the end how do we handle the `inverse_transform` ? In particular how do we handle the case where validation might be different before `transform` and `inverse_transform` ?

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,32 @@
...
+        self.validators = [
+            lambda x: True
+        ]
+
+    def add_validator(self, validator: Callable[[T], bool]):
+        self.validators.append(validator)
+        return self
+
+    def validate(self, data: T) -> bool:
+        for validator in self.validators:
```
hrzn, line 20: If you want you could also go fancy like `raise_if_not(map(lambda v: v(data), self.validators).all(), ...)`

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,32 @@
...
+T = TypeVar('T')
+
+
+class BaseTransformer(Generic[T]):
+    def __init__(self):
+        self.validators = [
+            lambda x: True
+        ]
+
+    def add_validator(self, validator: Callable[[T], bool]):
```
hrzn, line 15: Would it be possible to consider specifying the validators when building the class (as arguments of `__init__()`? This way it would be easier to see Transformers as immutable things that are not modified once built :)

darts/preprocessing/base_transformer.py
Kostiiii: Sorry again for including implementations detail to this proposal they might be misleading. I would say passing `inverse=True` to call actually look pretty nice, wdyt? We could set validators for inverse and all logic would be handled in `__call__`.

darts/preprocessing/base_transformer.py
Kostiiii: Thats fancy :P, hmm I probably shouldn't include implementations details in this proposal but just interface.

darts/preprocessing/base_transformer.py
Kostiiii: Of course, its not hard to do and if we agree on this (making transformer immutable) then add_validator would return new transformer. +1 for specifying validators in init.

Kostiiii:
> @Kostiiii It would help me for reviewing if you were giving a couple of examples of how transformers and pipelines would be used in practice (in cases with inverse transforms as well).

Good point, I will prepare some and ping you when they are ready.

TheMP:
> > @Kostiiii It would help me for reviewing if you were giving a couple of examples of how transformers and pipelines would be used in practice (in cases with inverse transforms as well).
> 
> Good point, I will prepare some and ping you when they are ready.

Very good point, how about including this in one of the jupyter notebooks?

Kostiiii:
> > > @Kostiiii It would help me for reviewing if you were giving a couple of examples of how transformers and pipelines would be used in practice (in cases with inverse transforms as well).
> > 
> > 
> > Good point, I will prepare some and ping you when they are ready.
> 
> Very good point, how about including this in one of the jupyter notebooks?

Okay so in the end there should be working example right? with implementation? Or just usage of pipeline/transformer interfaces?

hrzn:
> > > > @Kostiiii It would help me for reviewing if you were giving a couple of examples of how transformers and pipelines would be used in practice (in cases with inverse transforms as well).
> > > 
> > > 
> > > Good point, I will prepare some and ping you when they are ready.
> > 
> > 
> > Very good point, how about including this in one of the jupyter notebooks?
> 
> Okay so in the end there should be working example right? with implementation? Or just usage of pipeline/transformer interfaces?

I think at this stage some examples would be enough. Then, once it's agreed and implemented, we can perhaps have a Jupyter notebook for this? wdyt @TheMP ?

darts/preprocessing/base_transformer.py
hrzn: you're right ;)

Kostiiii:
> > > > > @Kostiiii It would help me for reviewing if you were giving a couple of examples of how transformers and pipelines would be used in practice (in cases with inverse transforms as well).
> > > > 
> > > > 
> > > > Good point, I will prepare some and ping you when they are ready.
> > > 
> > > 
> > > Very good point, how about including this in one of the jupyter notebooks?
> > 
> > 
> > Okay so in the end there should be working example right? with implementation? Or just usage of pipeline/transformer interfaces?
> 
> I think at this stage some examples would be enough. Then, once it's agreed and implemented, we can perhaps have a Jupyter notebook for this? wdyt @TheMP ?

Hey all, examples (and more proposals :) ) are in `preprocess.md`
Ping @TheMP @hrzn @endrjuskr @pennfranc @guillaumeraille @radujica @Droxef

hrzn:
A few comments:
* I would keep only this way of building a Pipeline: `pipeline = Pipeline(transformers, common_validators)`, and discard the way using `add_transformers()` etc. This way `Pipeline` can be immutable and only defined by the transformers they contain (both pipelines and transformers should be immutable).
* Why do we need validators? I would let each transformer perform their own validation and let them have the responsibility for validation (not the users). But maybe I missed a point?
* I would remove `Pipeline.resolve_n_steps()` and only keep `Pipeline.resolve()`. I think users who only want to do the first n steps should simply build another smaller pipeline or call something like `my_pipeline[0:n].resolve()`.
* I would actually rename `Pipeline.resolve()` into `transform()`, and impose that Pipelines consist in a sequence of `Transformer`'s (so they have `fit` and `transform` methods).
* We can be inspired by sklearn: https://scikit-learn.org/stable/modules/compose.html#pipeline
* `Pipeline.inverse_transform()` could return an exception if not all of its `Transformers` have and `inverse_transforms` method.

@Kostiiii let me know if these considerations make sense? I probably missed a few points and some usage patterns ;) However I think we should strive for simplicity and perhaps only start with simplistic Transformers and Pipelines (and maybe add complexity later on).

Kostiiii:
* 👍 
* There was a talk about validators for each step (for example we always want values to be non negative etc)
* 👍 + maybe bunch of nice to have helpers like slicing support, iters?  
* 👍 what about transformers that will be fitted only on train data? does it mean if we want fitted transformer we first need to fit it on train and later set same transformer for different pipeline for val set?
* looks neat
* makes 100% sense

By simplistic you mean operating on single timeseries, right? @hrzn

hrzn:
> * 👍
> * There was a talk about validators for each step (for example we always want values to be non negative etc)
> * 👍 + maybe bunch of nice to have helpers like slicing support, iters?
> * 👍 what about transformers that will be fitted only on train data? does it mean if we want fitted transformer we first need to fit it on train and later set same transformer for different pipeline for val set?
> * looks neat
> * makes 100% sense
> 
> By simplistic you mean operating on single timeseries, right? @hrzn

* 
* validators: I would vote for leaving them out for now. Each transformer can make its own checks and raise proper exceptions (e.g. when some values are zero etc).
* +1 for adding slicing/indexing on pipelines, but it doesn't have to be in the first implementation
* I think we can have `fit()` and `transform()` methods on Pipeline (and maybe even `predict()`), similar to sklearn. This way, you can do
```
my_pipeline.fit(train_series)
my_pipeline.transform(test_series)
```
does it make sense?
* By simplistic I mean single time series yes, but also removing everything we are not too sure would be useful, like validators, and keeping all logic and signatures as simple as possible for the users.

Droxef:
* If the `Pipeline` should be immutable, I don't think the data belong to this class. The approach `fit`/`transform` seems more accurate. So we could `fit`and `transform` a time series, and `inverse_transform` the prediction of a model easily.
* I may be wrong, but I don't think it is necessary to accept the different splits. validation and test don't need to be transformed, if they are compared with the `inverse_transform` of train. And I am not sure how much the inverse splitter is interesting. Personnaly, I would keep it outside the `Pipeline`.
* for the validators, it seems that they are more transformer-specific more than pipeline-specific, so I agree with @hrzn 
* If we only keep transformers in the pipeline, `predict` has no meaning?

Kostiiii:
> > * 👍
> > * There was a talk about validators for each step (for example we always want values to be non negative etc)
> > * 👍 + maybe bunch of nice to have helpers like slicing support, iters?
> > * 👍 what about transformers that will be fitted only on train data? does it mean if we want fitted transformer we first need to fit it on train and later set same transformer for different pipeline for val set?
> > * looks neat
> > * makes 100% sense
> > 
> > By simplistic you mean operating on single timeseries, right? @hrzn
> 
> * validators: I would vote for leaving them out for now. Each transformer can make its own checks and raise proper exceptions (e.g. when some values are zero etc).
> * +1 for adding slicing/indexing on pipelines, but it doesn't have to be in the first implementation
> * I think we can have `fit()` and `transform()` methods on Pipeline (and maybe even `predict()`), similar to sklearn. This way, you can do
> 
> ```
> my_pipeline.fit(train_series)
> my_pipeline.transform(test_series)
> ```
> 
> does it make sense?
> 
> * By simplistic I mean single time series yes, but also removing everything we are not too sure would be useful, like validators, and keeping all logic and signatures as simple as possible for the users.

This makes sense, just one question for clarification.
Does `pipeline.fit(train)` would do something like this 
```
ts = train
for transformers in pipeline.transforems:
    transformer.fit(ts)
    ts = transformer.transform(ts)
```
?

Kostiiii:
@Droxef I guess predict could be run on last object in pipeline, so the last transformer should have predict function in such case. 1st point makes sense to me and looks definitely better. 2nd I am not sure about that but you use darts more than me so you have better overview what could be used/helpful and what would be unnecessary/useless. For now anyway we will 99% go with simplistic approach so this is off the table anyway.

grll:
> > > * 👍
> > > * There was a talk about validators for each step (for example we always want values to be non negative etc)
> > > * 👍 + maybe bunch of nice to have helpers like slicing support, iters?
> > > * 👍 what about transformers that will be fitted only on train data? does it mean if we want fitted transformer we first need to fit it on train and later set same transformer for different pipeline for val set?
> > > * looks neat
> > > * makes 100% sense
> > > 
> > > By simplistic you mean operating on single timeseries, right? @hrzn
> > 
> > 
> > 
> > * validators: I would vote for leaving them out for now. Each transformer can make its own checks and raise proper exceptions (e.g. when some values are zero etc).
> > * +1 for adding slicing/indexing on pipelines, but it doesn't have to be in the first implementation
> > * I think we can have `fit()` and `transform()` methods on Pipeline (and maybe even `predict()`), similar to sklearn. This way, you can do
> > 
> > ```
> > my_pipeline.fit(train_series)
> > my_pipeline.transform(test_series)
> > ```
> > 
> > 
> > does it make sense?
> > 
> > * By simplistic I mean single time series yes, but also removing everything we are not too sure would be useful, like validators, and keeping all logic and signatures as simple as possible for the users.
> 
> This makes sense, just one question for clarification.
> Does `pipeline.fit(train)` would do something like this
> 
> ```
> ts = train
> for transformers in pipeline.transforems:
>     transformer.fit(ts)
>     ts = transformer.transform(ts)
> ```
> 
> ?

I am not sure to follow maybe I am missing something but how can the transformer be immutable and have a fit method with side effect to it like in sklearn?

hrzn:
> > > * 👍
> > > * There was a talk about validators for each step (for example we always want values to be non negative etc)
> > > * 👍 + maybe bunch of nice to have helpers like slicing support, iters?
> > > * 👍 what about transformers that will be fitted only on train data? does it mean if we want fitted transformer we first need to fit it on train and later set same transformer for different pipeline for val set?
> > > * looks neat
> > > * makes 100% sense
> > > 
> > > By simplistic you mean operating on single timeseries, right? @hrzn
> > 
> > 
> > 
> > * validators: I would vote for leaving them out for now. Each transformer can make its own checks and raise proper exceptions (e.g. when some values are zero etc).
> > * +1 for adding slicing/indexing on pipelines, but it doesn't have to be in the first implementation
> > * I think we can have `fit()` and `transform()` methods on Pipeline (and maybe even `predict()`), similar to sklearn. This way, you can do
> > 
> > ```
> > my_pipeline.fit(train_series)
> > my_pipeline.transform(test_series)
> > ```
> > 
> > 
> > does it make sense?
> > 
> > * By simplistic I mean single time series yes, but also removing everything we are not too sure would be useful, like validators, and keeping all logic and signatures as simple as possible for the users.
> 
> This makes sense, just one question for clarification.
> Does `pipeline.fit(train)` would do something like this
> 
> ```
> ts = train
> for transformers in pipeline.transforems:
>     transformer.fit(ts)
>     ts = transformer.transform(ts)
> ```
> 
> ?

I think so yes

hrzn:
@grll Indeed you are right, the transformers wouldn't be immutable but store some state (and so the pipeline too). I was too quick in mentioning immutability :/

Kostiiii:
I think we have enough for start. I will keep this open for a little while longer if anybody wants to bring some point etc but soon I will open pr with real stuff.

PR #138, opened 2020-07-13
Examples/m4 competition

Droxef:
M4 competition Benchmark

Summary
Proposition for adding M4 competition benchmark to examples. Add scripts to download dataset, create the timeseries and reproduce the results. Add a notebook to put together the different parts and and visualize some results.


darts/models/theta.py
```diff
@@ -95,7 +95,7 @@ def fit(self, series: TimeSeries, component_index: Optional[int] = None):
             new_ts = remove_from_series(ts, self.seasonality, model=self.season_mode)
 
         # SES part of the decomposition.
-        self.model = hw.SimpleExpSmoothing(new_ts.values()).fit()
+        self.model = hw.SimpleExpSmoothing(new_ts.values()).fit(initial_level=0.2)
```
TheMP, line 98: Do we want to change this in this PR?

examples/M4_competition/evaluate_ensembling.py
```diff
@@ -0,0 +1,287 @@
...
+                    seasonOut = season[-m:].shift(m)
+                    seasonOut = seasonOut.append_values(seasonOut.values())[:len(test)]
+            # model selection
+            naiveSeason = NaiveSeasonal(K=m)
+            naive2 = NaiveSeasonal(K=1)
+            ses = ExponentialSmoothing(trend=None, seasonal=None, seasonal_periods=m)
+            holt = ExponentialSmoothing(seasonal=None, damped=False, trend='additive', seasonal_periods=m)
+            damp = ExponentialSmoothing(seasonal=None, damped=True, trend='additive', seasonal_periods=m)
+
+            # season_mode = ["additive", "multiplicative"]
```
TheMP, line 157: Should this be here if it's commented out?

examples/M4_competition/evaluate_theta_methods.py
```diff
@@ -0,0 +1,133 @@
...
+        forecast = forecast - shiftdata
+    forecast = forecast * seasonality
+    if (forecast.univariate_values() < 0).any():
+        indices = seasonality.time_index()[forecast < 0]
+        forecast = forecast.update(indices, np.zeros(len(indices)), inplace=True)
+    return forecast
+
+
+def train_4theta(ts, n):
+    # season_mode = ["additive", "multiplicative"]
```
TheMP, line 63: Should this also be here?

examples/M4_competition/FourTheta.py
```diff
@@ -0,0 +1,342 @@
...
+        # fitted_values = TimeSeries.from_times_and_values(train_series.time_index(), model.fitted_values)
+        error = metric(model.fitted_values, train_series.univariate_values())
+        if error < min_error:
+            min_error = error
+            best_param_combination = param_combination_dict
+    logger.info('Chosen parameters: ' + str(best_param_combination))
+    return model_class(**best_param_combination)
+
+
+class FourTheta(UnivariateForecastingModel):
```
TheMP, line 53: This class is already included in the previous PR and merged, right? Or is it some different implementation?

darts/models/theta.py
Droxef: Sorry, it was something that I forgot to clean. 
Same for the comments after that

examples/M4_competition/FourTheta.py
Droxef: The difference is that it uses a modified version of gridsearch that do not create a TimeSeries.
Thus the execution time of the notebook drops from 1h to 5-10 minutes

hrzn: I know it is a pity that using TimeSeries increases the gridsearch time, but if we release this we cannot have a separate implementation, and we should really use the implementation in `darts/models`, even if it takes much longer to run.

PR #139, opened 2020-07-13
hotfix/manual-frequency-override (#patch)

pennfranc:
### Summary
'freq' parameter of constructor now leads to overriding the automatic frequency detection in case of missing dates when passed

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -33,7 +33,7 @@ class TimeSeries:
         df
             The actual time series, as a pandas DataFrame with a proper time index.
         freq
-            Optionally, a Pandas offset alias representing the frequency of the DataFrame. 
+            Optionally, a Pandas offset alias representing the frequency of the DataFrame.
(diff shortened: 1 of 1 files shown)


darts/timeseries.py
```diff
@@ -934,39 +937,46 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
         return TimeSeries(series, self.freq())
 
     @staticmethod
-    def _fill_missing_dates(series: pd.DataFrame) -> pd.DataFrame:
+    def _fill_missing_dates(series: pd.DataFrame, freq: Optional[str] = None) -> pd.DataFrame:
```
TheMP, line 940: How would that work in case of business days freq etc.? 

Since we already had questions about freq I think it either deserves an Enum with the list of options or at least link to pandas list of freq's to be clear what is supported.

pennfranc: In case of business day frequency, you can simply pass `freq='B'` to the constructor to make sure that's the one it uses. Regarding the pandas list of offset aliases, I linked to it in the docstring of the constructor.

PR #150, opened 2020-07-16
Features/indexing

grll:
Fixes #DARTS-95.

### Summary

Add better indexing support and column names to TimeSeries.


darts/timeseries.py
```diff
@@ -20,6 +20,7 @@
 class TimeSeries:
     def __init__(self,
                  df: pd.DataFrame,
+                 columns: Optional[Union[List[str], str]] = None,
```
Kostiiii, line 23: Super cool feature :) Could you put `columns` as last keyword argument? That way when this will be merged it won't break code for peoples that are using `freq` as 3rd argument or `fill_missing_dates` as 4th argument

grll: I already fixed it in many places in the library but true I could also make it as 4th argument for final users

darts/timeseries.py
```diff
@@ -31,6 +32,9 @@ def __init__(self,
         ----------
         df
             The actual time series, as a pandas DataFrame with a proper time index.
+        columns
```
pennfranc, line 35: Minor nitpick: I think the name `column_names` instead of `columns` might be a bit more descriptive of what this argument represents.

grll: while it is true that `column_names` is more descriptive I don't know if we want to deviate from pandas `columns` syntax ?

pennfranc: Ah, good point!

darts/timeseries.py
```diff
@@ -492,12 +513,12 @@ def from_series(pd_series: pd.Series,
...
+        return TimeSeries(pd.DataFrame(pd_series), freq, fill_missing_dates, column)
 
     @staticmethod
     def from_dataframe(df: pd.DataFrame,
-                       time_col: Optional[str],
-                       value_cols: Union[List[str], str],
+                       time_col: Optional[str] = None,
+                       value_cols: Optional[Union[List[str], str]] = None,
                        freq: Optional[str] = None,
                        fill_missing_dates: Optional[bool] = True) -> 'TimeSeries':
```
grll, line 523: As a bonus `from_dataframe` allow now to take only a single parameter (the dataframe) and infer everything (indices = time values) and columns become the multivariate values

darts/timeseries.py
```diff
@@ -31,6 +32,9 @@ def __init__(self,
         ----------
         df
             The actual time series, as a pandas DataFrame with a proper time index.
+        columns
+            The columns name in the same order as in the underlying DataFrame. If columns name are not unique they will
```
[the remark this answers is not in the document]
grll: the default behavior is to infer the column names from the dataframe. That's what we get with the `from_dataframe` function but since TimeSeries are immutable, to preserve the column upon operations we need to pass the names as parameter during creation

darts/timeseries.py
```diff
@@ -37,13 +38,30 @@ def __init__(self,
...
+                return True
+            else:
+                return len(set(columns)) == len(columns)
+
+        if columns is not None and columns_are_unique(columns):
+            if isinstance(columns, str):
+                columns = [columns]
+            raise_if_not(len(columns) == df.shape[1], "`{}` columns specified when `{}` was expected from `df` "
+                                                      "shape".format(len(columns), df.shape[1]))
+            df.columns = columns
```
hrzn, line 62: Here and in line 64 we are changing the DataFrame provided by the user. I think we should not touch it, but rather do everything on the "internal" dataframe `df_`, which is a copy of `df` created on line 74.

darts/timeseries.py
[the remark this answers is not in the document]
hrzn: I don't understand this, because when creating the TimeSeries, we are anyway making an internal copy `df_` of the dataframe. So any change of the columns on the original dataframe should not impact the TimeSeries that is already built. However I might miss something?

darts/timeseries.py
```diff
@@ -929,8 +957,21 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
         raise_if_not(self.has_same_time_as(other), 'The two TimeSeries must have the same time index.', logger)
         raise_if_not(self._df.shape == other._df.shape, 'The two TimeSeries must have the same shape.', logger)
 
-        series = combine_fn(self._df, other.pd_dataframe())
-        return TimeSeries(series, self.freq())
+        df_a = self._df
+        df_b = other._df
```
hrzn, line 961: That's probably a little faster than calling `other.pd_dataframe()`, but we now rely on `combine_fn` having no side effect on the dataframes it receives in arguments. I think that's OK in this context, given that _combine_from_pd_ops() is private but worth keeping in mind..

darts/timeseries.py
```diff
@@ -929,8 +957,21 @@ def _combine_from_pd_ops(self, other: 'TimeSeries',
...
+                pd_serie_a = df_a.iloc[:, 0]
+                pd_serie_b = df_b.iloc[:, 0]
+                pd_serie = combine_fn(pd_serie_a, pd_serie_b)
+                series_df = pd_serie.to_frame()
+            # multivariate case
+            else:
+                raise_if(len(set(df_a.columns.to_list() + df_b.columns.columns.to_list())) != len(df_a.columns),
+                         "Column name in each TimeSeries must match one to one.")
+                series_df = combine_fn(df_a, df_b)
+        return TimeSeries(series_df, self.freq(), columns=series_df.columns)
```
hrzn, line 974: If both `df_a` and `df_b` are None that will throw a not nice error saying `series_df` doesn't exist.

hrzn: In addition, to respect types, you should pass `self.freq_str()` instead of `self.freq()`. I even wonder if that might not lead to errors, and now that I speak of it, I see we have this issue at several places in the code.

darts/timeseries.py
```diff
@@ -1159,38 +1200,57 @@ def __copy__(self, deep: bool = True):
...
+
+        `pd.DatetimeIndex` is also supported and will return the corresponding value(s) at the provided time indices.
+
+        .. warning::
+            slices use pandas convention of including both ends of the slice.
+
+        """
+        def use_iloc(key: Any) -> TimeSeries:
+            """return a new TimeSeries from a pd.DataFrame using iloc indexing."""
+            return TimeSeries.from_dataframe(self._df.iloc[key], freq=self.freq())
```
hrzn, line 1220: again `self.freq_str()`

darts/timeseries.py
```diff
@@ -1159,38 +1200,57 @@ def __copy__(self, deep: bool = True):
...
-            # Verify that values in item are really in index to avoid the creation of NaN values
-            if isinstance(item, (np.ndarray, pd.DatetimeIndex)):
-                check = np.array([elem in self.time_index() for elem in item])
-                if not np.all(check):
-                    raise_log(IndexError("None of {} in the index".format(item[~check])), logger)
-
-            return TimeSeries(self._df.loc[item, :], self.freq())
-        elif isinstance(item, str):
-            return self._df[[pd.Timestamp(item)]]
+    def __getitem__(self, key: Union[pd.DatetimeIndex, List[str], List[int], List[pd.Timestamp], str, int,
```
hrzn, line 1203: cool :)

darts/timeseries.py
grll: Ah indeed I agree

darts/timeseries.py
grll: 1) you are right and its even worst because it is also the case when either one or the other is None

2) I used the syntax previously here but you are right I can fix that as well

darts/timeseries.py
grll: btw relying on sort_index() to create a copy of the df might make it a bit unclear that it is a copy

darts/timeseries.py
pennfranc: Good point about the `freq` parameter. I actually started this trend of passing `self.freq()`, sorry!

grll: 1. I just removed the condition since already from type they should not be None. The error will then be when calling .shape if they happens to be None

2. Fixed it

darts/timeseries.py
grll: opinions @TheMP @pennfranc @Droxef ?

darts/timeseries.py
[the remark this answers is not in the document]
grll: no my bad, the new implementation just keeps the column parameter as a way to override this default behavior of using the underlying DataFrame columns

darts/timeseries.py
pennfranc: Yes actually this made me notice a problem we currently have I think. In line 62 you change the columns of the original data frame right? This means that the original data frame that the user provides gets changed.

grll: @pennfranc  You are right I modified this behavior in the latest commits, my question was more about using ?? 
`sort_index()` to return a new DF I was personally not sure it was returning a copy or a view in that case.

darts/timeseries.py
```diff
@@ -541,7 +567,8 @@ def from_dataframe(df: pd.DataFrame,
     def from_times_and_values(times: pd.DatetimeIndex,
                               values: np.ndarray,
                               freq: Optional[str] = None,
-                              fill_missing_dates: Optional[bool] = True) -> 'TimeSeries':
+                              fill_missing_dates: Optional[bool] = True,
+                              column: Optional[str] = None,) -> 'TimeSeries':
```
pennfranc, line 571: This setup with the `column` parameter assumes that we are creating a univariate time series, right? But technically we can also use this function to create a multivariate time series by passing a 2- 
dimensional numpy array as `values`. So we can just change the data type to also allow for a list of strings to support multivariate time series creation, right?

grll: you are right I would have to check that everything works with 2d array

darts/timeseries.py
pennfranc: Yes good point, I think a comment clarifying at which point we are creating a copy of the df might be nice.

hrzn: +1 for a comment to make it clear. An option could also be to split it in using both `df.copy()` first, and then `df_.sort_index(inplace=True)`.

darts/timeseries.py
hrzn: I think you'll have to use an argument similar to the `init` method: `columns: Optional[Union[List[str], str]]`.

darts/timeseries.py
```diff
@@ -21,7 +21,8 @@ class TimeSeries:
     def __init__(self,
                  df: pd.DataFrame,
                  freq: Optional[str] = None,
-                 fill_missing_dates: Optional[bool] = True):
+                 fill_missing_dates: Optional[bool] = True,
+                 columns: Optional[Union[List[str], str]] = None):
```
hrzn, line 25: I'm still not convinced that we need this argument to the init method. I think it's enough to simply re-use the column names of the dataframe, which would simplify everything. For instance the checks that column names are unique and of the correct size etc; Pandas can take care of that, and removing it for darts would improve separation of concerns IMO. It would still be good to keep this argument in `from_times_and_values()` method though, in order to build a DF with correct column names before building the TimeSeries.
WDYT?

TheMP: @grll what was teh final decision regarding this comment from @hrzn ?

grll: I agree that separation of concerns is good but we still need to enforce a few more things that is not guaranteed by pd.DataFrame such as uniqueness and str format. I created a private method in the latest commit to do that as it is used in a couple of places.

darts/timeseries.py
grll: I changed it to rely more on pandas by accepting a `pd._typing.Axes` as columns parameter type. It's up to the user to come with either a list of columns or a pandas index...

PR #152, opened 2020-07-17
clean check_seasonality function

xfiderek:
Summary

* Cleaned and speeded up finding maxima of ACF


hrzn:
Cool simplification, thanks! Out of curiosity: do you know how much of a speedup that brings?

xfiderek:
> Cool simplification, thanks! Out of curiosity: do you know how much of a speedup that brings?

Maybe it is not that much, order of magnitude is the same.

I've run cell five times before and after improvement and chosen worst case. Here are results: 

before

after

PR #156, opened 2020-07-20
Fix Develop

Droxef:
Fix PR #123 

### Summary

Correct Bugs from PR #123


darts/backtesting/backtesting.py
```diff
@@ -268,7 +268,7 @@ def forecasting_residuals(model: ForecastingModel,
     first_index = series.time_index()[model.min_train_series_length]
 
     # compute fitted values
-    p = backtest_forecasting(series, model, first_index, fcast_horizon_n, True, verbose=verbose)
+    p = backtest_forecasting(series, model, first_index, fcast_horizon_n, trim_to_series=True, verbose=verbose)
```
TheMP, line 271: should this be a part of this PR?

.github/workflows/merge.yml
```diff
@@ -4,13 +4,15 @@ on:
...
     branches:
       - develop
+  pull_request:
+    branches:
+      - develop
 
 jobs:
   tests:
-    runs-on: ${{ matrix.os }}
+    runs-on: macos-latest
```
TheMP, line 13: following the ticket from JIRA we agreed to:
"Run notebooks+unit tests+mac tests on PR and only unit tests on push"

shouldn't this workflow run all the test then and there should be a separate workflow for the push to the feature branch?

Droxef: That was what I had before e77e94fcdbe2131f6cb549f4c3fc5abfe2af49e9
The problem is that on push and on pull-request triggers at the same time and we have duplicate tests.
I propose this solution to keep the chosen behavior 
(PR and develop workflow run together on push devevelop and pull-request on develop, but on push on other branches, only develop workflow runs)

Droxef: I didn't find a way to execute only the PR workflow on pull-request.
So if duplicate tests isn't bothering anyone, I can revert this.
What do you think?

darts/backtesting/backtesting.py
Droxef: It is a part of fixing the develop branch after searching the cause of bugs. But this is not caused by the PR #123 
So yes, it doesn't belong to this PR

.github/workflows/merge.yml
TheMP: It can be acceptable in the beginning to have second send of tests running during PR, as long as we don't run all the test suites on every push (it would take too long to execute).

darts/backtesting/backtesting.py
pennfranc: Wait so where exactly did this cause a bug? Because shouldn't line 274 make sure that the two series have the same indices anyways? I mean I think the change is good, because it should be (very) slightly more efficient. But I believe it would be good to have clearer scopes in individual PRs (don't get me wrong though, I'm guilty of this too).

darts/tests/test_4theta.py
```diff
@@ -50,7 +50,7 @@ def test_theta(self):
         fourtheta.fit(series)
         forecast_theta = theta.predict(20)
         forecast_fourtheta = fourtheta.predict(20)
-        self.assertTrue((forecast_theta - forecast_fourtheta <= 1e-12).all()[0])
+        self.assertTrue((forecast_theta - forecast_fourtheta <= 1e-10).all()[0])
```
pennfranc, line 53: Did you test whether the output of the models is deterministic in this context?

.github/workflows/merge.yml
Kostiiii: IMO we should have every test that is needed in `pull-requests` otherwise the tests that are 'exclusive' only for push won't run on pull requests from forks. Like right now no tests were run on two prs from contributors.

darts/backtesting/backtesting.py
Droxef: My bad, there is no bug, because this function is only for univariate time series. I revert it.
Nonetheless, this parameter shouldn't be here, even if it does nothing for now.

darts/tests/test_4theta.py
Droxef: You are right, the output is not deterministic, thus explaining the problem with one of the test (the precision error was doubled)

hrzn:
Is it possible to open two separate PRs: one for the workflows and one for the changes to the model/backtesting?

darts/models/exponential_smoothing.py
```diff
@@ -9,15 +9,16 @@
...
 logger = get_logger(__name__)
 
 
 class ExponentialSmoothing(UnivariateForecastingModel):
     def __init__(self,
-                 trend: Optional[str] = 'additive',
+                 trend: Optional[ModelMode] = ModelMode.ADDITIVE,
                  damped: Optional[bool] = False,
-                 seasonal: Optional[str] = 'additive',
+                 seasonal: Optional[ModelMode] = ModelMode.ADDITIVE,
```
pennfranc, line 21: Are the underlying string values of the enum still supported as argument values?

Droxef: No, the raw strings are not suported anymore. We only accept the Enum members.
But to be backward-compatible, it can be interesting to accept both.

PR #157, opened 2020-07-21
Generation/pass frequency

Kostiiii:
Added freq passing to couple of timeseries generating function to allow creating smaller timeseries + update tests to check if it possible to create timeseries with those lengths.

### DIFF
--- a/darts/tests/test_timeseries_generation.py
+++ b/darts/tests/test_timeseries_generation.py
@@ -12,56 +12,66 @@ from ..utils.timeseries_generation import (
 class TimeSeriesGenerationTestCase(unittest.TestCase):
 
     def test_constant_timeseries(self):
-
         # testing parameters
-        length = 100
         value = 5
 
-        # testing for constant value
-        constant_ts = constant_timeseries(value=value, length=length)
-        value_set = set(constant_ts._df.values.flatten())
-        self.assertTrue(len(value_set) == 1)
+        def test_routine(length):
+            # testing for constant value
+            constant_ts = constant_timeseries(value=value, length=length)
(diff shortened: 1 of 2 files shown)


darts/tests/test_timeseries_generation.py
```diff
@@ -12,56 +12,66 @@
...
-                               (end_value - start_value) / (length - 1))
+        def test_routine(length):
+            # testing for start value, end value and delta between two adjacent entries
+            linear_ts = linear_timeseries(start_value=start_value, end_value=end_value, length=length)
+            self.assertEqual(linear_ts.values()[0][0], start_value)
+            self.assertEqual(linear_ts.values()[-1][0], end_value)
+            self.assertAlmostEqual(linear_ts.values()[-1][0] - linear_ts.values()[-2][0],
+                                   (end_value - start_value) / (length - 1))
+
+        for length in range(2, 50):
```
TheMP, line 41: seems to be doing much more iterations now, how much slower is it?

TheMP: Also - do we need to go over all interegs from 2 to 50 or is it enough to pick up start, middle and end point?

Kostiiii: I feel like we don't have to since this tests are really fast. But it is probably enough to test few options here so will change it to your proposition

Kostiiii: > I feel like we don't have to worry about it so much since this tests are really fast. But it is probably enough to test few options here so will change it to your proposition

Kostiiii: For the record github actions were around 30 seconds slower so in the end it was worth it to change it.

PR #159, opened 2020-07-21
Add publishing docker image by gradle in release workflow

xfiderek:
Summary
* Refactored publishing docker image in release workflow
* Previously images were being published by github action
* Now they're published using gradle

Other Information
Workflow tested on mock repository in both possible cases (hotfix/non hotfix), but note that during tests different image (simple dockerfile) was being published on different repository (with different credentials).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,15 +126,15 @@ jobs:
           DOCKER_HUB_USER: ${{ secrets.DOCKER_HUB_USER }}
           DOCKER_HUB_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
 
-      - name: "4-1. Publish hotfix docker image"
-        if: contains(github.event.head_commit.message, '#hotfix')
+      #check build.gradle file for explanation of next steps
+      - name: "4. Publish image with tag corresponding to current version"
         run: |
-          ./gradlew dockerTagsPush -P version=${{ steps.bump_dry.outputs.tag }}
+          ./gradlew dockerPushVersion -P version=${{ steps.bump_dry.outputs.tag }}
 
-      - name: "4-2. Publish non-hotfix docker image"
+      - name: "5. Publish image with tag 'latest' if not hotfix"
         if: ${{ !contains(github.event.head_commit.message, '#hotfix') }}
         run: |
-          ./gradlew dockerPush -P version=${{ steps.bump_dry.outputs.tag }}
+          ./gradlew dockerPushLatest
 
 
   deploy-docs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -121,7 +121,10 @@ jobs:
           DRY_RUN: true
 
       - name: "3. Login to docker hub"
-        run: "echo '${{ secrets.DOCKER_HUB_TOKEN }}' | docker login --username ${{ secrets.DOCKER_HUB_USER }} --password-stdin"
+        run: docker login -u $DOCKER_HUB_USER -p $DOCKER_HUB_TOKEN
+        env:
+          DOCKER_HUB_USER: ${{ secrets.DOCKER_HUB_USER }}
+          DOCKER_HUB_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
 
       - name: "4-1. Publish hotfix docker image"
         if: contains(github.event.head_commit.message, '#hotfix')
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -120,22 +120,19 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DRY_RUN: true
 
-      - name: "3. Determine docker tags"
-        run: echo ::set-env name=DOCKER_TAGS::$(echo latest, ${{ steps.bump_dry.outputs.tag }})
+      - name: "3. Login to docker hub"
+        run: "echo '${{ secrets.DOCKER_HUB_TOKEN }}' | docker login --username ${{ secrets.DOCKER_HUB_USER }} --password-stdin"
 
-      - name: "3-1. Determine docker tags for hotfix"
+      - name: "4-1. Publish hotfix docker image"
         if: contains(github.event.head_commit.message, '#hotfix')
-        run: echo ::set-env name=DOCKER_TAGS::$(echo ${{ steps.bump_dry.outputs.tag }})
+        run: |
+          ./gradlew dockerTagsPush -P version=${{ steps.bump_dry.outputs.tag }}
 
-      - name: "4. Build and publish docker to docker hub"
-        uses: elgohr/Publish-Docker-Github-Action@2.14
-        with:
-          name: unit8/darts
-          username: ${{ secrets.DOCKER_HUB_USER }}
-          password: ${{ secrets.DOCKER_HUB_TOKEN }}
-          tags: "${{ env.DOCKER_TAGS }}"
-          cache: true
-          buildoptions: "--compress -q"
+      - name: "4-2. Publish non-hotfix docker image"
+        if: ${{ !contains(github.event.head_commit.message, '#hotfix') }}
+        run: |
+          ./gradlew dockerPush -P version=${{ steps.bump_dry.outputs.tag }}
+
 
   deploy-docs:
     runs-on: ubuntu-latest
--- a/build.gradle
+++ b/build.gradle
@@ -16,8 +16,14 @@ task build {
 // docker & docker run
 docker {
     name "unit8/darts"
-    //it will also automatically create image with tag 'latest'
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -120,22 +120,19 @@ jobs:
...
 
-      - name: "3-1. Determine docker tags for hotfix"
+      - name: "4-1. Publish hotfix docker image"
         if: contains(github.event.head_commit.message, '#hotfix')
-        run: echo ::set-env name=DOCKER_TAGS::$(echo ${{ steps.bump_dry.outputs.tag }})
+        run: |
+          ./gradlew dockerTagsPush -P version=${{ steps.bump_dry.outputs.tag }}
+
+      - name: "4-2. Publish non-hotfix docker image"
+        if: ${{ !contains(github.event.head_commit.message, '#hotfix') }}
```
TheMP, line 132: what happens in case of #patch in commit message?

TheMP: actually also - what happens in case of hotfix itself?

.github/workflows/release.yml
```diff
@@ -120,22 +120,19 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           DRY_RUN: true
 
-      - name: "3. Determine docker tags"
-        run: echo ::set-env name=DOCKER_TAGS::$(echo latest, ${{ steps.bump_dry.outputs.tag }})
+      - name: "3. Login to docker hub"
+        run: "echo '${{ secrets.DOCKER_HUB_TOKEN }}' | docker login --username ${{ secrets.DOCKER_HUB_USER }} --password-stdin"
```
TheMP, line 124: don't we have to pass the secret with env like here? https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#example-using-bash

endrjuskr: Can you do `docker login --username ${{ secrets.DOCKER_HUB_USER }} --password ${{ secrets.DOCKER_HUB_TOKEN }}"` ? In such case you will leave pasword in logs

build.gradle
```diff
@@ -16,7 +16,8 @@ task build {
 // docker & docker run
 docker {
     name "unit8/darts"
-    tags version, "latest"
+    //it will also automatically create image with tag 'latest'
+    tag version, "unit8/darts:${version}"
```
endrjuskr, line 20: will it create three tags? why we need `tag version` ?

endrjuskr: Let's explicit say latest, rather than assume it is automatic

.github/workflows/release.yml
```diff
@@ -120,22 +120,19 @@ jobs:
...
+      - name: "4-1. Publish hotfix docker image"
         if: contains(github.event.head_commit.message, '#hotfix')
-        run: echo ::set-env name=DOCKER_TAGS::$(echo ${{ steps.bump_dry.outputs.tag }})
+        run: |
+          ./gradlew dockerTagsPush -P version=${{ steps.bump_dry.outputs.tag }}
+
+      - name: "4-2. Publish non-hotfix docker image"
+        if: ${{ !contains(github.event.head_commit.message, '#hotfix') }}
+        run: |
+          ./gradlew dockerPush -P version=${{ steps.bump_dry.outputs.tag }}
```
endrjuskr, line 134: What's the difference between `dockerPush` and `dockerTagsPush`? It is not intuitive what each method does

.github/workflows/release.yml
Kostiiii: Maybe versioning logic should be documented somewhere (not for this pr) as it looks like it is not easy to deduce, at least for me.

.github/workflows/release.yml
xfiderek: > don't we have to pass the secret with env like here? https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#example-using-bash

yes, probably you are right. I will change it soon

xfiderek: > Can you do `docker login --username ${{ secrets.DOCKER_HUB_USER }} --password ${{ secrets.DOCKER_HUB_TOKEN }}"` ? In such case you will leave pasword in logs

According to [docker login](https://docs.docker.com/engine/reference/commandline/login/#login-to-a-self-hosted-registry) documentation using --password-stdin avoids logging credentials, but correct me if I misunderstood it.

.github/workflows/release.yml
xfiderek: > what happens in case of #patch in commit message?

There was not any specific behaviour for #patch commit before my changes as far as I can see

xfiderek: > actually also - what happens in case of hotfix itself?

In hotfix we skip publishing docker image with tag 'latest' since hotfix (according to my conversation with @endrjuskr ) refers to change in one of previous versions

build.gradle
xfiderek: > will it create three tags? why we need `tag version` ?

It will create two tags -  'latest' and 'x.y.z' (where 'x.y.z' is release version)

```version``` is variable that is passed to gradle after resolving current release tag by github action. So if release tag is 0.3.0 ```version``` will be replaced by 0.3.0, and gradle will build docker image with two tags, 'latest' and '0.3.0'. Maybe there is better way to parse current version to gradle. If so, I am open to any further suggestions.

.github/workflows/release.yml
xfiderek: > What's the difference between `dockerPush` and `dockerTagsPush`? It is not intuitive what each method does

As you can see above in gradle.build file in docker plugin we specify only tag ```version```, but I've mentioned before that gradle automatically creates also tag 'latest'. `dockerPush` pushes image with all **created** tags, wheareas `dockerTagsPush` publish image only with **specified** tags, so `dockerTagsPush` will push image only with one tag - ```version```, while `dockerPush` will push image with both ```version``` and 'latest' tags. I've disscussed value of ```version``` variable in previous comments.

.github/workflows/release.yml
endrjuskr: It is when you cat a file and then pipe it. In our case we echo it, so it will be visible. In a solution that I proposed it will as well because it inject it to command

endrjuskr: You need to use env - https://github.community/t/github-actions-docker-login/16053/2

.github/workflows/release.yml
endrjuskr: can we use always dockerTagsPush? to make it more explicit?

build.gradle
endrjuskr: so why we need this - "unit8/darts:${version}" ? Is it a formatter?

.github/workflows/release.yml
pennfranc: This is definitely something we have to concretely specify somewhere. Because the two latest hotfixes changed the latest release..

build.gradle
xfiderek: I also think that it is a bit unclear, but I will do my best to explain it. Basically "tags" is deprecated and now it is replaced by "tag". So, we can add any number of following lines:   
`tag x, y`
Now, `x` is internal name, some sort of alias, whereas `y` is `image:tag` that corresponds to image name and demanded tag, so for example we can specify following line: 
`tag unicorn, unit8/darts:0.3.0`  
Purpose of this `key, value` correspondence is to refer to created tags by keys. For example, we can then call:  
`./gradlew dockerPushUnicorn`   
And image with tag **0.3.0** (unit8/darts:0.3.0) will be pushed. It can be useful if we have multiple tags and want to push only subset of them. I really recommend reading [Task section in plugin's README](https://github.com/palantir/gradle-docker#tasks).

.github/workflows/release.yml
xfiderek: > can we use always dockerTagsPush? to make it more explicit?

We can create two tags, one for `latest` and one for current `version` and use a `dockerPush${version}` and optionally `dockerPushLatest`  (See [here](https://github.com/unit8co/darts/pull/159#discussion_r459334192)). In such case only thing that we should agree on is how to pass `version` variable to gradle. I see two possibilities:
1. Pass `version` to gradle as property from CLI in workflow (current implementation)
2. Set `version` as enviromental variable in workflow and use `System.env.version` in gradle to retrieve it. 

What do you think?

.github/workflows/release.yml
endrjuskr: Hotfixe is something that applies to older version that we still support. For instance v1 and v2. In our cases it was more like patch or minor, so that is why we released latest.

.github/workflows/release.yml
endrjuskr: I would recommend passing it as property. Reading something from env in gradle is not visible at first and might confuse people

PR #161, opened 2020-07-21
Refactor/fit args

grll:
Fixes #DARTS-164.

### Summary

- [x] remove `target_indices` parameter to all fit methods on multivariate models
- [x] remove `component_index` parameter to all fit method on univariate models
- [x] introduce `covariate_series` and `target_series` parameters to replace previous syntax
- [x] adapt existing code base to pass tests

Note: backtesting test are not passing but they will soon be refactored so I didn't invest time on that. 

New vs Old API:

```python
# old syntax:
multivariate_model.fit(multivariate_series, target_indices=[0, 1])

# new syntax:
multivariate_model.fit(multivariate_series, multivariate_series[["0", "1"]])

# old syntax:
univariate_model.fit(multivariate_series, component_index=2)

# new syntax:
univaraite_model.fit(multivariate_series["2"])
```


darts/models/forecasting_model.py
```diff
@@ -101,55 +100,33 @@ class UnivariateForecastingModel(ForecastingModel):
...
         ----------
-        series
-            The training time series on which to fit the model.
-        target_indices
-            A list of integers indicating which component(s) of the time series should be used
-            as targets for forecasting.
+        covariate_series
+            The training time series on which to fit the model (can be multivariate or univariate).
+        target_series
+            The target values used ad dependent variables when training the model
```
hrzn, line 128: The target can also be multivariate or univariate. I would unify the comments here.

darts/models/forecasting_model.py
```diff
@@ -101,55 +100,33 @@ class UnivariateForecastingModel(ForecastingModel):
...
-                     "If a multivariate series is given as input to this multivariate model, please"
-                     " provide a list of integer indices `target_indices`"
-                     " that indicate which components of this series should be predicted", logger)
-
-        raise_if_not(all(idx >= 0 and idx < series.width for idx in target_indices), "The target indices "
-                     "must all be between 0 and the width of the TimeSeries instance used for fitting - 1.", logger)
-
-        self.target_indices = target_indices
-        super().fit(series)
+        raise_if_not(len(covariate_series) == len(target_series), "covariate_series and target_series musth have same "
```
hrzn, line 130: You should probably compare the time indexes here instead of the lengths.

darts/models/torch_forecasting_model.py
```diff
@@ -261,46 +265,63 @@ def _create_from_cls_and_kwargs(cls, kws):
...
+            A series of target (dependent) value(s) predicted using the covariate_series if None specified, use the
+            covariate series as target.
         val_series
-            Optionally, a validation time series, which will be used to compute the validation loss
-            throughout training and keep track of the best performing models.
+            Optionally, 2 validation time series (covariate and target), which will be used to compute the validation
+            loss throughout training and keep track of the best performing models.
         verbose
             Optionally, whether to print progress.
         target_indices
```
hrzn, line 286: to remove

darts/models/torch_forecasting_model.py
```diff
@@ -261,46 +265,63 @@ def _create_from_cls_and_kwargs(cls, kws):
 
     @random_method
     def fit(self,
-            series: TimeSeries,
-            val_series: Optional[TimeSeries] = None,
-            verbose: bool = False,
-            target_indices: Optional[List[int]] = None) -> None:
+            covariate_series: TimeSeries,
+            target_series: Optional[TimeSeries] = None,
+            val_series: Optional[Tuple[TimeSeries]] = None,
```
hrzn, line 270: How about splitting in two:
```
val_covariate_series: Optional[TimeSeries]
val_target_series: Optional[TimeSeries]
```
That would be more consistent with the first two arguments.

grll: I did it like this first but having a Tuple makes it more easy to check that the length provided is either None or 2 as opposed to having two separate args

hrzn: I would say that it's more important to have a good signature here, even if the check is slightly more complex behind the scenes.

darts/models/forecasting_model.py
```diff
@@ -24,29 +24,33 @@ class ForecastingModel(ABC):
...
-    def fit(self, series: TimeSeries) -> None:
+    def fit(self) -> None:
         """ Fits/trains the model on the provided series
 
-        Parameters
-        ----------
-        series
-            the training time series on which to fit the model
+        Implements behavior that should happen when calling the `fit` method of every forcasting model regardless of
+        wether they are univariate or multivariate.
```
pennfranc, line 52: whether (sorry)

grll: No problem :) you can directly propose edits for such typo

darts/models/forecasting_model.py
```diff
@@ -24,29 +24,33 @@ class ForecastingModel(ABC):
...
         """
-        raise_if_not(len(series) >= self.min_train_series_length,
```
pennfranc, line 46: As far as I can see this is currently not being tested for the univariate model.

darts/models/forecasting_model.py
grll, line 52:
```suggestion
        whether they are univariate or multivariate.
```

darts/models/torch_forecasting_model.py
```diff
@@ -50,21 +50,23 @@ def __init__(self,
 
         Parameters
         ----------
-        series
+        covariate_series
             The time series to be included in the dataset.
+        target_series
+            The time series used has target.
```
pennfranc, line 56: as?

darts/models/forecasting_model.py
grll: good catch I moved this check in the parent class

pennfranc:
Overall I really like this change, just a couple of things I'm not quite sure about:
1. I kind of feel like the names of `target_series` and `covariate_series` should be switched. I think `covariate_series` should be the optional one since we always want to predict a time series, but optionally we want to include other time series as features that won't be predicted, right?
2. Kind of related to the above, personally I don't see why we don't simply include the target series in the training data by default. Because if it's not included by default, the user will have to manually make sure it is included in both arguments (`target_series` and `covariate_series`) at the same time (if he has other time series components that he just wants to include as features and not predict.
I hope this makes sense..

darts/models/forecasting_model.py
pennfranc: I didn't know you could do this!

grll:
> Overall I really like this change, just a couple of things I'm not quite sure about:
> 
> 1. I kind of feel like the names of `target_series` and `covariate_series` should be switched. I think `covariate_series` should be the optional one since we always want to predict a time series, but optionally we want to include other time series as features that won't be predicted, right?
> 2. Kind of related to the above, personally I don't see why we don't simply include the target series in the training data by default. Because if it's not included by default, the user will have to manually make sure it is included in both arguments (`target_series` and `covariate_series`) at the same time (if he has other time series components that he just wants to include as features and not predict.
>    I hope this makes sense..
> 
> Edit: Ok looking at your example above I think I get why you did it this way, this provides the user with all possible combinations and with an example it's pretty intuitive! I still think `covariate_series` should be the optional one though and set to `target_series` if not present.
> The thing is though, and maybe that's just me, the name **co**variate to me somehow suggests that it is **additional**, meaning on top of `target_series`. I guess that's why I would have included `target_series` in the training data by default.
> 
> @hrzn I agree that it requires a bit more effort to list all the fit arguments now, but I think this is a price worth paying.

I renamed `covariate_timeseries` into `training_timeseries` maybe it helps understand what does what, wdyt ?

hrzn (approved):

PR #163, opened 2020-07-22
Implement map method for TimeSeries

### LINKED ISSUE #121: Implement `map` method for TimeSeries class
**Missing functionality**
Although adding or multiplying a scalar value to every entry in a TimeSeries instance is very easy, when it comes to more complex element-wise operations (e.g. taking the sine function value of every entry in the series) there is no easy way to create a new TimeSeries instance, given a TimeSeries instance and operation.

**Proposed solution**
The most straight-forward way to implement this functionality would be implement a `map` function that takes as arguments a TimeSeries and a function, and returns a new TimeSeries instance (pretty much like the built-in python function).

Lovkush-A:
Fixes #121 

### Summary
An implementation for map is added. It uses pandas' applymap method.
A jupyter notebook is included to show the tests conducted.

### Other Information
I have added a couple of TODOs:
* check if typing is correct
* check if we need to add any error checking/raises of potential errors


pennfranc:
@Lovkush-A Thanks for your PR! I like your approach but have a couple of comments:
1. As you pointed out in the comment, some checks might be needed here. Since we're directly accessing pandas DataFrame columns, it might be good to check whether the given columns are within range first. To do this, you could use our `raise_if_not` function from the `logging` module.
2. I like your tests. But I think it would be better to implement them as unit tests rather than a notebook.

Lovkush-A:
@pennfranc 
1. Done.
2. Done. (At least I think it is done. I do not know how to test the tests...!)

darts/timeseries.py
```diff
@@ -880,6 +880,41 @@ def is_within_range(self,
         index = self.time_index()
         return index[0] <= ts <= index[-1]
 
+
+    def map(self,
+            fn: Callable[[float], float],
+            cols: Optional[Union[List[int], int]] = None) -> 'TimeSeries':
+        #TODO check if typing is correct
+        """
+        Applies the function fn to all values in this TimeSeries, or, to only those
```
hrzn, line 889: You can write `Applies the function fn elementwise to all values`.

darts/timeseries.py
```diff
@@ -880,6 +880,41 @@ def is_within_range(self,
         index = self.time_index()
         return index[0] <= ts <= index[-1]
 
+
+    def map(self,
+            fn: Callable[[float], float],
```
hrzn, line 885: For the type hints you can use `Callable[[np.number], np.number]` to be consistent with type check we do upon TimeSeries creation.

darts/timeseries.py
```diff
@@ -880,6 +880,41 @@ def is_within_range(self,
...
+            cols: Optional[Union[List[int], int]] = None) -> 'TimeSeries':
```
hrzn, line 886: @grll you'll probably have to change this as part of you refactoring

darts/timeseries.py
```diff
@@ -880,6 +880,41 @@ def is_within_range(self,
...
+
+        Returns
+        -------
+        TimeSeries
+            A new TimeSeries instance
+        """
+        new_dataframe = self.pd_dataframe() 
+        
+        if cols is None:
+            new_dataframe = new_dataframe.applymap(fn)
```
hrzn, line 908: In this case (when `cols is None`), you don't need the (expensive) call to `self.pd_dataframe()` on line 905 because `applymap()` already returns a new DF instance. I would call `self.pd_dataframe()` only just before line 915.

darts/timeseries.py
```diff
@@ -880,6 +880,41 @@ def is_within_range(self,
...
+        else:
+            if isinstance(cols, int):
+                cols = [cols]
+            raise_if_not(all([0 <= index and index < self.width for index in cols]),
+                     'The indices in `cols` must be between 0 and the number of components of the current '
+                     'TimeSeries instance - 1, {}'.format(self.width - 1), logger)
+            new_dataframe[cols] = new_dataframe[cols].applymap(fn)
+        return TimeSeries(new_dataframe, self.freq())
```
hrzn, line 916: `self.freq_str()`

darts/tests/test_map.py
```diff
@@ -0,0 +1,39 @@
...
+    series_2 = TimeSeries(df_2, 'D')
+    series_01 = TimeSeries(df_01, 'D')
+    series_012 = TimeSeries(df_012, 'D')
+
+    @classmethod
+    def setUpClass(cls):
+        logging.disable(logging.CRITICAL)
+
+    def test_map(self):
+        self.assertEqual(series_0, series.map(fn,0))
```
hrzn, line 34: space after `,`: `(fn, 0)`

darts/tests/test_map.py
```diff
@@ -0,0 +1,39 @@
+import unittest
+import numpy as np
+import pandas as pd
+import logging
+
+from ..timeseries import TimeSeries
+
+class MapTestCase(unittest.TestCase):
```
hrzn, line 8: I think the content of this class belongs rather to a method `test_map()` inside `TimeSeriesTestCase` in `test_timeseries.py`.

darts/tests/test_map.py
```diff
@@ -0,0 +1,39 @@
...
+    @classmethod
+    def setUpClass(cls):
+        logging.disable(logging.CRITICAL)
+
+    def test_map(self):
+        self.assertEqual(series_0, series.map(fn,0))
+        self.assertEqual(series_0, series.map(fn, [0]))
+        self.assertEqual(series_2, series.map(fn, 2))
+        self.assertEqual(series_01, series.map(fn, [0,1]))
+        self.assertEqual(series_012, series.map(fn, [0,1,2]))
```
hrzn, line 38: space after `,`: `[0, 1, 2]`

darts/tests/test_map.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        logging.disable(logging.CRITICAL)
+
+    def test_map(self):
+        self.assertEqual(series_0, series.map(fn,0))
+        self.assertEqual(series_0, series.map(fn, [0]))
+        self.assertEqual(series_2, series.map(fn, 2))
+        self.assertEqual(series_01, series.map(fn, [0,1]))
```
hrzn, line 37: space after `,`: `[0, 1]`

darts/tests/test_map.py
```diff
@@ -0,0 +1,39 @@
...
+        self.assertEqual(series_012, series.map(fn, [0,1,2]))
+        self.assertEqual(series_012, series.map(fn))
```
hrzn, line 39: I think you could add a sanity test case to check, for instance, that `series_01` is *not* equal to series.map(fn)

darts/tests/test_map.py
Lovkush-A: Done.

darts/tests/test_map.py
Lovkush-A: Done.

darts/tests/test_map.py
Lovkush-A: Done.

darts/tests/test_map.py
Lovkush-A: Done. (Also found a couple of other lists/tuples with missing spaces).

darts/tests/test_map.py
Lovkush-A: Done.

darts/timeseries.py
Lovkush-A: Done.

darts/timeseries.py
Lovkush-A: Done. (Specifically, added word 'elementwise')

darts/timeseries.py
Lovkush-A: Done.

darts/timeseries.py
Lovkush-A: Done.

darts/timeseries.py
grll: yes we would need to replace `int` indexing  with `str` indexing  but nothing too serious here

darts/timeseries.py
```diff
@@ -880,6 +880,39 @@ def is_within_range(self,
         index = self.time_index()
         return index[0] <= ts <= index[-1]
 
+
+    def map(self,
+            fn: Callable[[np.number], np.number],
+            cols: Optional[Union[List[int], int]] = None) -> 'TimeSeries':
+        """
+        Applies the function fn elementwise to all values in this TimeSeries, or, to only those
```
pennfranc, line 888: One more small thing: In the docstring, when explicitly referring to the name of a function argument, we usually put ` ticks around the name of the argument. So in this case for `fn` and for `cols` this would be nice.
Once that is done, I think we can approve this PR.

Lovkush-A: Done.

PR #164, opened 2020-07-22
Fix: package type information

Seirdy:
Follow the steps outlined in PEP-0561 to package type information,
allowing other type-checked Python code that imports `darts` to benefit
from better static analysis. See the PEP:
https://www.python.org/dev/peps/pep-0561/#packaging-type-information

- Include a `py.typed` marker file in package root
- Include `py.typed` in package_data


setup.py
```diff
@@ -33,6 +33,9 @@ def read_requirements(path):
       license='Apache License 2.0',
       packages=find_packages(),
       install_requires=read_requirements('requirements/main.txt'),
+      package_data={
```
TheMP, line 36: thx for this addon @Seirdy! we will merge it once we sort out our workflow CI setup

hrzn (approved):
thanks!

PR #167, opened 2020-07-27
Implement pipeline, base_transformer and transformer utils.

Kostiiii:
This introduces:
* BaseTransformer: abstract base for transformers
* Pipeline: Chaining of transformers (trying to be similar to/inspired from scikit-learn) 
* Transformers utils: Utils to create transformers from functions


TheMP:
DO you have some examples of how would the preprocessing look like now compared to before?

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+        any of this methods it should mark this with appropriate property (`reversible`, `fittable`, `can_predict`).
+
+        Parameters
+        ----------
+        validators
+            List of validators that will be called before transform
+        validator_fns
+            List of tuples of validating function and optional string with reason for validating.
+            From this will be created list of validators and it will be appended to validators.
+        reversible
```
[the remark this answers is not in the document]
TheMP: EDIT: Just saw the code below to filter for fittable - maybe a dummy fit that doesn't change anything would be easier for users than remembering to pass this flag? Or can this flag be somehow automatically set if the fit function is overriden? I'm just worried about a scenario when someone implements fit() but forgets to set the right flag

darts/preprocessing/utils.py
```diff
@@ -0,0 +1,139 @@
...
+from darts import TimeSeries
+from darts.logging import raise_if_not, raise_if, get_logger
+from darts.preprocessing.base_transformer import BaseTransformer
+
+logger = get_logger(__name__)
+T = TypeVar('T')
+_callable_with_args_kwargs = Callable[[T, Tuple[Any, ...], Dict[str, any]], T]
+
+
+class _WrapperTransformer(BaseTransformer[T]):
```
TheMP, line 18: How would this wrapper be used? What is the typical usecase?

Kostiiii:
> Sorry, wanted to request changes but clicked approve :)

so close :/

darts/preprocessing/base_transformer.py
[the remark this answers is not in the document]
Kostiiii: This is part I was spend quite some time to think about. Automatically detection if fit is overriden is possible - I didn't go with it because I felt like it was kinda tricky. I like this approach because, a) when user derives from this class and create custom transformer he has to explicitly mark what this transformer is capable of. b) if transformer is created from utility functions this properties are set automatically. But of course if you feel like setting this automatically is better we could go that way.

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+            Object on which validation functions will be run.
+
+        Returns
+        -------
+        True
+            If successful.
+        """
+        for validator in self._validators:
+            reason = validator.reason if validator.reason else ""
+            raise_if_not(validator(data), f"Validation failed for {self.name}\n{reason}", logger)
```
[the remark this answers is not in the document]
Kostiiii: Valid point. hmm I was mainly thinking about situation when quick running validation are put first and long running exhausting ones at the end. But it makes sense to collect all errors at once

darts/preprocessing/pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+        Pipeline combines multiple transformers chaining them together.
+
+        Parameters
+        ----------
+        transformers
+            List of transformers.
+        names
+            Optional list of names for transformers must have same length as transformers and names should be unique.
+            If no names are specified transformer name will be set to str value of his index in original list.
+        deep
```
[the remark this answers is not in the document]
Kostiiii: It is for use cases when we don't want to alter transformers passed to pipeline. Maybe it is not needed

darts/preprocessing/utils.py
Kostiiii: I would like this transformer to remains 'private' and be used only by functions `transformer_from_ts_functions` and `transformer_from_values_functions`. So no user would use this class directly. Usecase of this is to be helper wrapper over provided functions. lets say we have only transform and inverse_transform functions and don't want to write transformer for them (like `transform = lambda: x + 1`, `inverse_transfrom = lambda x: x - 1`) then we could just pass this function to `transformer_from_ts_functions` or `transformer_from_values_functions` and have transformer with approppriate functions set.

darts/preprocessing/pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+        }
+        self._names = names
+
+        self._args_for_transformers = defaultdict(tuple)
+        self._kwargs_for_transformers = defaultdict(dict)
+
+        self._reversible = all((t.reversible for t in self._transformers))
+        self._can_predict = len(self._transformers) and self._transformers[-1].can_predict
+
+    def set_transformer_args_kwargs(self, key: Union[str, int], *args, **kwargs):
```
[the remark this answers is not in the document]
Kostiiii: Do you mean similar function but on transformer? That would be simper, will add it here. Do you want this to be removed?

darts/preprocessing/pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+            logger.warning("Empty pipeline created")
+
+        if names:
+            raise_if_not(len(names) == len(self._transformers), "names should have same length as transformer", logger)
+            raise_if_not(len(names) == len(set(names)), "names should be unique", logger)
+        else:
+            names = list(map(str, range(len(self._transformers))))
+
+        self._name_2_idx = {
+            str(name): idx for idx, name in enumerate(names)
```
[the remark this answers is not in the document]
Kostiiii: Name property of transformer? It might be okay but what about situation when same couple transformers share same name? names could be removed and we could just use indexes to simplify it

Kostiiii:
> DO you have some examples of how would the preprocessing look like now compared to before?

I wouldn't say it is that much different (at least not yet). For now the only new thing are possibility to create transformers from functions and to chain them in pipeline. The feeling that I get from this task is that after this is done we will have clearer and structured way to go around preprocessing. And with more and more transformers, or preprocessing utils adjusted as transformers the way to do this would became more visible. Regarding examples:  What do you think about examples in documentation, maybe at __init__ level for preprocessing package?

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+class BaseTransformer(Generic[T], ABC):
+    def __init__(self,
+                 validators: Optional[List[Validator]] = None,
+                 validator_fns: Optional[List[Tuple[Callable[[T], bool], Optional[str]]]] = None,
+                 reversible: bool = False,
+                 can_predict: bool = False,
+                 fittable: bool = False):
+        """
+        Abstract class for transformers. All deriving classes have to implement only one function `transform`.
+        It also have `inverse_transform`, `fit` and `predict` left unimplemented. If child of this class implements
```
pennfranc, line 24:
```suggestion
        It also has `inverse_transform`, `fit` and `predict` left unimplemented. If a child of this class implements
```

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+        any of this methods it should mark this with appropriate property (`reversible`, `fittable`, `can_predict`).
```
pennfranc, line 25:
```suggestion
        any of these methods, it should mark them with the appropriate property (`reversible`, `fittable`, `can_predict`).
```
(sorry for being annoying)

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Parameters
+        ----------
+        validators
+            List of validators that will be called before transform
+        validator_fns
+            List of tuples of validating function and optional string with reason for validating.
+            From this will be created list of validators and it will be appended to validators.
+        reversible
+            Flag indicating whether this transformer have implemented inverse_transform
```
pennfranc, line 35:
```suggestion
            Flag indicating whether this transformer implements inverse_transform
```

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+        can_predict
+            Flag indicating whether this transformer have implemented predict
```
pennfranc, line 37:
```suggestion
            Flag indicating whether this transformer implements predict
```

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+        fittable
+            Flag indicating whether this transformer have implemented fit
```
pennfranc, line 39:
```suggestion
            Flag indicating whether this transformer implements fit
```

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+        Returns
+        -------
+        T
+            Inverse transformed data.
+        """
+        raise NotImplementedError(f"inverse_transform not implemented for transformer {self.name}")
+
+    def predict(self, data: T, *args, **kwargs) -> T:
+        """
+        Predict function. Not implemented in base class.
```
pennfranc, line 155: Can you give an example of a transformer that has a `predict` method? Maybe an additional sentence in the docstring that explains the role of the `predict` functions could be useful, since it's not immediately clear to me

darts/preprocessing/base_transformer.py
Kostiiii: Thanks for pointing this and all the rest mistakes/typos. Don't be sorry these are all valuable comments to me, thanks again

darts/preprocessing/utils.py
pennfranc: I think this makes sense, being able to pass functions as arguments instead of implementing them in a subclass, kind of like we already do with validators, right?

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        Abstract class for transformers. All deriving classes have to implement only one function `transform`.
+        It also have `inverse_transform`, `fit` and `predict` left unimplemented. If child of this class implements
+        any of this methods it should mark this with appropriate property (`reversible`, `fittable`, `can_predict`).
+
+        Parameters
+        ----------
+        validators
+            List of validators that will be called before transform
+        validator_fns
```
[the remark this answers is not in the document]
Kostiiii: Agree, looks better without it

darts/preprocessing/base_transformer.py
Kostiiii: Something to put on radar, but here should be some discusion how do we want to treat pipeline with predict at last "transformer". Do we run transform and all that stuff on it or just prediction? More detailed documentation is also valid point. Will add it

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
TheMP: How does altering transformers work? Can you give an example?

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
TheMP: That was my question more or less - I was wondering if it's possible to drop name property as they don't add many usability (we can name the class for that) and run on just indexing

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
TheMP: yes, I wonder why do we need setters altogether - Transformers could take init() parameters to constructors and then, after creation be immutable - no need to set things inside (unless Im missing some point here)

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
Kostiiii: We could do that, it would be simpler.

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
Kostiiii: One remark because of fit, how @grll mentioned in the previous discussion, transformers won't be immutable. I like your approach it is much cleaner. correct me if I am wrong, then transform and other functions would not need any additional parameters and I think for most use cases that would be fine. This would simplify a lot in implementation but one thing that sticks to me is that we then cannot modify behaviour of transformer. If transformer does something based on some alpha hyper parameter then to experiment with different alpha values we would need to create new transformer and new pipeline with that transformer.

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
Kostiiii: By calling fit on transformer, then transformer in most cases will have different state than before fit.

darts/preprocessing/pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+        self._args_for_transformers = defaultdict(tuple)
+        self._kwargs_for_transformers = defaultdict(dict)
+
+        self._reversible = all((t.reversible for t in self._transformers))
+        self._can_predict = len(self._transformers) and self._transformers[-1].can_predict
+
+    def set_transformer_args_kwargs(self, key: Union[str, int], *args, **kwargs):
+        """
+        Set arguments for `transform` and `inverse_transform` functions got transformer identified by either index
+        or name. Raises value error if transformer can't be find by key.
```
xfiderek, line 66: Should not it be something like: _Set arguments for `transform` and `inverse_transform` functions **for** transformer identified by either index or name. Raises value error if transformer can't be **found** by key._

darts/preprocessing/pipeline.py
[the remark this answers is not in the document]
xfiderek: I am not sure if i understand correctly but i think we can leverage `Transformer`'s mutability if we want to change some of its parameters. In such case we can simply store this transformer as variable before passing to `Pipeline` and if needed we can change its parameters using created variable instead of `set_transfomer_args_kwargs` method of `Pipeline`. It will probably mean moving this method to `Transformer` class but still I think it is more convenient than doing it in pipeline (we could then get rid of `names` argument if I am correct).  What do you think?

darts/preprocessing/base_transformer.py
LeoTafti: I am not a big fan of having a `predict` method. I believe it would make the API cleaner to restrict transformers and pipelines to preprocessing only, and then call `model.predict()` on the transformed data

darts/preprocessing/base_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+        -------
+        T
+            Transformed data.
+        """
+        return self.fit(data).transform(data, *args, **kwargs)
+
+    def __call__(self, data: T, *args, fit: bool = False, inverse: bool = False, **kwargs) -> T:
+        """
+        Calling transformer will run validation and perform transformation or
+        inverse_transformation on data. If fit flag is set transformer will first be fitted.
```
LeoTafti, line 213: Is this really needed ?

I find it less intuitive/clear to write `transformer(data, inverse=true)` than `transformer.inverse_transform(data)`

darts/preprocessing/pipeline.py
```diff
@@ -0,0 +1,248 @@
...
+            return Pipeline([self._transformers[key]], [self._names[key]], deep=True)
+        elif isinstance(key, str):
+            raise_if_not(key in self._name_2_idx.keys(),
+                         f"No transformer named '{key}' available transformers {self._name_2_idx.keys()}", logger)
+            return Pipeline([self._transformers[self._name_2_idx[key]]], [key], deep=True)
+        elif isinstance(key, slice):
+            return Pipeline(self._transformers[key], self._names[key], deep=True)
+        raise_if(True, "Key must be int, str or slice", logger)
+
+    def __call__(self, data: TimeSeries, inverse: bool = False) -> TimeSeries:
```
LeoTafti, line 214: Same comment here, I don't see the advantage of this considering we would then have to specify `inverse`

LeoTafti:
Notable changes:

- Removed `predict()` method from `BaseTransformer` (and thus also from `Pipeline`)
- I removed the "auto-validation" when calling `transform` / `inverse_transform`. This auto-validation wasn't consistently applied and I think we can let the user decide when to run `validate()` (e.g. is it needed before calling `fit` ? or do we really want to call `validate()` twice when doing fit and then transform ?)
- Removed `names` from `Pipeline` as suggested in the review, as this considerably simplifies the API. I think we can however keep a `name` attribute in `BaseTransformer`, it makes error messages or even simple prints of Pipeline / Transformers much more readable and insightful. Note that removing `names` means we cannot access a specific transformer in a pipeline using the `pipeline["transformer_x"]` notation anymore, but I still think it's worth it
- Removed `__call__()` overrides in `Pipeline` and `BaseTransformer`.

darts/preprocessing/base_transformer.py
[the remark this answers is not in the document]
LeoTafti: I also think keeping those flags is fine, but I'd remove their default value. Then any sublcass calling `super().__init__` would have to specify the flags.

LeoTafti:
> * I removed the "auto-validation" when calling `transform` / `inverse_transform`. This auto-validation wasn't consistently applied and I think we can let the user decide when to run `validate()` (e.g. is it needed before calling `fit` ? or do we really want to call `validate()` twice when doing fit and then transform ?)

I'm actually questioning myself on that point and considering putting it back.

I see the following benefits:
- It would be more consistent to always run `validate()` before `transform()` and `inverse_transform()`, both in `BaseTransformer` and in `Pipeline`. Otherwise only `BaseTransformers` have a `validate()` method that can be run separately, since a `Pipeline.validate()` method would have to also perform transformations to validate the obtained time series at each stage…
- It avoids user error (ie. the user forgetting to call `validate()` even though the user had specified validators)
- It simplifies the API by allowing to remove `validate()` from `BaseTransformer` (or at least to make it "private", ie. `_validate()`)

The main drawbacks would be:
1. This makes it impossible to call `validate()` before `fit()`
2. This could result in more calls to `validate()` than needed, which could possibly be costly

"1." above raises a question though: I believe validators were added to perform validation before calling `transform()` in the first place (docstring states `validators : List of validators that will be called before transform`) and as such `ScalerWrapper` makes use of a validator to check that `fit()` vas called before `transform()`. Isn't it true though that we might want to run the validation before calling `fit()` too, for "fittable" models ?

I'd suggest to automatically perform the validation step when calling `fit()` too for fittable models, and maybe define an abstract `class FittableTransformer(BaseTransformer)` which adds the `_fit_called` member and performs the "fit() called before transform() or inverse_transform()" check itself from within these functions, any opinions on that ?

darts/dataprocessing/utils.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    return _create_wrapper_data_transformer(_get_parent_classes(inverse_transform, fit),
+                                            transform,
+                                            inverse_transform,
+                                            fit,
+                                            name)
+
+
+def data_transformer_from_values_functions(transform: _callable_with_args_kwargs[np.ndarray],
+                                           inverse_transform: Optional[_callable_with_args_kwargs[np.ndarray]] = None,
```
LeoTafti, line 105: I'm not sure yet whether we really need both "versions" of the same function (namely `data_transformer_from_ts_functions()` and `data_transformer_from_values_functions()`, nor do I really like the functions' names. I might change my mind implementing common transformers, but leaving this comment here for the record.

grll: What do you think about calling it a `Transformation` ?

grll: or maybe keeping `transformer` in the name makes actually more sense I dont know like in sklearn you have `FunctionTranformer`: https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html#sklearn.preprocessing.FunctionTransformer

darts/dataprocessing/transformers/missing_values_filler.py
```diff
@@ -0,0 +1,50 @@
...
+from darts.logging import get_logger, raise_if, raise_if_not
+
+logger = get_logger(__name__)
+
+
+class MissingValuesFiller(BaseDataTransformer[TimeSeries]):
+    def __init__(self,
+                 fill: Union[str, float] = 'auto',
+                 name: str = "MissingValuesFiller",
+                 validators: Optional[List[Validator]] = None):
```
LeoTafti, line 21: I'm open to discussion on the fill parameter type. I've also considered making it optional, and default to calling `auto_fillna()` if not provided / `None`, what do you think is best ?

Also I'm wondering two things about `fillna()` and `auto_fillna()` in `statistics.py`:
- I know they're named after the corresponding pandas methods, but I don't think the name is supper obvious… I'd suggest refactoring both methods into a single one, maybe called `fill_missing_values()` which would then also take a `fill: Union[str, float] = 'auto'` parameter, what do you think ?
- Shouldn't we prefix them with an underscore (make them "private"), to encourage using the transformer instead  ?

darts/dataprocessing/transformers/scaler_wrapper.py
```diff
@@ -2,56 +2,57 @@
...
+from darts.dataprocessing.transformers import FittableDataTransformer, InvertibleDataTransformer
+
+from darts.timeseries import TimeSeries
+from darts.logging import get_logger, raise_log
 from sklearn.preprocessing import MinMaxScaler
 
 logger = get_logger(__name__)
 
 
-class ScalerWrapper:
```
LeoTafti, line 12: Actually couldn't we simply call it `Scaler` ?

I think that the fact that it's a wrapper around existing scaler is clear enough from the docstring and constructor ?

hrzn: Agree

darts/dataprocessing/pipeline.py
```diff
@@ -0,0 +1,166 @@
...
+        return Pipeline(transformers, deep=True)
+
+    def __iter__(self) -> Iterator[BaseDataTransformer[TimeSeries]]:
+        """
+        Returns
+        -------
+        Iterator
+            List of data transformers
+        """
+        return self._transformers
```
hrzn, line 146: I think that's an `Iterable` (and not an `Iterator`).

hrzn (commented):
It looks very nice! My main comment is that I'm not too convinced by the utility of the functions in `utils.py`, and would vote for dropping that (but I'm waiting to be convinced otherwise).

Also @LeoTafti I think it would help greatly if you could come up with one or two examples of typical useage of these transformers and pipelines (that could perhaps be a notebook).

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -0,0 +1,101 @@
...
+        True
+            If successful.
+        """
+        # Collect all validation errors (if any) and throw them all at once
+        fail_reasons = []
+        for validator in self._validators:
+            if not validator(data):
+                fail_reasons.append(validator.reason)
+
+        raise_if(len(fail_reasons) != 0,
```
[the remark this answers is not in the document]
LeoTafti: Indeed a `ValueError` probably isn't the best, I'll look around and change it.

darts/dataprocessing/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        def transform(self, data: T, *args, **kwargs) -> T:
+            return self._transform(data, *args, **kwargs)
+
+        if InvertibleDataTransformer in parent_classes:
+            def inverse_transform(self, data: T, *args, **kwargs) -> T:
+                return self._inverse_transform(data, *args, **kwargs)
+
+        if FittableDataTransformer in parent_classes:
+            def fit(self, data: T) -> 'Type[BaseDataTransformer[T]]':
+                self._fit(data)
```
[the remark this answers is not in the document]
LeoTafti: I think you're correct. These might not actually be very useful (except maybe for writing tests). I'll remove them and adapt the tests accordingly

darts/dataprocessing/pipeline.py
```diff
@@ -0,0 +1,166 @@
...
+        data
+            TimeSeries to fit on.
+        """
+        for transformer in filter(lambda t: isinstance(t, FittableDataTransformer), self._transformers):
+            transformer.fit(data)
+
+    def fit_transform(self, data: TimeSeries) -> TimeSeries:
+        """
+        For each data transformer in pipeline first fit the data if transformer is fittable then transform data using
+        fitted transformer. The transformed data is then passed to next transformer.
```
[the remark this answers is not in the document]
LeoTafti: Very good catch! I think the `fit_transform()` implementation is actually the correct one, and `fit()`should interleave calls to `transform()` too. I'll change it.

darts/dataprocessing/transformers/missing_values_filler.py
LeoTafti: Cool.

I also think we can remove the `missing_values.py` file and:
- Move `_fill_missing_values()` inside of the `TimeSeries` class. New syntax would be `series._fill_missing_values(fill=…)` instead of `_fill_missing_values(series, fill=…)`

- Rename `na_ratio()` into `missing_values_ratio()` and move it to `statistics.py`

WDYT ?

LeoTafti: Also questioning this:

> Shouldn't we prefix them with an underscore (make them "private") or even remove them once the transformer is implemented, to encourage using the transformer instead ?

I think that it's more intuitive to be able to simply call `fill_missing_values()` when that is the only transformation the user wants to make, instead of `MissingValuesFiller().transform()`. Of course having the transformer is still very useful because it allows using it in a `Pipeline`.

Thus I think that the best way to go about it is to provide both options.

hrzn: Agree with your EDIT ;)

darts/dataprocessing/pipeline.py
LeoTafti: I just looked, and to make `Pipeline` _iterable_ it needs to provide an `__iter__` method which returns an `Iterator`. The types are correct, but the return value isn't. I'll fix it.

darts/dataprocessing/transformers/base_data_transformer.py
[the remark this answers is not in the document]
LeoTafti: Didn't find a perfectly matching error type. The one I think conveys best the meaning would be an `AssertionError`, but that could also be a bit misleading in the sense that it's wouldn't be raised by `assert` statement, and an `AssertionError` might be a bit too "strong".

I think we can either leave it as such, or create a custom error. What do you think ?

darts/dataprocessing/pipeline.py
hrzn: Maybe something like `return._transformers.__iter__()`

darts/dataprocessing/pipeline.py
```diff
@@ -0,0 +1,178 @@
...
+        Parameters
+        ----------
+        data
+            TimeSeries to fit on.
+        """
+
+        # Find the last fittable transformer index
+        # No need to fit (and thus transform) after this index, possibly saving a fair bit of time
+        last_fittable_idx = -1
+        for idx, transformer in enumerate(self._transformers):
```
hrzn, line 55: Wouldn't it be enough to simplify and just have something like:

```
for transformer in self._transformers:
    if isinstance(transformer, FittableDataTransformer):
        transformer.fit(data)
    data = transformer.transform(data)
```

darts/dataprocessing/pipeline.py
LeoTafti: Oops I thought I had done exactly this, sorry!

darts/dataprocessing/pipeline.py
LeoTafti: Well I thought about this, and decided to go with the longer version to improve performance, but it might not be worth it?

The problem with the solution you propose (which is actually the same as calling `self.fit_transform(data)` and discarding the result) is that it runs at least one superfluous transform (the last one), and possibly more.

If for example the only the first transformer in the pipeline is fittable, there is no need to transform the data at all to fit other transformers further in the pipeline (because there are none). I'm not sure exactly how costly a transformation can be, but I thought it'd be nicer to avoid those unnecessary transforms. What do you think ?

hrzn: OK, fair enough, agree :)

hrzn (approved):
gj @LeoTafti !

darts/dataprocessing/utils.py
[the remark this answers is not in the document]
LeoTafti: Something occured to me over the weekend looking at common transforms [here](https://otexts.com/fpp2/transformations.html). These transforms don't need to store state (and thus don't need a `fit()` method) and still would be very useful.

This made me realize a few things:
1. Sateless transformers performing basically the same function as a call to `map()` can be very useful. This way, such a call could be chained inside of a `Pipeline`, and provided that the function used with `map()` has an inverse, could also be easily inverted as part of a pipeline.
2. It could be very useful to be able to `map()` a `TimeSeries` with a `function(datapoint_timestamp, datapoint)` instead of only `function(datapoint)`.
3. For multivariate `TimeSeries`, we probably also want to add some way of specifying a subset of target series to apply a transform to, in order to be able to create more complex pipelines. It would however complexify them a fair bit, and make implementing one's own transformers more error prone. I can think of a couple ways to go about this, I think it might be worth it to discuss it offline.

To implement 1., since`WrapperDataTransformer` was quite convoluted and we only need a non-invertible and an invertible wrapper, my proposition is to add two classes instead, one that derives from `BaseTransformer` and one that derives from `InvertibleTransformer`. I'm not sure about how to name them yet, what do you say about `FunctionWrapper` and `InvertibleFunctionWrapper` ?

TheMP (approved):

PR #175, opened 2020-08-05
Feature/ensembling groe

Droxef:
Final contribution from M4 competition benchmark
Implements Darts-176

### Summary

Implements:
* GROE cross-validation function
* Combination base class
* Combinaton using GROE
* Unittests for all three


darts/utils/cross_validation.py
```diff
@@ -0,0 +1,109 @@
...
+from ..logging import get_logger, raise_if_not, raise_log
+from typing import Union, Callable, Optional
+import numpy as np
+
+from ..models.forecasting_model import ForecastingModel
+
+logger = get_logger(__name__)
+
+
+def generalized_rolling_origin_evaluation(ts: TimeSeries, model: ForecastingModel,
```
TheMP, line 17: can you add a few unit tests please?

LeoTafti:
I agree that CombinationModel class would be nice to have.

Remarks:

- There is some overlap between RegressionModel and CombinationModel, maybe RegressionModel could be made a subclass of CombinationModel ?
- Readability of generalized_rolling_origin_evaluation() in cross_validation.py could be improved

darts/models/combination_model.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import List
+
+from ..timeseries import TimeSeries
+from ..logging import get_logger, raise_log, raise_if_not
+from ..models.forecasting_model import ForecastingModel
+
+logger = get_logger(__name__)
+
+
+class CombinationModel(ABC):
```
hrzn, line 16: How about calling it `EnsembleModel` ?

darts/models/groe_combination_model.py
```diff
@@ -0,0 +1,75 @@
...
+        self.weights = None
+
+    def update_groe_params(self, **groe_kwargs):
+        if "n_evaluations" in groe_kwargs:
+            self.n_evaluations = groe_kwargs.pop("n_evaluations")
+        self.groe_kwargs = groe_kwargs
+
+    def fit(self, train_ts: TimeSeries):
+        super().fit(train_ts)
+        self.criterion = []
```
hrzn, line 58: I would call this `self.criteria`

darts/models/groe_combination_model.py
```diff
@@ -0,0 +1,75 @@
...
+
+        raise_if(np.inf in self.criterion,
+                 "Cannot evaluate one of the models on this TimeSeries. Choose another fallback method",
+                 logger)
+
+        if 0. in self.criterion:
+            self.weights = np.zeros(len(self.criterion))
+            self.weights[self.criterion.index(0.)] = 1.
+        else:
+            score = 1 / np.array(self.criterion)
```
hrzn, line 71: `scores`

hrzn (commented):
It would be nice to have such a functionality, so let's keep this in the backlog for a future release. Before we merge though, it would be nice to uniformize the `generalized_rolling_origin_evaluation` to make it similar (or even better: compatible) with the other `*Model.backtest()` functions.

LeoTafti:
I'm not fully convinced by the choice of some of the default values in `groe()` (`generalized_rolling_origin_evaluation()`). I must say that I haven't read [the paper](https://arxiv.org/ftp/arxiv/papers/1503/1503.03529.pdf) carefully (yet), but I tried looking up where those value came from and couldn't find them

LeoTafti:
Still remains to add (or at least would be very nice) one concrete implementation of `EnsembleModel` which would wrap around ensemble models which provide `fit()` and `predict()` methods (ie. do something similar to `StandardRegressionModel`) which allows to use models from scikit-learn (e.g. `LinearRegression` or others from `sklearn.ensemble`

darts/utils/cross_validation.py
```diff
@@ -0,0 +1,129 @@
...
+from darts.timeseries import TimeSeries
+from darts.logging import get_logger, raise_if_not, raise_if
+from darts.models.forecasting_model import ForecastingModel
+import darts.metrics as metrics
+from darts.utils import get_index_at_point
+
+logger = get_logger(__name__)
+
+
+def generalized_rolling_origin_evaluation(series: TimeSeries,
```
hrzn, line 19: @LeoTafti Couldn't we replace this by the new backtest() method you're working on?

darts/models/ensemble_model.py
```diff
@@ -0,0 +1,66 @@
...
+        raise_if_not(all(isinstance(model, ForecastingModel) for model in models),
+                     "All models must be instances of darts.models.ForecastingModel",
+                     logger)
+        super().__init__()
+        self.models = models
+
+    def fit(self, training_series: TimeSeries, target_series: Optional[TimeSeries] = None) -> None:
+        super().fit(training_series, target_series)
+
+        for model in self.models:
```
hrzn, line 35: Not sure the models should be fit here.
If you take the example of GROE, the GROE function fits them all anyway, so no need to call model.fit() here. In general I think it looks like it should depend on the implementation, and hence not be in the base class.

darts/models/groe_ensemble_model.py
```diff
@@ -0,0 +1,77 @@
...
+                 **groe_kwargs):
+        """
+        Implementation of an EnsembleModel using GROE to compute the weights.
+
+        The weights are a function of the loss function of the GROE cross-validation scheme.
+        The weights for each constituent model's output are computed as the inverse of the
+        value of the loss function obtained by applying GROE on that model, normalized such
+        that all weights add up to 1.
+
+        Disclaimer: This model constitutes an experimental attempt at implementing ensembling using
```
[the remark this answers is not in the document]
LeoTafti: Probably not, no.

darts/utils/cross_validation.py
LeoTafti: Good point, they would indeed be almost identical.

The most notable difference I see is that `groe()` allows to specify a maximum number of prediction steps via `n_prediction_steps`. If we want to keep this (but do we ? I'd say no), I could move this functionality over to `backtest()`.

darts/models/ensemble_model.py
LeoTafti: Good catch, I totally missed that this was fitting the models twice.

I think we can indeed remove the `model.fit()` calls from the `EnsembleModel` abstract base class, but I might want to add a note in the `EnsembleModel.fit()` docstring to remember users inherinting from the class that this doesn't call `fit()` on the individual models and that it's the child's responsibility to call do so in its own `fit()` method. What do you think ?

hrzn: Yes, that sounds good!

darts/utils/cross_validation.py
LeoTafti: I did some more thinking and thought that a nicer way would be to provide a `stop` parameter instead.

But this might actually not be needed at all: it's already possible (and takes only one line) to trim the end of the series first before calling `historical_forecasts()` or `backtest()` to achieve the same result.

Since `historical_forecasts()` and `backtest()` signatures are already long/complicated enough imo, I'd be for leaving this out. What do you think ?

LeoTafti:
Following our discussion with @hrzn:
- I removed `generalised_rolling_mean_evaluation()` (aka `groe()`). Prefer using the new `backtest()` method instead
- I thus also removed `GROEEnsembleModel` model. The idea of having an ensemble model which uses error scores (which would now be obtained from the `backtest()` function) to attribute weights to its constituent models is nice, but the implementation of such a model using `backtest()` might not be trivial. Set aside, possible for a future PR.

darts/utils/cross_validation.py
hrzn: Yes, sounds good

darts/models/baselines.py
```diff
@@ -97,3 +99,26 @@ def predict(self, n: int):
...
+        super().__init__(models)
+
+    def fit(self, training_series: TimeSeries, target_series: Optional[TimeSeries] = None) -> None:
+        super().fit(training_series, target_series)
+
+        for model in self.models:
+            if isinstance(model, UnivariateForecastingModel):
+                model.fit(training_series)
+            else:
+                model.fit(training_series, target_series)
```
LeoTafti, line 121: This isn't very clean, but I had to resort to this because calling `model.fit(training_series, target_series)` on a univariate forecasting model raises an error.

 This hints too towards removing the distinction between `UnivariateForecastingModel` and `MultivariateForecastingModel` as suggest in #231.

hrzn: Yes, don't worry too much about it, it will disappear if everything goes as planned :)

hrzn (approved):

Issue #176, opened 2020-08-12
Make setup lighter


plorenzatto:
**Is your feature request related to a current problem? Please describe.**
Currently the default setup of has ~2.4 GB of dependencies. This make it unusable for AWS lambdas or other resource constrained environments.

**Describe proposed solution**
Change the setup so that the heaviest dependencies can be optionally installed:
Eg:
```
pip install u8darts[core] # Bare minimun
pip install u8darts[torch] # ~1.5G (1,4G torch +74M caffe2 + (14+3.4)M tensorboard)
pip install u8darts[fbprophet] # ~500M (31M  fbprophet + 474M pystan)
pip install u8darts[statsmodels] # 42M 
```

TheMP:
Hi @plorenzatto, 

thank you for bringing this up - this is a very good point that's been also in our backlog for a while already.

Is it something you would like to maybe contribute yourself/are you okay with waiting until we deliver this on our side?

plorenzatto:
I'm planning to work on and send a PR when it's ready.

I have some questions about this:
* Which functionalities/modules must absolutely work with the default installation?
* How did you originally thought of solving this issue?

plorenzatto:
@TheMP I'm working on #177 on this. Could you give me some input on the points that I mentioned in the previous message?

pennfranc:
@plorenzatto Thanks a lot for contributing, I think your proposed changes could be very useful! I think making torch and fbprophet optional like you suggested makes sense. But I'm a bit unsure when it comes to statsmodels, since in addition to being used in 3 models, it is also used in some functions from our `statistics` module.

I believe these parts should work without any compromises in the core version:
- `TimeSeries` class
- `backtesting` module (without TorchForecastingModel support)
- at least some models from the `models` module
- `preprocessing` module
- `utils` module (except the torch part)
- `metrics` module

Pretty much everything except some of the more fancy forecasting models for now I believe.

plorenzatto:
I agree with the general idea @pennfranc.

Indeed `statsmodels` is a dependency of `metrics` via `statistics` so better to keep it in the main requirements.

I have some doubts:
* `darts/tests/test_dataset.py` depends on `..models.torch_forecasting_model._TimeSeriesSequentialDataset` not sure if it's meant to be `torch` specific or not.
* `backtesting.explore_models` uses `AutoARIMA` and `Prophet` by default. Can we remove them?

pennfranc:
Good points:
[...]

PR #177, opened 2020-08-14
WIP Moved some deps to extras_require

plorenzatto:
Addresses #176 

### Summary

This PR makes some dependencies optional in order to make the base install much lighter.

TODOs:

- [x] Move deps from `requirements/main.txt` to `extras_require` arg in `setup.py`
- [x] Document the change
- [ ] Run tests after installing optional dependency
- [x] Bump version


plorenzatto:
@hrzn @TheMP @pennfranc Should I bump the major version?

pennfranc:
@plorenzatto There is no need for you to bump the version, we will handle that part!

requirements/main.txt
```diff
@@ -1,14 +1,8 @@
...
-ipywidgets==7.5.1
\ No newline at end of file
+numpy>=1.18.1
+scipy>=1.4.1
+statsmodels>=0.11.1
+matplotlib>=3.2.1
+pandas>=0.23.1
+ipython>=7.14.0
+tqdm>=4.46.0
+holidays>=0.10.2
```
plorenzatto, line 8: @pennfranc Should I use strict version numbers or allow greater or equal versions?

plorenzatto:
@pennfranc The tests are failing because of missing deps:
* One is sklearn. It seems that it was installed along one of the other now optional deps. I'll add it to the main requirements.
* The other issue is that the tests+docs pipeline assume that the now extra deps are already installed and cached so it needs to be changed. Could you give me some input on how to do this?

darts/models/__init__.py
```diff
@@ -3,15 +3,37 @@
...
+
+try:
+    from .rnn_model import RNNModel
+except ModuleNotFoundError:
+    logger.warning("Support Torch based models not available. To enable it install darts[torch]")
+
+try:
+    from .tcn_model import TCNModel
+except ModuleNotFoundError:
+    logger.warning("Support Torch based models not available. To enable it install darts[torch]")
```
plorenzatto, line 29: @pennfranc I placed all the extras in `try` blocks so that the import interface is kept.  
I did something similar by moving `AutoARIMA` outside `arima.py`. This should have the effect that if the correct deps are installed then `from arima import *` should import the same forecaster classes.

README.md
```diff
@@ -123,9 +123,16 @@ If what you want to tell us is not a suitable github issue, feel free to send us
 
 ## Installation Guide
 
-### Preconditions
+The default installation supports `ARIMA `, `StandardRegressionModel`, `NaiveMean`, `NaiveSeasonal`, `NaiveDrift`, `Theta`, `FFT` and `ExponentialSmoothing`.
```
pennfranc, line 126:
```suggestion
The default installation supports the following forecasting models: `ARIMA `, `StandardRegressionModel`, `NaiveMean`, `NaiveSeasonal`, `NaiveDrift`, `Theta`, `FFT` and `ExponentialSmoothing`.
```

darts/models/__init__.py
pennfranc: @plorenzatto Sorry for getting back to you so late!
Separating `AutoARIMA` and `ARIMA` seems like a good move to me given the different dependencies. The `try` block structure also seems like a good idea as far as I can see (thoughts @hrzn?).

pennfranc, line 29:
```suggestion
    logger.warning("Support of Torch based models not available. To enable it install darts[torch]")
```
(also above)

darts/models/__init__.py
```diff
@@ -3,15 +3,37 @@
...
+
+from .arima import ARIMA
 from .baselines import NaiveMean, NaiveSeasonal, NaiveDrift
-from .prophet import Prophet
 from .exponential_smoothing import ExponentialSmoothing
-from .rnn_model import RNNModel
-from .tcn_model import TCNModel
-from .theta import Theta, FourTheta
 from .fft import FFT
+from .theta import Theta
```
pennfranc, line 36: Any reason for removing `FourTheta` from the imports?

darts/models/auto_arima.py
```diff
@@ -0,0 +1,63 @@
...
+from pmdarima import AutoARIMA as PmdAutoARIMA
+
+from .forecasting_model import UnivariateForecastingModel
+from ..timeseries import TimeSeries
+from ..logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class AutoARIMA(UnivariateForecastingModel):
```
pennfranc, line 23: You still need to remove the class from `arima.py`, right?

requirements/main.txt
pennfranc: We switched to equal versions recently!

pennfranc:
> @pennfranc The tests are failing because of missing deps:
> 
> * One is sklearn. It seems that it was installed along one of the other now optional deps. I'll add it to the main requirements.
> * The other issue is that the tests+docs pipeline assume that the now extra deps are already installed and cached so it needs to be changed. Could you give me some input on how to do this?

Regarding your second point, one way to solve this should be to add the extra dependencies to `darts/requirements/dev.txt`.

darts/models/__init__.py
plorenzatto: I missed that one I'll put it back.

hrzn (commented):
Thank you for this proposal! It'll be good to have the possibility to install with only Python dependencies when needed.
However, I think it would be better for user experience that installing `darts` installs the full thing (with all deps), and installing `darts[light]` installs only Python dependencies. And then `darts[pmdarima]`, `darts[torch]` and `darts[fbprophet]` would correspond to `darts[light]` augmented with the corresponding dependencies. 
WDYT about that?

hrzn:
@TheMP @endrjuskr @radujica Do you know by any chance if there is a way with pip/setuptools to change our dependencies, so that
* `pip install u8darts` installs "everything" (like now)
* `pip install u8darts[light]` installs only the bare minimum, and then
* `pip install u8darts[torch]` installs the minimum + PyTorch (for example)?
I looked briefly but didn't find a way to do it. Perhaps you guys know better :D

hrzn:
I will be closing this one as we addressed the issue in another PR (also adapting the unit and CI tests). Thanks @plorenzatto !

Issue #171, opened 2020-07-29
Transformer Forecasting Model


Splizard:
With the recent popularity of GPT3, it got me thinking about how a Transformer model could be applied to time-serires forecasting.
Could this sort of  model be put on the 'wishlist' for Darts? 

There was a discussion on Kaggle recently about this and it links to more information about the suject.
https://www.kaggle.com/c/m5-forecasting-accuracy/discussion/142833

I believe the Attention mechanisim of the model is capable of discovering certain patterns that other models struggle with and this could be a better model for certain domains of time-series forecasting.

There is a pytorch implementation here:
https://github.com/maxjcohen/transformer

a recent paper on the subject: 
Deep Transformer Models for Time Series Forecasting:The Influenza Prevalence Case
https://arxiv.org/pdf/2001.08317

pennfranc:
This indeed looks like a very cool possibility, thanks for putting this on our radar! This could be a subclass of our `TorchForecastingModel` class.  If you're up for the challenge, you can always consider giving this a shot yourself too (check out [this](https://github.com/unit8co/darts/issues/122#issuecomment-655344527) post for guidelines regarding implementing a new model). For now we will put this on our backlog.

inverniz:
This seems interesting to me as well! I am currently giving it a try (still working on it locally). I will let you know how it goes!

LeoTafti:
Will be part of the next release!

PR #181, opened 2020-09-14
Transformer

inverniz:
Following the discussion in #171, this pull request proposes a simple implementation of an encoder-decoder transformer architecture for time series. The code is reasonably clean and could be merged. However, I'd like to discuss my approach before doing that so I opened a draft pull request.

#### Brief description of the _transformer_

This pull request implements a deep learning model called _TransformerModel_. _TransformerModel_ leverages the well-known state-of-the-art transformer architecture introduced in 'Attention is All you Need' by Vaswani et al. (2017).

Transformers are often used in Natural Language Processing. However, attempts at using this architecture with time series have been made (see #171).

The model inherits from darts' _TorchForecastingModel_ and is based on PyTorch's _nn.Transformer_ module.

The training flow is the following:
* The input time series have to be embedded in a higher-dimensional space. When doing NLP, tokens are usually vectorized using _nn.Embedding_. In the case of time series, I opted for a simple fully connected layer (_nn.Linear_)
* A positional encoding is added to the embedded time series in order to integrate information about the position of the values in the sequence
* Given n embedded data points _data_=(x<sub>1</sub>, ..., x<sub>n</sub>) and m embedded prediction labels _target_=(x<sub>n+1</sub>, ..., x<sub>n+m</sub>), the encoder-side of the transformer receives as input _encoder\_input_ = _data_ = (x<sub>1</sub>, ..., x<sub>n</sub>), while the decoder-side receives _decoder\_input_= (x<sub>n</sub>, ..., x<sub>n+m-1</sub>) (see Figure 1 of [this paper](https://arxiv.org/pdf/2001.08317.pdf) for a visual representation). A suitable mask is given to the decoder to make sure that prediction at time _n_ does not have access to information related to step _n+1_ or successive.

The inference flow is the same as the training flow, with the difference that:
* Given n embedded data points _data_=(x<sub>1</sub>, ..., x<sub>n</sub>) and no embedded prediction labels, the encoder-side of the transformer receives as input _encoder\_input_ = _data_ = (x<sub>1</sub>, ..., x<sub>n</sub>), while the decoder-side receives _decoder\_input_= (x<sub>n</sub>). 

#### Tests
basic tests have been written. They are slightly modified versions of the RNN tests. Coverage is 89%.

#### Examples

A notebook is available. Again, it is a tweaked version of the RNN notebook. The performance of the transformer and that of the RNN seem to be in the same magnitude when using the 'Air Passenger' dataset (8.41% MAPE after 290 epochs for RNN, 9.69% after 590 epochs for the transformer). However, in-depth analysis to understand if/when transformer performs better than other approaches has yet to be made. The 'Sunpost dataset' hasn't been tested yet since it is more computationally demanding and ideally requires GPU access to be processed quickly.

### Other Information

A few design choices that should be discussed: 
* The transformer architecture is non-recurrent. As such, if we want to forecast the 10 next data points in a time series, we have to perform 10 decoder forward passes. Currently, the _predict_ method has a _use\_full\_output\_length_ attribute that makes sense when using the _RNNModel_, but not with _TransformerModel_, since it cannot do inference with output_length > 1. I have reimplemented the predict method so that it raises a ValueError when _use\_full\_output\_length_ is set to true using _TransformerModel_. If you have other suggestions, please let me know. 
* Several methods such as _\_train_, _evaluate\_validation\_loss_ and others were reimplemented with the only change being the signature of the model's forward pass. That's because the transformer model needs both data *and* labels as input in order to compute _input\_encoder_ and _input\_decoder_ during training. Also, a flag was added to the forward pass indicating whether we are doing inference or training, so that encoder and decoder inputs can be computed accordingly. If you see a better way to do this, please let me know. 
* A subclass of _TimeSeriesSequentialDataset_ called _TimeSeriesSequentialTransformerDataset_ had to be created. That's because the original class returned only the target features of the labels. When using transformers with multivariate time-series, we need to have all labels features to compute _decoder_input_.
* If transformer models were to be fully integrated into darts, there would be a name conflict with the scaler used for preprocessing.

The positional encoding implementation presented in [Pytorch's transformer tutorial](https://pytorch.org/tutorials/beginner/transformer_tutorial.html) has been integrated into the code since it is supposed to be optimal. 

Please do not hesitate to come back to me with feedback.


hrzn:
Many thanks @inverniz ! That looks very promising. We'll review it in depth.

hrzn (changes requested):
Many thanks for this contribution. I think it's great, and we should merge it. My main comment, before we do this, would be to remove most of the methods of `TransformerModel` and rely on `TorchForecastingModel` instead. There are two things currently preventing this:
* At training time the model requires both data and targets => I think this can be implemented as an option in `TorchForecastingModel`.
* The `predict()` method currently requires `use_full_output_length=False` => I suspect we could generalise de decoder so that it can output more than one time step at a time (I've put more detailed comments in the code). Or, if that's not really the case for some reason, we could at least force building the `TransformerModel` with `output_length=1`, which I think would solve this problem.

inverniz:
> Many thanks for this contribution. I think it's great, and we should merge it. My main comment, before we do this, would be to remove most of the methods of `TransformerModel` and rely on `TorchForecastingModel` instead. There are two things currently preventing this:
> 
> * At training time the model requires both data and targets => I think this can be implemented as an option in `TorchForecastingModel`.
> * The `predict()` method currently requires `use_full_output_length=False` => I suspect we could generalise de decoder so that it can output more than one time step at a time (I've put more detailed comments in the code). Or, if that's not really the case for some reason, we could at least force building the `TransformerModel` with `output_length=1`, which I think would solve this problem.

Thank you for your feedback. I will take care of this asap.

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+            Tensor containing the (point) prediction at the last time step of the sequence.
+        """
+
+        super(_TransformerModule, self).__init__()
+
+        self.input_size = input_size
+        self.output_size = output_size
+        self.output_length = output_length
+
+        self.encoder = nn.Linear(input_size, d_model)
```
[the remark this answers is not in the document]
erbas: this is standard terminology, it would be a mistake to change it

inverniz:
_TransformerModel_ has been completely reworked in light of the fact that it was indeed enough to generalize the last decoder layer in order to perform multi-step ahead forecasting. This made me realize that the the setup I used previosly, which was taken from [this paper](https://arxiv.org/pdf/2001.08317.pdf), was in fact not optimal for the forecasting use-case. 
The previous setup worked as follows: 
- Encoder input: (x<sub>1</sub>, ..., x<sub>input_length</sub>) 
- Decoder input during training: (x<sub>input_length</sub>, ..., x<sub>input_length + output_length</sub>)
- Decoder input during inference: (x<sub>input_length</sub)
- Number of time steps output by the model is 1

The current setup works as follows:
- Encoder input: (x<sub>1</sub>, ..., x<sub>input_length</sub>) 
- Decoder input: (x<sub>input_length</sub>)
- Number of time steps output by the model is  _output_length_

In other words, I am now using the _output_length_ parameter to control how many time steps the model will ouput (its intended usage), instead of using it to decide how many time steps to feed to the decoder. This setup makes more sense because I don't see any real need to feed more than one time step to the decoder when working with time series. It is probably enough to input the whole time series in the encoder, and then leverage the cross-attention mechanism implemented in the decoder to extract context information (much like in your implementation of RNNs).

I'll summarize the content of the new commits:
- Multi-step ahead forecasting is now accessible
- since the forward pass doesn't need to access the labels anymore, there's no need to reimplement inherited methods in -- -TransformerModel_ nor to touch _TorchForecastingModel_
- The Multi-step ahead forecasting has been tested on the sun spots dataset and produces results in line with the RNN
- tests have been updated (coverage 100%)

As a side note, I see that some test unrelated to transformer are failing. I suppose that's because they were already failing in the _develop_ branch when I branched from it right?

darts/models/transformer_model.py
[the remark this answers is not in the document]
inverniz: While I agree it could be confusing (since the transformer also has an encoder and a decoder), it's the same terminology used in https://pytorch.org/tutorials/beginner/transformer_tutorial.html, so I also opted to leave it like this

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+                 num_decoder_layers: int = 6,
+                 dim_feedforward: int = 2048,
+                 dropout: float = 0.1,
+                 activation: str = "relu",
+                 custom_encoder: Optional[Any] = None,
+                 custom_decoder: Optional[Any] = None,
+                 random_state: Optional[Union[int, RandomState]] = None,
+                 **kwargs):
+
+        """ Transformer Model.
```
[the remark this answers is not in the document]
inverniz: done

examples/Transformer-examples.ipynb
```diff
@@ -0,0 +1,883 @@
+{
```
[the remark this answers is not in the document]
inverniz: I managed to test on the monthly sunspots dataset

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+                                          num_encoder_layers=num_encoder_layers,
+                                          num_decoder_layers=num_decoder_layers,
+                                          dim_feedforward=dim_feedforward,
+                                          dropout=dropout,
+                                          activation=activation,
+                                          custom_encoder=custom_encoder,
+                                          custom_decoder=custom_decoder)
+
+        self.tgt_mask = self.transformer.generate_square_subsequent_mask(output_length)
+        self.decoder = nn.Linear(d_model, output_size)
```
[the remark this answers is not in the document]
inverniz: Yes this was done in the new version (see the main post below)

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+        """
+
+        super(_TransformerModule, self).__init__()
+
+        self.input_size = input_size
+        self.output_size = output_size
+        self.output_length = output_length
+
+        self.encoder = nn.Linear(input_size, d_model)
+        self.positional_encoding = PositionalEncoding(d_model, dropout)
```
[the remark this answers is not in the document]
inverniz: Since the _positional encoding_ class was not meant to be exposed to the user, I hid it. Also, the transformer only needs _input_length_ values out of the _max_len_ that are produced, so I initialized _max_len_ to that value

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+
+def _get_checkpoint_folder(work_dir, model_name):
+    return os.path.join(work_dir, CHECKPOINTS_FOLDER, model_name)
+
+
+def _get_runs_folder(work_dir, model_name):
+    return os.path.join(work_dir, RUNS_FOLDER, model_name)
+
+
+class _TimeSeriesSequentialTransformerDataset(_TimeSeriesSequentialDataset):
```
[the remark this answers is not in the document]
inverniz: In the previous setup, I needed _get_item_ to return all the features from the label, not only the 'target features'. In the new setup, I don't need this anymore, so I reverted to using `_TimeSeriesSequentialDataset`

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+        else:
+            decoder_src = encoder_src[-1:, :, :]
+            tgt_mask = None
+
+        return encoder_src, decoder_src, tgt_mask
+
+    def forward(self, src, tgt, training=True):
+        encoder_src, decoder_src, tgt_mask = self._create_transformer_inputs(src, tgt, training)
+
+        encoder_src = self.encoder(encoder_src) * math.sqrt(self.input_size)
```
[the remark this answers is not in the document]
inverniz: It's a normalization factor (see section 3.2.1 of the 'Attention is all you need' paper)

darts/models/transformer_model.py
```diff
@@ -0,0 +1,477 @@
...
+        self.decoder = nn.Linear(d_model, output_size)
+
+    # if Training = 'True', we are in training mode:
+    # given src=(x_0, ..., x_n) and tgt=(x_n+1, ..., x_n+m), then
+    # encoder_src = src and decoder_src = (x_n, x_n+1, .. x_n+m-1)
+    # if Training = 'False', then we are in inference mode:
+    # given src=(x_0, ... x_n) and tgt=(), then
+    # encoder_src = src and decoder_src = (x_n)
+    def _create_transformer_inputs(self, src, tgt, training):
+        encoder_src = src.permute(1, 0, 2)
```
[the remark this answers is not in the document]
inverniz: done

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Parameters
+        ----------
+        model
+            a custom PyTorch module with the same specifications as
+            `darts.models.transformer_model._TransformerModule` (default=None).
+        input_size
+            The dimensionality of the TimeSeries instances that will be fed to the fit function (default=1).
+        input_length
+            Number of time steps to be output by the forecasting module (default=1).
```
hrzn, line 226: This is inaccurate (we also have this typo, which we should eventually correct, in the RNN docstring). It's the number of steps that will be input to the transformer module.

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+        num_decoder_layers
+            the number of decoder layers in the decoder (default=6).
+        dim_feedforward
+            the dimension of the feedforward network model (default=2048).
+        dropout
+            Fraction of neurons affected by Dropout (default=0.1).
+        activation
+            the activation function of encoder/decoder intermediate layer, 'relu' or 'gelu' (default='relu').
+        custom_encoder
+            a custom transformer encoder provided by the user (default=None)
```
hrzn, line 246: I would say "a custom user-provided encoder module for the transformer"
(this is because "provided by the user" sounds weird, as the user will likely be the person reading this docstring).

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+        custom_decoder
+            a custom transformer decoder provided by the user (default=None)
```
hrzn, line 248: "a custom user-provided decoder module for the transformer"

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+                 output_size: int = 1,
+                 d_model: int = 512,
+                 nhead: int = 8,
+                 num_encoder_layers: int = 6,
+                 num_decoder_layers: int = 6,
+                 dim_feedforward: int = 2048,
+                 dropout: float = 0.1,
+                 activation: str = "relu",
+                 custom_encoder: Optional[Any] = None,
+                 custom_decoder: Optional[Any] = None,
```
hrzn, line 199: I think `Optional[nn.Module]` would be more accurate in the above 2 lines.

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+        predictions = out[0, :, :]
+        predictions = predictions.view(-1, self.output_length, self.output_size)
+
+        return predictions
+
+
+class TransformerModel(TorchForecastingModel):
+    @random_method
+    def __init__(self,
+                 model: nn.Module = None,
```
hrzn, line 186: `Optional[nn.Module]`

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+                 output_size: int,
+                 d_model: int,
+                 nhead: int,
+                 num_encoder_layers: int,
+                 num_decoder_layers: int,
+                 dim_feedforward: int,
+                 dropout: float,
+                 activation: str,
+                 custom_encoder: nn.Module = None,
+                 custom_decoder: nn.Module = None,
```
hrzn, line 78: `Optional[nn.Module]` (twice)

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+        """ PyTorch module implementing a Transformer to be used in `TransformerModel`.
+
+        PyTorch module implementing a simple encoder-decoder transformer architecture.
+
+        Parameters
+        ----------
+        input_size
+            The dimensionality of the TimeSeries instances that will be fed to the fit function.
+        input_length
+            Number of time steps to be output by the forecasting module.
```
hrzn, line 89: input to (instead of output by)

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+                 custom_decoder: nn.Module = None,
+                 ):
+        """ PyTorch module implementing a Transformer to be used in `TransformerModel`.
+
+        PyTorch module implementing a simple encoder-decoder transformer architecture.
+
+        Parameters
+        ----------
+        input_size
+            The dimensionality of the TimeSeries instances that will be fed to the fit function.
```
hrzn, line 87: We have this typo in several places, but in fact it should be "the fit and predict functions"

darts/models/transformer_model.py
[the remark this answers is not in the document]
hrzn: OK, not a big fan of this terminology, but fair enough ;)

darts/models/transformer_model.py
```diff
@@ -0,0 +1,281 @@
...
+                                          dropout=dropout,
+                                          activation=activation,
+                                          custom_encoder=custom_encoder,
+                                          custom_decoder=custom_decoder)
+
+        self.tgt_mask = self.transformer.generate_square_subsequent_mask(output_length)
+        self.decoder = nn.Linear(d_model, output_length * output_size)
+
+    def _create_transformer_inputs(self, data):
+        # '_TimeSeriesSequentialDataset' stores time series in the
```
hrzn, line 148: thanks :)

hrzn (commented):
That looks good, thanks @inverniz! I have a few very minor comments. 

Also, before we can merge, it would be good to adapt the unit tests to make them conditional on Torch being installed. You can have a look at a (admittedly not super clean) way for doing that in the RNN test cases on the latest version of the develop branch. This is needed because in the upcoming release we'll have Torch as an optional dependency, so some users will have Darts installed without Torch.

inverniz:
Thanks again for the feedback. I have addressed the three issues (docstring corrections, argument types corrections, unit tests conditional on Torch) in the two latest commits. Should I also merge the latest changes from the develop branch, or will you take care of that?

hrzn (approved):
🚀

darts/models/__init__.py
```diff
@@ -12,6 +12,7 @@
 from .theta import Theta, FourTheta
 from .arima import ARIMA
 from .fft import FFT
+from .transformer_model import TransformerModel
```
hrzn, line 15: @inverniz Could you move this import to after line 30 (inside the try/except)? That should make the tests pass.

inverniz: @hrzn I pulled your changes, moved the import line, and updated the TransformerModel tests to align with the most recent RNN tests. Locally, I still have three tests failing that appear to be unrelated to my changes. Please see the attached screenshot. The three tests seem to fail because performance of a few model is below a certain threshold. I am running the tests by launching the coverage.sh script from the terminal.
Schermata 2020-10-08 alle 10 43 58

inverniz: After looking at the test logs of the PR, it seems that the failing task is "coverageTest_core". Indeed, I see that the overall coverage is only 77% (I suppose you want 80 or more) because TransformerModel has only 5% test coverage. This is weird because locally I have 87% overall coverage and 100% coverage for TransformerModel (see screenshot). Maybe I still have to import the tests for TransformerModel somewhere? 
Schermata 2020-10-08 alle 13 16 19

TheMP: Hi @inverniz, what is hitting you might be caused by the fact that we accepted a PR that breaks down the code into models to develop, which might change test/coverage results. Can you make sure you are up-to-date with the develop branch and retrigger the tests?

Sorry for the issues and pls ping me if it doesn't help.

LeoTafti: Hi @inverniz,

I'm not quite sure yet why the first tests are failing but we'll take a deeper look.

Concerning the coverage tests, it's due to a mistake on our side with the way we run them in the `build.gradle` file. I'm working on fixing this and will report back to you.

inverniz: HI @LeoTafti and @TheMP, thank you for your inputs. I am also looking into the issue. It seems to me that the latest version of the develop branch was successfully merged in my PR (the last commit in the develop branch is 2 days old, @hrzn did the merge yesterday), so that shouldn't be the issue. I will let you know when I find out more.

LeoTafti: I believe it worked because we got lucky, test coverage when running `./gradelew coverageTest_core` barely reached the minimum 80% if you look at [this run](https://github.com/unit8co/darts/runs/1214649410) and the same behavior observed for the `TransformerModel` can be seen for other files which require dependencies not included in the `core` flavor of the library as well (e.g. line 49 of the same run states `darts/models/rnn_model.py                        44     43     2%   7-172`).

inverniz: I see that #212 has been merged. I suppose I should update the branch and trigger the tests again, correct?

LeoTafti: Correct! It should work now, ping me if that's not the case

LeoTafti:
Looks like all checks are passing now @inverniz !

One (last) problem though, I tried to run the `Transformer-examples` notebook and noticed that it was using functions and classes that we now renamed / refactored. I tried to correct that, but now it seems like the notebook is failing for another reason. Maybe you could have an idea why ?

(Btw this also made me realize that checking that the examples run correctly should be part of the develop workflow. I'll address this separately)

inverniz:
Thanks for fixing the doc build issue @LeoTafti! I will have a look at the updated jupyter notebook asap!

inverniz:
I corrected a few signatures in the code and was able to run everything just fine. Everything should be ok now.

LeoTafti:
Awesome work @inverniz thanks for your contribution! 🎉 This PR has been merged to develop and will be part of the next release (coming soon in 1-2 weeks).

inverniz:
Great to hear that! I will be happy to contribute in the future as well, since there are definitely ways to optimize the proposed transformer architecture to make it more suitable to time series processing!

PR #185, opened 2020-09-17
Feature/prophet example

LeoTafti:
### Summary

Adds a notebook with a simple FB Prophet example

### Other Information

Would be nicer with a daily time series dataset which allows to model holiday effect


examples/Prophet-examples.ipynb
```diff
@@ -0,0 +1,234 @@
...
+    "\n",
+    "series.plot(label='actual')\n",
+    "forecast.plot(label='prediction')\n",
+    "plt.title('MAPE = {:.2f}%'.format(mape(series, forecast)))\n",
+    "plt.legend()"
+   ]
+  },
+  {
+   "source": [
+    "## Modelling hollidays and special events\n",
```
pennfranc, line 157:
```suggestion
    "## Modelling holidays and special events\n",
```

examples/Prophet-examples.ipynb
```diff
@@ -0,0 +1,234 @@
...
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2,
+ "cells": [
+  {
+   "source": [
+    "# Facebook Prophet\n",
+    "\n",
+    "In this notebook, we show examples of how Facebook Prophet can be used with darts. This model is especially suited for time series with strong seasonal effects and allows to specify hollidays or special events.\n",
```
pennfranc, line 28:
```suggestion
    "In this notebook, we show examples of how Facebook Prophet can be used with darts. This model is especially suited for time series with strong seasonal effects and allows to specify holidays or special events.\n",
```

hrzn (commented):
Thanks for creating the notebook!

I see this message when trying to run the Prophet notebook:
image
Setting the kernel makes it work. But perhaps have you specified a kernel in the notebook?

Otherwise, the section concerning handling holidays is empty (an example would be nice, if you find one where it helps or changes the forecast).

Finally, there is still the Latex example cell at the end of the notebook.

LeoTafti:
> I see this message when trying to run the Prophet notebook:
> image
> Setting the kernel makes it work. But perhaps have you specified a kernel in the notebook?

Indeed I had to specify a kernel for it to work in VSCode, I should've reverted back to the original one, my bad.

> Otherwise, the section concerning handling holidays is empty (an example would be nice, if you find one where it helps or changes the forecast).

Yes definitely, but I didn't find an appropriate dataset looking around quickly and then had more pressing changes to make :). I'll come back to it later – closing this PR for now.

PR #187, opened 2020-09-18
Fix/backtest sanity checks

### LINKED ISSUE #186: ForecastingModel.backtest() fails if forecast_horizon isn't provided
https://github.com/unit8co/darts/blob/2977f4f666e255b617ad980eab81d6ff6aad9db3/darts/models/forecasting_model.py#L148-L157

`n.forecast_horizon` at line 157 fails when `hasattr(n, 'forecast_horizon') == false`, ie. every time no explicit value is passed when calling `backtest()`

Error produced:
```
AttributeError: 'types.SimpleNamespace' object has no attribute 'forecast_horizon'
```

LeoTafti:
Fixes #186.

### Summary

Modified the sanity checks to use either the provided forecast_horizon parameter or the default value.
Also made a small improvement to the `_with_sanity_checks()` decorator

### Other Information

Had to move `_backtest_sanity_checks()`'s definition after `backtest()` for `signature(self.backtest)` to work, which is a little less nice. We should be able to undo that once we address https://github.com/unit8co/darts/pull/125#discussion_r489286538

### DIFF
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -11,6 +11,7 @@ where :math:`y_t` represents the time series' value(s) at time :math:`t`.
 
 from typing import Optional, Tuple, Union, Any, Callable
 from types import SimpleNamespace
+from inspect import signature
 from itertools import product
 from abc import ABC, abstractmethod
 import numpy as np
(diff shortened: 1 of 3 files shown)


darts/models/forecasting_model.py
```diff
@@ -292,6 +235,68 @@ def backtest(self,
 
         return forecast
 
+    def _backtest_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
```
pennfranc, line 238: Did you make any changes to this function except moving it?

darts/utils/utils.py
```diff
@@ -110,6 +111,7 @@ def fit(self, a, b=0, c=0):
             ...
     """
     def decorator(method_to_sanitize: Callable[[A, B], T]) -> Callable[[A, B], T]:
+        @wraps(method_to_sanitize)
```
pennfranc, line 114: How does this change the behavior of `_with_sanity_checks` exactly?

pennfranc (approved):

hrzn (approved):
thanks!

LeoTafti:
Solved by #189

Issue #188, opened 2020-09-18
[BUG] ForecastingModel.backtest: Can bypass sanity checks


LeoTafti:
**Describe the bug**
Invalid optional arguments to `ForecastingModel.backtest` can bypass sanity checks when provided as positional arguments

**To Reproduce**
```
NaiveDrift().backtest(linear_series, None, start=0.7, forecast_horizon=-1)
NaiveDrift().backtest(linear_series, None, 0.7, -1)
```

**Expected behavior**
The first line above is caught by the sanity checks and raise `ValueError: 'The provided forecasting horizon must be a positive integer.'` as expected

For the second line however the negative `forecast_horizon` param clears the sanity checks and produces an error further down the line:
```
NaiveDrift().backtest(linear_series, None, 0.7, -1)
  File ".../darts/utils/utils.py", line 120, in sanitized_method
    return method_to_sanitize(self, *args, **kwargs)
  File ".../darts/models/forecasting_model.py", line 217, in backtest
    pred = self.predict(forecast_horizon, **predict_kwargs)
  File ".../darts/models/baselines.py", line 98, in predict
    forecast = np.linspace(last, last_value, num=n)
  File "<__array_function__ internals>", line 5, in linspace
  File ".../lib/python3.8/site-packages/numpy/core/function_base.py", line 115, in linspace
    raise ValueError("Number of samples, %s, must be non-negative." % num)
ValueError: Number of samples, -1, must be non-negative.
```
**System (please complete the following information):**
 - Python version: 3.8.5
 - darts version: develop branch, commit #2977f4f

pennfranc:
Good catch! That's definitely a drawback of the current sanity checks implementation. It mostly relies on arguments being passed as keywords right now.. Do you think this could be a possible solution: https://stackoverflow.com/a/831164?

PR #189, opened 2020-09-22
Refactor/backtest checks

LeoTafti:
### Summary

- Changed `RegressionModel.backtest()` `start` parameter type to match `ForecastingModel.backtest()` signature.
- Factored out some code used in both forecasting and regression model to perform sanity checks on arguments and fixed a couple of bugs with these checks.
- Added a check on `start` to make sure it doesn't correspond to the first timestamp in the series, since this would leave no points in the training set and make it impossible to make a first prediction.
- Added a couple more unit tests for both forecasting and regression model `backtest()` methods.

### Things I'm unsure about

- I set the default value of `start` for `StandardRegressionModel` to be equal to `train_n_points`. I think that this is the most intuitive choice.
- I'm not sure which default value to give to `start` in the general case (ie. in `RegressionModel.backtest()`). I set it to 1, since that is the smallest possible index which makes sense.
- I put `_backtest_general_checks()` and `_backtest_convert_start()` in `utils.py` since I couldn't think of a better place to put code in common for both ForecastingModel and RegressionModel, but there might be a better place.

Ran tests locally, all green.


LeoTafti:
Also what would think about giving a warning when `StandardRegressionModel.fit()` is called with `train_n_points` larger than the number of available training points in `training_series`?

LeoTafti:
Fixes #188

darts/utils/utils.py
```diff
@@ -110,9 +117,84 @@ def fit(self, a, b=0, c=0):
...
+    # check that trim_to_series and start together form a valid combination
+    trim_to_series = n.trim_to_series
+
+    if trim_to_series:
+        raise_if_not(start + series.freq() * forecast_horizon in series,
+                     '`start` timestamp is too late in the series to make any predictions with'
+                     '`trim_to_series` set to `True`.', logger)
+
+
+def _backtest_convert_start(start: Union[pd.Timestamp, float, int], training_series: TimeSeries) -> pd.Timestamp:
```
hrzn, line 194: I think this method might also be useful outside of backtesting as well. You could rename it to reflect this. For instance to something like `get_ts_at_point(point: Union[pd.Timestamp, float, int], series: TimeSeries) -> pd.Timestamp`.

In fact, I think later on (not for this PR) we could even think of having such "fancy indexing" natively supported in `TimeSeries`, as well as in some `TimeSeries` methods such as `split_after`, `split_after`, `slice`, etc.

darts/models/standard_regression_model.py
```diff
@@ -70,3 +70,24 @@ def predict(self, features: List[TimeSeries]):
         super().predict(features)
         y = self.model.predict(self._get_features_matrix_from_series(features))
         return TimeSeries(pd.DataFrame(y, index=features[0].time_index()), self.train_target.freq())
+
+    def backtest(self,
```
hrzn, line 74: I don't think it's worth to rewrite the backtest() method here just to add the default `start = self.train_n_points`.

With this model, `self.train_n_points` can be set to `None`, in which case the model will just retrain on whatever is available. I think we could assume that this is the "default", and remove this backtest() implementation. We should however slightly change the `__init__` function so that the argument becomes `train_n_points: Optional[int]=None`.

darts/models/regression_model.py
```diff
@@ -84,11 +84,41 @@ def predict(self, features: List[TimeSeries]) -> TimeSeries:
...
+        """Method to be overriden in subclass for model specific sanity checks"""
+        pass
+
+    @_with_sanity_checks("_backtest_sanity_checks", "_backtest_model_specific_sanity_checks")
     def backtest(self,
                  feature_series: Iterable[TimeSeries],
                  target_series: TimeSeries,
-                 start: pd.Timestamp,
-                 forecast_horizon: int,
+                 start: Union[pd.Timestamp, float, int] = 1,
```
hrzn, line 120: Not too sure about having a default value for this parameter. Or if we have one, we should be consistent with ForecastingModel and put 0.5. (although I think 0.7 on both models would be better, actually)

darts/tests/test_backtesting.py
```diff
@@ -115,6 +140,24 @@ def test_backtest_regression(self):
...
+        # Using a too small start value
+        with self.assertRaises(ValueError):
+            StandardRegressionModel(15).backtest(features, target, start=0, forecast_horizon=3)
+
+        with self.assertRaises(ValueError):
+            StandardRegressionModel(15).backtest(features, target, start=0.01, forecast_horizon=3)
+
+        # Using StandardRegressionModel's start default value
+        pred = StandardRegressionModel(15).backtest(features, target, forecast_horizon=3)
+        self.assertEqual(r2_score(pred, target), 1.0)
```
hrzn, line 159: 👍   for the tests :)

hrzn (commented):
Nice work :)

darts/utils/utils.py
LeoTafti: I think so as well.

Being able to at least split a `TimeSeries` according to some ratio or index would indeed be really nice, and this method could help there.

darts/models/standard_regression_model.py
LeoTafti: > With this model, self.train_n_points can be set to None, in which case the model will just retrain on whatever is available. I think we could assume that this is the "default", and remove this backtest() implementation.

Indeed, `train_n_points` can totally be made Optional assuming this default behavior.

However I still think that it would be somewhat misleading to silently train on less than `train_n_points` when `train_n_points` has been explicitly specified. What do you think ?

darts/models/regression_model.py
LeoTafti: I wasn't too sure either.
The problem I see with making it 0.5 (or 0.7) as in ForecastingModel is that in the both model don't "work" with the same input series:

- In the ForecastingModel case, the series `backtest()` is "working" on (ie. `training_series`) is typically the "whole" series dataset (e.g. from 2000-01-01 to 2020-01-01)
- Whereas RegressionModel uses  `features_series` which are mostly already forecasts of other models (and would e.g. go each go from **2010**-01-01 to 2020-01-01.

I think the typical use case of RegressionModel would thus have a `start` param pointing to the very beginning of `feature_series`, and not at 50% or 70%

hrzn:
> Also what would think about giving a warning when `StandardRegressionModel.fit()` is called with `train_n_points` larger than the number of available training points in `training_series`?

That's a good idea

darts/models/standard_regression_model.py
hrzn: That's a good point. Like you suggest, we should probably display a warning when `train_n_points` has been specified and the model is fit on fewer points. Then in `backtest` the warning would also be displayed as it calls the fit() method.

darts/models/regression_model.py
hrzn: Fair point, however there is still the notion of training and validation set. So I think it remains intuitive (at least to me) that you would train regression models on the first 70% of the data available (so from 2010 to 2017) and then validate on the recent past only. Training on 1 data point is a bit awkward. So I'd vote for putting 70% everywhere or, if you really don't like it, to drop this default value altogether from `RegressionModel.backtest()`.

LeoTafti: Fair enough, I think we should go for 70% by default everywhere then.

hrzn (approved):
Thanks @LeoTafti !

PR #196, opened 2020-10-01
[WIP] Feature/lighter install

hrzn:
We just need to fix the matrix CI tests before merging.


requirements/core.txt
```diff
@@ -1,14 +1,9 @@
...
 matplotlib==3.2.2
-fbprophet==0.6
 pandas==1.1.0
-torch==1.5.1
-tensorboard==2.2.2
 ipython==7.15.0
 tqdm==4.46.1
 holidays==0.10.2
-plotly==4.9.0
-ipywidgets==7.5.1
```
pennfranc, line 14: Do we need to run some checks to see whether removing `plotly` and ìpywidgets` will not result in any unexpected behavior? I'm just thinking that there must have been a reason for why we added those..

.github/workflows/develop.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         id: cache
         with:
           path: ~/.cache/pip
-          key: tests-${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('requirements/main.txt', 'requirements/dev.txt') }}
+          key: tests-${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('requirements/core.txt', 'requirements/dev.txt') }}
```
pennfranc, line 34: The test cases need more than just the core requirements, right? (Since they touch all models)

TheMP (approved):

requirements/core.txt
hrzn: I think they are old remnants. I'll trust the unit tests (and mostly the notebook examples tests) on that one :)

.github/workflows/develop.yml
hrzn: Right, that's corrected.

Issue #192, opened 2020-09-28
Unified API  : `model.params` attribute


norihitoishida:
Hello! Firstly, I wanted to thank you for this wonderful library.

Unified API to get model's params might make darts more convenient.
For example, to get params of `Exponential Smoothing` model, We have to code like this:
```
model = ExponentialSmoothing()
model.fit(train)

print(model.model.model.params) # {'smoothing_level': 0.5789473661331209, 'smoothing_slope': ...
```

So, I want to implement ~~`model.get_params()` method like [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html#sklearn.svm.SVC.get_params).~~ `model.params` attribute with property like statsmodels and fbprophet.

I would like to get your opinion before send PR. Thanks!

- [ ] Models
     - [ ] abstract `params()` function in the `ForecastingModel` superclass
     - [ ] ARIMA
     - [ ] AutoARIMA
     - [ ] Baseline Models
          - [ ] NaiveDrift
          - [ ] NaiveMean
          - [ ] NaiveSeasonal
     - [x] ExponentialSmoothing
     - [ ] FFT
     - [x] Prophet
     - [ ] StandardRegressionModel
     - [ ] TCNModel
     - [ ] Theta
     - [ ] TorchForecastingModel (RNNModel)
- [ ] Preprocessing
     - [ ] Scaler wrapper

LeoTafti:
Hello @norihitoishida, thanks for your suggestion and offer to help! I think that would be a nice addition to our library.

Make sure to check out [our guidelines](https://github.com/unit8co/darts#contribute) first and to include a couple unit tests. Thanks!

norihitoishida:
Thanks @LeoTafti !

hrzn:
+1 for that, it's a good idea

norihitoishida:
Related to #194, Unittest fails in my development environment, so I will prioritize resolving that issue.

norihitoishida:
I found it more intuitive to use the `model.params` attribute instead of the `model.get_params()` method (following the fbprophet and statsmodels).

norihitoishida:
Sorry, due to my personal health issues, I have found that it will take some time to complete this task. 
Let me close this once.

hrzn:
Take care @norihitoishida

norihitoishida:
Thank you for your kindness : )

PR #197, opened 2020-10-01
Add params attribute to ExponentialSmoothing

norihitoishida:
Related to #192.

### Summary
Add `model.params` attribute to ExponentialSmoothing()

```
# Before
model = ExponentialSmoothing()
model.fit(train)
print(model.model.model.params) # {'smoothing_level': 0.5789473661331209, 'smoothing_slope': ...

# After
model = ExponentialSmoothing()
model.fit(train)
print(model.params) # {'smoothing_level': 0.5789473661331209, 'smoothing_slope': ...
```

### Other Information
Should I add any unittest for related to this PR?
I would like to get your opinions.
Thanks!


pennfranc (commented):
Thanks for implementing this! I think it might be good to implement this attribute as a function `params()` with a `@property` attribute. That way we have a consistent flow of how the parameters are collected, (namely inside that function), and how they can be accessed (by calling that function as an attribute). We could also implement an abstract `params` function in the `ForecastingModel` superclass. What do you think?

norihitoishida:
Thanks @pennfranc !
> I think it might be good to implement this attribute as a function `params()` with a `@property` attribute. 

Nice idea! I'll add that soon.

> We could also implement an abstract `params()` function in the `ForecastingModel` superclass. 

I agree with you.
I'd like to work on that once I finish implementing params in all models.

norihitoishida:
Sorry, I couldn't know why the docs test fails...
Could someone please help me?

darts/models/exponential_smoothing.py
```diff
@@ -53,6 +53,12 @@ def __init__(self,
             `statsmodels.tsa.holtwinters.ExponentialSmoothing.fit()`.
             See `the documentation
             <https://www.statsmodels.org/stable/generated/statsmodels.tsa.holtwinters.ExponentialSmoothing.fit.html>`_.
+
+        Attributes
+        ----------
+        params
+            The parameters for the Exponential Smoothing model.
```
LeoTafti, line 60: > Sorry, I couldn't know why the docs test fails...
> Could someone please help me?

It seems like these lines are causing buildDocs to fail. Document directly in `params()` instead, e.g.:
```
    @property
    def params(self) -> Dict[Any, Any]:
        """
        The parameters for the Exponential Smoothing model.
        """
        return self._params
```

Though you might want to do that "one level higher" by defining and document a `params()` method in `ForecastingModel`, as @pennfranc suggested.

norihitoishida:
Close this PR and I will made a single PR related to #192.

PR #198, opened 2020-10-01
Add params attr and func to Prophet

norihitoishida:
Related to #192.

### Summary
Add `model.params` attr and func to Prophet()

```
# Before
model = Prophet()
model.fit(train)
print(model.model.params) # OrderedDict([('k', array([[0.46408934]])), ('m', ...

# After
model = Prophet()
model.fit(train)
print(model.params) # OrderedDict([('k', array([[0.46408934]])), ('m', ...
```


darts/models/prophet.py
```diff
@@ -93,3 +96,10 @@ def predict(self, n: int) -> TimeSeries:
 
         forecast = predictions['yhat'][-n:].values
         return self._build_forecast_series(forecast)
+
+    @property
+    def params(self) -> Dict[Any, Any]:
```
norihitoishida, line 101: The return type is collections.OrderedDict, and comment is wrong. (Exponential Smoothing > Prophet)
I'll fix it later...

norihitoishida:
if anyone could tell me why the test is not progressing please?

LeoTafti:
> if anyone could tell me why the test is not progressing please? (test3.6, 3.7, 3.8)

It's is most likely a mistake on our side setting up the CI workflows, we'll look at it. In the meantime, you can safely ignore those.

LeoTafti:
On another topic, it'd be easier for us to review your contribution related to #192 if you made the changes for all models in a single branch (maybe feature/model-params ?) and opened a single PR. Thanks!

norihitoishida:
Thanks for your reply, @LeoTafti !
Also thanks for letting me know how to put out a PR so that I can reduce the review burden on you.
I'll open a single PR.

PR #211, opened 2020-10-07
Fix: Release workflow

LeoTafti:
### Summary

Changes the release workflow substantially.

Instead of triggering automatically on pushes to master and creating a draft release, the new release workflow is triggered manually (from the GitHub "Actions" tab in the repo) and requires to provide a "bump type" argument which must be one of `#major, #minor, #patch` (if something else is provided, the release is aborted without making any changes). There was no need for a draft release step anymore, so I removed it.

I also removed tests from the release workflow, and instead trigger the merge PR workflow on pushes to master.

To make a release, the new workflow is thus:
- Open a PR from develop to master.
- Merge PR (no need to put `#major`, … in the commit message anymore). This triggers the "darts merge PR workflow" which runs the full test suite. **Wait for them to successfully run** (should be the case since they should have already run on the develop branch), and then…
- Go to the "Actions" tab and select the "darts release workflow".
- Click "Run workflow" and provide the bump type parameter `#major`, `#minor` or `#patch`

Everything else should happen automatically and a new release published.

This has the big advantage over the last workflow to be less error-prone (no more problems with forgetting to put the bump type "hastag" in the commit message, …), and to allow pushing minor changes to master without triggering the release workflow (which is the problem we had when updating the readme).
The previous workflow also had a flaw in the sense that it made a commit to master changing the version number in config files before creating the draft pull request, rendering the "draft" release step useless since master had already been altered.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - develop
+      - master
 
 jobs:
   tests:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,46 +1,14 @@
 name: darts release workflow
 
 on:
-  push:
-    branches:
-      - master
+  workflow_dispatch:
+    inputs:
+        bump_type:
+          description: "Bump type (#major, #minor, #patch)"
+          required: true
 
 jobs:
-  tests:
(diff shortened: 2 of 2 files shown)


hrzn (approved):
Looks neat :)

TheMP (approved):
Remember to update Quip ;]

PR #212, opened 2020-10-08
fix(build.gradle): omit files from coverage check report depending on…

LeoTafti:
Fixes coverage tests failing with some flavors of the installation. See #181.

### Summary

Not a very clean solution at all (hence the draft pull request), but seems to work at least locally to exclude some files from the report (and coverage % calculation) depending on the flavour installed.

I couldn't come up with a way to omit files depending on their dependency to the flavour installed, other than naming them each in the `build.gradle` file. Would welcome any suggestion !

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,7 +84,7 @@ task installLocally(type:Exec) {
 
 // Tests
 void coverageTestSteps(flavour) {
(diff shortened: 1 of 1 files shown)


TheMP (commented):
Is it maybe possible to simply run coverage tests only for u8darts[all] and simply skip for the rest?

LeoTafti:
It is, thanks for pointing this out. It makes sense to me too, I'll change it.

hrzn (approved):

TheMP (approved):
much less worrying now, thanks ;]

PR #214, opened 2020-10-13
Feature/map transformers

LeoTafti:
### Summary

- Added two new concrete implementations of `DataTransformers`: `Mapper` and `InvertibleMapper` allow to easily perform the equivalent of a `map()` function on a TimeSeries, and can be made part of a `Pipeline`
- Extended `map()` on `TimeSeries` to accept functions which use both a value and its timestamp to compute a new value e.g.`f(timestamp, datapoint) = new_datapoint`
- Removed `cols` parameter from `map()`. Using indexing on `TimeSeries` is preferred.

### Other Information

I noticed that the addition of `MissingValuesFiller` was missing in the changelog. I added it there instead of creating a new PR just for this.


LeoTafti:
Small problem: The new signature of `map()` expects for `fn` either a `Callable[[np.number], np.number]` or a `Callable[[pd.Timestamp, np.number], np.number]`. I couldn't find an easy way of checking at run-time whether `fn` was of the first type or of the second, except for checking the number of arguments using `len(inspect.signature(fn).parameters)`.

(in particular, isinstance(fn, Callable[[np.number], np.number]) doesn't work, it raises `TypeError: Subscripted generics cannot be used with class and instance checks`)

The problem with this solution is that `inspect.signature()` doesn't work with built-in functions (e.g. `math.sin()`) or some numpy functions (e.g. `np.sin()`). A simple trick to solve this is to wrap them in a lambda, e.g. `lambda x: math.sin(x)`

Other solutions I couldn't think of would be:
- Add a `with_timestamp:bool` parameter to `map()` and `Mapper`/`InvertibleMapper` constructors, but I don't like the fact that this could easily be forgotten if we make it a keyword arg, is a breaking change if we make it a positional arg, and is somewhat redundant since the function in itself should convey this information.
- Have two separate `map()` methods (maybe `map()` and `map_with_timestamp()`). But then there is still the need for some way of deciding which one to call form `Mapper` and `InvertibleMapper`…

If there is another alternative I'm missing I'd love to hear it! Otherwise I'd say that the "lambda" trick is the best alternative provided it is well documented (I'd add a line to the `map()` function and `Mapper` / `InvertibleMapper` docstrings, and in the changelog).

TheMP:
> Small problem: The new signature of `map()` expects for `fn` either a `Callable[[np.number], np.number]` or a `Callable[[pd.Timestamp, np.number], np.number]`. I couldn't find an easy way of checking at run-time whether `fn` was of the first type or of the second, except for checking the number of arguments using `len(inspect.signature(fn).parameters)`. In particular, isinstance(fn, Callable[[np.number], np.number]) doesn't work, it raises `TypeError: Subscripted generics cannot be used with class and instance checks`
> 
> The problem with this solution is that `inspect.signature()` doesn't work with built-in functions (e.g. `math.sin()`) or some numpy functions (e.g. `np.sin()`). A simple trick to solve this is to wrap them in a lambda, e.g. `lambda x: math.sin(x)`
> 
> Other solutions I couldn't think of would be:
> 
> * Add a `with_timestamp:bool` parameter to `map()` and `Mapper`/`InvertibleMapper` constructors, but I don't like the fact that this could easily be forgotten if we make it a keyword arg, is a breaking change if we make it a positional arg, and is somewhat redundant since the function in itself should convey this information.
> * Have two separate `map()` methods (maybe `map()` and `map_with_timestamp()`). But then there is still the need for some way of deciding which one to call form `Mapper` and `InvertibleMapper`…
> 
> If there is another alternative I'm missing I'd love to hear it! Otherwise I'd say that the "lambda" trick is the best alternative provided it is well documented (I'd add a line to the `map()` function and `Mapper` / `InvertibleMapper` docstrings, and in the changelog).

@LeoTafti why do we  need to check for num of arguments/signature? In the end the code gets directly passed to transform function anyways

LeoTafti:
We need to know in `map()` because if it's an "f(x)=y" function `map()` internally uses `pandas.DataFrame.applymap()` (faster), but if it's an "f(timestamp, x)=y" function `map()` use `pandas.DataFrame.apply()` (slower)

darts/dataprocessing/transformers/mappers.py
```diff
@@ -0,0 +1,84 @@
...
+                 name: str = "InvertibleMapper",
+                 validators: Optional[Sequence[Validator]] = None):
+        """
+        Data transformer to apply a function and its inverse to a time series (similar to calling `series.map()`)
+
+        Parameters
+        ----------
+        fn
+            Either a function which takes a value and returns a value ie. f(x) = y
+            Or a function which takes a value and its timestamp and returns a value ie. f(timestamp, x) = y
```
hrzn, line 63: You should also document `inverse_fn`.

hrzn (commented):
@LeoTafti It would be nice to avoid having to use lambdas wherever we could go around it. I believe it should work for builtin functions (and e.g. `math.sin`):
```
In [1]: len(signature(math.sin).parameters)
Out[1]: 1
```

Numpy `ufunc` functions work differently though, but it's also possible to get their number of inputs/outputs (https://numpy.org/doc/stable/reference/ufuncs.html).

So perhaps a possible logic could be along the lines of 

```
if isinstance(fn, np.ufunc):
    if fn.nin == 1 and np.nout == 1:
        ... map() ...
    else:
        # raise error about number of arguments, maybe advise using a lambda

elif len(inspect.signature(fn).parameters) == 1:
    ... map() ...

else:
    # raise error, and advise using a lambda
```

darts/dataprocessing/transformers/mappers.py
LeoTafti: > @LeoTafti It would be nice to avoid having to use lambdas wherever we could go around it. I believe it should work for builtin functions (and e.g. `math.sin`):
> 
> ```
> In [1]: len(signature(math.sin).parameters)
> Out[1]: 1
> ```
Something weird is happening there. I also get this on my machine, but [this run on GitHub](https://github.com/unit8co/darts/runs/1246785396) failed due to `math.sin` and I had to wrap it inside a lambda…

hrzn: that's weird indeed. I tried on Python 3.8, maybe 3.6 causes issues?

LeoTafti: Good call, doesn't work on Python 3.6 but it does work with 3.7 and 3.8…

hrzn: +1 for going the way you recommend

hrzn (approved):

PR #215, opened 2020-10-13
feat(data processing example): Added example notebook

LeoTafti:
### Summary

Adds an example notebook to showcase use of data transformers and pipelines.

### Other Information

This is a draft PR, because I also wanted to add a small example about how to use `Validator` but couldn't come up with a nice use case… Any suggestion ? Or maybe `Validator` isn't really needed at all ?


hrzn (approved):
Nice! Couple of small comments:

* You could perhaps add a note that by default the Scaler uses the MinMaxScaler but other sklearn scalers could be used.
* For the mappers I would specify they act elementwise
* The example with adjusting for the number of days is nice. You could maybe just explain what's the deal with the lambda's two arguments (when can we use one or two arguments, how does it work).

LeoTafti:
What about this ?

> I also wanted to add a small example about how to use Validator but couldn't come up with a nice use case… Any suggestion ? Or maybe Validator isn't really needed at all ?

LeoTafti:
Update on `Validator`: After discussing the mater, `Validator` will be removed from the API. I'll take care of this in a dedicated PR

PR #216, opened 2020-10-15
Feature/box cox

LeoTafti:
### Summary

Adds a new data transformer to perform Box-Cox transformation.
See [here](https://otexts.com/fpp2/transformations.html#mathematical-transformations) for more details.

### Other Information

Draft PR, as I'm trying to decide how exactly to go about adding the bias adjustment and would like feedback.

The problem is that in contrary to what's done in the example of the textbook, the BoxCox transformation step here is completely separate from the forecasting model, and thus cannot by default tell the difference between actual data and forecasted data to compute the variance of the forecast and invert that with a bias adjustment factor.

This makes among other things handling multivariate time series is tricky too: what if we have forecast data only for some of the variables ?

Also I'm not quite sure to understand how to compute the _"h-step forecast variance"_ mentioned in the textbook linked above. I've also found [this link](https://robjhyndman.com/hyndsight/backtransforming/) but it uses $upper and $lower which I don't think we have. I've tried looking around for implementations of such a bias adjustment, but couldn't find anything else than this.

Unless I'm missing something, considering how much more complicated this makes BoxCox, I'd be for leaving this bias adjustment feature out, and keep a simple BoxCox transformer. What do you think ?


hrzn:
+1 for leaving the bias adjustment out. I think it would already be nice to have the simple (context free) transform in place.

Another addition for later could be the "auto lambda" feature they have in R as `BoxCox.lambda()`.

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -0,0 +1,102 @@
...
+        Parameters
+        ----------
+        data
+            The time series to fit on
+        lmbda
+            If None given, will automatically find an optimal value of lmbda (for each dimension
+            of the time series) using `scipy.stats.boxcox_normmax` with `method=optim_method`
+            If a single float is given, the same lmbda value will be used for all dimensions of the series.
+            Also allows to specify a different lmbda value for each dimension of the time series by passing
+            a sequence of values.
```
hrzn, line 53: Even though you talk about it above, I would also add a docstring entry for `optim_method`.

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        super().fit(data)
+
+        raise_if(not isinstance(optim_method, str) or optim_method not in ['mle', 'pearsonr'],
+                 "optim_method parameter must be either 'mle' or 'pearsonr'",
+                 logger)
+
+        if lmbda is None:
+            # Compute optimal lmbda for each dimension of the time series
+            lmbda = data.pd_dataframe().apply(boxcox_normmax, method=optim_method)
```
hrzn, line 67: Not much to do with this PR, but should we plan to have an option in `pd_dataframe()` to avoid copying the dataframe, for those internal use cases where you really know what you are doing? That would improve performance.

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -0,0 +1,102 @@
...
+        elif isinstance(lmbda, Sequence):
+            raise_if(len(lmbda) != len(data.pd_dataframe().columns),
```
hrzn, line 69: You should call `data.width` instead; it will be much better than calling `pd_dataframe()`.

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -0,0 +1,102 @@
...
+                     "lmbda should have one value per dimension (ie. column or variable) of the time series",
+                     logger)
+        else:
+            # Replicate lmbda to match dimensions of the time series
+            lmbda = [lmbda] * len(data.pd_dataframe().columns)
```
hrzn, line 74: again use `width`

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self._lmbda = lmbda
+
+        return self
+
+    def transform(self, data: TimeSeries, *args, **kwargs) -> TimeSeries:
+        super().transform(data, *args, **kwargs)
+
+        def _boxcox_wrapper(col):
+            idx = data.pd_dataframe().columns.get_loc(col.name)  # get index from col name
```
hrzn, line 84: that might really scale poorly, if it's doing a full dataframe copy for each column 🙀

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -0,0 +1,102 @@
...
+            return boxcox(col, self._lmbda[idx])
+
+        return TimeSeries.from_dataframe(data.pd_dataframe().apply(_boxcox_wrapper))
+
+    def inverse_transform(self, data: TimeSeries, *args, **kwargs) -> TimeSeries:
+        super().inverse_transform(data, *args, *kwargs)
+
+        def _inv_boxcox_wrapper(col):
+            idx = data.pd_dataframe().columns.get_loc(col.name)  # get index from col name
```
hrzn, line 93: beyond the fact that it's currently copying the full DF for each column, isnt' there a cleaner way than using the name to get the column index? I'm sure there is a way to iterate over the column indices directly.

darts/dataprocessing/transformers/boxcox.py
LeoTafti: Oops forgot it, I'll add one.

darts/dataprocessing/transformers/boxcox.py
LeoTafti: I actually thought about this when writing these lines… I think we can already do this, by accessing `series._df`.

The reason I went with `pd_dataframe()` is because I wasn't actually working from within the `TimeSeries` class, but if you think that it's okay to access the dataframe directly from here, I also agree. I'll change it.

darts/dataprocessing/transformers/boxcox.py
LeoTafti: > isnt' there a cleaner way than using the name to get the column index?

Good question, from within `apply()` I don't think so.

It's possible to iterate over the columns with a for loop instead and select the corresponding column (and perform a map on it). Something like:

```
for index in range(0, data.width):
   columnSeries = data._df.iloc[: , index]
   columnSeries.map(…)
```

Two concerns regarding this solution though:
- This could be less efficient than `apply()` for time series with many dimensions (though it depends on the implementation of `apply()`), since it iterates on the columns sequentially when there really isn't the need to enforce that.
- I'll have to check, but I think calling `iloc` this way returns a view, and calling `pandas.Series.map()` might actually modify the dataframe. So we'd still need to make at least one copy of the dataframe.

There are two other solutions I could think of:
1. Have lmbda be a `dict` that would map column names to lambda value. But some transforms loose the name of the column, so this is likely to fail when using `BoxCox` as part of a `Pipeline`.
2. Rename columns with their index instead of whatever name they have (can be done in place without copying anything), use `apply()` and then revert to the original names.

I don't think that 1. is very good, but 2. might actually be better than using the column name to compute an index. WDYT ?

hrzn: OK fair enough, then I would even say we could leave it using your current approach (it doesn't look super clean, but neither does 1. or 2.). I would still use `df_` in most places instead of `pd_dataframe()`, because you know what you're doing here so there's little risk and it can have a big performance impact.

hrzn (approved):

PR #218, opened 2020-10-19
Feat/N-BEATS

pennfranc:
This PR contains the implementation of the core N-BEATS model. Please note that there are several things that have to be added to replicate the results from the paper (most notably multi-timeseries training and ensembling).

One thing that is a bit problematic is the fact that `N-BEATS` is strictly speaking a univariate model (unless further expanded) even though `TorchForecastingModel` is a subclass of `MultivariateForecastingModel`. Currently the workaround I am using is to throw an error when fit is called on a multivariate time series.

Other than that, the model seems to be functioning reasonably well, and I think further improvements can happen in separate PRs.


darts/models/nbeats.py
```diff
@@ -0,0 +1,410 @@
...
+            input_length,
+            output_length
+        )
+
+        kwargs['input_length'] = input_length
+        kwargs['output_length'] = output_length
+
+        super().__init__(**kwargs)
+
+    @random_method
```
pennfranc, line 403: I am overriding fit to (crudely) fix the multivariate/univariate issue here.

darts/models/nbeats.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        super(_Stack, self).__init__()
+
+        self.input_length = input_length
+        self.output_length = output_length
+
+        if g_type == GType.GENERIC:
+            self.blocks_list = [
+                _Block(num_layers, layer_width, expansion_coefficient_dim, input_length, output_length, g_type)
+                for i in range(num_blocks)
```
hrzn, line 198: [nitpicking] You could write `for _ in range(num_blocks)`

darts/models/nbeats.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        super(_Block, self).__init__()
+
+        self.num_layers = num_layers
+        self.layer_width = layer_width
+        self.g_type = g_type
+        self.relu = nn.ReLU()
+
+        # fully connected stack before fork
+        self.linear_layer_stack_list = [nn.Linear(input_length, layer_width)]
```
hrzn, line 104: (random comment, not for this PR)
It would be nice to optionally support multihead attention with positional encoding here, instead of fully-connected.

hrzn (approved):
Really nice :)
Next step: training on multiple time series.

PR #219, opened 2020-10-20
fix(readme): outdated instruction in contribute

Digma:
### Summary

Contribute guidelines were outdated since last release. The instruction to install pip packages references `main.txt` which does not exists anymore in place of `core.txt` and the other optional dependencies.

### Other Information

I did not find a way with pip to use something like `requirements/*.txt` so I added a file that has all the dependencies. Another way to fix it would have been to list all dependencies in the pip command (an updated version of the current instruction). Feel free to let me know if you prefer that approach

Another open question is if we want by default for the dev environment to also install the optional dependencies. I would expect it to be the case, so all the project tests can be run locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Before working on a contribution (a new feature or a fix) make sure you can't fi
 1. Create an issue, describe how you would attempt to solve it, and if possible wait for a discussion.
 2. Fork the repository.
 3. Clone the forked repository locally.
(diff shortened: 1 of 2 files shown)


hrzn:
Thanks @Digma :)
And I also vote for installing all dependencies in dev environment by default.

hrzn (approved):

LeoTafti:
Thanks @Digma :)

PR #222, opened 2020-10-22
Feature/regression ensemble model

LeoTafti:
### Summary

Adds a concrete ensemble model implementation which uses any regression model to ensemble individual model predictions.

### Other Information

In order to correctly implement `RegressionEnsembleModel` when some forecasting models were subclasses of `TorchForecastingModel`, I needed a way to "retrain" these forecasting models from scratch, ie. as if they had never been trained before.

I initially wanted to add a `reset()` function instead, but this proved to be much more complicated than I thought.

Instead, I added a new `untrained_model()` method to `TorchForecastingModel`, which allows to get the exact same model as the one which was initially created, such that this code:

```python
model = RNNModel()
model.fit(series)
forecast1 = model.predict(20)

model = model.untrained_model()  # here
model.fit(series)
forecast2 = model.predict(20)
```

… will result in `forecast1 == forecast2`, which isn't the case if the line marked with "`# here`" isn't present.

I also had to slightly modify the `@random_method` decorator in `utils/torch.py` for this to work (non-breaking change).


darts/models/regression_ensemble_model.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.regression_model = regression_model
+
+    def fit(self, training_series: TimeSeries, target_series: Optional[TimeSeries] = None) -> None:
+        # TODO: Factor this out, same logic is used in 3 places already. Need to find an appropiate name
+        if target_series is None:
+            target_series = training_series
+        raise_if_not(all(training_series.time_index() == target_series.time_index()),
+                     "training and target series must have same time indices.",
+                     logger)
```
LeoTafti, line 60: If anyone has a suggestion for this, I'd like to hear it. I couldn't come up with a nice name. Best I could do was `get_train_target()`, but …

Note a method like this existed in an earlier version of `MultivariateForecastingModel` and was called `make_fitable_series()` which I didn't understand well (and since it was only used in one place, I removed it, but now need it back).

darts/models/torch_forecasting_model.py
```diff
@@ -151,7 +156,7 @@ def __init__(self,
                  lr_scheduler_cls: torch.optim.lr_scheduler._LRScheduler = None,
                  lr_scheduler_kwargs: Optional[Dict] = None,
                  loss_fn: nn.modules.loss._Loss = nn.MSELoss(),
-                 model_name: str = "torch_model_run",  # TODO uid
+                 model_name: str = "torch_model_run",  # TODO: uid
```
LeoTafti, line 159: Just realized this hadn't been done when working on `TorchForecastingModel`. I'll add it to the backlog, since right now having multiple `TorchForecastingModel` with the default name would create conflicting checkpoints…

darts/models/regression_ensemble_model.py
```diff
@@ -0,0 +1,100 @@
...
+        # wrap provided regression_model in a StandardRegressionModel (if not already the case)
+        if isinstance(regression_model, StandardRegressionModel):
+            # raise exception if train_n_points value is ambiguous
+            model_train_n_points = regression_model.train_n_points
+            raise_if(model_train_n_points is not None and regression_train_n_points != model_train_n_points,
+                     "Provided StandardRegressionModel.train_n_points parameter doesn't match specified"
+                     " regression_train_n_points parameter.",
+                     logger)
+
+            regression_model.train_n_points = regression_train_n_points
```
hrzn, line 54: I guess that's not needed as you've checked above they are the same.

darts/models/regression_ensemble_model.py
```diff
@@ -0,0 +1,100 @@
...
+                 "regression_train_n_points parameter too big (greater or equal"
+                 " the number of points in training_series)",
+                 logger)
+        forecast_training = training_series[:-self.regression_model.train_n_points]
+        forecast_target = target_series[:-self.regression_model.train_n_points]
+
+        regression_target = target_series[-self.regression_model.train_n_points:]
+
+        # fit the forecasting models
+        super().fit(forecast_training, forecast_target)
```
hrzn, line 79: That's not a very good way to fit them. Would be better to call `model.fit()` directly IMO.

darts/models/regression_ensemble_model.py
```diff
@@ -0,0 +1,100 @@
...
+        raise_if_not(all(training_series.time_index() == target_series.time_index()),
+                     "training and target series must have same time indices.",
+                     logger)
+
+        # spare train_n_points points to serve as regression target
+        raise_if(len(training_series) <= self.regression_model.train_n_points,
+                 "regression_train_n_points parameter too big (greater or equal"
+                 " the number of points in training_series)",
+                 logger)
+        forecast_training = training_series[:-self.regression_model.train_n_points]
```
hrzn, line 73: It would be nice to rely on the new `historical_forecasts()` method here, and make it return the time series of historical forecasts, which you would then use as features. wdyt?

darts/models/regression_ensemble_model.py
```diff
@@ -0,0 +1,100 @@
...
+        for model in self.models:
+            predictions.append(model.predict(self.regression_model.train_n_points))
+
+        # train the regression model on the individual models' predictions
+        self.regression_model.fit(train_features=predictions, train_target=regression_target)
+
+        # prepare the forecasting models for further predicting by fitting
+        # them with the entire data
+
+        if TORCH_AVAILABLE:
```
hrzn, line 92: You could perhaps simply test whether `model` has the attribute `untrained_model`, which would avoid the Pytorch test, and also could later be useful with other non-torch models.

darts/models/regression_ensemble_model.py
```diff
@@ -0,0 +1,100 @@
...
+            # Neural-Network based models need to be retrained from scratch.
+            self.models = [model.untrained_model() if isinstance(model, TorchForecastingModel) else model
+                           for model in self.models]
+
+        super().fit(training_series, target_series)
```
hrzn, line 97: also here, would be better to call `model.fit()` directly.

darts/models/regression_ensemble_model.py
LeoTafti: There's a catch:

if `model_train_n_points` is `None`, then `regression_train_n_points != model_train_n_points` isn't evaluated – and in that case it makes sense to set `regression_model.train_n_points` to whichever value was specified by `regression_train_n_points`. Then the rest of the class can assume that`regression_model.train_n_points` will be accurate and only use this one.

That's why I've implemented it that way, but maybe it's not the super obvious?

darts/models/regression_ensemble_model.py
LeoTafti: Agreed, it's basically linked to the discussion [here](https://github.com/unit8co/darts/pull/175#discussion_r517421190). I'll change it.

darts/models/regression_ensemble_model.py
hrzn: OK, then perhaps it'd be nice to add a small comment, so we are not tempted to remove the line later on :)

darts/models/regression_ensemble_model.py
LeoTafti: This got me thinking quite a bit :)

I like the idea, but the implementation might be tricky and may require to either make quite a few assumptions for the user, e.g. which values to use for `stride`, `forecast_horizon`, `retrain`, … or complicate the use of `RegressionEnsembleModel` quite a bit.

And even then it would still make one pretty big assumption if we let `last_points_only=True`: the regression would only be trained on (and thus only care about) the last points of the models' forecasts, rendering the class almost useless if you care about the whole forecast horizon.

This made me think that the same thing can be said about the current implementation though, it’s gives weight to every point in the forecast horizon, so `StandardRegressionModel` possibly isn’t as good as it could be if you only care about the last point of a prediction. This doesn't seem as bad to me though, as the regression learns which model performs "best" overall but you can still use it only caring about the last point predictions it makes.

To keep it nice and simple **I’d be for leaving it as such (at least for now)**

Nevertheless I think it could be doable and useful to latter have maybe another class which would use `historical_forecasts()` and a regression model, that would take pretty much the same parameters as `historical_forecasts()`.

It could handle the case of `last_points_only=False` by enforcing `forecast_horizon = stride` such that we can still create a single time series per model of the historical forecasts, and use these as feature time series to train the regression model. The behavior of the current `RegressionEnsembleModel` above would then be a special case of this model, with `forecast_horizon = regression_train_n_points`, `start=len(series)-forecast_horizon` and `last_points_only=False`.

hrzn: OK makes sense, agree, we can leave it as is at least for now :)

hrzn (approved):
🚀

darts/models/regression_ensemble_model.py
LeoTafti: `target_series` will likely be removed by #231 , so I won't be addressing this here anymore.

PR #225, opened 2020-10-27
feat(transformer): added disclaimer

pennfranc:
- Added a disclaimer describing the current limitations of our Transformer model with regard to the `tgt` variable that serves as input for `torch.nn.Transformer`.

- Removed a line that creates the tgt mask, since it is not currently being used.


LeoTafti (approved):

LeoTafti:
Closing this since I branched of from this PR and solved docs missing in Sphinx for decorated functions.

PR #227, opened 2020-10-27
feat(plot): change matplotlib default style

Digma:
Changes default plotting style to increase contrast and make plot lighter. Current plot are pretty dark and I feel it hurts the contrast

### Summary

End result:
Screenshot 2020-10-27 at 13 55 56

Changes:
* Changed font to default `sans-serif` which makes numbers more legible
* Background is now white
* Axes, values and grid are gray so visually the plotlines take visual precedence over them
* Axes label are bold so they stand out compared to axis value

### Other Information

The white background is 1 option, one could also think about a style closer to ggplot / seaborn with a gray-ish background but only on the plot area, similar to something like this:

Screenshot 2020-10-27 at 13 56 01

Open to comments and to add other improvements!

### TODO

- [x] Agree on desired look
- [ ] Re-run all notebooks?
- [ ] Update documentation images?


hrzn:
I would think that by default we should target something as clear/readable as possible (and not necessarily something mimicking Unit8 style), so I'm a bit more in favour of Gael's proposal as a default.
However, couldn't we imagine a Unit8 mode activated like `matplotlib.style.use('unit8')` and that would go "full Unit8" :D If we have such a style sheet, we could even provide this style sheet for other internal projets where we need branding.

Digma:
@hrzn  I can have a look at how it would work to have some custom optional style. 

In the meantime, we could try to agree on the default. I think I have a slight preference for white as the contrast is a bit higher. I think the advantage of gray is that it shows the plot area but I feel with the grid we already it is enough. I tried to find the reason why seaborn and ggplot went for gray background but I could not find any resources on that

LeoTafti:
Alright white it is then :)

hrzn (approved):

Digma:
Let me know if there is something I need to do / update (e.g. notebooks) to have a mergeable PR

LeoTafti:
Re-running the notebooks would be nice. Be aware though that the Transformer model notebook takes a while to run (~1 hour or so for me).

If you can also recreate the plot used in the readme it'd be nice, otherwise I'll do it eventually :) 

Thanks!

PR #230, opened 2020-10-29
fix(backtesting): forecast_horizon and stride parameter

LeoTafti:
### Summary

Modified `ForecastingModel.backtest()` and `RegressionModel.backtest()` to save the last `stride` points of each forecast instead of only the last one.

Added small illustrating diagram to the docs.  The diagram is in `docs/source/static/backtest_diagram.png`

### Other Information

* Fixed tests.
* Small tweak to NaiveDrift model (was copying the last point of the training set as first point of the prediction, now not the case anymore).
* Re-ran notebooks.

The output of the backtest function looks less smooth but is more representative of the model performance.


darts/models/forecasting_model.py
```diff
@@ -244,18 +242,26 @@ def backtest(self,
...
                 pred = self.predict(forecast_horizon, **predict_kwargs)
             else:
                 pred = self.predict(forecast_horizon, input_series=train, **predict_kwargs)
-            values.append(pred.values()[-1])  # store the N-th point
-            times.append(pred.end_time())  # store the N-th timestamp
 
-        forecast = TimeSeries.from_times_and_values(pd.DatetimeIndex(times), np.array(values),
-                                                    freq=training_series.freq() * stride)
+            if pred_time == start:  # use the full forecast in the first loop
+                values.extend(pred.values())
```
hrzn, line 254: I'm not sure about that.
I think the first prediction in the backtest should be for `pred_time + forecast_horizon`, otherwise some of the points in the resulting series will effectively have been obtained with a shorter horizon, which is not the behavior we want I think.

LeoTafti: Mmhhh that's a fair point.

I'm not sure about the `pred_time + forecast_horizon` part though. Looking at the small diagram, I think we can simply treat the first prediction like the next ones, only recording the last `stride` points.

LeoTafti: Is this what you suggested ?

backtest_diagram

pennfranc: Just to clarify what the current desired functionality of backtest is: Every point in the time series returned by `backtest`  should correspond to one call to `predict`. Also, if `retrain` is set to `True`, every point in the returned time series also corresponds to one call to `train`. All points in the returned time series have the same forecast horizon, i.e. the distance to the last training point is always equal to the forecast horizon. Stride simply tells us how many `freq` units apart two adjacent entries in the returned time series are. So in that sense it does not make a lot of sense (as far as i can see) to  record the last `stride` points, since we're only interested in the last predicted point in every iteration. Or did I misunderstand your proposal?

LeoTafti: Oh I see, I think I actually misunderstood the original `backtest()` functionality.

I erroneously thought that the fact only one point per forecast was recorded might have been due to later introducing a `stride` parameter.

I think plotting the return value of `backtest()` for `stride != 1` is a bit misleading then, because it links points spaced with more than 1 time step together (which made me think we actually wanted to have points in between, but didn't have them, hence my proposal)

I can see how the current implementation of `backtest()` makes sense then, though I'm not sure if it's the best way to backtest forecasting models ? Maybe if instead of returning a series, backtest returned a score, it might be able to more accurately / efficiently assess the performance of a model ?

LeoTafti: Because right since it only records the last point of each predictions, `backtest()` actually measures how well the model performs _at a given point in the future_ (ie. `forecast_horizon` time steps in the future). I'm not sure this is accurate either…

pennfranc: Why do you think it's misleading? The new frequency of the time series will be reflected in the `TimeSeries.freq` field. And time series object with different frequencies are quite common. But maybe we need to emphasize the fact that the output frequency is dependent on the stride paramter more.

You're right in that it would be faster to get a score directly to evaluate a model. However, you would lose the intermediate result, which are the predictions. You might want to keep the predictions for visualizations, or users might want to use a special kind of scoring that we cannot foresee. I think requiring the user to write one more line of code to compare the backtest forecast with the original series is not that big of the deal. But we could certainly add a wrapper called `backtest_score` or something like that that would all do it in one go, if you think this would better streamline the process. I'm pretty indifferent when it comes to that.

pennfranc: Yep, that was the initial idea though. Because that way we introduce consistency in that regard. You can test how good your model performs for different horizons.

LeoTafti: I'll have to read up to see what's done elsewhere, but I think a better way of getting a sense of whether a model performed well or not (keeping both `forecast_horizon` and `stride` parameters) might be to compute at each step (every `stride` points) the error between the prediction (of `forecast_horizon` points) and the actual series, and take the average of all error scores. This average could even then be divided by the number of points per forecast (`forecast_horizon`) to get a final score which could then be used to compare the performance of a model with different values of `forecast_horizon`

What do you think ?

pennfranc: I believe that could be a valuable addition to the library, but I don't think it should replace the current backtest functionality. This is because to me there is a lot of merit in being able to fix the time horizon between the forecast and the training series to get a more consistent reading. But going along with your idea, wouldn't it make more sense to keep all the `forecast_horizon` points of each forecast? After all, `stride` could be larger than `forecast_horizon`. I think the two should be treated independently.
If your proposal could be added to the current backtest functions without changing too much code, we could possibly introduce it by adding another parameter to backtest that acts as a toggle between the two modes. Do you think that might work?

LeoTafti: Thinking about it, I think we can get both functionalities in `backtest()` in a clean way, as long as we're okay with a small restriction.

We could indeed have a parameter to toggle between the two, and then either perform the current functionality as is, or go with a special case of what I'm proposing where we'd enforce `stride == forecast_horizon`. In this case we'd also be able to return a time series of all individual predictions (since they do not overlap – no need to return a score here) and this time series could be used to get a sense of the model's performance overall, instead of only at a given time step in the future like it currently does.

What would you think about that ?

hrzn: I'm not too sure I'm following what you have in mind for the second functionality that you're proposing @LeoTafti .

In general, I think backtesting should return a TimeSeries - it's very easy / trivial to get an accuracy metric number afterwards once we have the time series.

I think we should think of it as the model being re-fit (if retrain=True) each `stride` points, and used to get a new `forecast_horizon` worth of point. If I understand well our situation, the open question is mainly how much of this prediction should we keep in the returned TimeSeries. We can keep only the last point (like now), or the last `stride` points, or maybe we need an additional (optional) parameter to define this?
In any case, by default I believe it's good to keep this last-point-only behavior, because when backtesting, most often you want to simulate the behavior you'd obtain if you were to use this model "live". And we you use a model live, you can (in most cases I think) re-train it whenever a new data point becomes available. So for a given horizon `H`, at forecast time `T`, you'd like to know how your model performs at `T+H`, but typically it's less interesting to know how it performs at `T+H-1`, because the forecast for `T+H-1` would in the real life have been obtained with the model trained at `T-1`, not at `T`. Does it make sense?

LeoTafti: We've discussed it with @pennfranc too, and he made the same point you did about keeping the last-point-only behavior, and I indeed agree that in many cases this does make a lot of sense.

However I'm not too sure about the "typically it's less interesting to know how it performs at `T+H-1`, because the forecast for `T+H-1` would in the real life have been obtained with the model trained at `T-1`, not at `T`" part. If you're predicting hourly energy consumption for example with a horizon of 6 hours, you'll most likely also care very much about the accuracy of the model for the upcoming 1, 2 or 3 hours. It seems likely to me that a model a time `T` could make a better prediction for the next 1, 2 and 3 hours than a models "one forecasting horizon ago" (ie at time `T-H+1`, `T-H+2`, `…`)

@pennfranc suggested that in this case it would be possible to train 3 separate models, one optimized for each forecasting horizon. But this doesn't scale very well: what if you want accurate daily-predictions for the next 30 days ? Do you train 30 models ? I think in many cases it would make more sense to rather train a single model that is overall more accurate over the whole 30-day period.

LeoTafti: And here's what my second proposition is:

backtest-proposition

This way since we're only making predictions for non-overlapping indices, we can still return a time series and it is indeed trivial to get an accuracy metric afterwards.

The problem with this is that it reduces quite drastically the number of train/test splits we can make, compared to doing this:

backtest-proposition-2

But then we wouldn't be able to return a time series anymore, only a score.

LeoTafti:
Closing as #234 will address this properly.

PR #233, opened 2020-11-02
Fix/missing values colnames

LeoTafti:
Fixes `fill_missing_value()` which would remove the underlying dataframe's column names when used with `fill='auto'`

### Other Information

Corrected the type hint of `TimeSeries.from_times_and_values()` since it was already accepting DataFrames too.


hrzn (approved):
Good catch, thanks!

PR #234, opened 2020-11-03
backtest refactor / new backtest

LeoTafti:
### Summary

Following this morning's discussion, I've started working on adapting the library to be able to keep both the current version of the `backtest()` function (which returns a time series created from the last point of each historical forecast) and a new `backtest()` function which would only return a score.

I moved most of the code from `backtest()` into a newly created method called `historical_forecasts()` which ressembles the old `backtest()` method, except that it returns a `List[TimeSeries]` by default corresponding to each historical forecast.

The idea is to use `historical_forecast()` in a new `backtest()` function, which will take an additional `metric` parameter and return the average error score over all of the historical forecasts.

The old `backtest()` functionality is still available, but:

```python
model.backtest(…)
```

is now replaced by:

```python
model.historical_forecasts(…, last_points_only=True)
```

I also renamed the `trim_to_series` parameter into `overlapp_series_end` – this new name will make more sense once the new `backtest()` function is implemented. The new behavior thus follows this:

new_backtest

*where `n` is the last iteration for which a forecast is computed*

### Other information

This is WIP, I've only refactored the old `backtest()` function in `ForecastingModel`. Before I go ahead and apply the same modification to RegressionModel.backtest() and change every call to backtest() I wanted to get your opinion on this, so please let me know if you have any suggestion!


darts/models/baselines.py
```diff
@@ -95,5 +95,5 @@ def predict(self, n: int):
         first, last = self.training_series.first_value(), self.training_series.last_value()
         slope = (last - first) / (len(self.training_series) - 1)
         last_value = last + slope * n
-        forecast = np.linspace(last, last_value, num=n)
+        forecast = np.linspace(last, last_value, num=n + 1)[1:]
```
hrzn, line 98: oh good catch

darts/models/forecasting_model.py
```diff
@@ -140,25 +140,30 @@ def _backtest_model_specific_sanity_checks(self, *args: Any, **kwargs: Any) -> N
...
-        """ Retrain and forecast values pointwise with an expanding training window over `series`.
+    def historical_forecasts(self,
+                             training_series: TimeSeries,
+                             target_series: Optional[TimeSeries] = None,
+                             start: Union[pd.Timestamp, float, int] = 0.7,
+                             forecast_horizon: int = 1,
+                             stride: int = 1,
+                             retrain: bool = True,
+                             overlapp_series_end: bool = False,
+                             last_points_only: bool = False,
```
hrzn, line 151: I would consider setting the default to `True`, because calling with `False` would mostly be for the backtest() function, and for users who know what they are doing. For many users it would be nice to just get the single `TimeSeries` of historical last-point forecasts I think.

darts/models/forecasting_model.py
```diff
@@ -195,16 +200,19 @@ def backtest(self,
             Optionally, if the model is an instance of `TorchForecastingModel`, this argument will be passed along
             as argument to the `predict` method of the model. Otherwise, if this value is set and the model is not an
             instance of `TorchForecastingModel`, this will cause an error.
-        trim_to_series
-            Whether the predicted series has the end trimmed to match the end of the main series
+        overlapp_series_end
+            Whether the forecasts can go beyond the series' end or not
```
hrzn, line 204: "Whether the returned forecasts ..."

hrzn (commented):
It looks good! I would just think that `last_points_only` could be `True` by default (we can discuss it if you think that's a bad idea)

LeoTafti:
I disagree :)

I think you make a fair point when you say that currently we'd be setting it to `True` in most places (except inside of the new `backtest()`) but I believe this is mostly due to the fact that we only had this option up to now.

The behavior I'd expect from a method named `historical_forecasts()` would be to return a list of forecasts, which is why I set the default to `True`. When `last_points_only` is explicitly set to `False`, then it's okay IMO to not return a list of `TimeSeries` and return a slightly "weirder" value instead.

But that might be only me ^^ let me know if you still think otherwise!

LeoTafti:
Mostly done, remains to update calls to `backtest()` in the example notebooks and the changelog.

The previous functionality is untouched: even `gridsearch()` still allows both behaviors (optimizing for overall best average error score, or only considering the last points of each forecasts)

LeoTafti:
The `buildDocs` job is working locally but failing on GitHub and I can't figure out why. This is also the case on #233

hrzn:
> I disagree :)
> 
> I think you make a fair point when you say that currently we'd be setting it to `True` in most places (except inside of the new `backtest()`) but I believe this could be mostly due to the fact that we only had this option up to now.
> 
> The behavior I'd expect from a method named `historical_forecasts()` would be to return a list of forecasts, which is why I set the default to `False`. When `last_points_only` is explicitly set to `True`, then it's okay IMO to not return a list of `TimeSeries` and return a slightly "weirder" value instead.
> 
> But that might be only me ^^ let me know if you still think otherwise!
> 
> EDIT: And many thanks for the very quick review

My main reason was to favour manipulating time series wherever possible (rather than lists of time series), and also because I expect many concrete use-cases to care about the last-point series only, because that just represents what is done in business so often. When the method returns a `TimeSeries` you can just do
```
mape(model.historical_forecasts(series), series)
```
which is really nice and simple. I think returning a list of series is also valuable, but adds one step of complexity. The users who don't really know what they are doing will have to understand how to use these lists, how they relate to the horizons they specified etc. I would vote for keeping always the simplest settings as defaults.

LeoTafti:
I think you got me convinced, I'll make the change.

One question though, what do you think should be the default behavior for `gridsearch()` then?

It uses (depending on the exact parameters given to the method) `backtest()` internally to compute an error score – should it by default call `backtest()` with `last_points_only=True` or `last_points_only=False`?

hrzn:
> I think you got me convinced, I'll make the change.
> 
> One question though, what do you think should be the default behavior for `gridsearch()` then?
> 
> It uses (depending on the exact parameters given to the method) `backtest()` internally to compute an error score – should it by default call `backtest()` with `last_points_only=True` or `last_points_only=False`?

That's a good question! I would tend to think that this could be a parameter for the user to choose when calling `gridsearch()`, and we could have it to `False` by default. We have a small issue though, as currently you are computing the average metric in `backtest()`, and taking the average is a choice that may not always be best. I think we could actually think of proposing a `reduction` parameter in the `backtest()` method (and in the `gridsearch() method as well), which would specify how to combine the different metrics on the different series. For instance, we could have `reduction=np.mean` or `reduction=np.median` or any function that can take an array/sequence and return a float. So for instance calling `backtest(metric='mape', reduction=np.median)` would return one scalar; the median MAPE obtained on each validation time series. We can leave the mean as default, and say that if `reduction=None`, then no reduction is done, and you receive the list of all metrics values over each of the validation series. WDYT?

LeoTafti:
I like your suggestion about adding a `reduction` parameter that would default to `np.mean` a lot. I actually thought about the fact that computing the average might not always be the best when writing it, but didn't think about allowing the user to choose. I'll make the change.

I think allowing `reduction=None` is nice too, the only catch would be that we would then have a `Union[float, List[float]]` return type for `backtest()` and I think we've tried to avoid them so far (although `historical_forecasts()` also has a `Union` return type, so…).

hrzn:
> I like your suggestion about adding a `reduction` parameter that would default to `np.mean` a lot. I actually thought about the fact that computing the average might not always be the best when implementing `backtest()`, but didn't think about introducing a `reduction` parameter. I'll make the change.
> 
> I think allowing `reduction=None` is nice too, the only catch would be that we would then have a `Union[float, List[float]]` return type for `backtest()` and I think we've tried to avoid them so far (although `historical_forecasts()` also has a `Union` return type, so…).

Yes, but I think it's OK in this case. Because to call it with `None` you have to know what you are doing and so you'll know what to do with the results.

LeoTafti:
Pretty much done. I made `last_points_only` default to `True` in both `ForecastingModel` and `RegressionModel` `historical_forecasts()` methods, but kept it `False` by default in `backtest()` and `gridsearch()`.

I removed _most_ instances where `last_points_only` was set explicitely to `True` from the example notebooks, though I have left them in a few places (2-3 calls) where I thought they helped understanding.

Also added the `reduction` param to `backtest()` and showcased its use in the darts-intro notebook, tell me what you think.

darts/models/forecasting_model.py
```diff
@@ -225,39 +233,159 @@ def backtest(self,
...
+            instances such as `RNNModel` and `TCNModel` support setting `retrain` to `False`.
+        overlapp_series_end
+            Whether the returned forecasts can go beyond the series' end or not
+        last_points_only
+            Whether to use the whole forecasts or only the last point of each forecast to compute the error
+        metric
+            A function that takes two TimeSeries instances as inputs and returns a float error value.
+        reduction
+            A function used to combine the individual error scores obtained when `last_points_only` is set to False
+            If explicitely set to `None`, the method will return a list of the individual error scores instead
```
hrzn, line 351:
```suggestion
            If explicitely set to `None`, the method will return a list of the individual error scores instead.
```

darts/models/forecasting_model.py
```diff
@@ -187,24 +193,26 @@ def backtest(self,
...
         use_full_output_length
             Optionally, if the model is an instance of `TorchForecastingModel`, this argument will be passed along
             as argument to the `predict` method of the model. Otherwise, if this value is set and the model is not an
             instance of `TorchForecastingModel`, this will cause an error.
-        trim_to_series
-            Whether the predicted series has the end trimmed to match the end of the main series
+        overlapp_series_end
+            Whether the returned forecasts can go beyond the series' end or not
+        last_points_only
+            Whether to keep and return the whole forecasts or only the last point of each forecast
```
hrzn, line 207:
```suggestion
            Whether to retain the last point of each historical forecasts. If set, the method returns a `TimeSeries` of 
            these historical point forecasts. If not set, the method returns a list of the historical `TimeSeries` forecasts.
```

darts/models/regression_model.py
```diff
@@ -107,35 +113,32 @@ def _backtest_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
...
+        _historical_forecasts_general_checks(target_series, kwargs)
+
+    @_with_sanity_checks("_historical_forecasts_sanity_checks")
+    def historical_forecasts(self,
+                             feature_series: Iterable[TimeSeries],
+                             target_series: TimeSeries,
+                             start: Union[pd.Timestamp, float, int] = 0.5,
+                             forecast_horizon: int = 1,
+                             stride: int = 1,
+                             overlapp_series_end: bool = False,
```
hrzn, line 125: `overlap`

darts/models/regression_model.py
```diff
@@ -107,35 +113,32 @@ def _backtest_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
...
-                 verbose=False) -> TimeSeries:
-        """ A function for backtesting `RegressionModel`'s.
-
-        This function computes the time series of historical predictions
-        that would have been obtained, if the current model had been used to predict `target_series`
-        using the `feature_series`, with a certain time horizon.
+        _historical_forecasts_general_checks(target_series, kwargs)
+
+    @_with_sanity_checks("_historical_forecasts_sanity_checks")
+    def historical_forecasts(self,
```
hrzn, line 119: So much code repetition between RegressionModel and ForecastingModel 😅

hrzn (approved):
Gj @LeoTafti

darts/models/forecasting_model.py
```diff
@@ -125,40 +130,41 @@ def _backtest_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
...
 
+    @_with_sanity_checks("_historical_forecasts_sanity_checks")
+    def historical_forecasts(self,
+                             training_series: TimeSeries,
+                             target_series: Optional[TimeSeries] = None,
+                             start: Union[pd.Timestamp, float, int] = 0.5,
+                             forecast_horizon: int = 1,
+                             stride: int = 1,
+                             retrain: bool = True,
+                             overlapp_series_end: bool = False,
```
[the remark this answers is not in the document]
LeoTafti: I went with `overlap_end` instead, overlap could be confusing when used with `last_points_only=False` since forecasts may overlap too.

PR #250, opened 2020-12-25
Fix/remove example notebook errors

pennfranc:
I noticed there are error messages in two of the example notebooks (also on the master branch). I reran the notebooks to get rid of them. Now we have a commit with huge diffs again, though.. is there a better way to do this maybe?


TheMP:
> I noticed there are error messages in two of the example notebooks (also on the master branch). I reran the notebooks to get rid of them. Now we have a commit with huge diffs again, though.. is there a better way to do this maybe?

After some digging I've found out this - https://app.reviewnb.com/unit8co/darts/pull/250/. Works for free in case of open-source repos so I activated it for darts. 

That tool only shows meaningful diffs for the sake of commenting, unfortunately no simple way to compress the diff list (and every image makes our git history way bigger). I only know of 2 ways:

1. What you did - re-run only the single cell to keep the change minimal.
2. There are tools to simply keep vanilla .py files instead of .ipynb, but we would have to force end-users to build notebooks themselves (with the assistance of tools)  - which unfortunately makes their life quite difficult.

TheMP (approved):

PR #255, opened 2021-01-08
Add note in README about clang error while installing

camilaagw:
Summary:
While following the **Installation Guide** I run into the `error: command 'x86_64-apple-darwin13.4.0-clang++' failed with exit status 1` when installing the library using pip inside the conda env `pip install 'u8darts[all]'`. It was fixed  by setting the variables `CC`and `CXX` (https://github.com/astropy/halotools/issues/813):
```
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++
```
I propose to include this fix as a note in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,13 @@ we also provide the following alternate lighter install options:
 * Install core + neural networks (PyTorch): `pip install 'u8darts[torch]'`
 * Install core + Facebook Prophet: `pip install 'u8darts[fbprophet]'`
 * Install core + AutoARIMA: `pip install 'u8darts[pmdarima]'`
-   
+
+Note: In case you run into errors with `clang` (the C compiler) set the variables `CC` and `CXX` as follows:
+```bash
+export CC=/usr/bin/clang
+export CXX=/usr/bin/clang++
+```
+
 ### Running the examples only, without installing:
 
 If the conda setup is causing too many problems, we also provide a Docker image with everything set up for you and ready-to-use python notebooks with demo examples.


README.md
```diff
@@ -175,7 +175,13 @@ we also provide the following alternate lighter install options:
 * Install core + neural networks (PyTorch): `pip install 'u8darts[torch]'`
 * Install core + Facebook Prophet: `pip install 'u8darts[fbprophet]'`
 * Install core + AutoARIMA: `pip install 'u8darts[pmdarima]'`
-   
+
+Note: In case you run into errors with `clang` (the C compiler) set the variables `CC` and `CXX` as follows:
+```bash
+export CC=/usr/bin/clang
```
hrzn, line 181: I guess these commands depend on the exact setup. Perhaps worth adding a note saying that this is only an example which should (most of the time) work out of the box on MacOS for example.

TheMP: I'm wondering if it's actually an issue with the darts [couldn't reproduce on my Mac], as it seems more connected to compiling one of the dependencies and general setup - I think we should either start a new page (Troubleshooting?) or let the users get their setups right.

camilaagw: I also tried to reproduce the same in my personal Mac (OS Catalina) and the installation went smoothly (no need to setup env variables). I only encountered this issue with a new MacBook (OS Catalina). I will try to investigate if others are having the same problem

camilaagw: Update: Mos (Mostafa Ajallooeian) tested the Darts installation with his new MacBook (OS Catalina), and he didn't experience any issue. I also tested the installation on another MacBook (OS Big Sur) and the installation was successful. I propose to close the PR as the issue seems to be specific to my setup

PR #260, opened 2021-01-27
feature: Add function to extract subseries from a series with missing…

camilaagw:
### Summary

Add function to extract sub-series from a series with missing values


hrzn (approved):
Very nice

PR #261, opened 2021-01-27
Rename example notebooks

camilaagw:
### Summary

Rename example notebooks so that they are easier to navigate from github: 
image

Note: During the renaming, I noted that the notebook NBEATS-examples is not present either in the `develop` or `merge` workflow for GitHub actions. Should it be included?


hrzn:
Could we slightly change the order?
1. darts-intro
2. multi-time-series-covariates
3. data-processing
... (and then the order you propose)

camilaagw:
@hrzn Perfect, I will rename in the order you propose.

Question: During the renaming I noted that the notebook NBEATS-examples is not present either in the develop or merge workflow for GitHub actions. Should it be included?

hrzn:
> @hrzn Perfect, I will rename in the order you propose.
> 
> Question: During the renaming I noted that the notebook NBEATS-examples is not present either in the develop or merge workflow for GitHub actions. Should it be included?

We left some of them out on purpose because these checks take a long time, so I think we can leave it like that for now.

PR #265, opened 2021-02-01
Feature/regressionensemblemodel example

camilaagw:
### Summary

Add an example of the usage of `RegressionEnsembleModel`  to darts-into.ipynb:

https://github.com/camilaagw/darts/blob/feature/regressionensemblemodel-example/examples/darts-intro.ipynb

As the notebook already contained an example of an ensemble model with an `RegressionModel` I decided to leave it (as it explains well the logic in ensembling models) and complement it with the `RegressionEnsembleModel `. Nevertheless I had  to adapt the example based on `RegressionModel`, to get a similar result to the one obtained with `RegressionEnsembleModel`.

Other direction to do this, could be to just leave the example with  `RegressionEnsembleModel ` and extend it with some additional plots.

### Other Information

While making the example for `RegressionModel` consistent with the one for `RegressionEnsembleModel `, I realised  that there is room for improvement in the way the predictions for the individual models are done in `RegressionEnsembleModel` (if my understanding is correct). 

In `RegressionEnsembleModel`, the predictions for the individual models are computed with a forecasting horizon equal to `train_n_poins`, but, what if at the end I want my ensemble model to do predictions for a forecasting horizon different than the `train_n_poins`?
For example, we can have a case in which the user has 20,000 training  points, he wants to make an ensemble model to predict the next point in the future, and he wants to train the regressor of the ensemble model in the last 4,000 points. According to this, `RegressionEnsembleModel` would fit each of the individual models on 16,000 points, and then it will call their `predict` method with 4,000 as the forecasting horizon. The forecasting horizon used to generate the predictions from the individual models, is totally different from the one he intends to use with the model.
Instead of doing this, it would be more appropriate to use a `historical_forecast`, with a forecasting horizon of one, to produce the 4,000 predictions of the individual models that will be aggregated under the regression model. 
Probably, the `fit` method of `RegressionEnsembleModel` needs to know in advance the forecasting horizon.


hrzn:
> ### Summary
> Add an example of the usage of `RegressionEnsembleModel` to darts-into.ipynb:
> 
> https://github.com/camilaagw/darts/blob/feature/regressionensemblemodel-example/examples/darts-intro.ipynb
> 
> As the notebook already contained an example of an ensemble model with an `RegressionModel` I decided to leave it (as it explains well the logic in ensembling models) and complement it with the `RegressionEnsembleModel `. Nevertheless I had to adapt the example based on `RegressionModel`, to get a similar result to the one obtained with `RegressionEnsembleModel`.
> 
> Other direction to do this, could be to just leave the example with `RegressionEnsembleModel ` and extend it with some additional plots.
> 
> ### Other Information
> While making the example for `RegressionModel` consistent with the one for `RegressionEnsembleModel `, I realised that there is room for improvement in the way the predictions for the individual models are done in `RegressionEnsembleModel` (if my understanding is correct).
> 
> In `RegressionEnsembleModel`, the predictions for the individual models are computed with a forecasting horizon equal to `train_n_poins`, but, what if at the end I want my ensemble model to do predictions for a forecasting horizon different than the `train_n_poins`?
> For example, we can have a case in which the user has 20,000 training points, he wants to make an ensemble model to predict the next point in the future, and he wants to train the regressor of the ensemble model in the last 4,000 points. According to this, `RegressionEnsembleModel` would fit each of the individual models on 16,000 points, and then it will call their `predict` method with 4,000 as the forecasting horizon. The forecasting horizon used to generate the predictions from the individual models, is totally different from the one he intends to use with the model.
> Instead of doing this, it would be more appropriate to use a `historical_forecast`, with a forecasting horizon of one, to produce the 4,000 predictions of the individual models that will be aggregated under the regression model.
> Probably, the `fit` method of `RegressionEnsembleModel` needs to know in advance the forecasting horizon.

I agree with your point, we should probably revise the `RegressionEnsembleModel` to make it behave like you describe.

darts/models/ensemble_model.py
```diff
@@ -34,13 +34,13 @@ def __init__(self, models: List[ForecastingModel]):
         super().__init__()
         self.models = models
 
-    def fit(self, training_series: TimeSeries) -> None:
+    def fit(self, series: TimeSeries) -> None:
```
hrzn, line 37: thanks :)

hrzn (approved):
Very nice, thanks @camilaagw

Issue #254, opened 2021-01-06
[BUG] Section "Running the examples only, without installing" in readme depends on Prophet which is not delivered along the docker image


ajallooeian:
**Describe the bug**
We have the section "Running the examples only, without installing" in the README.md. Following the instructing and installing with gradlew, the given Jupyter notebook `examples/darts-intro.ipynb` has a second cell:
```
from darts.models import (
    NaiveSeasonal,
    NaiveDrift,
    Prophet,
    ExponentialSmoothing,
    ARIMA,
    AutoARIMA,
    StandardRegressionModel,
    Theta,
    FFT
)
```
But prophet is not installed by default and you get:
```
ImportError: cannot import name 'Prophet' from 'darts.models' (/home/jovyan/work/darts/models/__init__.py)
```

**To Reproduce**
Follow the instruction on the section "Running the examples only, without installing" on a fresh environment. 

**Expected behavior**
The example to run without a need to manual installation. Or add a flag/condition to skip prophet if it is not installed. 

**System (please complete the following information):**
 - Python version: na
 - darts version: 0.5.0

ajallooeian:
The same behaviour for `AutoARIMA`

camilaagw:
@ajallooeian Julien fixed this issue. If you have some time, could you please check it works for you?

PhilipMachineLearning:
Still an issue for me

iamMaverick:
Still an issue.

![image](https://user-images.githubusercontent.com/25539320/179456946-df11bdef-94b5-4b52-9e78-722146f890aa.png)

hrzn:
> Still an issue. @hrzn
> 
> ![image](https://user-images.githubusercontent.com/25539320/179456946-df11bdef-94b5-4b52-9e78-722146f890aa.png)

Please consult the [install guide section explaining how to enable Prophet](https://github.com/unit8co/darts/blob/master/INSTALL.md#enabling-support-for-facebook-prophet). In the next version (0.21.0) we will reinstate prophet as a dependency and this won't be necessary anymore.

PR #269, opened 2021-02-02
Fix/doc

hrzn:
I tested this one by triggering the release workflow, and the docs were built successfully (see here: https://unit8co.github.io/darts/generated_api/darts.html).

It fixes a bunch of smaller issues as well.

Once merged on master I will produce again a patch release to obtain (i believe) version 0.6.2.


Dockerfile
```diff
@@ -23,6 +23,6 @@ ADD . /home/jovyan/work
 
 WORKDIR /home/jovyan/work
 
-RUN pip install -q .
+RUN pip install -q '.[all]'
```
hrzn, line 26: I think this should fix https://github.com/unit8co/darts/issues/254

TheMP (approved):

Dockerfile
```diff
@@ -23,6 +23,6 @@ ADD . /home/jovyan/work
...
 RUN pip install -q -r requirements/dev.txt
```
camilaagw, line 27: Are the `requirements/dev.txt` really needed?

Dockerfile
camilaagw: I tested this locally, and it does not work. Instead `RUN pip install  .['all']` is working for me

PR #272, opened 2021-02-08
Extend `split_after` and `split_before` to accept float and integers

camilaagw:
### Summary

Add `split` method to timeseries, to split it based on a proportion

The method is intended to be called in the following way:
```
train_series, val_series = series.split(proportion=0.7)
```

Note: For consistency with the methods `split_before` and `split_after` I decided to create a method instead of a function.


darts/timeseries.py
```diff
@@ -304,6 +304,30 @@ def copy(self, deep: bool = True) -> 'TimeSeries':
...
+        Returns
+        -------
+        Tuple[TimeSeries, TimeSeries]
+            A tuple of two time series. The first time series contains the first samples up to the `proportion` ,
+            and the second contains the remaining ones (1-`proportion`).
+        """
+        if (proportion < 0) or (proportion > 1):
+            raise_log(ValueError('Proportion must be between 0.0 and 1.0'))
+
+        split_idx = round(len(self)*proportion)
```
hrzn, line 327: I think you should rely on `utils.get_timestamp_at_point()` here - in fact, you could even have your method accept `Union[pd.Timestamp, float, int]`, which would be quite nice.

darts/timeseries.py
```diff
@@ -304,6 +304,30 @@ def copy(self, deep: bool = True) -> 'TimeSeries':
         else:
             return TimeSeries(self._df, self.freq_str())
 
+    def split(self, proportion: float) -> Tuple['TimeSeries', 'TimeSeries']:
+        """
+        Splits he TimeSeries in two, based on a given proportion.
```
hrzn, line 309: I think here you should specify what is the boundary behavior. What happens when I call `series.split(0.5)` and `series` has length 5? Do I get back 2 + 3 or 3 + 2?
In fact you could perhaps just augment the signature of `split_before()` and `split_after()` in order to support float (and int - see other comment below).

camilaagw: I think augmenting the signature of `split_before()` and `split_after()` is a really good idea. It is also a good idea to use the function `utils.get_timestamp_at_point()` to stick to DRY.  In order to use it, it was necessary to move `utils.get_timestamp_at_point()` to be a method of `TimeSeries` to avoid circular dependencies, so I had to do some refactoring.

Let me know if you agree with this approach. I am still hesitating if the `integer` passed should be an index or the number of samples

darts/timeseries.py
camilaagw: I agree! See my response in the next comment.

darts/timeseries.py
hrzn: > I think augmenting the signature of `split_before()` and `split_after()` is a really good idea. It is also a good idea to use the function `utils.get_timestamp_at_point()` to stick to DRY. In order to use it, it was necessary to move `utils.get_timestamp_at_point()` to be a method of `TimeSeries` to avoid circular dependencies, so I had to do some refactoring.
> 
> Let me know if you agree with this approach. I am still hesitating if the `integer` passed should be an index or the number of samples

I would vote for an index. Number of points might make sense in some context, but I think it might be confusing if it returns something different than the corresponding `__getitem__()`

hrzn (approved):
Looks nice :)

PR #273, opened 2021-02-08
Fix #271: Historical forecasts does not return last point

### LINKED ISSUE #271: [BUG]: historical_forecasts() only returns N-1 predictions - last point missing
Looking at this notebook, for example:
https://unit8co.github.io/darts/examples/05-RNN-examples.html

one can see, for example in the plot produced in cell [12], that in the backtest_series the last point is missing (i.e. the prediction for the last point of the 'actual' time series is missing). 

Unless I misunderstood the historical_forecast() method, this seems to be a small bug.

Could you please, check/fix this issue?
Thanks

camilaagw:
Fixes #271

### Summary

Small fix on historical forecasts so that it returns the last point of the series.

Note: Also refactored code for readability.


hrzn (commented):
Thanks for the fix @camilaagw 
Could you perhaps add a unit test to capture this?

camilaagw:
> Thanks for the fix @camilaagw
> Could you perhaps add a unit test to capture this?

I added a couple of assertions in `test_backtest_forecasting`. Let me know if we should add more, or if they should be added somewhere else

hrzn (approved):
Thanks!

PR #279, opened 2021-02-27
Feature/upgrade to statsmodels v0.12.2

camilaagw:
### Summary

- Updated the statsmodels version to 0.12.2 as ` statsmodels.tsa.arima.model.ARIMA` had some bugs in v0.11.1
- Replace `statsmodels.tsa.arima_model.ARIMA` and `statsmodels.tsa.arima_model.ARMA`  for `statsmodels.tsa.arima.model.ARIMA` and modified the ARIMA constructor to allow for seasonal components. 
- Because `statsmodels.tsa.arima.model.ARIMA` has different defaults than its predecessors, introduced the `trend` parameter to be able to reproduce previous results
- Specify  the `initialization _method` parameter as `estimated` in SimpleExpmoothing to get a behaviour consistent with the legacy one
- 
### Other Information

With the upgrade, there are some differences in the performance of `ARIMA` and `ExponentialSmoothing`. For `ExponentialSmoothing` the difference in performance seemed to be just bad luck as with other datasets it exhibited a better behaviour. For `ARIMA(1, 1, 1)` I still don't know what is the cause of the difference in performance ( see darts/tests/test_local_forecasting_models.py)


camilaagw:
The tests for Python 3.8 work locally.  I am wondering, is there some cache in the CI?

hrzn:
@camilaagw thanks for that!
It seems few tests are still not passing. For instance performance of Theta/4Theta has perhaps changed? (see: https://github.com/unit8co/darts/pull/279/checks?check_run_id=2004254644)
Also it seems there's an issue in intro notebook: https://github.com/unit8co/darts/pull/279/checks?check_run_id=2004254338

hrzn (approved):
Thanks @camilaagw !

PR #285, opened 2021-03-08
Feature/diff

tneuer:
### Summary

Implements the TimeSeries.diff() method wrapping around pandas.DataFrame.diff(). Often used for making TimeSeries stationary.

### Other Information

Additionally tests were added, the possibility to difference multiple times with the "n" parameter and to retain the "na" from pandas.DataFrame.diff() it its functionality wants to be reproduced.


darts/timeseries.py
```diff
@@ -669,6 +669,39 @@ def shift(self, n: int) -> 'TimeSeries':
         new_series.index = new_time_index
         return TimeSeries(new_series, self.freq_str())
 
+    def diff(self, n: Optional[int] = 1,
```
hrzn, line 672: You can put each argument on a different line (including `self`).

darts/timeseries.py
```diff
@@ -669,6 +669,39 @@ def shift(self, n: int) -> 'TimeSeries':
...
+        -------
+        TimeSeries
+            A TimeSeries constructed after differencing.
+        """
+        if not isinstance(n, int) or n < 1:
+             raise_log(ValueError("'n' must be a positive integer >= 1."))
+        if not isinstance(periods, int):
+             raise_log(ValueError("'periods' must be an integer."))
+
+        temp_df = self.pd_dataframe(copy=True)
```
hrzn, line 698: Not sure you need the copying if `diff()` already returns a copy and doesn't work in-place (to be confirmed but I guess that's the case).

darts/timeseries.py
```diff
@@ -669,6 +669,39 @@ def shift(self, n: int) -> 'TimeSeries':
         new_series.index = new_time_index
         return TimeSeries(new_series, self.freq_str())
 
+    def diff(self, n: Optional[int] = 1,
+                periods: Optional[int] = 1,
+                dropna: Optional[bool] = True) -> 'TimeSeries':
+        """
+        Returns a differenced time series with the goal to make a time series stationary.
```
hrzn, line 676: I would slightly rephrase: `Returns a differenced time series. This is often used to make a time series stationary.`

hrzn (approved):
Few small comments but looks good otherwise, thanks!

darts/timeseries.py
tneuer: Done

darts/timeseries.py
tneuer: Done

darts/timeseries.py
tneuer: And done

PR #286, opened 2021-03-08
value_cols now no longer has to be provided if not necessary

tneuer:
### Summary

TimeSeries.from_dataframe() had to take both time_col and value_cols as input, where time_col had be excluded from value_cols. This should be done automatically because time_col is never also a value_col. Now it suffices to give a time_col and Darts takes care of the rest. value_cols can still be provided if the user wants to.


hrzn (approved):
Thanks!

PR #292, opened 2021-03-12
Feature/dummy index

tneuer:
### Summary

This new function allows the creation of a darts.TimeSeries from any `pd.DataFrame` or `pd.Series` if no time stamp is available. Might be useful for longitudinal data measured in steps. An artificial time stamp is added starting at 1970-01-01 00:00:00 with a frequency of 1 second.

### DIFF
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -62,12 +62,16 @@ class ForecastingModel(ABC):
         self.training_series = series
         self._fit_called = True
 
-        if hasattr(self, "trend"):
-            raise_if(
-                self.trend and self.trend != "c" and self.training_series.has_dummy_index,
-                "'trend' is not None. Dummy indexing is not supported in that case.",
-                logger
-            )
+        if series.has_dummy_index:
+            self._supports_dummy_index()
+
+    def _supports_dummy_index(self) -> bool:
+        """ Checks if the forecasting model supports a dummy index.
+
+        By default, returns True. Needs to be overwritten by models that do not support
+        dummy indexing and raise meaningful exception.
+        """
+        return True
 
     @abstractmethod
     def predict(self, n: int) -> TimeSeries:
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -22,7 +22,7 @@ import numpy as np
 import pandas as pd
 
 from ..timeseries import TimeSeries
-from ..logging import get_logger, raise_log, raise_if_not
+from ..logging import get_logger, raise_log, raise_if_not, raise_if
 from ..utils import (
     _build_tqdm_iterator,
     _with_sanity_checks,
@@ -61,6 +61,13 @@ class ForecastingModel(ABC):
                      .format(len(series), str(self), self.min_train_series_length))
         self.training_series = series
         self._fit_called = True
+
+        if hasattr(self, "trend"):
+            raise_if(
+                self.trend and self.trend != "c" and self.training_series.has_dummy_index,
+                "'trend' is not None. Dummy indexing is not supported in that case.",
+                logger
+            )
 
     @abstractmethod
     def predict(self, n: int) -> TimeSeries:
--- a/darts/tests/test_local_forecasting_models.py
+++ b/darts/tests/test_local_forecasting_models.py
@@ -48,8 +48,10 @@ try:
     from ..models import AutoARIMA
     models.append((AutoARIMA(), 13.7))
     extended_models.append(AutoARIMA())
+    PMDARIMA_AVAILABLE = True
 except ImportError:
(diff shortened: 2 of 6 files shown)


hrzn (commented):
For this change, I think that we should go the extra step and modify the constructor and the other methods as well, because IMO our objectives here should be that users could easily create time series with “any” DataFrame without worrying if there’s no DateTimeIndex in it.

darts/models/forecasting_model.py
```diff
@@ -62,6 +62,13 @@ def fit(self, series: TimeSeries) -> None:
         self.training_series = series
         self._fit_called = True
 
+        if hasattr(self, "trend"):
+            raise_if(
+                self.trend and self.trend != "c" and self.training_series.has_dummy_index,
```
hrzn, line 67: I think here instead of check for `self.trend` we should probably call a function `self.supports_dummy_index()` or something like that. This way the super-class doesn't have to know about the specifics of the implementations (e.g. checking `self.trend != 'c'`. It is rather in `AutoARIMA` that we should do this check. We can have a default `ForecastingModel.supports_dummy_index()` always returning `True`, which could be overridden in `AutoARIMA`.

darts/tests/test_local_forecasting_models.py
```diff
@@ -131,3 +132,15 @@ def test_exogenous_variables_support(self):
                 model.fit(self.ts_gaussian, exog=self.ts_gaussian[:-1])
             with self.assertRaises(ValueError):
                 model.fit(self.ts_gaussian[1:], exog=self.ts_gaussian[:-1])
+
+    def test_dummy_series(self):
```
hrzn, line 138: 👍

darts/tests/test_local_forecasting_models.py
```diff
@@ -131,3 +134,16 @@ def test_exogenous_variables_support(self):
...
+        values = np.random.uniform(low=-10, high=10, size=100)
+        ts = TimeSeries(pd.DataFrame({"V1": values}), dummy_index=True)
+
+        varima = VARIMA(trend="t")
+        with self.assertRaises(ValueError):
+            varima.fit(series=ts)
+
+        if PMDARIMA_AVAILABLE:
```
hrzn, line 146: Tests were failing, I added this one.

darts/timeseries.py
```diff
@@ -41,25 +42,38 @@ def __init__(self,
         fill_missing_dates
             Optionally, a boolean value indicating whether to fill missing dates with NaN values
             in case the frequency of `series` cannot be inferred.
+        dummy_index
+            Optionally, if no date time index is present or even available. Creates a dummy index to be
```
hrzn, line 46: I would rephrase: "Optionally, if no date time index is present, this flag will instruct Darts to build a dummy time index in order to obtain a valid TimeSeries. This option can be used in cases where the time index doesn't matter.".

darts/timeseries.py
```diff
@@ -725,14 +740,15 @@ def from_series(pd_series: pd.Series,
         TimeSeries
             A TimeSeries constructed from the inputs.
```
hrzn, line 741: You should also add the documentation for dummy_index here.

darts/timeseries.py
```diff
@@ -758,20 +774,21 @@ def from_dataframe(df: pd.DataFrame,
         TimeSeries
             A univariate or multivariate TimeSeries constructed from the inputs.
```
hrzn, line 775: Here too, you should document dummy_index

darts/timeseries.py
```diff
@@ -806,6 +823,28 @@ def from_times_and_values(times: pd.DatetimeIndex,
             df.columns = columns
         return TimeSeries(df, freq, fill_missing_dates)
 
+    def create_dummy_index(self, start="19700101", freq="S") -> 'TimeSeries':
```
hrzn, line 826: Probably this method should be private (i.e. rename `_create_dummy_index`)

darts/timeseries.py
```diff
@@ -832,7 +871,15 @@ def plot(self,
                 kwargs['figure'] = plt.gcf()
                 if 'label' in kwargs:
                     kwargs['label'] = label + '_' + str(i)
-            self.univariate_component(i).pd_series().plot(*args, **kwargs)
+            if self.has_dummy_index:
```
hrzn, line 874: 👍

darts/timeseries.py
tneuer: Done

darts/timeseries.py
tneuer: Done

darts/timeseries.py
tneuer: Done

darts/timeseries.py
tneuer: Done

hrzn (approved):

PR #294, opened 2021-03-14
Feature/Kalman filter 

mpetros:
Implements DARTS-19 Kalman filter

### Summary

Simple Kalman filter model class implementation using Timeseries. Introducing filtering class model to abstract filter type prediction method signature for a model

### Other Information

-- Variables in Kalman filter are intentionally named as they are since those are standard variable names across many implementations, literature and examples.


darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
...
+import numpy as np
+import pandas as pd
+
+from ..timeseries import TimeSeries
+from ..logging import get_logger, raise_log, raise_if_not
+
+logger = get_logger(__name__)
+
+
+class FilteringModel(ABC):
```
hrzn, line 29: I'm not sure `FilteringModel` is the best name. To me filtering time series really sounds too much like removing low frequency or high frequency components. How about something like `SmoothingModel` ?

darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
+"""
+Filtering Model Base Class
+------------------------------
+
+A filtering model captures the measured values of a time series as a function of the past as follows:
+
+.. math:: y_{t+1} = f(y_{t+1}, y_t, y_{t-1}, ..., y_1),
```
hrzn, line 7: This is the same definition as the forecasting model.
What do we want these "smoothing models" to do? In my understanding they should rather do something like 
constructing a new series `z` of the same length as `y`, where each element `z_t` is a function of the whole series `y`, no?

darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
...
+            A target time series. The model will be trained to forecast this time series.
+        """
+        raise_if_not(len(series) >= self.min_train_series_length,
+                     "Train series only contains {} elements but {} model requires at least {} entries"
+                     .format(len(series), str(self), self.min_train_series_length))
+        self.training_series = series
+        self._fit_called = True
+
+    @abstractmethod
+    def filter(self) -> TimeSeries:
```
hrzn, line 60: I think we should probably also call this one `predict()`

darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
...
+        self._fit_called = False
+
+    @abstractmethod
+    def fit(self, series: TimeSeries) -> None:
+        """ Trains the model on the provided series
+
+        Parameters
+        ----------
+        series
+            A target time series. The model will be trained to forecast this time series.
```
hrzn, line 51: `A target time series. The model will be fit on this time series.`

darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
...
+    @property
+    def min_train_series_length(self) -> int:
+        """
+        Class property defining the minimum required length for the training series.
+        This function/property should be overridden if a value higher than 3 is required.
+        """
+        return 3
+
+
+class MovingAverage(FilteringModel, ABC):
```
hrzn, line 87: Good example

darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
...
+    
+    """ Moving average class which implements Filtering Model and shows
+    a simple example of filtering data using moving average.
+    """
+
+    def __init__(self, window):
+        self.values = [.0 for _ in range(window)]
```
hrzn, line 94: I wouldn't keep this as class variables. I think it should be re-initialised every time `filter()` (or `predict()`) is called.

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+        Q : float
+            Process uncertainty/noise multiplicator
+        kf : KalmanFilter
+            Instance of Kalman Filter (for advanced configuration)
+        """
+        super().__init__()
+        if kf is None:
+            self.dim_x = dim_x
+            self.dim_z = dim_z
+            self.kf = KalmanFilter(dim_x=dim_x, dim_z=dim_z)
```
hrzn, line 53: Shouldn't all of this be done on `predict()` ?

darts/models/filtering_model.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self):
+        # The series used for training the model through the `fit()` function.
+        # This is only used if the model has been fit on one time series.
+        self.training_series: Optional[TimeSeries] = None
+
+        # state; whether the model has been fit (on a single time series)
+        self._fit_called = False
+
+    @abstractmethod
+    def fit(self, series: TimeSeries) -> None:
```
hrzn, line 45: Is there a need for a `fit()` method for these models?

hrzn (commented):
I added a few comments, but before I review more, could you please re-open the PR into the `develop` branch instead of `master`? (We keep `master` for releases).

darts/models/filtering_model.py
mpetros: There is a little change in indexing of the formula. Imho any filter/smoothing function looks at current and past values in general. Thats why i said t+1=t+1. I will add more description to avoid confusion with predicting formula

darts/models/filtering_model.py
mpetros: agree. Renamed in the next iteration

darts/models/filtering_model.py
mpetros: predicting model "predicts", filtering model would "filter", now renamed to smoothing model, is "smooth()" the most accurate one? feel free to overrule and I call it predict.

darts/models/filtering_model.py
mpetros: corrected, thx

darts/models/filtering_model.py
mpetros: just to be able to pass the data and not consume CPU time yet?

darts/models/kalman.py
mpetros: This are all initial parameters.  Or am I not getting something?

darts/models/filtering_model.py
mpetros: this was meant to be FIFO for moving average

mpetros:
> I added a few comments, but before I review more, could you please re-open the PR into the `develop` branch instead of `master`? (We keep `master` for releases).

changed within this PR to keep comments

darts/models/kalman.py
hrzn: What I meant is that `self.kf` will contain some state, and this state should be reset every time we want to re-fit the model on a new series (the hyper parameters can be kept, but the model should be re-init). Similar e.g. to what we are doing with ARIMA, ExponentialSmoothing etc.

I'm saying `predict()` (and not `fit`) because I don't think we need a `fit` function for `SmoothingModel`s. (Also I'm having second thoughts I think `FilteringModel` with a `filter()` function [and not a `predict()` one] might have been better in fact 😅 )

darts/models/smoothing_model.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy as np
+import pandas as pd
+
+from ..timeseries import TimeSeries
+from ..logging import get_logger, raise_log, raise_if_not
+
+logger = get_logger(__name__)
+
+
+class SmoothingModel(ABC):
```
hrzn, line 30: After second thoughts, I would actually call `FilteringModel`. Initially a filtering model made me think about lowpass/bandpass filters, but in fact I think Kalman filter can be see as a lowpass filter (and moving average too), so I think you initial name was more appropriate (sorry for that!)

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+        more precise values compared to noise interfered data
+        
+        Parameters
+        ----------
+        dim_x : int
+            Size of the Kalman state vector
+        dim_y : int
+            Size of the measurement vector
+        x : numpy array
+            filter state estimate
```
hrzn, line 37: Perhaps say it is initial state estimate

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class Kalman(SmoothingModel, ABC):
+
+    def __init__(
+            self, 
+            dim_x: int = 1, 
+            dim_z: int = 1, 
+            x = np.array([[1.]]),
+            F = np.array([[1]]),
```
hrzn, line 20: This could be None by default, and initialised to a default matrix of appropriate dimensions (depending on `dim_x`). Similar for H and P.

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+from .smoothing_model import SmoothingModel
+from ..timeseries import TimeSeries
+
+
+class Kalman(SmoothingModel, ABC):
+
+    def __init__(
+            self, 
+            dim_x: int = 1, 
+            dim_z: int = 1,
```
hrzn, line 18: Should this be called `dim_y` (following the docstring)?
Also if I understand well it should match the dimensionality of the TimeSeries it's used for, so I think it would be nice to lazily set this variable upon calling `filter(ts)` to the value of `ts.width`.

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        super().__init__()
+        if kf is None:
+            self.dim_x = dim_x
+            self.dim_z = dim_z
+            self.kf = KalmanFilter(dim_x=dim_x, dim_z=dim_z)
+            self.kf.F = F
+            self.kf.H = H
+            self.kf.x = x
+            self.kf.P *= 1000.
```
hrzn, line 57: Where do these 4 magic lines come from?

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+    def fit(self, series: TimeSeries):
+        super().fit(series)
+
+    def smooth(self):
+        return self.training_series.map(self._kalman_iteration)
+
+    def _kalman_iteration(self, observation):
+        self.kf.predict()
+        self.kf.update(observation)
+        return self.kf.x[0][0]
```
hrzn, line 76: Does this work with multivariate series?

darts/models/kalman.py
```diff
@@ -0,0 +1,81 @@
...
+        
+        Parameters
+        ----------
+        dim_x : int
+            Size of the Kalman state vector
+        dim_y : int
+            Size of the measurement vector
+        x : numpy array
+            filter state estimate
+        F : numpy array
```
hrzn, line 38: It would be nice to state what are the expected dimensions of these matrices (e.g. `dim_x x dim_x`)

darts/models/smoothing_model.py
```diff
@@ -0,0 +1,109 @@
...
+
+    """ The base class for smoothing models. It defines the *minimal* behavior that all smoothing models have to support.
+        The signatures in this base class are for "local" models handling only one series and no covariates.
+        Sub-classes can handle more complex cases.
+    """
+    @abstractmethod
+    def __init__(self):
+        # The series used for training the model through the `fit()` function.
+        # This is only used if the model has been fit on one time series.
```
hrzn, line 39: Does this sentence still apply? Is there the possibility to fit some filtering models on several series?

darts/models/smoothing_model.py
```diff
@@ -0,0 +1,109 @@
+"""
+Smoothing Model Base Class
+------------------------------
+
+A smoothing model more accurate value from measured values of a time series
+Using the current value and historic values as follows:
+
+.. math:: y_{t+1} = f(y_{t+1}, y_t, y_{t-1}, ..., y_1),
```
hrzn, line 8: This would mean that these models always represent current values as a function of the past only. I think that's not accurate (e.g. moving average at position t+1 also depends of signal at position t+2). This equation was useful for forecasting models but perhaps not really needed for filtering models...
If you want to keep it perhaps rewrite it as something like \hat{y}_{t} = f(y_1, ..., y_t, ... y_T).

darts/models/filtering_model.py
hrzn: Actually on second thoughts I think you were right and `filter()` might be the way to go :D sorry for that...

darts/models/filtering_model.py
hrzn: Yes FIFO is good, but I wouldn't keep the values as a class variable. Instead I think you should just re-init the array when `filter()` is called.

darts/models/filtering_model.py
hrzn: I don't think that's really something that would be needed in practice. I really think all we need is a `predict(series: TimeSeries)` function. Adding a `fit()` that does nothing complexifies the API (and `fit()` would be a poor name for a function that does nothing except storing the data).

hrzn (commented):
I still have a few comments. Mainly:
* I think we should have cleaner treatment of the KF parameters: inferring the measurements dimensions based on the TimeSeries dimension, and creating default matrices of appropriate dimensions if no matrix is specified.
* `FilteringModel` was in fact a better name I think (sorry!) and we should have only a `filter(series: TimeSeries)` function, no `fit()` function
* Could you make sure it properly handles multivariate TimeSeries' ?

darts/models/smoothing_model.py
mpetros: no, problem, reverted.

darts/models/kalman.py
mpetros: sure.

darts/models/kalman.py
mpetros: ok, understood.

darts/models/filtering_model.py
mpetros: I am keeping just the "filter" function, to keep the naming in sync with model name. fit function remvoed and kf re-initialized.

darts/models/kalman.py
mpetros: sure for dim_x. H and P have the "state" dimensions not "measure" dimensions. I will make a comment about it, however in real-life kf should be passed with a custom configuration. I will mention that in doc string.

darts/models/kalman.py
mpetros: corrected.

darts/models/smoothing_model.py
mpetros: not anylonger. Thanks for the hint.

darts/models/smoothing_model.py
mpetros: is this valid? its past and current value?

darts/models/filtering_model.py
mpetros: ok, I see.

darts/models/kalman.py
```diff
@@ -0,0 +1,95 @@
...
+    def filter(self, series: TimeSeries):
+
+        if self.kf is None:
+            # default Kalman configuration
+            dim_z = series.width
+            self.kf = KalmanFilter(dim_x=self.dim_x, dim_z=dim_z)
+            self.kf.F = self.F
+            self.kf.H = self.H
+            self.kf.x = self.x
+            self.kf.P *= self.P
```
hrzn, line 78: Here you are treating `self.P` as a scalar, not a matrix.

darts/models/kalman.py
```diff
@@ -0,0 +1,95 @@
...
+            self.kf.Q[-1,-1] *= self.Q
+            self.kf.Q[4:,4:] *= self.Q
```
hrzn, line 80: where does this come from?? (It will have no effect for any process of dimensionality less than 4)
Why not just using `self.Q` which is supposed to be a matrix of correct dimensions?

darts/models/kalman.py
hrzn: Did you check what happens with multivariate series here?

darts/models/kalman.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def _kalman_iteration(self, observation):
+        self.kf.predict()
+        self.kf.update(observation)
+        return self.kf.x[0][0]
+
+
+    @property
+    def min_train_series_length(self) -> int:
+        return 10
```
hrzn, line 95: Is there a reason why 10 here?

hrzn (commented):
There are still a few things that I think should be adapted before we can merge. In particular the way the parameters are treated is a bit weird (different defaults in different places, some matrices and scalars mixed up). I will try to propose a PR into your branch with some fixes.

darts/models/kalman.py
mpetros: this was just an educated guess (filtering less than 10 measurements doesn't add much value)

mpetros: this was just an educated guess - filtering under 10 values wouldn’t  really add any value?

darts/models/kalman.py
```diff
@@ -0,0 +1,95 @@
...
+from abc import ABC
+
+from typing import Optional
+from filterpy.kalman import KalmanFilter
+
+from .filtering_model import FilteringModel
+from ..timeseries import TimeSeries
+
+
+class Kalman(FilteringModel, ABC):
```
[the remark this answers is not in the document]
mpetros: sure

darts/models/kalman.py
mpetros: I have an issue with the "self.training_series.map" -> no idea how to best map a function returning multivariate series?

hrzn:
@mpetros Just a note on this PR - I think it might greatly benefit from the new probabilistic TimeSeries (this way the series returned by `KalmanFilter.filter()` can be probabilistic (i.e., take the covariance matrix into account) instead of being the mean state only. For this reason I want to wait a little more that we merge https://github.com/unit8co/darts/pull/350/ and then (slightly) rework this one to make it probabilistic.

darts/models/filtering_model.py
tomasvanpottelbergh: Why does this class and `KalmanFilter` inherit from `ABC`? These are not supposed to be abstract classes I think?

hrzn (approved):

PR #296, opened 2021-03-15
Fix/Minor tweaks for fit and predict functions

pennfranc:
I was playing around with the new `GlobalForecastingModel` and ran into a couple of issues when predicting with a model that was trained with covariates. I changed 3 lines, which helped. I don't have a grasp of all the underlying code anymore though, so this might come from a place of ignorance.


darts/models/forecasting_model.py
```diff
@@ -599,6 +599,8 @@ def fit(self,
             super().fit(series)  # handle the single series case
         if covariates is not None:
             self._expect_covariates = True
+            self.training_series = series
```
hrzn, line 602: Normally this should be done only when the model is fit on one TimeSeries. Here `series` might be a `Sequence[Series]`, in which case we don't want to store it (in such cases the user is forced when calling `predict()` to specify which series needs to be forecasted).

Storing the (single) series as well as setting `self._fit_called` is handled by the line just above:
```
super().fit(series)
```

My guess is that your problems were due to trying to use a model this way:
```
my_model.fit(a_sequence_of_series)
pred = my_model.predict()
```
Instead of:
```
my_model.fit(a_sequence_of_series)
pred = my_model.predict(a_series_to_forecast)
```

Could you confirm if that's the problem you experienced? If yes it's not a bug and there's no need for a correction :)

hrzn (commented):
Nice to see you playing with Darts again @pennfranc :)

darts/models/forecasting_model.py
pennfranc: Thanks for the clarification! But what about the case when we fit a model on just one TimeSeries but include covariates? Currently it looks like we have to pass the covariates again, as well as the target series.

I encountered the issue when running the TCN example notebook. After training the model with 
```
model_air.fit(series=train, covariates=train_year, val_series=val, val_covariates=val_year, verbose=True)
```
I call `model_air.predict(12)` which gives me an error message saying
> For global models, if `predict()` is called without specifying a series, the model must have been fit on a single training series.

even though it has technically been fit on a single training series. If this is just an imprecise error message, is there a reasoning behind requiring to pass both the train and covariate series again during prediction time after training on a single series + covariates? Or am I just completely confused? Haha

hrzn: You are right indeed, good catch! We should store the `series` when it's a single `TimeSeries` even if covariates are used. Your proposal goes in the right direction, however here you're storing `series` even if it's a Sequence, which we shouldn't do. I think the logic should look more like:
```
if covariates is not None:
            self._expect_covariates = True
            if isinstance(series, TimeSeries):
                    self.training_series = series
                    self._fit_called = True
```

pennfranc: Makes sense, I added the condition like you suggested. Do you think it would make sense to also save the covariates when training on a single series for the sake of consistency and simplicity?

hrzn: Yes, I think that's a good idea

hrzn: @pennfranc do you want to do this last little change? I think we could merge it afterwards.

pennfranc: @hrzn I know I'm pretty late, but I added the last change including test cases

hrzn: Nice, thanks!

hrzn (approved):

PR #308, opened 2021-04-09
Batch prediction for Torch models

mkos:
### Summary

Until now, prediction in Torch-based models was done on a sample-by-sample basis, which is inefficient when one wants to predict larger amount of samples. This PR introduces updates to `fit_from_dataset()` that allows to use batched prediction and ability to explicitly provide tensors (Torch and NumPy) instead sequence of timeseries object.

### Other Information

* The change is backward compatible and no updates to your existing code is required.
* On a 145k in case of RNN model the prediction time dropped from around 1h to under 2s.


mkos:
Wrong branch, will recreate.

PR #309, opened 2021-04-09
 Batch prediction for Torch models

mkos:
### Summary

Until now, prediction in Torch-based models was done on a sample-by-sample basis, which is inefficient when one wants to predict larger amount of samples. This PR introduces updates to `predict_from_dataset()` that allows to use batched prediction and ability to explicitly provide tensors (Torch and NumPy) instead sequence of timeseries object.

### Other Information

* The change is backward compatible and no updates to your existing code is required.
* On a 145k in case of RNN model the prediction time dropped from around 1h to under 2s.


darts/models/torch_forecasting_model.py
```diff
@@ -401,70 +405,202 @@ def predict(self,
...
+            ts_forecasts = predictions_tensor.cpu().detach().numpy()
+
+        else:
+            raise_log(
+                AttributeError('Unsupported data type in `input_series_dataset`. Available data types are: '
+                               'TimeSeriesInferenceDataset, torch.Tensor or NumPy array'))
+
+        return ts_forecasts
+
+    def _produce_prediction(self, in_dataset: torch.utils.data.DataLoader, n: int, verbose: bool = False) -> torch.Tensor:
```
hrzn, line 579: +1

darts/models/torch_forecasting_model.py
```diff
@@ -401,70 +405,202 @@ def predict(self,
...
         dataset = SimpleInferenceDataset(series, covariates)
-        predictions = self.predict_from_dataset(n, dataset)
+        predictions = self.predict_from_dataset(n, dataset, verbose=verbose)
         return predictions[0] if called_with_single_series else predictions
 
     def predict_from_dataset(self,
                              n: int,
-                             input_series_dataset: TimeSeriesInferenceDataset
-                             ) -> Sequence[TimeSeries]:
+                             input_series_dataset: Union[TimeSeriesInferenceDataset, torch.Tensor, np.ndarray],
```
hrzn, line 413: Unfortunately I don't think it's a good idea to accept tensors of Numpy arrays here. By doing this, we would loose a lot of the nice "all included" services that Darts provides. Users would have to worry about building tensors or arrays, which goes against the principles of Darts (the only thing users have to know are `TimeSeries`). Users that don't want to shouldn't even have to know that we are using Pytorch under the hood.

The good news is that this can be done very easily. You already built the most importance piece of the puzzle, which is the batched `_produce_prediction()` method. This is very good. Now in order to have fast inference using `TimeSeriesInferenceDataset` I think you should (roughly) follow the following steps:

* Change the `SimpleInferenceDataset.__getitem__()` so that it returns `np.ndarray` instead of `TimeSeries`. Creating all these `TimeSeries` is likely to be the time consuming thing here. Probably doing something like
```
    def __getitem__(self, idx: int) -> Tuple[np.ndarray, Optional[np.ndarray]]:
        return (self.series[idx].values(copy=False), None) if self.covariates is None else (self.series[idx].values(copy=False), self.covariates[idx].values(copy=False))
```
is enough. Accessing the values without copying will be fast because it uses Numpy views.

* Instantiate a `TimeSeriesTorchDataset` (like we are doing for training). There's nothing more to do here, it should work out of the box.

* Create your torch `DataLoader` from this `TimeSeriesTorchDataset` (that too should work out of the box).

That should remove quite a bit of code from what you currently have, but most importantly will allow users to not have to worry about tensors and array, but just about `TimeSeries`.

darts/tests/test_global_forecasting_models.py
```diff
@@ -113,3 +113,160 @@ def test_covariates(self):
                 mape_err = mape(self.ts_pass_val, pred)
                 self.assertTrue(mape_err < err, 'Model {} produces errors too high (several time '
                                                 'series with covariates). Error = {}'.format(model_cls, mape_err))
+        
+
```
hrzn, line 117: one too many blank line

darts/tests/test_global_forecasting_models.py
```diff
@@ -113,3 +113,160 @@ def test_covariates(self):
                 mape_err = mape(self.ts_pass_val, pred)
                 self.assertTrue(mape_err < err, 'Model {} produces errors too high (several time '
                                                 'series with covariates). Error = {}'.format(model_cls, mape_err))
+        
+        
+        def test_predict_from_dataset_unsupported_input(self):
+            # an exception should be thrown if an unsupported type is passed
+            UNSUPPORTED_TYPE = 'UNSUPPORTED_TYPE'
```
hrzn, line 120: should be lowercase in a function

darts/tests/test_global_forecasting_models.py
```diff
@@ -113,3 +113,160 @@ def test_covariates(self):
...
+            # just need to test this with one model
+            model_cls, kwargs, err = models_cls_kwargs_errs[0]
+            model = model_cls(input_chunk_length=IN_LEN, output_chunk_length=OUT_LEN, **kwargs)
+            model.fit([self.ts_pass_train, self.ts_pass_train_1])
+            
+            with self.assertRaises(AttributeError):    
+                model.predict_from_dataset(n=1, input_series_dataset=UNSUPPORTED_TYPE)
+            
+
```
hrzn, line 129: remove blank line

darts/tests/test_global_forecasting_models.py
```diff
@@ -113,3 +113,160 @@ def test_covariates(self):
...
+            model_cls, kwargs, err = models_cls_kwargs_errs[0]
+            model = model_cls(input_chunk_length=IN_LEN, output_chunk_length=OUT_LEN, **kwargs)
+            model.fit([self.ts_pass_train, self.ts_pass_train_1])
+            
+            with self.assertRaises(AttributeError):    
+                model.predict_from_dataset(n=1, input_series_dataset=UNSUPPORTED_TYPE)
+            
+                
+        def test_multi_ts_numpy_and_tensor(self):
+            # testing the predictions with np.ndarray/torch.Tensor as input. The function should return a np.ndarray/torch.Tensor as well,
```
hrzn, line 131: Some lines are too long (max 120 characters)

darts/tests/test_global_forecasting_models.py
```diff
@@ -113,3 +113,160 @@ def test_covariates(self):
...
+                          covariates=[self.time_covariates_train])
+                
+                pred = model.predict_from_dataset(n=PRED_LEN, input_series_dataset=input_array)
+                
+                # checking the overall returned shape
+                expected_shape = (input_array.shape[0], PRED_LEN, multivariate_array.shape[1]) # (n_samples, prediction_length, prediction_width)
+                self.assertEqual(expected_shape, pred.shape, 'Model prediction size not as expected.'
+                                 'Expected shape: {}, current shape: {}'.format(expected_shape, pred.shape))
+             
+        def test_multivariates_and_covariates_torch(self):
```
hrzn, line 245: I guess you should be able to remove most of these tests if you implement the changes I suggest and drop tensors and array support :D

hrzn (changes requested):
Thanks a lot guys, this is a step in the good direction!
However, we unfortunately cannot support this signature of `predict_from_dataset()` with tensors and arrays. I've suggested another (simpler!) path for making inference fast supporting only the `TimeSeriesInferenceDataset`s.

darts/models/torch_forecasting_model.py
```diff
@@ -401,70 +411,143 @@ def predict(self,
...
-            in_sequence[:, -roll_size:, :] = out[:, :roll_size, :]
+        predictions_tensor = self._produce_prediction(pred_loader, n, verbose=verbose)
 
-            # take only last part of the output sequence where needed
-            out = self.model(in_sequence)[self.first_prediction_index:]
+        # postprocessing
 
-            prediction.append(out[0, :, :])
+        predictions_array = predictions_tensor.cpu().detach().numpy()
+        ts_forecasts = []
```
hrzn, line 509: this doesn't seem to be used.

hrzn (commented):
Looks good! I just got a few minor comments. Before you merge, you'll need to solve the conflict with `develop` and make sure you include the changes that were merged from https://github.com/unit8co/darts/pull/310/files.

hrzn (approved):
thanks!

PR #311, opened 2021-04-11
restore compatibility with Python 3.6

hrzn:
In an earlier PR I removed compatibility with Python 3.6 in order to follow Numpy 1.20+ recommendation. However due to some issues with Python 3.9 on MacOS, I had to restore Numpy 1.19.5, and so I'm restoring Python 3.6 compatibility here.


hrzn:
Well, I see now that scipy requires 3.7+, so I'll close this PR and we'll drop Python 3.6 here too.

PR #315, opened 2021-04-19
basic implementation of walk_forward_split() and train_test_split()

mkos:
Fixes #.

### DIFF
--- a/darts/tests/utils/test_model_selection.py
+++ b/darts/tests/utils/test_model_selection.py
@@ -1,0 +1,104 @@
+from ..base_test_class import DartsBaseTestClass
+from darts.utils.model_selection import train_test_split
+from darts.utils.timeseries_generation import constant_timeseries
+
+
+def make_dataset(rows, cols):
+    return [constant_timeseries(i, cols) for i in range(rows)]
+
+
+def verify_shape(dataset, rows, cols):
+    return (
+        len(dataset) == rows and
+        all(len(row) == cols for row in dataset)
+    )
+
+
+class ClassTrainTestSplitTestCase(DartsBaseTestClass):
+    # test 1
+    # test parameters for axis = 0
+    def test_parameters_for_axis_0(self):
+        train_test_split(make_dataset(2, 10), axis=0, test_size=1)
+
+        # expecting no exception
+        self.assertTrue(True)
+
+    # test 2
+    # test parameteres for axis = 1
+    def test_parameters_for_axis_1_no_n(self):
+        with self.assertRaises(AttributeError):
+            train_test_split(make_dataset(1, 10), axis=1, horizon=1)
+
+    def test_parameters_for_axis_1_no_horizon(self):
+        with self.assertRaises(AttributeError):
+            train_test_split(make_dataset(1, 10), axis=1, n=1)
+
+    # test 3
+    # test empty data input
+    def test_empty_dataset(self):
+        with self.assertRaises(AttributeError):
+            train_test_split([])
+
+    # test 4
+    # number of samples too small, horizontal split
+    # input is 1 timeseries of length 10, axis = 0, default split
+    def test_horiz_number_of_samples_too_small(self):
+        with self.assertRaises(AttributeError):
+            train_test_split(make_dataset(1, 10))
+
+    # test 5
+    # sunny day horizontal split
+    # input is 4 timeseries of length 10, axis = 0 default split
+    def test_sunny_day_horiz_split(self):
+        train_set, test_set = train_test_split(make_dataset(4, 10))
+
+        self.assertTrue(
+            verify_shape(train_set, 2, 10) and
+            verify_shape(train_set, 2, 10)
+        )
+
+    # test 6
+    # timeseries too short, vertical split
+    # input is 2 timeseries of length 1, axis = 1, n = 10, horizon = 10, default split .25
+    def test_timeseries_too_short_vertical_split(self):
+        self.assertTrue(False)
+
+    # test 7
+    # sunny day vertical split
+    # input is 2 timeseries of length 250, axis = 1, n = 70, horizon = 50, default split .25
+    def test_sunny_day_vertical_split(self):
+        train_set, test_set = train_test_split(make_dataset(2, 250), axis=1, n=70, horizon=50)
+
+        self.assertTrue(
+            verify_shape(train_set, 2, 200) and
+            verify_shape(train_set, 2, 150)
+        )
+
+    # test 8
+    # training set insufficient size
+    # input is 2 timeseries of lenght 250
+    def test_training_set_insufficient_size(self):
+        with self.assertRaises(AttributeError):
+            pass
+
+    # test 9
+    # test_split absolute number horizontal split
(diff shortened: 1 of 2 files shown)


darts/utils/model_selection.py
```diff
@@ -0,0 +1,79 @@
+from typing import Sequence, Optional, Union
+from ..timeseries import TimeSeries
+
+
+def train_test_split(
+        data: Sequence[TimeSeries],
```
hrzn, line 6: Could you implement support for `Union[TimeSeries, Sequence[TimeSeries]]` ?

darts/utils/model_selection.py
```diff
@@ -0,0 +1,79 @@
...
+        test_size: Optional[Union[float, int]] = 0.25,
+        axis: Optional[int] = 0,
+        n: Optional[int] = 0,
+        horizon: Optional[int] = 0,
+        ) -> Sequence[Sequence[TimeSeries]]:
```
hrzn, line 11: Should be `Tuple[Sequence[TimeSeries]]` ? Or actually even `Union[Tuple[TimeSeries], Tuple[Sequence[TimeSeries]]]` to support splitting simple `TimeSeries`.

darts/utils/model_selection.py
```diff
@@ -0,0 +1,79 @@
...
+            index = test_size
+
+        return data[:index], data[index:]
+
+    elif axis == 1:
+
+        if horizon == 0 or n == 0:
+            raise AttributeError("You need to provide non-zero `horizon` and `n` parameters when axis=1")
+
+        train_set = []
```
hrzn, line 57: This assumes that the datasets fit into memory.
Another option to scale to bigger datasets here could be to build our own Sequence implementation. Something like:
```
class MyTimeSeriesSequence(Sequence):
    def __init__(data):
        self.data = data

    def __getitem__(idx: int):
        ts = self.data[idx]
        ts_length = len(ts)
        ...
        return ts[:train_end_index], ts[test_start_index:]

...

return MyTimeSeriesSequence(data)
```

In this example it would make the return type of the method `Sequence[Tuple]` instead of `Tuple[Sequence]`, but it can be changed if `Tuple[Sequence]` makes more sense (wdyt about this?)

darts/utils/model_selection.py
```diff
@@ -0,0 +1,79 @@
...
+
+            test_start_index = train_end_index - n
+
+            train_set.append(ts[:train_end_index])
+            test_set.append(ts[test_start_index:])
+
+        return train_set, test_set
+
+    else:
+        raise AttributeError()
```
hrzn, line 79: Maybe an error message explaining what `axis` should be?

darts/tests/utils/test_model_selection.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_parameters_for_axis_0(self):
+        train_test_split(make_dataset(2, 10), axis=0, test_size=1)
+
+        # expecting no exception
+        self.assertTrue(True)
+
+    # test 2
+    # test parameteres for axis = 1
+    def test_parameters_for_axis_1_no_n(self):
+        with self.assertRaises(AttributeError):
```
hrzn, line 29: why should this fail?

darts/tests/utils/test_model_selection.py
```diff
@@ -0,0 +1,104 @@
...
+            train_test_split(make_dataset(1, 10), axis=1, horizon=1)
+
+    def test_parameters_for_axis_1_no_horizon(self):
+        with self.assertRaises(AttributeError):
```
hrzn, line 33: why should this fail? you have a default value for horizon.

darts/tests/utils/test_model_selection.py
```diff
@@ -0,0 +1,104 @@
...
+    # test 10
+    # test_split absolute number vertical split
+    # input is 4 timeseries of length 10, axis = 1, test_split = 2
+    def test_test_split_absolute_number_vertical(self):
+        train_set, test_set = train_test_split(make_dataset(4, 10), axis=1, test_size=2, n=1, horizon=2)
+
+        self.assertTrue(
+            verify_shape(train_set, 4, 8) and
+            verify_shape(train_set, 4, 4)
+        )
```
hrzn, line 104: Might be nice to add a test case for checking the train / val shapes when the timeseries have unequal lengths.

hrzn (commented):
You still have some failing tests.

darts/utils/model_selection.py
mkos: done

darts/tests/utils/test_model_selection.py
mkos: Since `n` and `horizon` are only valid parameters for `axis=1` they are set by default to 0 to cover the case of `axis=0`. But if the `axis=1` and any of these is set to 0, then it's an error.

darts/tests/utils/test_model_selection.py
mkos: Since `n` and `horizon` are only valid parameters for `axis=1` they are set by default to 0 to cover the case of `axis=0`. But if the `axis=1` and any of these is set to 0, then it's an error.

darts/utils/model_selection.py
mkos: How about there's going to be additional parameter where you can provide your own implementation of sequence that defaults to `list()`?

darts/tests/utils/test_model_selection.py
mkos: done.

darts/utils/model_selection.py
mkos: (I have implemented this behavior)

hrzn: That would make it quite complex to use for users no? It means they would have to implement their own class. Whereas if you implement the class for them they don't have to worry about it. Is there a use-case where a class looking like the one I'm proposing above wouldn't work?

mkos: Everything is fine with your implementation. I was always hesitant to implement things that may not be used at all, I'm a big fan of "less is more". I'm assuming, based on what you said about use cases, that in 90% of the cases `list()` will be completely sufficient. For the remaining 10% of the cases, where large datasets are used, I'm assuming these will be the power users that know how to implement it. I'm also a big proponent of flexibility with sane defaults. And I can imagine some exotic cases where custom implementation will be required (eg. storing data on the disk, using a joblib to manage memory or using tensors as an internal repr etc.).

darts/utils/model_selection.py
```diff
@@ -0,0 +1,79 @@
...
+            ts_length = len(ts)
+            train_end_index = ts_length - horizon
+
+            if 0 < test_size < 1:
+                test_size = int((ts_length - horizon) * (test_size))
+                print('test_size: ', test_size)
+            if train_end_index < n:
+                raise AttributeError("Test size is too small: training timeseries is of 0 size")
+
+            test_start_index = ts_length - horizon - n - test_size - 1
```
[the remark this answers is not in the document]
mkos: - add an option for type of vertical cut `simple` and `model-aware` (default)
- change `n` to `input_size`

mkos:
Will open regular PR.

PR #316, opened 2021-04-20
Feature/random forest

tneuer:
The branch was originally created to enable the usage of the random forest algorithm. I used this chance to greatly simplify the RegressionModel base class as discussed, so the branch name does not really entail everything that was done.

Now the RegressionModel inherits from the ForecastingModels and enables it to use all its features. Any scikit-learn algorithm can be passed to the `model` argument in the constructor. This restructuring caused some problems down the line for ensemble models and some tests failed. I corrected all of them to fit in the general framework of Darts.

The StandardRegressionModel and RandomForest are basically wrappers around RegressionModel for the specific case of certain scikit-learn regressors.

### Summary

Restructuring of RegressionModel, StandardRegressionModel and EnsembleModel.


darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
-        # state
+        Parameters
+        ----------
+        lags : Union[int, list]
+            Number of lagged target values used to predict the next time step. If an integer is given
+            the last `lags` lags are used (inclusive). Otherwise a list of integers with lags.
+        lags_exog : Union[int, list, bool]
+            Number of lagged exogenous values used to predict the next time step. If an integer is given
+            the last `lags_exog` lags are used (inclusive). Otherwise a list of integers with lags. If False,
+            the value at time `t` is used which might lead to leakage. If True the same lags as for the
```
hrzn, line 53: Perhaps give a little more explanation after "which might lead to leakage": when does it (or does it not) lead to leakage.

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
+        elif isinstance(self.lags_exog, list):
+            for lag in self.lags_exog:
+                raise_if(not isinstance(lag, int) or (lag < 0), (
+                    "Every element of `lags_exog` must be a positive integer. Given: {}."
+                    .format(self.lags_exog))
+                )
+
+        self.model = model
+        if self.lags is not None and self.lags_exog is None:
+            self.max_lag = int(np.max(self.lags))
```
hrzn, line 107: you can also simply use the built-in `max()` instead of `np.max()`.

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
+            TimeSeries object containing the exogenous values.
         """
+        super().fit(series, exog)
+        raise_if(exog is not None and self.lags_exog is None,
+            "`exog` not None in `fit()` method call, but `lags_exog` is None in constructor. "
+        )
+        raise_if(exog is None and self.lags_exog is not None,
+            "`exog` is None in `fit()` method call, but `lags_exog` is not None in constructor. "
+        )
+        self.target_column = series.columns()[0]
```
hrzn, line 131: (for later): could we extend this to work on all columns?
One idea could be, for the models that simply predict a scalar, to copy the model N times, and fit the N models to the N columns. Some models by be able to do vector regression, but not sure if there's a way to detect this somehow.

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
+            training_data = pd.concat([training_data, lagged_data], axis=1)
 
-    def _historical_forecasts_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
-        """Sanity checks for the historical_forecasts function
+        if self.lags_exog is not None:
+            for col in exog.columns():
+                lagged_data = self._create_lagged_data(
+                    series=exog[col], lags=self.lags_exog
+                )
+                training_data = pd.concat([training_data, lagged_data], axis=1)
```
hrzn, line 186: Not sure if I'm right, but it could be that calling `pd.concat()` repeatedly is less efficient than just storing the dataframes in a list and calling `pd.concat()` once at the end on the whole list.

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
+            if str(self.target_column) in self.exog_columns:
+                series = TimeSeries(series.pd_dataframe().rename({self.target_column: "Target"}))
+                self.target_column = series.columns()[0]
+        self.training_data = self._create_training_data(series, exog)
+        self.train_x =  self.training_data.pd_dataframe().drop(self.target_column, axis=1)
+        self.train_y =  self.training_data.pd_dataframe()[self.target_column]
+        self.nr_exog = exog.width*len(self.lags_exog) if exog is not None else 0
+
+        self.model.fit(
+            X=self.train_x,
```
hrzn, line 143: This assumes that the arguments are name `X` and `y`. Maybe that's fine (for sure it'll work for sklearn), but perhaps we could even make it more "agnostic" and not used named parameters here?

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
+        lagged_series = series.pd_dataframe(copy=True)
+        target_name = lagged_series.columns[0]
+        for lag in lags:
+            new_column_name = target_name + "_lag{}".format(lag)
+            lagged_series[new_column_name] = lagged_series[target_name].shift(lag)
+        lagged_series.dropna(inplace=True)
+        lagged_series.drop(target_name, axis=1, inplace=True)
+        return lagged_series
+
+    def predict(self, n: int, exog: Optional[TimeSeries] = None, **kwargs):
```
hrzn, line 219: Maybe write a docstring for this one?

darts/models/regression_model.py
hrzn: In addition it might be nice to explain how the target and exog time axes relate to each other. I.e. that for fitting they need to have matching time axes, but for predicting `exog` has to start 1 timestamp after the end of the target.

darts/timeseries.py
```diff
@@ -977,7 +980,7 @@ def append_values(self,
             raise_if_not(index[-1] == index[0] + self.freq(),
                          'Appended index must have the same frequency as the current one.', logger)
         values = values[new_indices]
-        new_series = pd.DataFrame(values, index=index)
+        new_series = pd.DataFrame(values, columns=self.pd_dataframe().columns, index=index)
```
hrzn, line 983: good catch.

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
+                    append_row = [[forecast[0], *exog.iloc[i, :].values.tolist()]]
+                    prediction_data = prediction_data.append_values(append_row)
+                else:
+                    prediction_data = prediction_data.append_values(forecast)
+                prediction_data = prediction_data.append_values(dummy_row)[1:]
+            forecasts[i] = forecast[0]
+        return self._build_forecast_series(forecasts)
+
+    def __str__(self):
+        return model.__str__()
```
hrzn, line 271:
```suggestion
        return self.model.__str__()
```

darts/models/ensemble_model.py
```diff
@@ -45,14 +45,14 @@ def fit(self, series: TimeSeries) -> None:
     def predict(self, n: int) -> TimeSeries:
         super().predict(n)
 
-        predictions = []
-        for model in self.models:
-            predictions.append(model.predict(n))
+        predictions = self.models[0].predict(n)
+        for model in self.models[1:]:
```
hrzn, line 49: only do this if there are >1 models otherwise it might fail.

darts/models/forecasting_model.py
```diff
@@ -251,13 +250,26 @@ def historical_forecasts(self,
...
+                        forecast = self.predict(n=forecast_horizon, series=train, covariates=train_cov)
+                    else:
+                        forecast = self.predict(n=forecast_horizon, covariates=train_cov)
+                else:
+                    if 'series' in predict_signature.parameters:
+                        forecast = self.predict(n=forecast_horizon, series=train, exog=train_cov)
+                    else:
+                        start = train.end_time() + train.freq()
+                        train_cov = covariates[start:start+(forecast_horizon-1)*train.freq()]
+                        forecast = self.predict(n=forecast_horizon, exog=train_cov)
```
hrzn, line 272: I guess this could fail if the method is called with `covariates` argument, but the model supports `exog`?

darts/models/forecasting_model.py
```diff
@@ -360,9 +372,13 @@ def backtest(self,
                                               verbose)
 
         if last_points_only:
-            return metric(series, forecasts)
+            return metric(series[start:][forecast_horizon-1:], forecasts)
```
hrzn, line 375: why this change?

darts/models/standard_regression_model.py
```diff
@@ -18,65 +16,31 @@
 class StandardRegressionModel(RegressionModel):
```
hrzn, line 16: Could we rename `LinearRegressionModel` maybe?

hrzn (commented):
Looks really good.
Maybe one thought: it would be nice to somehow document `exog` vs `covariates` somewhere. Also it would be nice if before merging you could update the `CHANGELOG.md`.

hrzn:
An additional comment: would it be possible to make a `LinearRegression` forecasting model, like you did with Random Forest but using linear regression? IMO linear regression will also often be a go-to method so it might be nice to have it ready.

I'm also thinking it might be time to reorganise a bit `darts.models`, and maybe introduce submodules like `darts.models.regression`, `darts.models.ensemble` and `darts.models.torch` because it starts to be a little messy. WDYT?

darts/models/random_forest.py
```diff
@@ -0,0 +1,66 @@
...
+
+logger = get_logger(__name__)
+
+
+class RandomForest(RegressionModel):
+    def __init__(self,
+                 lags: Union[int, list] = None,
+                 lags_exog: Union[int, list, bool] = None,
+                 n_estimators: Optional[int] = 100,
+                 max_depth: Optional[int] = None,
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/regression_model.py
```diff
@@ -13,300 +13,259 @@
...
     _build_tqdm_iterator,
     _with_sanity_checks,
     _historical_forecasts_general_checks
 )
-from .. import metrics
 
 logger = get_logger(__name__)
 
 
 # TODO: Extend this to a "DynamicRegressiveModel" class, which acts on List[List[TimeSeries]].
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/regression_model.py
tneuer: I changed the phrasing a little. Do you think it would help to give an example in the documentation of `RegressionModel`.

darts/models/regression_model.py
tneuer: Done

darts/models/regression_model.py
tneuer: I explained how the input of `exog` should look like in the documentation of `predict()`. It might be quite strict as it is now but hopefully it will be alright for most users. Unfortunately now it is not possible for a user to make predictions in the time frame during training which might be useful for some users to get training errors. Do you think this needs to be fixed?

darts/models/regression_model.py
tneuer: Hm you mean regression on multiple different targets, not repetitions of the same time series, right?
I don't know which (scikit-learn) models support multiple regression but I think for now it's easier to leave the responsibility for that on the user-side. Otherwise we need to infer quite a lot from the input shape. Probably that's more complicated to deal with from our side.

darts/models/regression_model.py
tneuer: You're right. Removed the named attribute.

darts/models/regression_model.py
tneuer: Done

darts/models/regression_model.py
tneuer: Done

darts/models/standard_regression_model.py
tneuer: Yep I thought about it and that's probably better. Just wasn't sure to drop support for a class called `StandardRegressionModel` completely.
I just the file and class name.

tneuer: No I have 20 more changed file but that's due to the renaming from StandardRegressionModel to LinearRegressionModel.

darts/models/regression_model.py
tneuer: Wow, yes you're right here. Just did some simple tests and it really makes a large difference even for smaller datasets. Changed it and I'll keep that in mind.

darts/models/ensemble_model.py
tneuer: Done

tneuer:
The linear regression model is implemented in the `LinearRegressionModel` now (`StandardRegressionModel` before). As soon as this branch is pulled in I'll implement [HistGradientRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html)(scikit's implementation of LightGBM) and [RidgeRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html). This should be possible in under an hour now that the `RegressionModel` is in place.

Also I think it would be a great step to make these different submodules. Probably then it wouldn't be too bad to have slightly different interfaces for these different models (like `torch` model covariates) because the user really clearly knows where he imports a model from. Then we just keep the API strictly consistent within these submodules and at least similar between submodules. In the future we can then try to unify them some more again, but this would be a really easy first step that helps with the conceptual differences.

darts/models/forecasting_model.py
tneuer: I refactored the code here a bit. I'm not sure if it's better at all. It's a little inconvenient that currently the RegressionModel is as far as I know the only model that checks if the exogenous variables are starting at time `last_train_timestamp+1`. That's why it needs to be treated separately.
It needs to check that because it reuses the last few values from training for prediction. For example, if you currently enter 12 values for the exog variable you get 12 values back. Maybe it would be better that if you use a lag of 12 and pass 12 values you only get 1 value back. The data is used to construct the lagged dataframe which in turn is used to make one prediction. 
I'll take a look how this is handled elsewhere to be consistent here.

darts/models/forecasting_model.py
tneuer: I think I changed this at some point because for me some tests were failing. I'm pretty sure now this was a fault of my new implementation.

I could change it back because it does not change the result, but now the first argument `series[start:][forecast_horizon-1:]` has the same length and time index as the second `forecasts` argument. Before I changed it the `series` had another time index than the forecasts and they were intersected in the `metric` itself. Let me know if I should change it back.

darts/models/regression_model.py
hrzn: I think we can merge without having this in place, but it would be nice to keep track of it in our roadmap.

darts/models/regression_model.py
hrzn: OK fair enough let's leave it like that for now.

darts/models/forecasting_model.py
hrzn: I think you could change it back. If we are confident the intersection is as expected, then I would find it clearer to rely on `metric` to do the intersection.

darts/models/standard_regression_model.py
hrzn: OK, I think it's OK to change the name :)

CHANGELOG.md
```diff
@@ -1,10 +1,23 @@
...
 
 [Full Changelog](https://github.com/unit8co/darts/compare/0.7.0...develop)
+### For users of the library:
+
+**Added:**
+- `RandomForest` algorithm implemented. Uses the scikit-learn `RandomForestRegressor` to predict future values from exogenous
+variables and lagged values of the target.
+
+**Changed**
+- `RegressionModel` is now a user exposed class. It acts as a wrapper around any regression model with a `fit()` and `predict`
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/forecasting_model.py
```diff
@@ -360,9 +374,13 @@ def backtest(self,
...
 
-        errors = [metric(series, forecast) for forecast in forecasts]
+        for forecast in forecasts:
+            start = forecast.start_time()
+            freq = forecast.freq()
+            series_snippet = series[start:start+(forecast_horizon-1)*freq]
```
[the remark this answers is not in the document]
tneuer: Yep I changed everything back to its old version.

darts/models/linear_regression_model.py
```diff
@@ -0,0 +1,49 @@
...
+            at the current time `t`. This might lead to leakage if for predictions the values of the exogenous
+            variables at time `t` are not known.
+        **kwargs
+            Additional keyword arguments passed to `sklearn.linear_model.LinearRegression`.
+        """
+        self.kwargs = kwargs
+        super().__init__(
+            lags=lags,
+            lags_exog=lags_exog,
+            model=LinearRegression(n_jobs=-1, fit_intercept=False, **kwargs)
```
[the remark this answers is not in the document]
tneuer: I copy+pasted the old definition, but you're right here. I removed the explicit choice of both arguments. Also `njobs` is not really that useful for a normal user because it only provides speed-up for large multi-target problems.

darts/models/linear_regression_model.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+    def __str__(self):
+        return 'LinearRegression(lags={}, lags_exog={})'.format(
+            self.lags, self.lags_exog
+        )
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/random_forest.py
```diff
@@ -0,0 +1,69 @@
...
+from sklearn.ensemble import RandomForestRegressor
+
+logger = get_logger(__name__)
+
+
+class RandomForest(RegressionModel):
+    def __init__(self,
+                 lags: Union[int, list] = None,
+                 lags_exog: Union[int, list, bool] = None,
+                 n_estimators: Optional[int] = 100,
```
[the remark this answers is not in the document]
tneuer: Hm, probably the benefit of leaving them explicitly in the signature is that users who are not really familiar with the scikit-learn implementation don't have to look up a third party documentation, but can only look at ours (also those arguments are shown when using tab-completion in most editors). In my opinion, n_estimators and max_depth are two of the most important hyper-parameters for `RandomForest`, that's why I used those two.

What do you think? Remove them and save the documentation or leave it as is?

darts/models/regression_ensemble_model.py
```diff
@@ -17,7 +20,7 @@ class RegressionEnsembleModel(EnsembleModel):
     def __init__(self,
                  forecasting_models: List[ForecastingModel],
                  regression_train_n_points: int,
-                 regression_model=LinearRegression(n_jobs=-1, fit_intercept=False)):
+                 regression_model=None):
         """
         Class for ensemble models using a regression model for ensembling individual models' predictions
```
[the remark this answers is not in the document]
tneuer: Yeah that's great. I also was a little confused in the beginning when rewriting parts of this class.

darts/models/random_forest.py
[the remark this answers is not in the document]
hrzn: Sounds good, let's leave them 👍

hrzn:
You have committed an empty `.Rhistory` file, could you remove?

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+            prediction_data = prediction_data.append_values(dummy_row)
+
+        if exog is not None:
+            required_start = self.training_series.end_time()+self.training_series.freq()
+            raise_if_not(exog.start_time() == required_start,
+                "`exog` first date must be equal to self.training_series.end_time()+1*freq. " +
+                "Given: {}. Needed: {}.".format(exog.start_time(), required_start)
+            )
+        if isinstance(exog, TimeSeries):
+            exog = exog.pd_dataframe()
```
hrzn, line 260: maybe you could use `copy=False` here, as it doesn't seem you are touching `exog` in this code, and it might make it (slightly) faster to avoid copying.

darts/tests/test_local_forecasting_models.py
```diff
@@ -46,7 +51,7 @@
 
 try:
     from ..models import AutoARIMA
-    models.append((AutoARIMA(), 13.7))
+    models.append((AutoARIMA(), 30.5))
```
hrzn, line 54: What happened to AutoARIMA? 😄

darts/tests/test_regression_models.py
```diff
@@ -1,100 +1,193 @@
...
-
-    for model in models:
-        model.fit(train_f, train_t)
-        prediction = model.predict(test_f)
-        current_r2 = r2_score(prediction, test_t)
-        test_case.assertTrue(current_r2 >= min_r2, "{} model was not able to denoise data."
-                             "A r2 score of {} was recorded.".format(str(model), current_r2))
-
-
 class RegressionModelsTestCase(DartsBaseTestClass):
```
hrzn, line 36: Could you perhaps add one or two test cases with some series containing NaNs in order to test this potential issue here?

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
 import numpy as np
 import pandas as pd
 
-from abc import ABC, abstractmethod
-from typing import List, Iterable, Union, Any, Callable
 
+from .. import metrics
```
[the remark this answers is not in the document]
tneuer: Removed

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+from inspect import signature
 from ..timeseries import TimeSeries
-from ..logging import raise_if_not, get_logger, raise_log
+from sklearn.linear_model import LinearRegression
+from .forecasting_model import ExtendedForecastingModel
+from typing import List, Iterable, Union, Any, Callable, Optional
```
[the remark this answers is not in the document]
tneuer: I think now everything is used. I'll try if I can make sublime check if imports are used.

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+from ..logging import raise_if, raise_if_not, get_logger, raise_log
 from ..utils import (
```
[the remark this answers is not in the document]
tneuer: See previous

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+                 lags: Union[int, list] = None,
+                 lags_exog: Union[int, list, bool] = None,
+                 model=None):
+        """ Regression Model
 
-        # Stores training date information:
-        self.train_features: List[TimeSeries] = None
-        self.train_target: TimeSeries = None
+        Can be used to fit any scikit-learn-like regressor class to predict the target
+        time series with lagged values.
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+                    "Every element of `lags` must be a strictly positive integer. Given: {}.".format(self.lags))
+                )
+
+        self.lags_exog = lags_exog
+        if self.lags_exog is True:
+            self.lags_exog = self.lags[:]
+        elif self.lags_exog is False:
+            self.lags_exog = [0]
+        elif isinstance(self.lags_exog, int):
+            raise_if_not(self.lags_exog > 0, "`lags_exog` must be strictly positive. Given: {}.".format(self.lags_exog))
```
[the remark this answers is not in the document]
tneuer: We still need to handle the `lags_exog=False` case, so I changed it to `elif self.lags_exog is False or self.lags_exog==0`.
Maybe it would also be a good idea to **not** support `type(lags_exog)==bool`. It is somewhat confusing probably and if a user wants the same lag for exog and endog he can very easily do something like 
```python
lags = [1, 2, 3, 4]
model = RegressionModel(lags=lags, lags_exog=lags)
```

Also should we maybe rename the `lags` argument to `lags_endog` or `lags_series`? Let me know what you think.

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+        series : TimeSeries
+            Target series.
+        exog : TimeSeries, optional
+            Exogenous variables.
 
         Returns
         -------
         TimeSeries
-            A series containing the predicted targets, of the same length as the features series
+            Data frame with lagged values.
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+        self.target_column = series.columns()[0]
+
+        # Prepare data
+        training_x = self._create_training_data(series=series, exog=exog)
+        training_y = series[training_x.time_index()]
+
+        # Fit model
+        if "series" in signature(self.model.fit).parameters:
+            self.model.fit(
+                training_y,
```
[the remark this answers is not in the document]
tneuer: Done

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
+            for i, col in enumerate(exog.columns()):
+                lagged_data = self._create_lagged_data(
+                    series=exog[col], lags=self.lags_exog
+                )
+                training_data_list.append(lagged_data)
 
-    def _historical_forecasts_sanity_checks(self, *args: Any, **kwargs: Any) -> None:
-        """Sanity checks for the historical_forecasts function
+        training_data = pd.concat(training_data_list, axis=1)
+        return TimeSeries(training_data.dropna(), freq=series.freq())
```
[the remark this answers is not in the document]
tneuer: Hm I could work around this by dropping the first `self.max_lags` values. Then probably most of the models down the road will complain. But this is probably better than silently removing `na`.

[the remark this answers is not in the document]
tneuer: I guess I don't even really need it at this place anymore because I'm calling the method `_create_lagged_data` from here. I'll handle it in there and just drop the first few rows.

darts/models/regression_model.py
```diff
@@ -13,300 +13,289 @@
...
-        If `last_points_only` is set to True, it will use only the last point of each historical forecast.
-
-        This always re-trains the models on the entire available history,
-        corresponding an expanding window strategy.
+        lagged_series = series.pd_dataframe(copy=True)
+        target_name = lagged_series.columns[0]
+        for lag in lags:
+            new_column_name = target_name + "_lag{}".format(lag)
+            lagged_series[new_column_name] = lagged_series[target_name].shift(lag)
+        lagged_series.dropna(inplace=True)
```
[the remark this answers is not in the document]
tneuer: See answer to previous.

darts/models/regression_model.py
tneuer: Done

darts/models/regression_ensemble_model.py
```diff
@@ -34,45 +40,42 @@ def __init__(self,
...
-                     "Provided StandardRegressionModel.train_n_points parameter doesn't match specified"
-                     " regression_train_n_points parameter.",
-                     logger)
-
-            # if it was None, set regression_model.train_n_points to regression_train_n_points
-            regression_model.train_n_points = regression_train_n_points
-        else:
-            regression_model = StandardRegressionModel(regression_train_n_points, regression_model)
+        if not isinstance(regression_model, (LinearRegressionModel, RandomForest)):
+            regression_model = RegressionModel(lags_exog=[0], model=regression_model)
```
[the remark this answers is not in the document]
tneuer: Hm, I like when I know of what type an object is when handling it. Even if we only use the `fit()` and `predict()` methods we from now on know that it's an RegressionModel and what we can do with it.
But yes I don't think it would make any difference. I'll add some more tests here just to make sure everything works.

darts/tests/test_local_forecasting_models.py
tneuer: Didn't pass the test at the time otherwise, but I'll check again what happened here. Maybe due to the version upgrade? Maybe they changed the way they perform their grid search and don't find the optimal solution anymore?

darts/models/regression_model.py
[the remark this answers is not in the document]
tneuer: So I changed the above two lines in question to:
```python
training_data = pd.concat(training_data_list, axis=1)[self.max_lag:]
return TimeSeries(training_data, freq=series.freq())
```
Now the first few rows are dropped because there aren't enough lags available but I keep all other nans. I also introduced a new test for this.

darts/models/regression_ensemble_model.py
```diff
@@ -34,45 +40,42 @@ def __init__(self,
...
         # fit the forecasting models
         for model in self.models:
             model.fit(forecast_training)
 
         # predict train_n_points points for each model
-        predictions = []
-        for model in self.models:
-            predictions.append(model.predict(self.regression_model.train_n_points))
+        predictions = self.models[0].predict(self.train_n_points)
+        for model in self.models[1:]:
```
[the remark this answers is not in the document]
tneuer: Wrote a test for it and it does not crash. For a list of length one the query `self.models[1:]` returns an empty list which will then be ignored in the for loop. I can change it to make it clearer but the code also works as is :)

darts/models/regression_ensemble_model.py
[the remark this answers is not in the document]
tneuer: I changed this whole section now to:
```python
if regression_model is None:
    regression_model = LinearRegressionModel(lags_exog=0, fit_intercept=False)

regression_model = RegressionModel(lags_exog=0, model=regression_model)
```
So we only rely on darts models in the ensemble regression (Of course one step down the line they depend on the corresponding scikit-learn models).

darts/tests/test_local_forecasting_models.py
tneuer: Okey I had an old version of pmdarima (1.6.0). With the current version in the requirements (1.8.0) I get a MAPE of 15.034 so I set the threshold to 15.5. 
So it is much better again but still not at the previous value of 13.7.

darts/tests/test_regression_models.py
tneuer: Do you mean the issue discussed before when dropping rows? For that I now included tests so that only the first rows are dropped for which not enough lags are available. 
Or do you mean some performance tests when using nans?

darts/models/regression_ensemble_model.py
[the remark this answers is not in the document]
tneuer: On a side-note. The lags_exog argument of the `LinearRegressionModel` could be anything without making a difference as the `RegressionModel` has its own lags_exog parameter which will be used. So even though it wouldn't make much sense, the following code runs without problem:

```python
if regression_model is None:
    regression_model = LinearRegressionModel(lags_exog=12, fit_intercept=False)

regression_model = RegressionModel(lags_exog=0, model=regression_model)
```
I will mention this in the notebook, but do you think it's okey like that or should we check for this and raise an error?

darts/models/regression_model.py
```diff
@@ -13,300 +13,286 @@
...
+            the last `lags` lags are used (inclusive). Otherwise a list of integers with lags is required.
+        lags_exog : Union[int, list, bool]
+            Number of lagged exogenous values used to predict the next time step. If an integer is given
+            the last `lags_exog` lags are used (inclusive). Otherwise a list of integers with lags is required.
+            If True `lags` will be used to determine lags_exog. If False, the values of all exogenous variables
+            at the current time `t`. This might lead to leakage if for predictions the values of the exogenous
+            variables at time `t` are not known.
+        model
+            A regression model that implements `fit()` and `predict()` methods.
+            Default: `sklearn.linear_model.LinearRegression(n_jobs=-1, fit_intercept=False)`
```
hrzn, line 54:
```suggestion
            Typically a scikit-learn model or a Darts `ExtendedForecastingModel` (models accepting `exog`)
            can be used here.
            Default: `sklearn.linear_model.LinearRegression(n_jobs=-1, fit_intercept=False)`
```

darts/models/regression_model.py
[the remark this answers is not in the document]
hrzn: I agree with your point about removing the `bool` - probably clearer to just specify the lags explicitly. 
I think `lags` is already not a bad name, I would say no need to rename.

darts/tests/test_regression_models.py
hrzn: I meant the issue discussed before when dropping rows, so I guess we are good.

Issue #293, opened 2021-03-12
Adding LightGBM


whiletruelearn:
**Is your feature request related to a current problem? Please describe.**
LightGBM won the M5 time series forecasting competition. It will be a good algorithm to add.
https://www.innovating-automation.blog/m5-forecasting-competition/

**Describe proposed solution**
Create a wrapper for the lightgbm library in `darts`

tneuer:
There are a lot of great algorithms out there we're still planning to implement in future releases, but you're completely right. LightGBM should definitely be high up on our priority list. Thanks for bringing this to our attention!

RNogales94:
I have code this feature, but when I try to use lightGBM for a simple example (Airline passenger forecast) the LightGBM predict a constant, because the model becomes too simple. 

I can share the code with a pull request but I think there's something wrong. I can show you what I have done by google meet or something before the PR if you wish

tneuer:
Hey @RNogales94, thanks for bringing up the issue again. We have also been working on this problem and a few general notes to update all interested folk:

- You are right that Tree-Based-Models tend to predict the mean (that's the operation performed on the leaf-level) and sometimes have troubles with time series, **especially** when you deal with a time series with non-zero trend (like the passenger data set). The user should perform differencing before passing any TimeSeries to a Tree-Based-Model (this feature has just be enabled in Darts by calling `.diff()` on any `TimeSeries`.)
- For now, we decided not to implement the LightGBM class from the `lightgbm` library, because it requires additional dependencies outside of python which cannot be downloaded via conda. On a Mac you need to perform [these](https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html#macos) steps to make `lightgbm` work and we already have so many Python dependencies that we decided against having even more out-of-Python dependencies which would break the `Darts` installation.
- In the near future we release models wrapping around Random Forest and  HistGradientBoostingRegressor from scikit-learn (it is currently in the PR and we need to review it properly).
- In the same release we enable the user to use Darts for every model exposing the .fit() and .predict() methods (so every scikit-learn model). You then could download LightGBM separately and then pass that model directly to the base class called (for now) `RegressionModel` to make everything work.

@RNogales94 I would still be eager to see how you implemented the model. Maybe we see something that would improve on our implementation. As it is still in the PR we could change and improve it without causing any problems. 
Would you mind making a PR for your implementation. It will probably not be accepted due to the mentioned concerns with `lightgbm` but I think that would still be the easiest way to review the code together.

[...]

PR #317, opened 2021-04-20
Adding LightGBM support

RNogales94:
Fixes #.

### Summary

Adding LightGBM as a separate module, following what you have done with prophet

### Other Information

Some parenthesis have been removed in order to follow Python standards.


tneuer:
Hello @RNogales94, I finally had some time to review your PR. It looks great and would have probably fit well into the context of our library.
Unfortunately as mentioned in [this issue](https://github.com/unit8co/darts/issues/293) we decided against having native LightGBM support for now due to the dependencies outside of Python. We will add support for scikit-learn's [HistGradientBoostingRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html) and [RandomForestRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestRegressor.html). We will also support basically all models with a `fit` and `predict` API in the `RegressionModel`. So I think a user could massage the API of LightGBM a little to make it work.
Again thanks of the effort, we hope to see another pull request at some point!

P.S.: We will also look at the parenthesis that you fixed. Probably your version is really closer to Python standards :)

PR #318, opened 2021-04-22
Feature/multi ts transformers

piaz97:
### Summary
Sequences of time series are supported as input to transformers and pipelines. Transformers and Pipeline support the `verbose` parameter for monitoring the operations and the `n_jobs` parameter for processing more samples in parallel.

### Details about the changes

- [transformers and pipeline] Added support to multiple time series as input;
- [transformers and pipeline] Added `verbose` and `n_jobs` parameters;
- [BoxCox Transformer] Moved `lambda` parameter from `fit` to constructor, so that it is now possible to use it in a Pipeline;
- [Pipeline] Added `partial` parameter to Pipeline's `inverse_transform`. If set to True, the inverse transformation will be applied even if not all the transformers are invertible, inverting only the invertible ones (e.g., inverting a Pipeline with a MissingValueFiller).
- [Scaler] Changed the default value of the `scaler` parameter in the constructor, now set to `None`, and properly initialized afterward.
- [Test] Added some test for covering the new functionalities, modified the tests related to BoxCox.
- [Sample Notebook] Updated the notebook _03-data-processing_, adding the new use cases.

**Note**: the changes regarding the BoxCox transformer are **not** backward compatible


darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -21,8 +24,38 @@ def __init__(self,
...
+        self._verbose = value
+
+    def set_n_jobs(self, value: int):
+        """
+        Sets the number of cores that will be used by the transformer while processing multiple time series. Set to `-1` 
+        for using all the available cores.
+        """
+
+        raise_if_not(isinstance(value, int), "n_jobs must be an integer")
+        self._n_jobs = value
```
hrzn, line 59: In the transformers you often have a pattern like that:

```
def transform():
        ...

        if isinstance(data, TimeSeries):
            return some_fn(data)
        else:
            def map_ts(ts):
                return some_fn(ts)

            iterator = _build_tqdm_iterator(data)

            transformed_data = Parallel(n_jobs=self._n_jobs)(delayed(map_ts)(ts) for ts in iterator)
```

This leads to quite a bit of code duplication, because all transformers now have to implement the (same) parallelism code. How about having the `Parallel` business done only in the super class(es)? Instead of `transform()` and `inverse_transform()` being abstract, you could have them call an abstract method of the form
```
@abstract
def ts_transformation(ts: TimeSeries) -> TimeSeries
    pass
    
@abstract
def ts_inverse_transformation(ts: TimeSeries) -> TimeSeries
    pass
```
They would do the same thing as the "old" transform & inverse_transform methods, but be called by the new transform & inverse_transform methods, which would no longer be abstract and provide parallelism.

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -61,35 +75,92 @@ def fit(self,
...
+        if isinstance(data, TimeSeries):
+            return _transform_single_series(data, self._lmbda)
+        else:
+            # multiple time series
+            iterator = _build_tqdm_iterator(zip(data, self._lmbda),
+                                            verbose=self._verbose,
+                                            desc="Transforming BoxCox",
+                                            total=len(data))
+
+            transformed_data = Parallel(n_jobs=self._n_jobs)(delayed(_transform_single_series)(ts, lmbda)
```
hrzn, line 138: Are you sure `transformed_data` is always of type `Sequence` ? If yes can it be any type of `Sequence` ?

hrzn (commented):
This is a really nice improvement.
Perhaps the only main thing I'm suggesting would be to slightly refactor so that only parent classes implement the parallelism.

darts/dataprocessing/transformers/boxcox.py
piaz97: `transformed_data` is always a `Sequence`, specifically, a `List`. I modified the types accordingly.

darts/dataprocessing/pipeline.py
```diff
@@ -14,8 +14,10 @@
 
 class Pipeline:
     def __init__(self,
-                 transformers: Sequence[BaseDataTransformer[TimeSeries]],
-                 copy: bool = False):
+                 transformers: Sequence[BaseDataTransformer],
+                 copy: bool = False,
+                 verbose: bool = None,
```
ymaktepi, line 19: This should probably default to `False`. Next line should probably default to the default of `transformer.n_jobs`.

ymaktepi: Mb, I read the comment and it makes sense to have a `None` default.

piaz97:
I have refactored the classes as suggested. Now the `transform()`, `inverse_transformed()` and `fit()` methods don't need to handle the parallelization code in the derived classes. To allow this, the developer must define, for each of the method `transform()`, `inverse_transformed()` and `fit()`:

- A static method `ts_<method>` defining the function to apply to a single TS. This must be a static method for avoiding parallelization issues with **joblib** (encountered with instance methods).
- (Just if necessary) A `_<method>_iterator` function creating the iterator yielding tuples of input to the `ts_<method>`. E.g., with the `Scaler` the `ts_transform()` method required an additional `transformer` argument. The developer will have to handle the logic behind these 'additional args'.

Furthermore, the base classes are no longer typed as generics, instead are now designed to handle `TimeSeries` and `Sequence[TimeSeries]`.

darts/tests/dataprocessing/transformers/test_mappers.py
```diff
@@ -32,21 +32,44 @@ def inverse_ts_func(ts, x):
...
 
     def test_mapper(self):
         transformed = self.plus_ten.transform(self.zeroes)
         self.assertEqual(transformed, self.tens)
 
+        # multiple time series
+        series_array = [self.zeroes, self.tens]
+        transformed = self.plus_ten.transform(series_array)
+        self.assertEqual(transformed, [self.tens, self.twenties])
+
```
ymaktepi, line 45: maybe it's too late on a friday but would that make sense?
```suggestion
        inputs = [
            self.zeroes,
            [self.zeroes, self.tens]
        ]
        
        expected_outputs = [
            self.tens,
            [self.tens, self.twenties]
        ]
   
        for to_transform, expected_output in zip(transformed_inputs, expected_outputs)
            transformed = self.plus_ten.transform(to_transform)
            self.assertEqual(transformed, expected_output)
```

you could also consider having a list of tuples (input, expected) instead of a `zip`, which would be a bit more readable. same comment applies for the other functions in that file.

darts/tests/dataprocessing/transformers/test_boxcox.py
```diff
@@ -45,3 +51,40 @@ def test_boxcox_inverse(self):
...
+        # lambda = None
+        box_cox = BoxCox()
+        transformed = box_cox.fit_transform([self.multi_series, self.multi_series])
+        back = box_cox.inverse_transform(transformed)
+        pd.testing.assert_frame_equal(self.multi_series._df, back[0]._df, check_exact=False)
+        pd.testing.assert_frame_equal(self.multi_series._df, back[1]._df, check_exact=False)
+
+    def test_boxcox_multiple_calls_to_fit(self):
+        """
+        This test checks whether calling the scaler twice is calculating new lambdas insted of
```
ymaktepi, line 79:
```suggestion
        This test checks whether calling the scaler twice is calculating new lambdas instead of
```

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -2,49 +2,162 @@
...
         """
-        Transform the data.
-        Not implemented in base class and has to be implemented by deriving classes.
+        The function that will be applied to each `TimeSeries` object once the `transform()` function is called. The
+        function must take as first argument a `TimeSeries` object, and return the transformed `TimeSeries` object.
+        If more parameters are added as input in the derived classes, the `_transform_iterator()` should be
+        redefined accordingly, to yield the necessary arguments to this function (See `_transform_iterator()` for
+        further details).
+
+        This method is not implemented in the base class and must be implemented in the deriving classes.
```
ymaktepi, line 78: a small explanation of why this method is static is probably needed (if not for users, at least for the library's developers)

darts/dataprocessing/transformers/fittable_data_transformer.py
```diff
@@ -2,88 +2,177 @@
...
+            Sequence of `TimeSeries` received in input.
 
         Returns
         -------
-        BaseDataTransformer[T]
-            Fitted data transformer (typically would be self)
+        Iterator[Tuple[TimeSeries]]
+            An iterator containing tuples of inputs for the `ts_fit()` method.
+
+        Examples
```
hrzn, line 91: 😍

darts/dataprocessing/transformers/scaler.py
```diff
@@ -26,73 +31,60 @@ def __init__(self, scaler=MinMaxScaler(feature_range=(0, 1)), name="Scaler"):
...
+
+    @staticmethod
+    def ts_fit(series: TimeSeries, transformer, *args, **kwargs) -> Any:
+        # fit_parameter will receive the transformer object instance
+        scaler = transformer.fit(series.values().reshape((-1, series.width)))
+        return scaler
+
+    def _fit_iterator(self, series: Sequence[TimeSeries]) -> Iterator[Tuple[TimeSeries, Any]]:
+        # generator which returns deep copies of the 'scaler' argument
+        new_scaler_gen = (deepcopy(self.transformer) for _ in range(len(series)))
```
hrzn, line 80: good one

darts/dataprocessing/transformers/scaler.py
```diff
@@ -26,73 +31,60 @@ def __init__(self, scaler=MinMaxScaler(feature_range=(0, 1)), name="Scaler"):
             will scale all the values of a time series between 0 and 1.
```
hrzn, line 31: Perhaps you could add a sentence here saying that if this transformer is applied on multiple series, then this scaler will be deep-copied for each series.

hrzn (approved):
Really great job @piaz97 
Very clean :)
I have a few very minor comments only. Also, before merging, could you make sure that the notebooks' outputs are fine? It seems you have some import errors in the notebook 03.

PR #319, opened 2021-04-23
fix(ci): merge pipeline now caches gradle download

ymaktepi:
Fixes #.


hrzn (approved):
Nice thanks 🦄 
Shall we also perhaps do it for the other workflows (not only merge)?

PR #323, opened 2021-05-05
WIP Feat/datasets

ymaktepi:
### Summary
Adds basic architecture for datasets lazy loading.

Usage:
```python
from darts.datasets import AirPassengers
ts: TimeSeries = AirPassengers.load()
```

### Other Information
For now, only CSV files are supported (via `DatasetLoaderCSV` instantiation).
The idea would be to add other derivations of the `DatasetLoader` abstract class when we need other file formats.
WIP status because we need to add other datasets and the possibility to add a custom preprocessing function to the `load` call for the user to prepare the data.


darts/datasets/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+"""
+    Overall usage of this package:
+    from darts.datasets import AirPassengers
+    ts: TimeSeries = AirPassengers.load()
+"""
+AirPassengersDataset = DatasetLoaderCSV(
+    metadata=DatasetLoaderMetadata(
+        "air_passengers.csv",
+        uri="https://raw.githubusercontent.com/unit8co/darts/master/examples/AirPassengers.csv",
```
hrzn, line 11: instead of leaving the CSV files in `examples/` could we put them in a (new) `datasets/` folder?

darts/datasets/dataset_loaders.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, metadata: DatasetLoaderMetadata, root_path: Path = None):
+        self._metadata: DatasetLoaderMetadata = metadata
+        if root_path is None:
+            self._root_path: Path = DatasetLoader._DEFAULT_DIRECTORY
+        else:
+            self._root_path: Path = root_path
+
+    def load(self) -> TimeSeries:
+        """
+        Load the dataset in memory, as a TimeSeries.
```
hrzn, line 48: It doesn't have to be supported right now, but it would be nice at some point to also support datasets that consist in a `Sequence[TimeSeries]` (e.g., M4 competition dataset).

darts/datasets/dataset_loaders.py
```diff
@@ -0,0 +1,147 @@
...
+        Returns
+        -------
+        """
+        if not self._is_already_downloaded():
+            raise DatasetLoadingException(f"Checking md5 checksum of a absent file: {self._get_path_dataset()}")
+
+        with open(self._get_path_dataset(), "rb") as f:
+            md5_hash = hashlib.md5(f.read()).hexdigest()
+            if md5_hash != self._metadata.hash:
+                raise DatasetLoadingException(f"Expected hash for {self._get_path_dataset()}: {self._metadata.hash}"
```
hrzn, line 84: cool :)

hrzn (approved):
Super nice!
It looks like a very good start already. It would be nice to refactor our examples (notebooks + README) to make them use when possible (but this can come in a future PR)

darts/datasets/__init__.py
ymaktepi: the goal is to move them (new branch, new repo, cloud hosting, cdn, whatever) indeed. for now we're using github for PoC.

README.md
```diff
@@ -43,15 +43,13 @@ Create a `TimeSeries` object from a Pandas DataFrame, and split it in train/vali
 
 ```python
 import pandas as pd
-from darts import TimeSeries
+from darts.datasets import AirPassengersDataset
 
-df = pd.read_csv('AirPassengers.csv', delimiter=",")
+df = AirPassengersDataset.load()
 series = TimeSeries.from_dataframe(df, 'Month', '#Passengers')
```
ymaktepi, line 49:
```suggestion
series = AirPassengersDataset.load()
```

examples/01-darts-intro.ipynb
```diff
@@ -37,6 +37,17 @@
...
+      "\u001b[0;31mModuleNotFoundError\u001b[0m: No module named 'darts.datasets'"
+     ]
```
ymaktepi, line 50: Weird

README.md
hrzn: In fact yesterday I started having second thoughts about changing the README to use Datasets. I'm thinking in the Readme it might be better to explicitly show how Darts can be easily used starting from a Dataframe. Otherwise users may think this dataset business is hiding some complexity.

Issue #314, opened 2021-04-19
[BUG] Bug when using multiple Scalers constructed with no arguments.


mkmkmk124:
### Describe the bug
After learning and transforming multiple time series data by using the fit_transform function of multiple Scaler objects constructed with no arguments, if we try to restore each time series by using the inverse_transform function, the original time series will not be restored except for the last learned time series. 

### To Reproduce
```
from darts import TimeSeries
import pandas as pd
import matplotlib.pyplot as plt
from darts.dataprocessing.transformers import Scaler

plt.figure();
timeseries1 = TimeSeries(pd.DataFrame([1,2,3,4,5],index=pd.date_range(start='1/1/2017', periods=5, freq='MS')))
timeseries2 = TimeSeries(pd.DataFrame([50,40,30,20,10],index=pd.date_range(start='1/1/2017', periods=5, freq='MS')))
timeseries1.plot(label="time_series_1")
timeseries2.plot(label="time_series_2")
plt.legend();
plt.title("Fig. 1");

plt.figure();
scaler_1, scaler_2 = Scaler(), Scaler()
timeseries1_scaled = scaler_1.fit_transform(timeseries1)
timeseries2_scaled = scaler_2.fit_transform(timeseries2)
timeseries1_scaled.plot(label="time_series1_scaled")
timeseries2_scaled.plot(label="time_series2_scaled")
plt.legend();
plt.title("Fig. 2");

plt.figure();
timeseries1_inversed = scaler_1.inverse_transform(timeseries1_scaled)
timeseries2_inversed = scaler_2.inverse_transform(timeseries2_scaled)
timeseries1_inversed.plot(label="time_series1_inversed")
timeseries2_inversed.plot(label="time_series2_inversed")

plt.legend();
plt.title("Fig. 3");
```
![fig1](https://user-images.githubusercontent.com/13403703/115214068-202f9b80-a13d-11eb-8b91-4d874313ac1a.png)
![fig2](https://user-images.githubusercontent.com/13403703/115214141-2e7db780-a13d-11eb-9a17-d021a9177e9e.png)
![fig3](https://user-images.githubusercontent.com/13403703/115214144-2f164e00-a13d-11eb-82f2-fe7654954b5f.png)

### Expected behavior
Fig. 1 and Fig. 3 are expected to have the same plot.

### Details to be corrected
Change the default value of the scaler argument in the constructor of the Scaler class to None, and implement the processing for the case of None inside the constructor.

Before: 
```
[...]

PR #325, opened 2021-05-06
[Fix] Default args

piaz97:
### Summary
This little PR, together with #318 and #316, fix the problem of default mutable arguments in functions and methods, which could cause wrong behaviors. Also, bug #314 is now fixed.

co-author: @tneuer


hrzn (approved):
👍

Issue #329, opened 2021-05-14
[BUG] Wrong implementation of Residual block in TCN model


vpozdnyakov:
**Describe the bug**
Wrong implementation of Residual block in TCN model

**To Reproduce**
In the original paper https://arxiv.org/pdf/1803.01271.pdf, there is a sequence 
`Conv -> ReLU -> Dropout -> Conv -> ReLU -> Dropout`

<img src='https://user-images.githubusercontent.com/12996098/118294608-8fea3980-b4e3-11eb-98e2-1fd656a6544e.png' width=400>

Here is your implementation in https://github.com/unit8co/darts/blob/develop/darts/models/tcn_model.py, class `_ResidualBlock`, method `forward`

```python
# first step
left_padding = (self.dilation_base ** self.nr_blocks_below) * (self.kernel_size - 1)
x = F.pad(x, (left_padding, 0))
x = self.dropout_fn(F.relu(self.conv1(x)))

# second step
x = F.pad(x, (left_padding, 0))
if self.nr_blocks_below < self.num_layers - 1:
    x = F.relu(x)
x = self.dropout_fn((self.conv2(x)))
```

That is equivalent to a sequence
`Conv -> ReLU -> Dropout -> ReLU -> Conv -> Dropout`

**Expected behavior**
Residual block is correctly implemented, for example:

```python
# first step
left_padding = (self.dilation_base ** self.nr_blocks_below) * (self.kernel_size - 1)
x = F.pad(x, (left_padding, 0))
x = self.dropout_fn(F.relu(self.conv1(x)))

# second step
x = F.pad(x, (left_padding, 0))
x = self.conv2(x)
if self.nr_blocks_below < self.num_layers - 1:
    x = F.relu(x)
x = self.dropout_fn(x)
```

**System (please complete the following information):**
 - darts version 0.7.0

pennfranc:
You're right, nice catch! Your suggestion should work just fine, I put it in PR #330.

hrzn:
This is fixed in 0.8.0

PR #330, opened 2021-05-14
Fix/TCN Residual Block ReLU Position

pennfranc:
Addresses #329. Since I had to rerun the TCN notebook, I also took the liberty of changing the covariates to more appropriate quantities.


hrzn (approved):
Good one, thanks!

pennfranc:
> Good one, thanks!

No worries, feel free to merge it anytime. But for some reason the 'docs' workflow fails. This also appears to be the case in other open PRs.

PR #338, opened 2021-05-27
Fix model retraining  for TorchForecastingModel

mkos:
Before, it wasn't very easy to train already trained model for more epochs. This fix brings new parameter, `epochs`, to `fit()` and `fit_from_dataset()` methods that allows for further training your model.


hrzn:
Thanks @mkos, that looks good. Could you re-open it into `develop` though? I think that's why tests are not running.

mkos:
🤦🏻

PR #341, opened 2021-06-02
[WIP] Fix/ForecastingModel and subclasses tweaks

pennfranc:
I combined a few small things in this PR. For one, I tried to make the documentation of the `GlobalForecastingModel` class clearer. I also made some tweaks to the fit functions. ForecastingModel instances that are not extended are now only accepting univariate data.

### DIFF
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -593,7 +593,6 @@ class GlobalForecastingModel(ForecastingModel, ABC):
     3. Covariates can be supported (multivariate or univariate).
     4. They can allow for multivariate target series and covariates.
     
-    Note that not all global forecasting models implement points 3 and 4.
     The name "global" stems from the fact that a training set of a forecasting model of this class is not constrained
     to a temporally contiguous, "local", time series.
 
@@ -719,7 +718,7 @@ class ExtendedForecastingModel(ForecastingModel, ABC):
         Parameters
         ----------
         series
-            The model will be trained to forecast this time series. Can be multivariate.
+            The model will be trained to forecast this time series. Can be multivariate if the model supports it.
         exog
             A time series of exogenous variables. This time series will not be forecasted, but can be used by
             some models as an input. Can be multivariate.
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -628,7 +628,7 @@ class GlobalForecastingModel(ForecastingModel, ABC):
         ----------
         series
             One or several target time series. The model will be trained to forecast these time series.
-            Can be multivariate.
+            The series may or may not be multivariate, but if multiple series are provided they must have the same number of components.
         covariates
             One or several covariate time series. These time series will not be forecast, but can be used by
             some models as an input. Can be multivariate.
@@ -760,4 +760,3 @@ class ExtendedForecastingModel(ForecastingModel, ABC):
         if self._expect_exog and len(exog) != n:
             raise_log(ValueError(f'Expecting exogenous variables with the same length as the'
                                  f' forecasting horizon ({n}).'))
-
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -35,7 +35,7 @@ logger = get_logger(__name__)
 
 class ForecastingModel(ABC):
     """ The base class for forecasting models. It defines the *minimal* behavior that all forecasting models have to support.
-        The signatures in this base class are for "local" models handling only one series and no covariates.
+        The signatures in this base class are for "local" models handling only one univariate series and no covariates.
         Sub-classes can handle more complex cases.
     """
     @abstractmethod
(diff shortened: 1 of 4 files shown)


darts/models/forecasting_model.py
```diff
@@ -604,31 +616,32 @@ def fit(self,
...
         """ Fits/trains the model on the provided series
 
-        Defines behavior that should happen when calling the `fit()` method of every forecasting model.
+        Defines behavior that should happen when calling the `fit()` method of every global forecasting model.
 
         Some models support training only on one time series, while others can handle a sequence.
         Similarly, some models can handle covariates.
 
-        Some covariates are know in the future, and others aren't. This is a property of the `TimeSeries`, which
+        Some covariates are known in the future, and others aren't. This is a property of the `TimeSeries`, which
```
hrzn, line 624: Actually we should remove this sentence. When writing it I had in mind to capture some information about "when a covariate is known" in the TimeSeries, but it's not done, and at the moment global models support only "historic" covariates.

darts/models/forecasting_model.py
```diff
@@ -604,31 +616,32 @@ def fit(self,
...
         may or may not be exploited by the models.
 
         Parameters
         ----------
         series
             One or several target time series. The model will be trained to forecast these time series.
+            Can be multivariate.
```
hrzn, line 631:
```suggestion
            The series may or may not be multivariate, but if multiple series are provided they must have the same number of components.
```

darts/models/forecasting_model.py
```diff
@@ -604,31 +616,32 @@ def fit(self,
...
         covariates
             One or several covariate time series. These time series will not be forecast, but can be used by
-            some models as an input.
+            some models as an input. Can be multivariate.
```
hrzn, line 634:
```suggestion
            some models as an input. The covariate(s) may or may not be multivariate, but if multiple covariates are provided they must have the same number of components.
```

darts/models/forecasting_model.py
```diff
@@ -680,6 +695,10 @@ def predict(self,
 class ExtendedForecastingModel(ForecastingModel, ABC):
     """ The base class for "extended" forecasting models, handling optional exogenous variables.
 
+    Extended forecasting models expand upon the functionality of `ForecastingModel` in 2 ways:
+    1. They introduce an optional `exog` time series parameter which can be used as a covariate.
```
hrzn, line 699: The main difference with global models here is that `exog` denotes "future" covariates - i.e., covariates which are known for the next `horizon` time steps at prediction time. Whereas global models handle "historic" covariates (i.e., covariates that have the same time axes as the input series). We kept the name `exog` here (and not `covariates`) to try making it clearer that it's not quite the same thing; and also because statsmodels calls it `exog` (for the future covariates).

darts/models/forecasting_model.py
```diff
@@ -700,19 +719,17 @@ def fit(self,
         Parameters
         ----------
         series
-            A time series. The model will be trained to forecast this time series.
+            The model will be trained to forecast this time series. Can be multivariate.
```
hrzn, line 722: Whether it can be multivariate or not depends on the models implementations. for instance ARIMA doesn't support multivariate (VARIMA does).

darts/models/forecasting_model.py
```diff
@@ -49,13 +49,15 @@ def __init__(self):
...
     def fit(self, series: TimeSeries) -> None:
-        """ Trains the model on the provided series
+        """ Trains the model on the provided univariate series.
 
         Parameters
         ----------
         series
             A target time series. The model will be trained to forecast this time series.
         """
+        if not isinstance(self, ExtendedForecastingModel):
```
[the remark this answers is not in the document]
pennfranc: The reason I'm not is because `GlobalForecastingModel` does not call the constructor of its superclass in this implementation. Do you think that's ok?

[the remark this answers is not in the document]
hrzn: Yes, but perhaps it wouldn't hurt to add `GlobalForecastingModel` to the check here, in case we change it one day and call the init method.

darts/models/forecasting_model.py
pennfranc: Given the upcoming PR introducing future covariates, I think we could leave this, what do you think? Unless there is an imminent release.

darts/models/forecasting_model.py
pennfranc: I see. Do you think we should rethink the naming once we introduce future covariates in the torch models?

darts/models/forecasting_model.py
```diff
@@ -585,13 +587,23 @@ def residuals(self,
 class GlobalForecastingModel(ForecastingModel, ABC):
     """ The base class for "global" forecasting models, handling several time series and optional covariates.
 
+    Global forecasting models expand upon the functionality of `ForecastingModel` in 4 ways:
+    1. Models can be fitted on many series (multivariate or univariate) with different indices.
+    2. The input series can be different from the series used to fit the model.
+    3. Covariates can be supported (multivariate or univariate).
+    4. They can allow for multivariate target series and covariates.
+    
+    Note that not all global forecasting models implement points 3 and 4.
```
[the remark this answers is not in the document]
pennfranc: I removed the line

darts/models/forecasting_model.py
pennfranc: I adapted the description

darts/models/forecasting_model.py
hrzn: Yes, if we can find a better naming I'm up for it 👍

hrzn (approved):

Issue #287, opened 2021-03-08
Parallelise gridsearch computation


whiletruelearn:
**Is your feature request related to a current problem? Please describe.**
Great library. Can we think of running different models in parallel while doing `gridsearch`.
Current search is sequential and takes a lot of time.

**Describe proposed solution**
Bring in some multiprocessing logic for efficient gridsearch operation.

**Describe potential alternatives**
we can thinking of using `ray.tune` , `joblib` or simple `multiprocessing`  for this feature.

**Additional context**

This will be a useful enhancement IMHO

tneuer:
Great suggestion! That would indeed be useful and speed up things. We'll see if we manage to implement this in the next release and keep you updated.
Thanks for the input and we're glad you like the library :)

hrzn:
+1, good idea

hrzn:
This is now part of v0.9.0.

PR #346, opened 2021-06-07
Feature/parallelise grid search

piaz97:
Addresses #287

### Summary

Parallelise grid search computation

### Detailed Information
- Added parallelization to class method `gridsearch` in  `ForecastingModel`.
- Changed default `model_name` constructor parameter for `TorchForecastingModel`, from `'torch_model_run'` to `'torch_model_run_' + str(os.getpid())` (as suggested in a TODO). In this way, models running at the same time will not overwrite each other state on the disk (e.g., saving and deleting old checkpoints).
- Added tests for checking whether the best parameter combination changes between running gridsearch sequentially and in parallel. 
- Added `n_jobs` parameter to select_best_model `method in the `FourTheta` model, which is passed to the underlying gridsearch


darts/models/torch_forecasting_model.py
```diff
@@ -104,7 +104,7 @@ def __init__(self,
                  lr_scheduler_cls: torch.optim.lr_scheduler._LRScheduler = None,
                  lr_scheduler_kwargs: Optional[Dict] = None,
                  loss_fn: nn.modules.loss._Loss = nn.MSELoss(),
-                 model_name: str = "torch_model_run",  # TODO: uid
+                 model_name: str = "torch_model_run_" + str(os.getpid()),
```
hrzn, line 107: We could also maybe add a timestamp, to make it even more unique across several runs happening e.g. few days appart. For instance something like "<date-HH:MM:SS>_torch_model_run_<PID>"

darts/tests/test_backtesting.py
```diff
@@ -220,6 +222,58 @@ def test_gridsearch(self):
         es_params = {'seasonal_periods': list(range(5, 10))}
         self.assertTrue(compare_best_against_random(ExponentialSmoothing, es_params, dummy_series))
 
+    @unittest.skipUnless(TORCH_AVAILABLE, "requires torch")
+    def test_gridsearch_n_jobs(self):
+        '''
+        Testing that running gridsearch with multiple workers returns the same best_parameters as the single worker run.
```
hrzn, line 228: very nice, thanks!

hrzn (approved):
👍

darts/models/torch_forecasting_model.py
piaz97: Sure, that sounds good!🔝 I added that with the following formatting `"YYYY-mm-dd_HH:MM:SS_torch_model_run_PID"`, e.g., 2021-06-14_09:53:32_torch_model_run_44607

hrzn: Perfect, thanks

PR #347, opened 2021-06-14
Fix/nbeats logtensorboard

piaz97:
Fixes crash when running NBEATS with `log_tensorboard = True`.

### Summary

Model training was crashing when run with `log_tensorboard = True`. The problem consisted in the last backcast layer in the last block of the last stack in the model, which was instantiated as a trainable part of the net, but was not considered during the backprop (since not connected to the final prediction and loss function). When tensordboard was accessing the gradient data of those parameters, the gradient was still set to `None`since never updated. 

Screenshot 2021-06-14 at 11 46 58

I fixed the bug by:
- in the NBEATS model, setting the last FC backcast linear layer and backcast_g `requires_grad` to `False`;
- in TorchForecastingModel, skipping accessing the gradient data of params with `requires_grad=False`


pennfranc:
Looks good to me, thanks!

hrzn (approved):
good one, thanks

PR #349, opened 2021-06-17
Fix/future covariates loader

pennfranc:
My previous implementation of future covariates did not work in the general case for any predict batch size because as we iterate through batches of the input time series, we also need to iterate through future covariates. 
To solve this I adapted the `SimpleInferenceDataset` class to now return 3 values: target_series_past, covariates_past, covariates_future. The last two values are optional.

I also had to adapt the last part of the `predict_from_dataset` method where the `TimeSeries` predictions are created. This is because the current version on develop assumes that all time series that are being predicted fit in one batch. I had to make sure the indices of the time series in each batch are known so that the right time indices can be retrieved from the `SimpleInferenceDataset`.

### DIFF
--- a/darts/models/torch_forecasting_model.py
+++ b/darts/models/torch_forecasting_model.py
@@ -637,7 +637,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         Optional[Sequence[TimeSeries]]
             Covariates required to predict horizons up to `self.output_chunk_length`, or `None` if the model
             does not require covariates.
-        Optional[torch.Tensor]
+        Optional[Sequence[TimeSeries]]
             Covariates required to predict horizons beyond `self.output_chunk_length`, or `None` if the model
             does require future covariates either because it does not require covariates at all or because
             `n <= self.output_chunk_length`.
--- a/darts/models/torch_forecasting_model.py
+++ b/darts/models/torch_forecasting_model.py
@@ -461,10 +461,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
                      'of the series this model has been trained on. Provided input dim = {}, '
                      'model input dim = {}'.format(in_dim, self.input_dim))
 
-        # create past input, past covariate and future covariate series
-        tgt_past_arr, cov_past_arr, cov_future_arr = self._prepare_predict_timeseries(n, series, covariates)
-
-        dataset = SimpleInferenceDataset(tgt_past_arr, cov_past_arr, cov_future_arr)
+        dataset = SimpleInferenceDataset(series, covariates, n, self.input_chunk_length, self.output_chunk_length)
         predictions = self.predict_from_dataset(n, dataset, verbose=verbose, batch_size=batch_size, n_jobs=n_jobs,
                                                 roll_size=roll_size)
         return predictions[0] if called_with_single_series else predictions
@@ -596,7 +593,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
                 batch_prediction = torch.cat(batch_prediction, dim=1)
                 batch_prediction = batch_prediction[:, :n, :]
                 batch_prediction = batch_prediction.cpu().detach().numpy()
-                
+
                 ts_forecasts = Parallel(n_jobs=n_jobs)(
                     delayed(self._build_forecast_series)(
                         batch_prediction[batch_idx],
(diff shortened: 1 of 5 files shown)


darts/models/torch_forecasting_model.py
```diff
@@ -71,26 +71,28 @@ def _cat_with_optional(tsr1: torch.Tensor, tsr2: Optional[torch.Tensor]):
...
 
-        if len(item) == 2:
+        if isinstance(self.ts_dataset, SimpleInferenceDataset):
             # the dataset contains (input_target, input_covariate) only
-            input_tgt = torch.from_numpy(item[0]).float()
-            input_cov = torch.from_numpy(item[1]).float() if item[1] is not None else None
-            return self._cat_with_optional(input_tgt, input_cov)
+            input_tgt = torch.from_numpy(item[0].values()).float()
+            input_cov = torch.from_numpy(item[1].values()).float() if item[1] is not None else None
+            future_cov = torch.from_numpy(item[2].values()).float() if item[2] is not None else np.array([])
```
pennfranc, line 84: I can't return `None` here, the data loader will not accept it. But the `np.array([])` workaround is a bit hacky, I'm still thinking about how to improve this. Open to suggestions.

darts/models/torch_forecasting_model.py
```diff
@@ -71,26 +71,28 @@ def _cat_with_optional(tsr1: torch.Tensor, tsr2: Optional[torch.Tensor]):
...
 
-    def __getitem__(self, idx: int) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
+    def __getitem__(self, idx: int) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
         """
         Cast the content of the dataset to torch tensors
         """
         item = self.ts_dataset[idx]
 
-        if len(item) == 2:
+        if isinstance(self.ts_dataset, SimpleInferenceDataset):
```
hrzn, line 80: maybe you could check that it's an `TimeSeriesInferenceDataset` instead? (slightly more general)

darts/models/torch_forecasting_model.py
```diff
@@ -71,26 +71,28 @@ def _cat_with_optional(tsr1: torch.Tensor, tsr2: Optional[torch.Tensor]):
...
             # the dataset contains (input_target, input_covariate) only
-            input_tgt = torch.from_numpy(item[0]).float()
-            input_cov = torch.from_numpy(item[1]).float() if item[1] is not None else None
-            return self._cat_with_optional(input_tgt, input_cov)
+            input_tgt = torch.from_numpy(item[0].values()).float()
```
hrzn, line 82: good catch, I guess the previous version wasn't working...

darts/models/torch_forecasting_model.py
hrzn: Indeed it's a bit hacky. Maybe we need two different Torch Dataset classes?

pennfranc: You mean one class that does return future covariates and one that does not?

darts/models/torch_forecasting_model.py
```diff
@@ -72,26 +72,31 @@ def _cat_with_optional(tsr1: torch.Tensor, tsr2: Optional[torch.Tensor]):
...
+            input_cov = torch.from_numpy(item[1].values()).float() if item[1] is not None else None
+            future_cov = torch.from_numpy(item[2].values()).float() if item[2] is not None else None
 
-        elif len(item) == 3:
+            if future_cov is not None:
```
pennfranc, line 87: @hrzn second attempt of solving this issue. The problem with this is we don't have a consistent return type, which is clearly not ideal. At the same time, duplicating the class for every possible scenario seems kind of inefficient too. I don't know yet how to best handle this.

darts/models/torch_forecasting_model.py
```diff
@@ -593,23 +597,28 @@ def predict_from_dataset(self,
                 batch_prediction = batch_prediction[:, :n, :]
                 batch_prediction = batch_prediction.cpu().detach().numpy()
                 
-                ts_forecasts = Parallel(n_jobs=n_jobs)(delayed(self._build_forecast_series)(prediction, input_series[0])
-                                                       for prediction, input_series in zip(batch_prediction,
-                                                                                           input_series_dataset))
+                ts_forecasts = Parallel(n_jobs=n_jobs)(
```
pennfranc, line 600: This change was actually already necessary before introducing future covariates. The previous version did not handle batches properly as far as I can tell. This is another sign that we need more exhaustive tests here.

darts/models/torch_forecasting_model.py
hrzn: Yes

darts/models/torch_forecasting_model.py
hrzn: I guess this hack can do it for now.

darts/models/torch_forecasting_model.py
```diff
@@ -593,23 +597,28 @@ def predict_from_dataset(self,
...
+    def _prepare_predict_timeseries(self, n: int, series, covariates):
         """
-        Creates a torch tensor `in_past` from `TimeSeriesInferenceDataset` instance for initial input to model
-        which includes the target series and covariate series if the model was trained with covariates.
-        Only past covariates are included, even if more are provided.
-
-        If the model is required to produce the forecast over multiple iterations, i.e. if
-        `n > self.output_chunk_length`, and if it was trained with covariates, then `cov_future` will
-        contain the future covariates up to `n` time steps into the future.
+        Creates 3 lists with `TimeSeries` instances required to produce model predictions.
```
hrzn, line 614: Good comment, thanks!

darts/models/torch_forecasting_model.py
```diff
@@ -622,74 +631,70 @@ def _prepare_predict_tensors(self, n: int, input_series_dataset: TimeSeriesInfer
...
+        raise_if_not(len(series) == len(covariates),
+                     'The number of target series must be equal to the number of covariates.')
+
+        for target_series, covariate_series in zip(series, covariates):
             raise_if_not(len(target_series) >= self.input_chunk_length,
-                            'All input series must have length >= `input_chunk_length` ({}).'.format(
+                         'All input series must have length >= `input_chunk_length` ({}).'.format(
                 self.input_chunk_length))
 
             # TODO: here we could be smart and handle cases where target and covariates do not have same time axis.
```
hrzn, line 659: I think we could perhaps remove this comment, as the logic you implemented does this (more or less)

darts/models/torch_forecasting_model.py
```diff
@@ -622,74 +631,70 @@ def _prepare_predict_tensors(self, n: int, input_series_dataset: TimeSeriesInfer
...
+        tgt_past_arr = []
+        cov_past_arr = []
         cov_future_arr = []
 
-        for target_series, covariate_series in input_series_dataset:
+        covariates = [None] * len(series) if covariates is None else covariates
+        raise_if_not(len(series) == len(covariates),
+                     'The number of target series must be equal to the number of covariates.')
+
+        for target_series, covariate_series in zip(series, covariates):
```
hrzn, line 654: I'm a bit scared about this loop for large datasets.

Before, we had a "shallow" wrapping of the `Sequence[TimeSeries]` into a `SimpleInferenceDataset`, which means that the data doesn't have to be all loaded at once in memory (e.g. if the Sequence[TimeSeries] implements lazy loading). Now we are looping and building Lists which means everything will be held in memory.

I guess it can be fine (as it's for inference only - likely less often subject to big datasets - it wouldn't work for training), but it would be nice to find a way to do this split lazily to be really scalable in the 1% of cases where it's needed :D

hrzn:
@pennfranc my main comment concerns the memory requirements of the solution (which proactively builds lists containing everything) for big datasets. Maybe give it some thoughts, and if it's too complex to go around it for now we can merge your PR. For now IMO it's still better to support future covariates, than to scale to 0.1% of the biggest datasets (but ideally we should support both).

darts/models/torch_forecasting_model.py
pennfranc: Done

darts/models/torch_forecasting_model.py
pennfranc: Adapted it and moved it to `SimpleInferenceDataset.__init__`

darts/models/torch_forecasting_model.py
pennfranc: Absolutely agree, I removed the loop and moved the logic to the `SimpleInferenceDataset.__getitem__` method.

pennfranc:
> @pennfranc my main comment concerns the memory requirements of the solution (which proactively builds lists containing everything) for big datasets. Maybe give it some thoughts, and if it's too complex to go around it for now we can merge your PR. For now IMO it's still better to support future covariates, than to scale to 0.1% of the biggest datasets (but ideally we should support both).

@hrzn I removed the loop and put all of the logic that was inside it into the `SimpleInferenceDataset.__getitem__` method, which should restore the lazy loading behavior. This was a really easy change and I think it's way better now! Let me know if the change makes sense to you.

darts/utils/data/simple_inference_dataset.py
```diff
@@ -11,17 +11,88 @@
...
+        be equal to the future covariates up to `n - self.output_chunk_length` time steps into the future.
+
+        Parameters
+        ----------
+        series
+            The target series that are to be predicted into the future.
+        covariates
+            Optionally, the corresponding covariates that are used for predictions. This argument is required
+            if the model was trained with covariates.
+        n
```
hrzn, line 39: Perhaps a couple of comment lines to explain why these three parameters are needed would be nice (for our future selves ;)

hrzn, line 39: Perhaps a couple of comment lines to explain why these three parameters are needed would be nice (for our future selves ;)

darts/utils/data/simple_inference_dataset.py
```diff
@@ -11,17 +11,88 @@
...
+
+        cov_past = cov_future = None
+        if covariate_series is not None:
+
+            # get first timestamp that lies in the future of target series
+            first_pred_time = target_series.end_time() + target_series.freq()
+
+            # isolate past covariates and add them to array
+            if covariate_series.end_time() >= first_pred_time:
+                cov_past = covariate_series.drop_after(first_pred_time)
```
hrzn, line 79: All of these commands (drop_after, indexing with [], etc...) create new `TimeSeries` objects, which means behind the scenes a lot of object creation, and data copying (remember the whole data is copied over whenever a series is created). I suspect the performance might be much better if we make __getitem__ return np.ndarray's instead, and then access the values using `series.values(copy=False)`, very much like we do within the training datasets. It's (slightly) uglier, but it has the big advantage of manipulating only Numpy views of the underlying arrays, which I think will be much faster in some cases.
WDYT?

pennfranc: Yeah I see what you mean. However, the problem is that we need the time indices at the time of prediction. I think that was the initial motivation between making `SimpleInferenceDataset` return `TimeSeries` instances instead of numpy arrays. Strictly speaking, we only need the time indices of the target series. So maybe we could have a mixed return type, only creating `TimeSeries` objects for the target series, but not for the covariates to at least save some resources? Or I guess we could separately emit the timestamp that indicates the start of the prediction and have the target and covariate values be numpy arrays?

darts/utils/data/simple_inference_dataset.py
pennfranc: Done

darts/utils/data/simple_inference_dataset.py
hrzn: You're right, we do need the time indices.
I would then propose that we leave it like that for now, and add a task in our backlog to revisit this at some later point of time. We could then do like you propose (emit only the start of the prediction Timestamp).

hrzn (approved):

PR #353, opened 2021-06-23
torch model initialization update

mkos:
@hrzn I had to expand an API of some of existing calls to account for removed `n_epochs` parameter and for new `epochs` parameter in`fit()` method. See the specific comments below

### DIFF
--- a/darts/models/forecasting_model.py
+++ b/darts/models/forecasting_model.py
@@ -154,6 +154,7 @@ class ForecastingModel(ABC):
                              forecast_horizon: int = 1,
                              stride: int = 1,
                              retrain: bool = True,
+                             retrain_epochs: int = 100,
                              overlap_end: bool = False,
                              last_points_only: bool = True,
                              verbose: bool = False) -> Union[TimeSeries, List[TimeSeries]]:
@@ -199,6 +200,10 @@ class ForecastingModel(ABC):
             Whether to retrain the model for every prediction or not. Currently only `TorchForecastingModel`
             instances such as `RNNModel`, `TCNModel`, `NBEATSModel` and `TransformerModel` support
             setting `retrain` to `False`.
+        retrain_epochs
+            If ``retrain`` is set to ``True``, retrain for so many epochs. Currently only `TorchForecastingModel`
+            instances such as `RNNModel`, `TCNModel`, `NBEATSModel` and `TransformerModel` use this parameter.
+            [Default: 100]
         overlap_end
             Whether the returned forecasts can go beyond the series' end or not
         last_points_only
@@ -254,12 +259,20 @@ class ForecastingModel(ABC):
                 train_cov = covariates.drop_after(pred_time)
 
             if retrain:
-                if covariates and "covariates" in fit_signature.parameters:
-                    self.fit(series=train, covariates=train_cov)
-                elif covariates and "exog" in fit_signature.parameters:
-                    self.fit(series=train, exog=train_cov)
+                if self.__class__.__bases__[0].__name__ == 'TorchForecastingModel':
+                    if covariates and "covariates" in fit_signature.parameters:
+                        self.fit(series=train, covariates=train_cov, epochs=retrain_epochs)
+                    elif covariates and "exog" in fit_signature.parameters:
+                        self.fit(series=train, exog=train_cov, epochs=retrain_epochs)
+                    else:
+                        self.fit(series=train, epochs=retrain_epochs)
                 else:
-                    self.fit(series=train)
+                    if covariates and "covariates" in fit_signature.parameters:
+                        self.fit(series=train, covariates=train_cov)
+                    elif covariates and "exog" in fit_signature.parameters:
+                        self.fit(series=train, exog=train_cov)
+                    else:
+                        self.fit(series=train)
 
             if covariates:
                 if 'covariates' in predict_signature.parameters:
@@ -301,6 +314,7 @@ class ForecastingModel(ABC):
                  forecast_horizon: int = 1,
                  stride: int = 1,
                  retrain: bool = True,
+                 retrain_epochs: int = 100,
                  overlap_end: bool = False,
                  last_points_only: bool = False,
                  metric: Callable[[TimeSeries, TimeSeries], float] = metrics.mape,
@@ -350,6 +364,10 @@ class ForecastingModel(ABC):
             Whether to retrain the model for every prediction or not. Currently only `TorchForecastingModel`
             instances such as `RNNModel`, `TCNModel`, `NBEATSModel` and `TransformerModel` support
             setting `retrain` to `False`.
+        retrain_epochs
+            If ``retrain`` is set to ``True``, retrain for so many epochs. Currently only `TorchForecastingModel`
+            instances such as `RNNModel`, `TCNModel`, `NBEATSModel` and `TransformerModel` use this parameter.
+            [Default: 100]
         overlap_end
             Whether the returned forecasts can go beyond the series' end or not
         last_points_only
@@ -373,6 +391,7 @@ class ForecastingModel(ABC):
                                               forecast_horizon,
                                               stride,
                                               retrain,
+                                              retrain_epochs,
                                               overlap_end,
                                               last_points_only,
                                               verbose)
@@ -392,6 +411,7 @@ class ForecastingModel(ABC):
                    series: TimeSeries,
                    covariates: Optional[TimeSeries] = None,
                    forecast_horizon: Optional[int] = None,
+                   epochs: int = 100,
                    start: Union[pd.Timestamp, float, int] = 0.5,
                    last_points_only: bool = False,
                    val_series: Optional[TimeSeries] = None,
@@ -448,6 +468,10 @@ class ForecastingModel(ABC):
             An optional covariate series. This applies only if the model supports covariates.
         forecast_horizon
             The integer value of the forecasting horizon used in expanding window mode.
+        epochs
+            train for so many epochs. Currently only `TorchForecastingModel`
+            instances such as `RNNModel`, `TCNModel`, `NBEATSModel` and `TransformerModel` use this parameter.
+            [Default: 100]
         start
             The `int`, `float` or `pandas.Timestamp` that represents the starting point in the time index
             of `training_series` from which predictions will be made to evaluate the model.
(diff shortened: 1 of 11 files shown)


darts/models/forecasting_model.py
```diff
@@ -154,6 +154,7 @@ def historical_forecasts(self,
                              forecast_horizon: int = 1,
                              stride: int = 1,
                              retrain: bool = True,
+                             retrain_epochs: int = 100,
```
mkos, line 157: @hrzn this `retrain_epochs` param is only used when `TorchForecastingModel` is being used. The default 100 epochs was a previous default for `n_epochs`.

darts/models/forecasting_model.py
```diff
@@ -254,12 +259,20 @@ def historical_forecasts(self,
                 train_cov = covariates.drop_after(pred_time)
 
             if retrain:
-                if covariates and "covariates" in fit_signature.parameters:
-                    self.fit(series=train, covariates=train_cov)
-                elif covariates and "exog" in fit_signature.parameters:
-                    self.fit(series=train, exog=train_cov)
+                if self.__class__.__bases__[0].__name__ == 'TorchForecastingModel':
```
mkos, line 262: @hrzn since `epoch` parameter to the `fit()` method is unique to the `TorchForecastingModels` I have founded my code based on base class name

darts/models/forecasting_model.py
```diff
@@ -301,6 +314,7 @@ def backtest(self,
                  forecast_horizon: int = 1,
                  stride: int = 1,
                  retrain: bool = True,
+                 retrain_epochs: int = 100,
```
mkos, line 317: @hrzn same as with `historical_forecasts`.

darts/models/forecasting_model.py
```diff
@@ -392,6 +411,7 @@ def gridsearch(model_class,
                    series: TimeSeries,
                    covariates: Optional[TimeSeries] = None,
                    forecast_horizon: Optional[int] = None,
+                   epochs: int = 100,
```
mkos, line 414: @hrzn same as with `historical_forecasts`. This time the name is different - would you say it should be `retrain_epochs` for consistency?

darts/models/forecasting_model.py
```diff
@@ -552,6 +590,7 @@ def _evaluate_combination(param_combination):
     def residuals(self,
                   series: TimeSeries,
                   forecast_horizon: int = 1,
+                  retrain_epochs: int = 100,
```
mkos, line 593: @hrzn same as with `historical_forecasts`.

darts/models/regression_ensemble_model.py
```diff
@@ -51,7 +51,19 @@ def __init__(self,
         self.regression_model = regression_model
         self.train_n_points = regression_train_n_points
 
-    def fit(self, series: TimeSeries) -> None:
+    def fit(self, series: TimeSeries, epochs: Optional[int]=100) -> None:
```
mkos, line 54: @hrzn this is a bit tricky because I went with a shortcut - you're unable to set different `epochs` value for any of the models in ensemble. Also `epochs` or `retrain_epochs`?

darts/models/forecasting_model.py
mkos: Also: do you want UTs for new params? I think they should be.

darts/models/forecasting_model.py
```diff
@@ -590,6 +633,7 @@ def residuals(self,
                                       forecast_horizon=forecast_horizon,
                                       stride=1,
                                       retrain=True,
+                                      retrain_epochs=retrain_epochs,
```
mkos, line 636: @hrzn same here. Note that this function is not covered by UT.

darts/models/forecasting_model.py
```diff
@@ -254,12 +259,20 @@ def historical_forecasts(self,
...
 
             if retrain:
-                if covariates and "covariates" in fit_signature.parameters:
-                    self.fit(series=train, covariates=train_cov)
-                elif covariates and "exog" in fit_signature.parameters:
-                    self.fit(series=train, exog=train_cov)
+                if self.__class__.__bases__[0].__name__ == 'TorchForecastingModel':
+                    if covariates and "covariates" in fit_signature.parameters:
+                        self.fit(series=train, covariates=train_cov, epochs=retrain_epochs)
+                    elif covariates and "exog" in fit_signature.parameters:
```
pennfranc, line 265: This case would never actually happen I think, since `TorchForecastingModel.fit` does not expect a `exog` parameter.

darts/models/forecasting_model.py
```diff
@@ -254,12 +259,20 @@ def historical_forecasts(self,
...
+                        self.fit(series=train, exog=train_cov, epochs=retrain_epochs)
+                    else:
+                        self.fit(series=train, epochs=retrain_epochs)
                 else:
-                    self.fit(series=train)
+                    if covariates and "covariates" in fit_signature.parameters:
```
pennfranc, line 270: At this point in time, this case is also never realized as far as I can tell, since only `TorchForecastingModel` instances expect a `covariates` parameter in their fit function. But it I guess it might be worth leaving it here, especially since if I understand correctly regression models will soon inherit `GlobalForecastingModel`.

darts/models/regression_ensemble_model.py
pennfranc, line 54: Should we also change the function signature of the superclass?

pennfranc: I guess one way to avoid this could be to keep the `TorchForecastingModel.epochs` attribute, which is optionally set if a number of epochs is provided in `TorchForecastingModel.__init__`, but is overridden if `epochs` is provided in fit. That way we can still create an ensemble with models requiring different numbers of epochs. Just a random idea, there might be a better way..

darts/models/torch_forecasting_model.py
```diff
@@ -118,6 +122,10 @@ def __init__(self,
         When subclassing this class, please make sure to set the self.model attribute
         in the __init__ function and then call super().__init__ while passing the kwargs.
 
+        Note: if you set custom name for your model and do some training (i.e. checkpoints are generated), then
```
pennfranc, line 125:
```suggestion
        Note: if you set a custom name for your model and do some training (i.e. checkpoints are generated), then
```

darts/models/torch_forecasting_model.py
```diff
@@ -118,6 +122,10 @@ def __init__(self,
...
+        during next model initialization with the same name, exisiting state from the last checkpoint will be loaded.
```
pennfranc, line 126:
```suggestion
        during the next model initialization with the same name, the existing state from the last checkpoint will be loaded.
```

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+        if os.path.exists(checkpoints_folder) and len(glob(os.path.join(checkpoints_folder, "checkpoint_*"))) > 0:
+            self.load()
+
+    def load(self, best=False):
+        ''' Loads the model from a checkpoint by given name and work directory.
+
+            Parameters
+            ----------
+            best
+                Load the best performing checkpoint checkpoint instead of the latest checkpoint (the default)
```
pennfranc, line 218:
```suggestion
                Load the best performing checkpoint instead of the latest checkpoint (the default)
```

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+        model = self.load_from_checkpoint(self.model_name, self.work_dir, best)
+        self.__dict__.update(model.__dict__)
+        self.set_params(**params)
+
+    @classmethod
+    def _get_param_names(cls):
+        """Get parameter names for the estimator"""
+        # fetch the constructor or the original constructor before
+        # deprecation wrapping if any
+        # source: scikit-learn's base.py
```
pennfranc, line 231: If this code is adapted from sklearn, should we maybe include the link to the file?

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+        # Extract and sort argument names excluding 'self'
+        return sorted([p.name for p in parameters])
+
+    def get_params(self, deep=True):
+        """
+        Get parameters for this estimator.
+        source: scikit-learn's base.py
+        Parameters
+        ----------
+        deep : bool, default=Tru
```
pennfranc, line 260: We might have to adapt the docstring to match our own docstrings here and for the other sklearn-based methods.

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+                deep_items = value.get_params().items()
+                out.update((key + '__' + k, val) for k, val in deep_items)
+            out[key] = value
+        return out
+
+    def set_params(self, **params):
+        """
+        Set the parameters of this estimator.
+        The method works on simple estimators as well as on nested objects
+        (such as :class:`~sklearn.pipeline.Pipeline`). The latter have
```
pennfranc, line 281: We have to adapt the docstring here

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+        out = dict()
+        for key in self._get_param_names():
+            value = getattr(self, key)
+            if deep and hasattr(value, 'get_params'):
+                deep_items = value.get_params().items()
+                out.update((key + '__' + k, val) for k, val in deep_items)
+            out[key] = value
+        return out
+
+    def set_params(self, **params):
```
hrzn, line 277: Could you make this method private?

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+                raise RuntimeError("scikit-learn estimators should always "
+                                   "specify their parameters in the signature"
+                                   " of their __init__ (no varargs)."
+                                   " %s with constructor %s doesn't "
+                                   " follow this convention."
+                                   % (cls, init_signature))
+        # Extract and sort argument names excluding 'self'
+        return sorted([p.name for p in parameters])
+
+    def get_params(self, deep=True):
```
hrzn, line 254: Could you make this method private?

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+                Load the best performing checkpoint checkpoint instead of the latest checkpoint (the default)
+
+        '''
+        params = {k:v for k,v in self.get_params() if k not in self.immutable_args}
+        model = self.load_from_checkpoint(self.model_name, self.work_dir, best)
+        self.__dict__.update(model.__dict__)
+        self.set_params(**params)
+
+    @classmethod
+    def _get_param_names(cls):
```
hrzn, line 227: This code seems to be used only from within `get_params()`. Could you maybe copy this method over into `get_params()` ?
Also please adapt all the sklearn things (and remove what's unnecessary).

darts/models/torch_forecasting_model.py
```diff
@@ -200,6 +205,131 @@ def __init__(self,
...
+
+        for key, sub_params in nested_params.items():
+            valid_params[key].set_params(**sub_params)
+
+        return self
+
+    def reset_model(self):
+        ''' Resets the model object.
+
+            WARNING: also removes all stored data like checkpoints and training history.
```
hrzn, line 322:
```suggestion
            WARNING: also removes all stored data like checkpoints and training history. If you want to keep the existing checkpoints and training history, create a new model instance with a different name.
```

hrzn (commented):
I'm not convinced by removing `n_epochs` from the constructor, for essentially two reasons:
*) It introduces extra complexity and parameters (and signature changes) in different parts of the code (e.g. backtesting etc)
*) The number of epochs to train a model for can actually be seen in some cases as just another hyper-parameters like any other, and a sort of "property" of the model, so IMO it is quite natural to accept it in the constructor.

I think we should go with a solution like what @pennfranc proposed in one of his comments; namely accepting the number of epochs in the constructor, and *optionally* also in the fit() method. This way: if it's not provided to fit(), then the class attribute is used. If it's provided to fit(), then this number of epochs is used instead (this will allow users who need to only quickly train for an extra N epochs). Doing this change should allow you to simplify your PR quite a bit. And it will also remove the need for the extra unit tests you were considering.

Otherwise the logic of never-by-default resetting a model (and so always doing "additional" training by default) is good IMO.

mkos:
Other PR will be opened with another implementation. Thanks for all the feedback.

PR #354, opened 2021-06-23
Feat/True RNN

pennfranc:
I created a new 'truly recurrent' RNN class that replaced the previous `RNNModel`.  The previous RNN class is now called `BlockRNNModel`. `RNNModel` is fully recurrent in the sense that, at prediction time, an output is computed using these inputs:
- the previous target value, which will be set to the last known target value for the first prediction,
  and for all other predictions it will be set to the previous prediction
- the previous hidden state
- the current covariates (if the model was trained with covariates)

A prediction with forecasting horizon `n` thus is created in `n` iterations of `RNNModel` predictions and requires `n` future covariates to be known. 

In addition to being better suited at predicting time series that are highly dependent on known future covariates, it should be quite easy to build on top of this to build the DeepAR probabilistic model.

To make all of this work I had to do the following:
- Introduce a `TorchForecastingModel.is_recurrent` attribute to be able to differentiate between this new model and the old block models.
- Adapt the `ShiftedDataset` class to allow for also shifting covariates forward (in addition to just the target), such that recurrent models can train on the latest covariates.
- Create a `TorchForecastingModel._produce_train_output` that specifies how the training output is obtained from the torch module. By default (for block models) this is simple calling the forward function. However I had to create this function so that I can override it in `RNNModel`.
- Move the existing code inside the `TorchForecastingModel.predict_from_dataset` loop into a new `TorchForecastingModel._predict_batch_block_model` function.
- Create a `TorchForecastingModel._predict_batch_recurrent_model` function that defines the prediction logic for recurrent models.
- Call the correct method within `TorchForecastingModel.predict_from_dataset`, depending on whether we have a block or recurrent model.

Further changes:
- I updated the [examples/05-RNN-examples.ipynb](https://github.com/unit8co/darts/blob/feat/true-rnn/examples/05-RNN-examples.ipynb) notebook to make use of the new model
- I also sneaked in an updated description of how Darts handles future covariates in the [examples/02-multi-time-series-and-covariates.ipynb](https://github.com/unit8co/darts/blob/feat/true-rnn/examples/02-multi-time-series-and-covariates.ipynb) notebook.

Still left to do:
- Add more tests
- Establish whether recurrent abstractions could be extended to other models (Transformer), and whether it might make sense to create new parent classes for recurrent vs. block models.


darts/models/block_rnn_model.py
```diff
@@ -0,0 +1,194 @@
+"""
```
pennfranc, line 1: These changes here are not really changes, I simply renamed the old RNN model. I did add a couple of paragraphs to the docstrings, however, so I would be happy if you could check those.

examples/05-RNN-examples.ipynb
```diff
@@ -37,7 +37,15 @@
    "metadata": {
```
hrzn, line 37: <p>I guess this title could be changed (there's no longer output_chunk_length)?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/354/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

darts/utils/data/simple_inference_dataset.py
```diff
@@ -92,14 +100,19 @@ def __getitem__(self, idx: int) -> Tuple[TimeSeries, Optional[TimeSeries], Optio
...
                 )
+                req_cov_string = 'n' if self.model_is_recurrent else 'n - output_chunk_length'
                 raise_if_not(covariate_series.end_time() >= last_required_future_covariate_ts,
-                             'All covariates must be known `n - output_chunk_length` time steps into the future')
+                             'All covariates must be known `{}` time steps into the future'.format(req_cov_string))
 
                 # isolate necessary future covariates and add them to array
-                cov_future = covariate_series.drop_before(first_pred_time - covariate_series.freq())
-                cov_future = cov_future[:self.n - self.output_chunk_length]
+                cov_future = covariate_series.drop_before(first_pred_time - (1 - int(self.model_is_recurrent))
```
hrzn, line 114: How come the non-recurrent models need to be split one step before?

darts/models/torch_forecasting_model.py
```diff
@@ -399,15 +402,24 @@ def predict(self,
...
-        only the last ``input_chunk_length`` time steps will be considered.
+        Recurrent models:
+        All predictions are produced in a recurrent way by taking as input
+        - the previous target value, which will be set to the last known target value for the first prediction,
+          and for all other predictions it will be set to the previous prediction
+        - the previous hidden state
+        - the current covariates (if the model was trained with covariates)
+        As a result, if covariates were used, `n` covariates have to be available into the future.
+
+        If some time series in the `series` argument have more time steps than the model was trained with,
```
hrzn, line 421:
```suggestion
        If some time series in the `series` argument have more than `input_chunk_length` time steps,
```

darts/models/rnn_model.py
```diff
@@ -42,79 +45,64 @@ def __init__(self,
...
 class RNNModel(TorchForecastingModel):
     @random_method
     def __init__(self,
-                 input_chunk_length: int,
-                 output_chunk_length: int,
                  model: Union[str, nn.Module] = 'RNN',
-                 hidden_size: int = 25,
-                 n_rnn_layers: int = 1,
-                 hidden_fc_sizes: Optional[List] = None,
+                 input_chunk_length: int = 12,
```
hrzn, line 102: [For later]: could we imagine having None here for recurrent models, which would mean something "use the longest sequence available"?

darts/models/torch_forecasting_model.py
```diff
@@ -541,71 +563,108 @@ def predict_from_dataset(self,
         with torch.no_grad():
             for batch_tuple in iterator:
 
-                input_series, indices = batch_tuple[0], batch_tuple[-1]
+                input_series = batch_tuple[0]
```
hrzn, line 566: Could you maybe add a comment to remind here that `input_series` already contains the past covariates concatenated to the target series

darts/models/torch_forecasting_model.py
```diff
@@ -541,71 +563,108 @@ def predict_from_dataset(self,
...
+            # update predictions depending on how many data points have been predicted
+            if prediction_length <= n - self.output_chunk_length - roll_size:
+                batch_prediction.append(out[:, :roll_size, :])
+                prediction_length += roll_size
+            elif prediction_length < n - self.output_chunk_length:
+                # if we produce have `n - output_chunk_length < #predictions < n` we want to only use
+                # the predictions and covariates necessary to exactly reach `n - output_chunk_length`,
+                # so that the final forecast produces exactly the right number of predictions to reach `n`
+                spillover_prediction_length = (prediction_length + roll_size) - (n - self.output_chunk_length)
+                roll_size -= spillover_prediction_length
```
hrzn, line 636: `roll_size` cannot become negative here? (I guess it works, just asking)

hrzn (approved):
This looks great @pennfranc ! First truly recurrent network in Darts :)

Couple of comments (besides the minor things I've put in other PR comments):
* I think you can merge it as such, before taking a stab at the probabilistic version. The probabilistic version would require making the loos "dynamic" and having a neat mechanism so that each NN model can optionally use an architecture with different number of outputs. I would say we could do all of this in a second stage after the present PR.

* Do we still need the `input_chunk_length`? I would have expected `training_length` to be enough. (maybe that's the only thing I'd ask you to clarify or change before we can truly merge).

darts/utils/data/simple_inference_dataset.py
pennfranc: In the shifted dataset, for recurrent models, I shift the covariates forward relative to the input series by one time step. This ensures that recurrent models have as input the most recent covariates when making a prediction (covariates with the same timestamp as the target). Because the RNN is trained that way, this shift has to be incorporated into the SimpleInferenceDatqset class as well, and this applies to future covariates too. This translates to this different cutoff point. You think there might be a better way of handling this? I do see that having two different ways of creating cov_future might not be ideal if someone is trying to understand the code.

darts/models/rnn_model.py
```diff
@@ -42,79 +45,64 @@ def __init__(self,
...
-        if self.name == "LSTM":
-            hidden = hidden[0]
-        predictions = hidden[-1, :, :]
-        predictions = self.fc(predictions)
-        predictions = predictions.view(batch_size, self.out_len, self.target_size)
+        # TODO: for one layer out should be equal to hidden, make sure that's true
+        predictions = self.V(out)  # TODO: make sure this matrix multiplication is correct
+
+        # predictions is of size (batch_size, input_length, target_size)
+        predictions = predictions.view(batch_size, -1, self.target_size)
```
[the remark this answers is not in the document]
pennfranc: Not always I think. During training it will be equal to `training_length`, at the first call of `forward` during predict it will be equal to `input_chunk_length`, and at every subsequent call to forward it will be equal to 1. 
Generally speaking our RNN module will produce an output with the same length as the input that was provided to it (in the time dimension). It's just that during predict, we only ever use the last of those outputs.

darts/models/rnn_model.py
pennfranc: Yes, I was also thinking about this. I created a corresponding Jira task. The main problem with removing the fixed length requirement is that we create tensors that contain several time series at once. So we would have to create torch tensors that contain time series with different lengths, which might be a bit tricky, but I am not sure.

darts/models/forecasting_model.py
```diff
@@ -214,8 +214,8 @@ def historical_forecasts(self,
             If `last_points_only` is set to False, a list of the historical forecasts.
         """
         if covariates:
-            raise_if_not(series.has_same_time_as(covariates),
-                         'The provided series and covariates must have the same time index.')
+            raise_if_not(series.end_time() <= covariates.end_time(),
```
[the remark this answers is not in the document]
pennfranc: Done

darts/models/rnn_model.py
```diff
@@ -42,79 +45,64 @@ def __init__(self,
...
-        """ Here, we apply the FC network only on the last output point (at the last time step)
+        """ Here, we apply the V matrix to every hidden state to produce the outputs
         """
-        if self.name == "LSTM":
-            hidden = hidden[0]
-        predictions = hidden[-1, :, :]
-        predictions = self.fc(predictions)
-        predictions = predictions.view(batch_size, self.out_len, self.target_size)
+        # TODO: for one layer out should be equal to hidden, make sure that's true
+        predictions = self.V(out)  # TODO: make sure this matrix multiplication is correct
```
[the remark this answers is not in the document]
pennfranc: removed the comments

darts/models/torch_forecasting_model.py
pennfranc: Done

darts/models/torch_forecasting_model.py
pennfranc: It cannot become negative since `spillover_prediction_length < roll_size`. Proof:
```
n - self.output_chunk_length > prediction_length (precondition of elif block)
```
implies
```
0 > (prediction_length) - (n - self.output_chunk_length)
```
implies
```
roll_size > (prediction_length + roll_size) - (n - self.output_chunk_length) = spillover_prediction_length
```

darts/models/torch_forecasting_model.py
```diff
@@ -205,6 +205,9 @@ def __init__(self,
         self.lr_scheduler_cls = lr_scheduler_cls
         self.lr_scheduler_kwargs = dict() if lr_scheduler_kwargs is None else lr_scheduler_kwargs
 
+        # by default models are block models (i.e. not recurrent)
+        self.is_recurrent = False
```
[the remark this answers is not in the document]
pennfranc: Makes sense to create a method. But could we also make it a normal method that returns `False` and can optionally be overridden by subclasses? I'm just thinking in terms of reducing code because the block model is still the general case. But maybe that would not be ideal from a software engineering point of view?

darts/utils/data/simple_inference_dataset.py
```diff
@@ -92,14 +100,19 @@ def __getitem__(self, idx: int) -> Tuple[TimeSeries, Optional[TimeSeries], Optio
...
+                # recurrent models need `n`
                 last_required_future_covariate_ts = (
-                    target_series.end_time() + (self.n - self.output_chunk_length) * target_series.freq()
+                    target_series.end_time() + (self.n - self.output_chunk_length * (1 - int(self.model_is_recurrent)))
+                    * target_series.freq()
                 )
+                req_cov_string = 'n' if self.model_is_recurrent else 'n - output_chunk_length'
                 raise_if_not(covariate_series.end_time() >= last_required_future_covariate_ts,
-                             'All covariates must be known `n - output_chunk_length` time steps into the future')
+                             'All covariates must be known `{}` time steps into the future'.format(req_cov_string))
```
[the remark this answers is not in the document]
pennfranc: Masking as in filling in default values for future covariates if they're missing?

darts/models/rnn_model.py
```diff
@@ -128,31 +116,33 @@ def __init__(self,
...
+        hidden_dim
             Size for feature maps for each hidden RNN layer (:math:`h_n`).
-        n_rnn_layers
-            Number of layers in the RNN module.
-        hidden_fc_sizes
-            Sizes of hidden layers connecting the last hidden layer of the RNN module to the output, if any.
         dropout
             Fraction of neurons afected by Dropout.
+        training_length
+            The length of 1 training sample time series.
```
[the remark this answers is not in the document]
pennfranc: Done. I also added a suggestion to choose `training_length > input_chunk_length`. Let me know if that makes sense to you

darts/models/rnn_model.py
```diff
@@ -161,25 +151,31 @@ def __init__(self,
...
+                               hidden_dim=self.hidden_dim,
                                dropout=self.dropout)
         else:
             model = self.rnn_type_or_module
         return model
+
+    def _build_train_dataset(self,
+                             target: Sequence[TimeSeries],
+                             covariates: Optional[Sequence[TimeSeries]]) -> ShiftedDataset:
+        return ShiftedDataset(target_series=target,
```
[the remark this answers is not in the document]
pennfranc: Currently recurrent models only make sense with the `ShiftedDataset`, so I added a check for that at the beginning of `fit_from_dataset()`.

darts/utils/data/simple_inference_dataset.py
```diff
@@ -92,14 +100,19 @@ def __getitem__(self, idx: int) -> Tuple[TimeSeries, Optional[TimeSeries], Optio
             if self.n > self.output_chunk_length:
```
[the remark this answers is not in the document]
pennfranc: Yep, but so far that will be automatically true since the only recurrent model so far is `RNNModel` which cannot have a `output_chunk_length` different from 1. But I added the check for future recurrent models.

examples/05-RNN-examples.ipynb
pennfranc: Ah yes, thanks, I missed that. Done.

pennfranc:
@hrzn Thanks for the detailed feedback! Your suggestion to add DeepAR probabilistic functionality in a subsequent PR sounds reasonable to me. To your point about `input_chunk_length`, I think at some point we can remove it like you suggested. For this we need a good way to dynamically zero-pad the shorter time series in `SimpleInferenceDataset`. But I think it would also make sense to do this in a separate PR.

darts/utils/data/simple_inference_dataset.py
hrzn: Thanks for the explanation! I think you can leave it like that, but maybe copy-paste what you wrote here into a comment in the code for our future selves :) thanks!

darts/models/rnn_model.py
hrzn: OK perfect, let's revisit at some point.

darts/models/torch_forecasting_model.py
hrzn: Beautiful, thanks!

darts/utils/data/simple_inference_dataset.py
pennfranc: Done

PR #357, opened 2021-06-25
Fix unintentional user data removal

mkos:
Add parameter `force` to `TorchForecastingModel`'s constructor, if `force=False` model will throw an exception if model data for given `work_dir` and `model_name` exists. If `force=True`, and the data exists, the model will initialize but won't allow fitting (because that would overwrite the data again) until `reset_model()` is run. Alternatively, the user can load the data with the `load_from_checkpoint()` class method.

Additionally, implemented a new logic for handling `n_epochs` in constructor and `epochs` in `fit()` and `fit_from_dataset()` - basically, if present, `epochs` in fit* methods takes precedence.


darts/models/torch_forecasting_model.py
```diff
@@ -327,19 +362,17 @@ def fit_from_dataset(self,
                          verbose: bool = False,
                          epochs: int = 0) -> None:
 
+        raise_if(self.checkpoint_exists and self.force,
```
pennfranc, line 365: Is there a case where someone would set `force=True` and then would not want to call `reset_model()`. I'm asking because it seems to me that that if someone does set `force=True`, they always want to reset the model, in which case we could automatically call `reset_model()` if `force=True`, instead of requiring the user to do it. But maybe I'm missing something. What do you think?

darts/models/torch_forecasting_model.py
```diff
@@ -746,16 +779,15 @@ def _load_untrained_model(self, folder):
...
 
-    def _prepare_tensorboard_writer(self, continue_training: bool = False):
+    def _prepare_tensorboard_writer(self):
         runs_folder = _get_runs_folder(self.work_dir, self.model_name)
         if self.log_tensorboard:
-            if not continue_training:
-                shutil.rmtree(runs_folder, ignore_errors=True)
+            if self.total_epochs == 0:
+                tb_writer = SummaryWriter(runs_folder, purge_step=self.total_epochs)
+            else:
```
pennfranc, line 787: The logic in this else block was executed in the case where `continue_training == False` before, which corresponded to the case where `epochs==0`, meaning we were training the model from scratch. Now it is executed when `self.total_epochs > 0`. Now it is executed when `self.total_epochs > 0`, which is the opposite case, no? It seems to me we are executing the same code in a different case, or am I missing something? Or, if I'm right, why is that the case?

darts/tests/test_torch_forecasting_model.py
```diff
@@ -0,0 +1,142 @@
...
+            self.assertEqual(model1.total_epochs, 40)
+
+        # n_epochs = 20, fit|epochs=None, total_epochs=10 - train for another 20 epochs
+        def test_train_from_10_to_20_no_fit_epochs(self):
+            model1 = RNNModel(10, 10, n_epochs=20, work_dir=self.temp_work_dir)
+
+            times = pd.date_range('20130101', '20130410')
+            pd_series = pd.Series(range(100), index=times)
+            series = TimeSeries.from_series(pd_series)
+            model1.fit(series, epochs=10)  # cheating here a little bit
```
pennfranc, line 115: why cheating?

darts/tests/test_torch_forecasting_model.py
```diff
@@ -0,0 +1,142 @@
...
+            pd_series = pd.Series(range(100), index=times)
+            series = TimeSeries.from_series(pd_series)
+            model1.fit(series)
+            self.assertEqual(model1.total_epochs, 20)
+
+            model1.fit(series)
+            self.assertEqual(model1.total_epochs, 40)
+
+        # n_epochs = 20, fit|epochs=None, total_epochs=10 - train for another 20 epochs
+        def test_train_from_10_to_20_no_fit_epochs(self):
```
pennfranc, line 109:
```suggestion
        def test_train_from_10_to_30_no_fit_epochs(self):
```?

Comment (pennfranc) [reply darts/tests/test_torch_forecasting_model.py:109]: Also, why call it '...no_fit...'?

File: darts/tests/test_torch_forecasting_model.py [hunk 5]
```diff
@@ -0,0 +1,142 @@
...
+            model1 = RNNModel(10, 10, n_epochs=20, work_dir=self.temp_work_dir)
+
+            times = pd.date_range('20130101', '20130410')
+            pd_series = pd.Series(range(100), index=times)
+            series = TimeSeries.from_series(pd_series)
+            model1.fit(series, epochs=15)
+            self.assertEqual(model1.total_epochs, 15)
+
+        # n_epochs = 20, fit|epochs=15, total_epochs=10 - train for 15 epochs
+        def test_train_from_0_to_20_fit_15_epochs(self):
```
pennfranc, line 132: I like the tests, but I don't fully grasp the logic behind the names here

darts/models/torch_forecasting_model.py
hrzn: I was doing my review, and before reading your comment @pennfranc I wrote this:

"
Would there be a common use-case for creating a model with `force=True` but then not calling `reset_model()`?
It not really, I would suggest we go ahead and reset the model when `force` is `True`. If users set this to `True` we can take this as a voluntary action IMO.
"

mkos: Then the entire PR wouldn't make too much sense, because it would make again deleting user's data automatic in some cases. 
How I see it is that we often run notebooks in `Run all` mode. If there is a code that was left with `force=True` that would delete all the data and that might have not been users intention. This little param may get lost in all the code. I case of my approach, user has a chance to either remove `force=True` or do `reset_model()`. With the automatic decision, she may end up with the result she didn't want.
And I understand that users should be careful what they execute, but in the reality they often aren't from different reasons, and this code is covering that case.

darts/tests/test_torch_forecasting_model.py
mkos: it's cheating, because the case is that user interrupted training with Ctrl-C after 10 epochs and this code simulates it. But I can ofc remove the comment.

darts/tests/test_torch_forecasting_model.py
mkos: because training is based on the `n_epochs` parameter instead of `fit`'s `epochs` param. So it's rather `...no_fit_epochs`

darts/tests/test_torch_forecasting_model.py
mkos: in the case above: 0 is the current number of epochs, number after `_to_` is the `n_epochs` parameter and then either `epochs` value for `fit`. Do you have other suggestion?

darts/models/torch_forecasting_model.py
mkos: Good catch! You're absolutely right.

darts/tests/test_torch_forecasting_model.py
pennfranc: Could you maybe write that in a comment? Something like 'this line creates a model in the same state as it would be in if training had been interrupted after 10 epochs'. At least for me it would clarifies things.

darts/tests/test_torch_forecasting_model.py
pennfranc: Ah I see, thanks for explaining! But it should still say `...10_to_30...`, no?

darts/tests/test_torch_forecasting_model.py
mkos: Sure.

darts/models/torch_forecasting_model.py
mkos: @pennfranc / @hrzn is this a satisfactory explanation or should we discuss it further?

darts/tests/test_torch_forecasting_model.py
mkos: the second number tells the value of `n_epochs` parameter, but you're right, this is somewhat confusing. I'll think how to better convey the test purpose.

darts/models/torch_forecasting_model.py
```diff
@@ -327,19 +362,17 @@ def fit_from_dataset(self,
                          verbose: bool = False,
                          epochs: int = 0) -> None:
```
hrzn, line 363: The docstring for `epochs` in `fit()` says
```
If your model is already trained but you want to train it even further, you can provide here the number of additional epochs.
```
I would change it to
```
If specified, will train the model for `epochs` (additional) epochs, irrespective of what `n_epochs` was provided to the model constructor.
```

darts/models/torch_forecasting_model.py
hrzn: @mkos I'm not very convinced TBH because in any case someone building the model with `force=True` will in all likelihood also be calling `reset_model()`. So it will be equally likely to forget this piece of code as well in a notebook and have the same problem when launching "run all". 

But more in general, I'm not convinced that we should proactively try to anticipate all the sort of programming and environment issues users might have; at least when it comes at the expense of clarity (as is the case here - it's confusing to have a `force` parameter that doesn't reset the model). I think having a clear & straightforward API is the best way to avoid errors. Of course some users might mess up their execution runs in the notebooks, but frankly this sort of issues happening with notebooks is a bit out of our realm I think. We have to assume users will use notebooks correctly, and provide the clearest possible API.

hrzn: I think the part we should keep is the overriding of `epochs` parameter in `fit()` when provided. Also we can keep the `reset_model()` method, which could be convenient in some cases. WDYT?

hrzn (changes requested):

darts/models/torch_forecasting_model.py
mkos: so to sum up:
1. If `force=False` and checkpoints exist - raise exception
2. If `force=True` and checkpoints exist - run reset_model() quietly
3. fit does not emit any warning.
correct?

darts/tests/test_torch_forecasting_model.py
mkos: fixed the naming convention.

darts/models/torch_forecasting_model.py
mkos: But we don't want it at all, do we? (based on the discussion above)

darts/models/torch_forecasting_model.py
pennfranc: @mkos you still have to do this one right?

darts/models/torch_forecasting_model.py
pennfranc: That sounds good to me

hrzn: Me too 👍

darts/models/torch_forecasting_model.py
hrzn: The conversation above was about how to use `force` and `reset_model()` when instantiating a model with an existing name. However once a given model is instantiated, it still makes sense to accept the `epochs` parameter in the `fit()` function in order to potentially override the `n_epochs` class parameter, no? I think it's a nice feature :)

mkos: oh, sorry - I didn't notice that it's about the docstring...

mkos:
@hrzn / @pennfranc all the requested changes are done.

darts/models/torch_forecasting_model.py
```diff
@@ -164,6 +166,9 @@ def __init__(self,
         torch_device_str
             Optionally, a string indicating the torch device to use. (default: "cuda:0" if a GPU
             is available, otherwise "cpu")
+        force
+            force model loading, even if the data does not exist. Note that you won't be able to train model initialized
```
hrzn, line 170: I think you should update this docstring. Something like this?
```
If set to `True`, any previously-existing model with the same name will be reset (all checkpoints will be discarded).
```

darts/models/torch_forecasting_model.py
```diff
@@ -115,7 +116,8 @@ def __init__(self,
                  work_dir: str = os.path.join(os.getcwd(), DEFAULT_DARTS_FOLDER),
                  log_tensorboard: bool = False,
                  nr_epochs_val_period: int = 10,
-                 torch_device_str: Optional[str] = None):
+                 torch_device_str: Optional[str] = None,
+                 force=False):
```
hrzn, line 120: How about `force_reset` ?

hrzn (commented):
Just got minor comments about docstring and the name of the parameter - after this I think you can merge (once you solve conflicts and have the tests checks in green).

darts/models/torch_forecasting_model.py
mkos: makes sense.

darts/models/torch_forecasting_model.py
mkos: done.

darts/tests/test_global_forecasting_models.py
```diff
@@ -234,15 +233,12 @@ def test_fit_with_fit_epochs(self, rmtree_patch, train_patch, save_patch):
 
                 model.fit(multiple_ts, epochs=epochs)
 
-                rmtree_patch.assert_called()
```
pennfranc, line 237: One last thing. This tested whether `shutil.rmtree` was called, right? Is there a way we could still have a test case that covers this behavior when `reset_model` is called? I hope this makes sense, if not sorry, I'm not too familiar with this `patch` functionality.

pennfranc: I'm just a bit paranoid about removing test cases as darts is getting bigger and bigger

mkos: yes, but with this specific testcase rmtree should not be called - is that what you want to check? Just to make sure, that was mine test for the previous implementation and for me it didn't make sense to keep it there as the test is about something else.

darts/tests/test_torch_forecasting_model.py
```diff
@@ -0,0 +1,127 @@
...
+
+        def test_create_instance_existing_model_with_name_no_fit(self):
+            model_name = 'test_model'
+            model1 = RNNModel('RNN', 10, 10, work_dir=self.temp_work_dir, model_name=model_name)
+            # no exception is raised
+
+            model2 = RNNModel('RNN', 10, 10, work_dir=self.temp_work_dir, model_name=model_name)
+            # no exception is raised
+
+        @patch('darts.models.torch_forecasting_model.TorchForecastingModel.reset_model')
```
mkos, line 43: @pennfranc this and the next test case cover `reset_model()` - is it sufficient for you or do you want to cover another scenario?

darts/tests/test_global_forecasting_models.py
pennfranc: Makes sense!

darts/tests/test_torch_forecasting_model.py
pennfranc: Ah yes sorry I missed this one, thanks!

darts/tests/test_torch_forecasting_model.py
```diff
@@ -0,0 +1,127 @@
+from .base_test_class import DartsBaseTestClass
```
hrzn, line 1: Thanks for all the tests !

hrzn (approved):
Thanks @mkos, nice improvement!

PR #358, opened 2021-06-28
Fixed the ZeroDivision error bug

### LINKED ISSUE #312: [BUG] too short time series give a ZeroDivisionError
**Describe the bug**
When training some models with a short time series (eg. 8 points),
a ZeroDivisionError occur.
```
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-29-7c617aa60865> in <module>
     12 model = NBEATSModel(input_chunk_length=1, output_chunk_length=1)
     13 
---> 14 model.fit(series)

~/.local/lib/python3.8/site-packages/darts/utils/torch.py in decorator(self, *args, **kwargs)
     63         with fork_rng():
     64             manual_seed(self._random_instance.randint(0, high=MAX_TORCH_SEED_VALUE))
---> 65             decorated(self, *args, **kwargs)
     66     return decorator

~/.local/lib/python3.8/site-packages/darts/models/torch_forecasting_model.py in fit(self, series, covariates, val_series, val_covariates, verbose)
    292         logger.info('Train dataset contains {} samples.'.format(len(train_dataset)))
    293 
--> 294         self.fit_from_dataset(train_dataset, val_dataset, verbose)
    295 
    296     @random_method

~/.local/lib/python3.8/site-packages/darts/utils/torch.py in decorator(self, *args, **kwargs)
     63         with fork_rng():
     64             manual_seed(self._random_instance.randint(0, high=MAX_TORCH_SEED_VALUE))
---> 65             decorated(self, *args, **kwargs)
     66     return decorator

~/.local/lib/python3.8/site-packages/darts/models/torch_forecasting_model.py in fit_from_dataset(self, train_dataset, val_dataset, verbose)
    345 
    346         # Train model
--> 347         self._train(train_loader, val_loader, tb_writer, verbose)
    348 
    349         # Close tensorboard writer

~/.local/lib/python3.8/site-packages/darts/models/torch_forecasting_model.py in _train(self, train_loader, val_loader, tb_writer, verbose)
    518 
    519             if epoch % self.nr_epochs_val_period == 0:
--> 520                 training_loss = total_loss / len(train_loader)
    521                 if val_loader is not None:
    522                     validation_loss = self._evaluate_validation_loss(val_loader)

ZeroDivisionError: division by zero
```
Seems like this is the case only for global models.
Tried with:

- NBEATSModel
- TCNModel
- RNNModel
- TransformerModel

**To Reproduce**
Run the following code snippet to get the error

```python
from darts.models.transformer_model import TransformerModel
import pandas as pd
from darts import TimeSeries

df = pd.DataFrame({
    'time': ['2010-01-01', '2012-01-01'], # change 2012 to 2019, and the error disappear
    'values': [1, 2],
})

df['time'] =  pd.PeriodIndex(df['time'], freq='Q').to_timestamp()
df.index = df['time']
df = df.resample('Q').sum().interpolate('time').reset_index()

series = TimeSeries.from_dataframe(df, time_col='time', value_cols=['values'])

model = TransformerModel(input_chunk_length=2, output_chunk_length=1)
model.fit(series)
```

**Expected behavior**
The model should train on short time series.

 - Python version: 3.8.8
 - darts version: 0.6.1

piaz97:
Fixes #312 .

### Summary

The bug was caused while training with a `batch_size` bigger than the available number of samples. Since we used torch `DataLoader` with `drop_last=True`, we were not training at all in this scenario. Later, while evaluating the loss, we were evaluating `training_loss = total_loss / len(train_loader)`, causing the 0 division error. Setting `drop_last=False`, which creates always at least one batch, solves this issue.

### Other Information

After correcting the bug, the test darts.tests.test_regression_ensemble_model.RegressionEnsembleModelsTestCase (test_torch_models_retrain) was failing. Diving into the code, this problem is caused while comparing predictions obtained in the two following ways:
1. we train a model and **predict**;
2. train the model on something else, reset the model calling the method `untrained_model()`; repeat step 1 obtaining other (different) **predictions**. We are currently doing this step 2. in `regression_ensemble_model.py`. 

The test fails since these two different predictions are even more different in magnitude than before. Notice that we obtain the test failure also with the old usage of DataLoader, simply changing the random seed used during the model instantiation. Thus, is not directly connected with this bug fix. The problem seems to be related to the additional steps done in 2., which results in slightly different predictions (even if we are starting with the same `random_state`).


pennfranc:
@piaz97 Thanks for fixing this and for the detailed explanations! Do you think it might be worth adding a test that covers this specific case? Also, I suppose we could investigate why it is that predictions obtained in the two ways you describe are different, and whether it would be easy to fix that.

pennfranc (approved):
Thanks for adding the test!

hrzn (approved):
Thanks!

PR #361, opened 2021-06-29
Feat/DeepAR

pennfranc:
Adds the option to make `RNNModel` probabilistic by setting the `__init__` argument `probabilistic` to `True`, which makes it almost equivalent to the DeepAR model described here: https://arxiv.org/pdf/1704.04110.pdf.

To achieve this I had to:
- move the logic of producing model outputs for training and prediction into two separate functions that can be overridden
- move the logic of computing the loss into a separate function that can be overridden
- let `RNNModel` override all of the above functions
- add a new factory function (`TimeSeries.from_times_and_values_probabilistic`) to create probabilistic time series more easily from numpy arrays
- introduce `num_samples` parameter to `TorchForecastingModel.predict` and adapt method to support probabilistic forecasts
- introduce a `LikelihoodModel` class to abstract different learnt probability distributions

Checkout DeepAR in action here: https://github.com/unit8co/darts/blob/feat/DeepAR/examples/09-DeepAR-examples.ipynb


darts/timeseries.py
```diff
@@ -345,6 +345,45 @@ def from_times_and_values(times: Optional[Union[pd.DatetimeIndex, pd.RangeIndex]
                                          fill_missing_dates=fill_missing_dates,
                                          freq=freq)
 
+    @staticmethod
+    def from_times_and_values_probabilistic(times: Optional[Union[pd.DatetimeIndex, pd.RangeIndex]],
```
hrzn, line 349: Actually could we re-use `from_times_and_values()` and have it behave differently depending whether the dimensionality of `values` is 3 (like here), 2 (like the current `from_times_and_values() - which we have no reason to keep dependent on DataFrame), or 1 (in which case we assume it's a deterministic series with 1 component).

hrzn:
The approach looks very good! Great job @pennfranc 👍

darts/timeseries.py
pennfranc: Good point, done! Doing this I was also able to simplify the `TimeSeries.from_values()` method. Let me know if the changes make sense

darts/models/torch_forecasting_model.py
```diff
@@ -205,6 +206,12 @@ def __init__(self,
         self.lr_scheduler_cls = lr_scheduler_cls
         self.lr_scheduler_kwargs = dict() if lr_scheduler_kwargs is None else lr_scheduler_kwargs
 
+        # by default models are block models (i.e. not recurrent)
+        self.is_recurrent = False
+
+        # by default models are deterministic (i.e. not probabilistic)
+        self.is_probabilistic = False
```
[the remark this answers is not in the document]
pennfranc: Sounds good but I think we still need an attribute, because the RNN could be probabilistic or not, so I guess we can return the attribute in the abstract method.. would that make sense?

darts/models/torch_forecasting_model.py
```diff
@@ -393,21 +403,31 @@ def predict(self,
                 batch_size: Optional[int] = None,
                 verbose: bool = False,
                 n_jobs: int = 1,
-                roll_size: Optional[int] = None
+                roll_size: Optional[int] = None,
+                num_samples: int = 1,
```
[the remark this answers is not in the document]
pennfranc: Our 2 efforts have now combined! Haha

darts/models/torch_forecasting_model.py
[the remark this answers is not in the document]
pennfranc: Update: I have now added the first version of the likelihood model abstraction (`darts.utils.likelihood_model.LikelihoodModel`), and so I am now using the `self.likelihood_model` attribute instead of `self.is_probabilistic`. Of course we could still create a `is_probabilistic()` method which returns False if `self.likelihood_model==None` and True otherwise. What do you think?

darts/timeseries.py
```diff
@@ -307,47 +307,56 @@ def from_series(pd_series: pd.Series,
...
             A TimeSeries constructed from the inputs.
         """
 
-        # TODO: make it more general, not always going via a DataFrame
         raise_if_not(isinstance(times, pd.RangeIndex) or isinstance(times, pd.DatetimeIndex),
                      'the `times` argument must be a RangeIndex or a DateTimeIndex. Use '
                      'TimeSeries.from_values() if you want to use an automatic RangeIndex.')
-        df = pd.DataFrame(values, index=times)
+
+        if not times.name:
```
hrzn, line 342: Sound logic, however, are you maybe mutating the time index provided in argument here?
You could maybe simply store a `time_name` variable here.

darts/timeseries.py
```diff
@@ -307,47 +307,56 @@ def from_series(pd_series: pd.Series,
...
-            df.columns = columns
+            coords[DIMS[1]] = columns
 
-        return TimeSeries.from_dataframe(df,
-                                         time_col=None,
-                                         value_cols=None,
-                                         fill_missing_dates=fill_missing_dates,
-                                         freq=freq)
+        xa = xr.DataArray(values,
+                          dims=(times.name,) + DIMS[-2:],
```
hrzn, line 356: So here you'd be using `time_name` instead of `times.name`

darts/timeseries.py
```diff
@@ -357,22 +366,12 @@ def from_values(values: np.ndarray,
...
-            xa = xr.DataArray(values,
-                              dims=DIMS,
-                              coords={DIMS[0]: time_index, DIMS[1]: columns})
-            return TimeSeries.from_xarray(xa, fill_missing_dates=fill_missing_dates, freq=freq)
-        else:
-            raise_log(ValueError(
-                      'The dimensionality of the provided array must be 2 or 3 (found: {})'.format(len(values.shape))),
-                      logger)
+        values_ = np.reshape(values, (len(values), 1)) if len(values.shape) == 1 else values
+        return TimeSeries.from_times_and_values(times=time_index,
```
hrzn, line 370: This is so much better like that! 👍

darts/utils/likelihood_models.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self):
+        """
+        Abstract class for a `Likelihood` model. It contains all the logic to compute the loss
+        and to sample the distribution, given the parameters of the distribution
+        """
+        pass
+
+    @abstractmethod
+    def _compute_loss(self, output: torch.Tensor) -> torch.Tensor:
```
hrzn, line 24: I think you inverted the parameters of `_compute_loss()` and `_sample()`

darts/utils/likelihood_models.py
```diff
@@ -0,0 +1,59 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Sequence, Optional, Tuple
+import torch
+import torch.nn as nn
+
+from ..timeseries import TimeSeries
+
+
+class LikelihoodModel(ABC):
```
hrzn, line 14: This is perfect, excellent way to represent probabilistic likelihoods. Great job!

darts/models/rnn_model.py
```diff
@@ -67,27 +72,32 @@ def __init__(self,
...
         self.target_size = target_size
+        self.output_size = (
+            likelihood_model._num_parameters * target_size if likelihood_model is not None else target_size
+        )
         self.name = name
+        self.likelihood_model = likelihood_model
 
         # Defining the RNN module
         self.rnn = getattr(nn, name)(input_size, hidden_dim, 1, batch_first=True, dropout=dropout)
+        # TODO: dropout is not used currently since the pytorch module doesn't adds it to the last layer and we have
```
hrzn, line 83: Should we make `num_layers` a parameter?

darts/models/rnn_model.py
```diff
@@ -139,6 +150,9 @@ def __init__(self,
             training. Generally speaking, `training_length` should have a higher value than `input_chunk_length`
             because otherwise during training the RNN is never run for as many iterations as it will during
             training. For more information on this parameter, please see `darts.utils.data.ShiftedDataset`
+        likelihood_model
+            Optionally, the likelihood model to be used for probabilistic forecasts.
```
hrzn, line 154: Maybe you could say here it expects a `darts.utils.LikelihoodModel` so that users reading the docstring know where to look to instantiate their likelihood.

hrzn (approved):
Super good stuff!

darts/models/rnn_model.py
pennfranc: Yes, actually I already did that and it's merged into develop. I'm just waiting for your timeseries refactoring to be merged into develop before merging this branch with the newest version of develop. I'm hoping to avoid some annoying merge conflicts that way

darts/utils/likelihood_models.py
```diff
@@ -0,0 +1,71 @@
...
+        Computes a loss from a model `output`, which represents the parameters of a given probability
+        distribution for every ground truth value in `target`, and the `target` itself.
+        """
+        pass
+
+    @abstractmethod
+    def _sample(self, output: torch.Tensor) -> torch.Tensor:
+        """
+        Samples a prediction from the probability distributions defined by the specific likelihood model
+        and the parameters given in `output`.
```
hrzn, line 35: It's slightly weird to name it `output`, because even though most of the time it's the output of a network, from the point of view of this module what's received here is more of an input. But this is really a detail, no need to change (unless you really want to).

pennfranc: True. How about `model_output` or `likelihood_parameters`?

darts/models/torch_forecasting_model.py
```diff
@@ -571,20 +587,25 @@ def predict_from_dataset(self,
                 input_series = batch_tuple[0]
                 cov_future = batch_tuple[1] if len(batch_tuple) == 3 else None
 
-                if self.is_recurrent:
-                    batch_prediction = self._predict_batch_recurrent_model(n, input_series, cov_future)
-                else:
-                    batch_prediction = self._predict_batch_block_model(n, input_series, cov_future, roll_size)
+                # repeat prediction procedure for every needed sample
+                batch_predictions = []
+                for i in range(num_samples):
```
hrzn, line 595: Actually, isn't there a way to "tile" the `input_series` and `cov_future`, in a way that would allow only a simple call to `predict_*_model()` using batching? That would make things faster.

darts/models/rnn_model.py
```diff
@@ -27,8 +28,10 @@ def __init__(self,
                  name: str,
                  input_size: int,
                  hidden_dim: int,
+                 num_layers: int,
                  target_size: int = 1,
-                 dropout: float = 0.):
+                 dropout: float = 0.,
+                 likelihood_model: Optional[LikelihoodModel] = None):
```
hrzn, line 34: I think this could be named simply `likelihood`

darts/utils/likelihood_models.py
```diff
@@ -0,0 +1,71 @@
...
+        Returns the number of parameters that define the probability distribution for one single
+        target value.
+        """
+        pass
+
+
+class GaussianLikelihoodModel(LikelihoodModel):
+
+    def __init__(self):
+        self.loss = nn.GaussianNLLLoss(reduction='sum')
```
hrzn, line 52: Why `sum` and not `mean` here? By default for the other deterministic models we are using `MSELoss()` which defaults to a `mean` reduction.

pennfranc: No good reason, I changed it

darts/utils/likelihood_models.py
hrzn: model_output is good 👍

PR #363, opened 2021-06-30
Feat/reorganized historical forecasting code

pennfranc:
I was trying to extend `historical_forecast` to support a new `num_samples` parameter in another branch, but the way the code was organized made that pretty hard. In this PR I tried to simplify things by using a more object-oriented approach. All in all I was able to reduce the lines of code used.
Also, I made some tests deterministic that were probabilistic before.


darts/models/forecasting_model.py
```diff
@@ -500,21 +478,14 @@ def gridsearch(model_class,
         # compute all hyperparameter combinations from selection
         params_cross_product = list(product(*parameters.values()))
 
-        # TODO: We should find a better object oriented way of handling covariates in GlobalForecastingModel
```
pennfranc, line 503: The idea was already here

hrzn: It's a great sign when some refactoring happens to be addressing some old TODOs... :D

hrzn (approved):
Much cleaner 👍

PR #370, opened 2021-07-06
Fix import preventing BlockRNN tests from running

tomasvanpottelbergh:
Fixes skipping of BlockRNN tests (see test logs).


pennfranc (approved):
Good catch, thanks!

Issue #366, opened 2021-07-04
Adding DeepTCN model


vpozdnyakov:
This issue is created for my future PR.

It would be great to add the probabilistic DeepTCN model described in https://arxiv.org/pdf/1906.04397.pdf

Mention #288

pennfranc:
Sounds great! Please take a look at the recent version of `RNNModel` in the develop branch to check out how we implemented our first probabilistic deep learning model. If possible, it would be cool if you could follow the same idea of overriding the `_compute_loss` and `_produce_predict_output` methods while utilizing the new `LikelihoodModel` class from `utils.likelihood_models`.

hrzn:
I would also add that you can propose a change to our current `TCNModel` if you judge it can be reused.
(Don't forget to branch off `develop` and not `master`).
Thanks!

pennfranc:
I merged merged your PR into develop, DeepTCN is going to be part of our next release :)

PR #372, opened 2021-07-06
Feat/DeepTCN

vpozdnyakov:
Here is my implementation of DeepTCN model that I mentioned in #366.

The main difference from the original DeepTCN is that there is no a decoder part. The decoder passes future covariates through MLP and blend it with output TCN sequence. In this case, `forward` should take input sequence, past covariates, future covariates. I'm not sure that it can be implemented without changes in base models, such as TorchForecastingModel. So I believe this implementation can be used without the decoder part. May it would be better to implement the decoder as a part of TorchForecastingModel in a separate PR?

Also there is an example ipynb, that was copy-pasted from the DeepAR notebook. Is it ok?

Also I'm not sure about tests. Could you suggest a list of tests that should be added?


pennfranc:
Cool stuff! I think this is already valuable, even without the decoder stuff. You are right that there would be some fundamental changes to be made in the parent class to support the decoder. I think it would be wise to plan this out on a high level first, without jumping straight into the implementation. So yeah I agree with your suggestion to keep this one as a standalone version , and tackle the decoder in a separate PR.. Copy-pasting the notebook is completely fine, as long as all the relevant parts are adapted to the TCN.

Also, I see that many of the changes here mirror the changes we had to make to the `RNNModel` class. So what do you think about moving some of the new methods such as `_compute_loss` and the calculation of `output_size` to the superclass? Of course it has to be done in a way that does not disrupt the non-probabilistic models, but that doesn't seem too hard. 

In terms of tests, we are also planning to add some for DeepAR soon, which we can probably mostly replicate for DeepTCN. In the meantime. In the meantime you can take a look at existing test cases for our neural network models. Before you write tests, make sure to merge with #371, to make sure tests are repeatable. This should be merged into develop very soon.

Thanks a lot for your contributions!

darts/models/torch_forecasting_model.py
```diff
@@ -937,3 +938,21 @@ def _get_best_torch_device(self):
     def _get_learning_rate(self):
         for p in self.optimizer.param_groups:
             return p['lr']
+
+class TorchParametricProbabilisticForecastingModel(TorchForecastingModel, ABC):
```
hrzn, line 942: This seems like a good idea - could you also maybe adapt RNNModel so it uses it as well?

vpozdnyakov:
@pennfranc your suggestion to create a superclass is very tempting! It can be a simple way to convert any neural network model into probabilistic model. Here is my draft of such superclass that is showed in TCN as an example. 

`TorchParametricProbabilisticForecastingModel` is a new class that describes parametric probabilistic models. "Parametric" means that these models are based on some predefined parametric distribution, say Gaussian. There can also be non-parametric models such as GAN, Normalizing flow, VAE and so on. The inheritance is described as follows
`TorchForecastingModel -> TorchParametricProbabilisticForecastingModel -> TCNModel`

I think that all probabilistic stuff can be defined in the high-level `TCNModel` without any modifications in the low-level `_TCNModule`. It reduces the number of modifications to convert actual model into probabilistic model.

But I don't understand where I can calculate the `output_dim`. I can't do it in `TorchParametricProbabilisticForecastingModel` since the `output_dim` is passed from `TorchForecastingModel` directly to `TCNModel` via `_create_model` method. So now it is calculated in `TCNModel`. From my point of view, there is no way to move the `output_dim` calculation to `TorchParametricProbabilisticForecastingModel` without changes in the `TorchForecastingModel`. Could you please tell me how it can be done?

darts/models/torch_forecasting_model.py
vpozdnyakov: Yes, I can, but firstly I want to figure out the problem with `output_dim` calculation that I mentioned in the PR comments.

pennfranc:
@vpozdnyakov I like the new class! And I see the problem with simply moving the calculation of `target_size` to the superclass. I think it's not super important, it should be fine to leave it like this for now.

vpozdnyakov:
I've replicated the changes in RNN model and add docstring for the superclass. I think that there is no need to add new tests except of those that can be added together with DeepAR tests.

pennfranc (approved):

hrzn:
Many thanks @vpozdnyakov !

PR #373, opened 2021-07-07
Feat/probabilistic statsmodels

hrzn:
* Introduces probabilistic forecasts for ARIMA and ExponentialSmoothing
* Adapt the metrics so that, if any of the TimeSeries is stochastic, the metric is computed on its median sample value. Before it was "silently" computing the metric on sample 0. Later, we can introduce a new feature to let users select how they want to aggregate samples for metrics computations if/when some aggregation is required.


darts/models/forecasting_model.py
```diff
@@ -554,7 +554,8 @@ def residuals(self,
         training series length required by the model and the gap introduced by `forecast_horizon`.
         Most commonly, unless otherwise specified, the term "residuals" implies a value for `forecast_horizon` of 1.
 
-        This method works only on univariate series and does not currently support covariates.
+        This method works only on univariate series and does not currently support covariates. It uses the median
+        prediction (when dealing stochastic forecasts having num_samples > 1).
```
pennfranc, line 558:
```suggestion
        prediction (when dealing with stochastic forecasts having num_samples > 1).
```

pennfranc:
Looks really cool! I have 2 remarks:
1. We need to either make the default value for `num_samples` equal to 1 for ARIMA, or we have to adapt the gridsearch backtest such that it is deterministic. Because right now this stochasticity is the reason for why the tests fail, as far as I can see.

2. Since you added `num_samples` to `ARIMA` and `ExponentialSmoothing`, we also have to add it to `ForecastingModel`. Because right now it is introduced at the level of `GlobalForecastingModel`. To do this, we need to do 3 things specifically:
    - have to move the is_probabilistic method from GlobalForecastingModel to ForecastingModel
    - we also have to move the `num_samples` check from the predict method of GlobalForecastingModel to the predict method of `ForecastingModel`
    - we have to introduce `num_samples` to the predict signature of all forecasting models, so that their signature does not differ from their parent class. If they are not probabilistic we will simply pass the `num_samples` parameter to the superclass constructor and do nothing else with it.
    
 Does that make sense?

darts/tests/test_backtesting.py
```diff
@@ -206,8 +208,6 @@ def test_backtest_regression(self):
         self.assertGreater(score, 0.9)
 
     def test_gridsearch(self):
-
-        np.random.seed(1)
```
pennfranc, line 210: Could it be that removing this makes the creation of the random walk time series below non-deterministic?

hrzn:
@pennfranc Makes total sense (I underestimated a bit what it would take to change the tests, so reverted `num_samples=1` by default).
I also agree with your suggestion to put `num_samples` in `ForecastingModel`; this is done.

darts/models/ensemble_model.py
```diff
@@ -42,7 +42,9 @@ def fit(self, series: TimeSeries) -> None:
         """
         super().fit(series)
 
-    def predict(self, n: int) -> TimeSeries:
+    def predict(self,
+                n: int,
+                num_samples: int = 1) -> TimeSeries:
         super().predict(n)
```
pennfranc, line 48:
```suggestion
        super().predict(n, num_samples)
```

darts/models/theta.py
```diff
@@ -296,7 +296,9 @@ def fit(self, series):
         if self.normalization:
             self.fitted_values *= self.mean
 
-    def predict(self, n: int) -> 'TimeSeries':
+    def predict(self,
+                n: int,
+                num_samples: int = 1) -> 'TimeSeries':
         super().predict(n)
```
pennfranc, line 302:
```suggestion
        super().predict(n, num_samples)
```

darts/models/fft.py
```diff
@@ -275,7 +275,9 @@ def fit(self, series: TimeSeries):
         # precompute all possible predicted values using inverse dft
         self.predicted_values = np.fft.ifft(self.fft_values_filtered).real
 
-    def predict(self, n: int):
+    def predict(self,
+                n: int,
+                num_samples: int = 1):
         super().predict(n)
```
pennfranc, line 281:
```suggestion
        super().predict(n, num_samples)
```

darts/models/exponential_smoothing.py
```diff
@@ -77,11 +78,19 @@ def fit(self, series: TimeSeries):
         hw_results = hw_model.fit(**self.fit_kwargs)
         self.model = hw_results
 
-    def predict(self, n):
+    def predict(self, n, num_samples=1):
         super().predict(n)
```
pennfranc, line 82:
```suggestion
        super().predict(n, num_samples)
```

pennfranc:
@hrzn There are some instances where `num_samples` isn't passed to `super().__init__()`, and the `Theta` and `NaiveSeasonal ` models don't yet accept the `num_samples` parameter in their predict function. But other then that I think things are looking good!

hrzn:
Thanks a lot @pennfranc I think this time we should be good (unless I missed something :D)

pennfranc (approved):
Looks good!

PR #380, opened 2021-07-12
Fix/requirements caching

ymaktepi:
This PR goes hand in hand with `feat/relax-requirements`.

Two changes:
the CI will install the latest versions of all the dependencies (via `pip-compile`) and run the tests

the CI also creates a `requirements-latest.txt`  file that will be used for caching the pip packages in order to speed up the testing process.


tomasvanpottelbergh:
Looks good! I imagine Python 3.8 has been removed to speed things up? I would potentially leave it in the matrix to catch strange dependency conflicts. What do others think?

.github/workflows/merge.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9]
-        flavour: ['core', 'all', 'pmdarima', 'torch', 'fbprophet']
+        python-version: [3.7, 3.9]
```
hrzn, line 15:
```suggestion
        python-version: [3.7, 3.8, 3.9]
```

^ restoring 3.8 for merge workflow, as discussed

hrzn (approved):
LGTM! If I understand well, this does not include committing `requirements-latest.txt`, right? I guess it might be nice to have a followup later on to do this.

.github/workflows/merge.yml
tomasvanpottelbergh: And we could then only test 3.9 for PRs maybe?

.github/workflows/develop.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9]
-        flavour: ['core', 'all']
+        python-version: [3.7, 3.9]
```
tomasvanpottelbergh, line 14:
```suggestion
        python-version: [3.9]
```

tomasvanpottelbergh (approved):

PR #390, opened 2021-07-15
Fix make_dists.sh not working with new build

tomasvanpottelbergh:
The `make_dists.sh` script broke with `build>=0.5.0` because of [this PR](https://github.com/pypa/build/pull/304).

This is fixed by including the requirements folder via `MANIFEST.in`. Also, a `pyproject.toml` file has been added for build consistency.


hrzn (approved):
Thanks! Does this change the way to build the package, or can we still simply use make_dists.sh ?

tomasvanpottelbergh:
This hasn't changed, just made sure make_dists.sh works again

PR #392, opened 2021-07-15
Fix some documentation issues

tomasvanpottelbergh:
- Fix rendering of table in README
- Switch from m2r to m2r2
- Add new examples to docs


hrzn (approved):
👍 thanks

PR #402, opened 2021-07-17
Feat/read from csv

xfiderek:
### Summary
* Added static method for creating Multivariate Timeseries from csv file. The method simply calls `from_dataframe` under the hood
* Mentioned feature in the first notebook


darts/timeseries.py
```diff
@@ -220,6 +220,50 @@ def _clean_component_list(columns) -> List[str]:
...
+    @staticmethod
+    def from_csv(filepath_or_buffer: pd._typing.FilePathOrBuffer,
+                 time_col: Optional[str] = None,
+                 value_cols: Optional[Union[List[str], str]] = None,
+                 fill_missing_dates: Optional[bool] = True,
+                 freq: Optional[str] = None,
+                 **kwargs,) -> 'TimeSeries':
+        """
+        Returns a deterministic TimeSeries instance built from a single csv file.
+        One column (or the DataFrame index) has to represent the time,
```
hrzn, line 232:
```suggestion
        One column can be used to represent the time (if not present, the time index will be an Int64Index),
```

darts/timeseries.py
```diff
@@ -220,6 +220,50 @@ def _clean_component_list(columns) -> List[str]:
...
+        and a list of columns `value_cols` has to represent the values for this time series.
```
hrzn, line 233:
```suggestion
        and a list of columns `value_cols` can be used to indicate the values for this time series.
```

hrzn (commented):
Thanks @xfiderek ! A few small suggested changes. The most important would be to set `fill_missing_dates` to `False` by default.

xfiderek:
Thanks @hrzn . I was unlucky enough to create this PR right before you made those changes :D. I will adjust the method

hrzn (approved):
Thanks!

PR #405, opened 2021-07-20
Fix/docker build

tomasvanpottelbergh:
The Docker build on release was failing because of a pystan>=3.

This is fixed by removing the manual `pip install pystan` command. At the same time, the Dockerfile was cleaned up and updated.


hrzn (approved):
Thanks 👍

PR #406, opened 2021-07-22
Fix/col bug data transformer

dumjax:
Fixes #

- Transformer scaler fix to keep column names
- new method with_column_renamed in TimeSeries

**Summary**
After transforming data with Scaler, the columns were disappearing and replaced by integer.

This Scaler is the only method to my knowledge using the method `TimeSeries.from_times_and_values` were the `columns` field should be used (with another function in `map` which is currently not used, with a forced flag of `auto`), and it wasn't.

A test has also been added to check if we keep the columns.

On top of that, an optional `column_name` has been added to all timeseries generation functions


darts/tests/dataprocessing/transformers/test_data_transformer.py
```diff
@@ -27,6 +29,9 @@ def test_scaling(self):
         series1_tr2 = transformer2.fit_transform(self.series1)
         series3_tr2 = transformer2.transform(self.series3)
 
+        # should keep columns
+        self.assertEqual(self.col_1, series1_tr1.columns)
```
pennfranc, line 33: Would this have captured the error from before? I'm just asking because as far as I know, integer columns are the default, so if scaler replaced the original column names with integer column names here we would not notice, right?

dumjax: Indeed you are right, I should set names different from integer first.

hrzn (commented):
Looks good! Actually I wonder if we couldn't also take the chance to improve `ForecastingModel._build_forecast_series()` which also drops the columns names, wdyt?

dumjax:
> Looks good! Actually I wonder if we couldn't also take the chance to improve `ForecastingModel._build_forecast_series()` which also drops the columns names, wdyt?

Sure, I can try

dumjax:
> Looks good! Actually I wonder if we couldn't also take the chance to improve `ForecastingModel._build_forecast_series()` which also drops the columns names, wdyt?

I added it

darts/utils/timeseries_generation.py
```diff
@@ -203,13 +208,14 @@ def random_walk_timeseries(length: int = 10,
...
-    return TimeSeries.from_times_and_values(times, values, freq=freq)
+    return TimeSeries.from_times_and_values(times, values, freq=freq, columns=pd.Index([column_name]))
 
 
 def holidays_timeseries(time_index: pd.DatetimeIndex,
                         country_code: str,
                         prov: str = None,
-                        state: str = None) -> TimeSeries:
+                        state: str = None,
+                        column_name: Optional[str] = '0') -> TimeSeries:
```
hrzn, line 218: [nitpick]: actually "holidays" might be a better name here?

hrzn (approved):

darts/utils/timeseries_generation.py
dumjax: Actually more generally we could create s specific name for each of these generations ('sine', 'constant' etc). What do you think?

hrzn: Good idea!

pennfranc: Yep makes sense! A lot of tests and notebooks will have to be changed though, just as a heads-up

dumjax: Yes, and in the tests, we have things such as 

series = linear_timeseries(start_value=1, length=12, freq='MS', start_ts=pd.Timestamp('2000-01-01'), end_value=12) 
zeroes = constant_timeseries(value=0.0, length=12, freq='MS', start_ts=pd.Timestamp('2000-01-01'))

def function(ts, x):
     return x - ts.month

new_series = series.map(function)
self.assertEqual(new_series, zeroes)` 

the assertEqual here doesnt work anymore, as the column names are different ('const' and 'linear'). It seems important to be able to quickly change names of the columns, by adding a .withColumnName property to the TImeSeries object to create a new timeserie with column name changed (immutability) so I can change the name of the column before testing the assertEqual.
wdyt?

dumjax: It would look like :

my_ts = my_ts.withColumnRenamed('old_col', 'new_col')

exactly like pyspark

darts/tests/dataprocessing/transformers/test_mappers.py
```diff
@@ -65,6 +66,17 @@ def test_mapper_with_timestamp(self):
 
         for to_transform, expected_output in test_cases:
             transformed = self.subtract_month.transform(to_transform)
+            if isinstance(to_transform, list):
+                tmp_t = []
```
hrzn, line 70: It looks like `tmp_t == transformed` no? And also that you could build `expected_output` with something like
```
expected_output = [o.with_column_renamed(o.components[0], t.components[0]) for (o, t) in zip(transformed, expected_output)]
```

darts/tests/test_backtesting.py
```diff
@@ -162,11 +162,19 @@ def test_backtest_regression(self):
...
-            {"0": "Value0", "1": "Value1", "2": "Value2"}, axis=1)
+        features = (
+            features
+            .with_column_renamed(features.components[0], "Value0")
+            .with_column_renamed(features.components[1], "Value1")
         )
 
+        features_multivariate = (
+            features_multivariate
+            .with_column_renamed(features_multivariate.components[0], "Value0")
```
hrzn, line 173: Would be cleaner to be able to call
`with_columns_renamed(features_multivariate.components, ["Value0", "Value1", "Value2"])` no? Could also be more efficient because every call copies the data over (as a new TimeSeries creation).

darts/timeseries.py
```diff
@@ -14,6 +14,7 @@
 from inspect import signature
 from collections import defaultdict
 from pandas.tseries.frequencies import to_offset
+from xarray.core import dataarray
```
hrzn, line 17: Is this needed?

darts/timeseries.py
```diff
@@ -1803,6 +1804,26 @@ def plot(self,
         plt.legend()
         plt.title(self._xa.name);
 
+    def with_column_renamed(self, col_name: str, col_name_new: str) -> 'TimeSeries':
```
hrzn, line 1807: Would be nice to accept either single strings, or a sequence of strings here for renaming several columns at once.

dumjax: I wanted to follow pyspark example, where you can't change more than one column but you do  anyway a chain of call like df = (
              df
              .with_column_renamed(a,b)
              .with_column_renamed(c,d)
              .other functions
              ...
         )

but I don't need to follow it stricto sensus.

darts/timeseries.py
dumjax: no, forgot to get rid of it

darts/tests/test_backtesting.py
dumjax: see my comment below, I can change it

darts/tests/dataprocessing/transformers/test_mappers.py
dumjax: yes indeed, I'll correct it

darts/timeseries.py
hrzn: OK, fair enough, let's leave it like that (at least for now) then :)

dumjax: I actually changed it, we can use lists of columns

darts/tests/test_backtesting.py
```diff
@@ -162,9 +162,14 @@ def test_backtest_regression(self):
         features_multivariate = (gaussian_series + sine_series).stack(gaussian_series).stack(sine_series)
         target = sine_series
 
-        features = TimeSeries.from_dataframe(features.pd_dataframe().rename({"0": "Value0", "1": "Value1"}, axis=1))
-        features_multivariate = TimeSeries.from_dataframe(features_multivariate.pd_dataframe().rename(
-            {"0": "Value0", "1": "Value1", "2": "Value2"}, axis=1)
+        features = (
```
pennfranc, line 165: Much nicer!

darts/timeseries.py
```diff
@@ -1803,6 +1803,43 @@ def plot(self,
         plt.legend()
         plt.title(self._xa.name);
 
+    def with_columns_renamed(self, cols_name: Union[List[str], str], cols_name_new: Union[List[str], str]) -> 'TimeSeries':
+        """
+        Changes a ts column name and returns a new TimeSeries instance
+        -------
```
pennfranc, line 1809: The docstring doesn't quite match the style we use. It needs a 'Parameters' and a 'Returns' section with appropriate parameter/return value descriptions after the general function description (see other docstrings)

darts/timeseries.py
```diff
@@ -1803,6 +1803,43 @@ def plot(self,
         plt.legend()
         plt.title(self._xa.name);
 
+    def with_columns_renamed(self, cols_name: Union[List[str], str], cols_name_new: Union[List[str], str]) -> 'TimeSeries':
```
pennfranc, line 1806: Sorry for being annoying here, but I think `col_names` might be be better than `cols_name`.

darts/timeseries.py
```diff
@@ -1803,6 +1803,43 @@ def plot(self,
...
+                                                    "Some column names in cols_name don't exist in the time series", logger)
+        
+        raise_if_not(len(cols_name) == len(cols_name_new), 'Length of col_name_new list should be'
+                                                    ' equal to length of col_name list', logger)
+
+
+        cols = self.components
+
+        for (o, n) in zip(cols_name, cols_name_new):
+            cols = [n if (c==o) else c for c in cols]
```
pennfranc, line 1829: Doesn't `cols` get overwritten here at every iteration of the loop?

pennfranc (approved):

darts/timeseries.py
pennfranc, line 1809: The docstring doesn't quite match the style we use. It needs a 'Parameters' and a 'Returns' section with appropriate parameter/return value descriptions after the general function description (see other docstrings)

pennfranc (changes requested):

pennfranc (approved):

PR #409, opened 2021-07-23
QOL Future Covariates + Cyclic Encoding

CompilerLuke:
The API to generate future covariates with datetime_attribute_timeseries was more cumbersome than it needed to be,
as it involved generating a pd.date_range manually.
Moreover cyclical encoding can be preferable to one-hot encoding, as it uses less memory.

### Summary

Modifies datetime_attribute_timeseries and holidays_timeseries by adding an until parameter, that lets you extend the time axis into the future. 

Modifies datetime_attribute_timeseries and TimeSeries.add_datetime_attribute by adding a cyclical encoding parameter.


darts/utils/timeseries_generation.py
```diff
@@ -206,10 +206,42 @@ def random_walk_timeseries(length: int = 10,
     return TimeSeries.from_times_and_values(times, values, freq=freq)
 
 
+def _extend_time_index_until(time_index:pd.DatetimeIndex,
+                             until: Union[int, str, pd.Timestamp],
+                             ) -> pd.DatetimeIndex:
+    if until == 0: return time_index
+
+    end = time_index[-1]
+    freq = time_index.freq
```
hrzn, line 215: If instead you take the `freq` attribute of a `TimeSeries`, it will be `1` if the series is integer-indexed (i.e., indexed with a `pandas.Int64Index` instead of a `pandas.DatetimeIndex`). This can be convenient for doing arithmetic on timestamps in a way that is agnostic to whether they are ints or Timestamps (e.g. what you're doing at line 221 would also work on integer-indexed series).

darts/utils/timeseries_generation.py
```diff
@@ -256,6 +294,11 @@ def datetime_attribute_timeseries(time_index: Union[pd.DatetimeIndex, TimeSeries
     one_hot
         Boolean value indicating whether to add the specified attribute as a one hot encoding
         (results in more columns).
+    cyclic
+        Boolean value indicating whether to add the specified attribute as a cyclic encoding.
+        (adds 2 columns, corresponding to sin and cos transformation)
+    until
+        Extend the time_index into the future either by 'int' time points or until timestamp(`str` or `pd.Timestamp`).
```
hrzn, line 301: This could be somewhat misleading if the series is integer-indexed.
In this case the `int` could also be interpreted as the last time step we want data for and the behaviour wouldn't match that of the time-indexed series. You could maybe add two parameters (one of which only could be defined): `until: Option[Union[int, str, pd.Timestamp]]` would always define the last timestamp and `add_length: Option[int]` would define an additional number of steps. WDYT?

hrzn (commented):
Looks good!

darts/utils/timeseries_generation.py
CompilerLuke: Now that I think of it, it makes sense to separate these two behaviours into two different parameters.

hrzn:
@CompilerLuke was there a reason why you chose to close this PR and open a new one?

CompilerLuke:
@hrzn  Sorry there was a mistake on my part with git. I informed @pennfranc of it.

hrzn:
No worries :)

PR #410, opened 2021-07-23
Feat/multivariate ensemble

mkos:
### Summary

Allows using global forecasting models in ensembles. Until now, only "local" forecasting models were allowed to be in an ensemble.

### Other Information

* you still cannot mix global and local models, obviously
* not only `EnsembleModel` was changed, but all exisition subclassess: `NaiveEnsembleModel` and `RegressionEnsembleModel`
* documentation was not yet updated, but if you spot a place where it needs to be, please create an issue
* two tests for `RegressionEnsembleModel` still do not pass (marked with `# TODO FIXIT` comment)
* The `NaiveEnsembleModel`'s `fit()` method's signature was changed to match the one from base class


darts/models/ensemble_model.py
```diff
@@ -24,33 +25,70 @@ class EnsembleModel(ForecastingModel):
...
+            return ts_sequence
 
     def predict(self,
                 n: int,
-                num_samples: int = 1) -> TimeSeries:
-        super().predict(n, num_samples)
+                series: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+                covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+                num_samples: int = 1,
+                ) -> Union[TimeSeries, Sequence[TimeSeries]]:
```
hrzn, line 73: Are there cases where this returns a `Sequence[TimeSeries]` ?

darts/models/ensemble_model.py
```diff
@@ -24,33 +25,70 @@ class EnsembleModel(ForecastingModel):
...
+                self.models[0].predict(n, series, covariates, num_samples))
+        else:
+            predictions = self.models[0].predict(n, num_samples)
 
-        predictions = self.models[0].predict(n, num_samples)
         if len(self.models) > 1:
             for model in self.models[1:]:
-                predictions = predictions.stack(model.predict(n, num_samples))
+                if self.is_global_ensemble:
+                    prediction = self._ts_sequence_to_multivariate_ts(
```
hrzn, line 86: Before we were stacking all the univariate forecasts of one target series into one TimeSeries because that could give us a convenient way to use regression models on this one TimeSeries. However now I think this aggregation should per-target-series and so for N targets you should end up with N predicted TimeSeries, each of which being the output of some ensembling logic, no?

darts/models/ensemble_model.py
mkos: It will return `Sequence[Timeseries]` when user provides multiple timeseries for predicition also in `Sequence[Timeseries]`. For each sample, one prediction will be generated. Hence `Sequence[Timeseries]`.

darts/models/rnn_model.py
```diff
@@ -96,6 +96,9 @@ def forward(self, x, h=None):
 
 
 class RNNModel(TorchParametricProbabilisticForecastingModel, DualCovariatesTorchModel):
+
+    uses_past_covariates = False
```
mkos, line 100: Note to self: search for other models like this. @hrzn is there a fast way to find models, which support both types of covariates in fit(), but they use only one?

hrzn: Here's the list:
* `PastCovariatesTorchModel`: only past
* `FutureCovariatesTorchModel`, `DualCovariatesTorchModel` and `DualCovariatesForecastingModel`: only future
* `MixedCovariatesTorchModel` and `SplitCovariatesTorchModel`: both

hrzn (commented):
Overall I think functionality wise it is correct, and the code is going in the right direction. I have a few comments though. Maybe the most important points (IMO) are:
* There are lots of checks for `is_global` etc which could be avoided by calling `_fit_wrapper()` and `_predict_wrapper`.
* `is_univariate` is the wrong name (single series != univariate), and actually I think we might get away without having to check this condition, just by doing something like `series = [series] if isinstance(series, TimeSeries) else series` upfront. These two points should reduce a lot the code and conditions and hopefully make things a bit clearer.
* We should not, if possible, rely on the stacking of `TimeSeries` to store the outputs of all the models. It was done like that in the past but it's not very clean I think (it's abusing the components dimension of `TimeSeries`). Also, not sure how it would play with stochastic series. I made a proposal in a comment for doing it differently.

darts/models/ensemble_model.py
```diff
@@ -23,50 +24,114 @@ class EnsembleModel(ForecastingModel):
...
+        raise_if(not self.is_global_ensemble and not isinstance(series, TimeSeries),
+                 "All models are of type darts.models.ForecastingModel which do not support covariates.",
+                 logger
+                 )
+        raise_if(not self.is_global_ensemble and past_covariates is not None,
+                 "All models are of type darts.models.ForecastingModel which do not support covariates.",
+                 logger
+                 )
+
+        self.is_univariate = isinstance(series, TimeSeries)
```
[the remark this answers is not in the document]
mkos: no, I think I've messed up naming.

darts/models/ensemble_model.py
```diff
@@ -23,50 +24,114 @@ class EnsembleModel(ForecastingModel):
...
+        if past_covariates is not None:
+            self.is_univariate_covariate = isinstance(past_covariates, TimeSeries)
+
+        raise_if(past_covariates is not None and (self.is_univariate != self.is_univariate_covariate),
```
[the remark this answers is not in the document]
mkos: No, I think the issue is with naming again, it is meant to be `is_single_series` and `is_single_series_covariate` and I think it makes sense, because if we provide single series as timeseries and multiple series as covariates that's wrong as well as opposite situation, right?

darts/models/ensemble_model.py
```diff
@@ -23,50 +24,114 @@ class EnsembleModel(ForecastingModel):
...
     """
-    def __init__(self, models: List[ForecastingModel]):
+    def __init__(self, models: Union[List[ForecastingModel], List[GlobalForecastingModel]]):
         raise_if_not(isinstance(models, list) and models,
                      "Cannot instantiate EnsembleModel with an empty list of models",
                      logger)
-        raise_if_not(all(isinstance(model, ForecastingModel) for model in models),
-                     "All models must be instances of darts.models.ForecastingModel",
+
+        is_local_ensemble = all(isinstance(model, ForecastingModel) and not isinstance(model, GlobalForecastingModel)
```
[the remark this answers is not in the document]
mkos: Yeah, but it covers a test case that is checking whether there are classes provided instead of objects, see `test_input_models_local_models()` in `test_ensemble_models.py`. Ok?

darts/models/regression_ensemble_model.py
```diff
@@ -58,25 +58,87 @@ def __init__(self,
...
+                past_forecast_covariates = self.past_covariate_series[:-self.train_n_points]
+            else:
+                past_forecast_covariates, _ = \
+                    self._split_multi_ts_sequence(self.train_n_points, past_covariates)
+        else:
+            past_forecast_covariates=None
+
+        if future_covariates is not None:
+            if self.is_univariate:
+                split_point = len(series) - self.train_n_points
```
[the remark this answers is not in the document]
mkos: I wasn't aware of that functionality.

darts/models/ensemble_model.py
```diff
@@ -23,50 +24,114 @@ class EnsembleModel(ForecastingModel):
...
+                        past_covariates=past_covariates, future_covariates=future_covariates, num_samples=num_samples)
+
+        if self.is_global_ensemble and not self.is_univariate:
+            predictions = self.models[0].predict(n=n, series=series,
+                        past_covariates=past_covariates, future_covariates=future_covariates, num_samples=num_samples)
+        else:
+            predictions = self.models[0].predict(n=n, num_samples=num_samples)
 
-        predictions = self.models[0].predict(n=n, num_samples=num_samples)
         if len(self.models) > 1:
```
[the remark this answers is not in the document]
mkos: done.

darts/models/ensemble_model.py
```diff
@@ -23,50 +24,114 @@ class EnsembleModel(ForecastingModel):
...
             for model in self.models[1:]:
-                predictions = predictions.stack(model.predict(n=n, num_samples=num_samples))
+                if self.is_global_ensemble and not self.is_univariate:
+                    prediction = model.predict(n=n, series=series,
+                        past_covariates=past_covariates, future_covariates=future_covariates, num_samples=num_samples)
+                    predictions = self._stack_ts_seq(predictions, prediction)
```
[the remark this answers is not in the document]
mkos: I don't think I understand: Note that I'm not stacking everything into one timeseries, but for each training timeseries I'm building a stack and train regression model with it. So let's assume that we have M models in ensemble and N timeserie samples, than this would return a N-sequence of M-stacked predictions. And that would be used for training the regression model. 

As far as you told me about the single timeserie case, this is the way to do it: use M-stacked predictions as covariates. In this case simply N=1.

Am I missing something?

darts/models/regression_ensemble_model.py
```diff
@@ -58,25 +58,87 @@ def __init__(self,
...
+            series: Union[TimeSeries, Sequence[TimeSeries]],
+            past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+            future_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None
+            ) -> None:
+
+        super().fit(series, past_covariates=past_covariates, future_covariates=future_covariates)
 
         # spare train_n_points points to serve as regression target
-        raise_if(len(self.training_series) <= self.train_n_points,
+        if self.is_univariate:
```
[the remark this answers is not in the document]
mkos: but then I will have to turn it back from `[series]` to `series` at the end, right? Note: currently there are far many checks because the code simplified from other reasons. Let me know if you still want me to work on it.

darts/models/regression_ensemble_model.py
```diff
@@ -89,9 +151,23 @@ def fit(self, series: TimeSeries) -> None:
...
+                    kwargs['future_covariates'] = future_covariates
+                model.fit(**kwargs)
+
+            else:
+                model.fit(self.training_series)
+
+    def ensemble(self,
+                 predictions: Union[TimeSeries, Sequence[TimeSeries]],
+                 series: Optional[Sequence[TimeSeries]] = None) -> Union[TimeSeries, Sequence[TimeSeries]]:
+        if self.is_univariate:
```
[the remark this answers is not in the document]
mkos: Why? Even if I use the trick with converting input to sequence I think I still need that check IMO:
```
        if self.is_single_series:
            predictions = [predictions]
            series = [series]

        ensembled =  [self.regression_model.predict(n=len(prediction), series=serie, future_covariates=prediction)
                      for serie, prediction in zip(series, predictions)]

        return ensembled[0] if self.is_single_series else ensembled
```

darts/models/ensemble_model.py
```diff
@@ -23,50 +24,114 @@ class EnsembleModel(ForecastingModel):
...
+                series: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+                past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+                future_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+                num_samples: int = 1,
+                ) -> Union[TimeSeries, Sequence[TimeSeries]]:
+
+        super().predict(n=n, series=series,
+                        past_covariates=past_covariates, future_covariates=future_covariates, num_samples=num_samples)
+
+        if self.is_global_ensemble and not self.is_univariate:
```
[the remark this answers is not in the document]
mkos: This is no longer a case since I started using `_predict_wrapper()`.

darts/models/rnn_model.py
mkos: cool, thanks!

mkos: done.

darts/tests/test_ensemble_models.py
```diff
@@ -38,6 +51,49 @@ def test_predict_ensemble(self):
...
+    def test_fit_multivar_ts_with_local_models(self):
+        naive = NaiveEnsembleModel([NaiveDrift(), NaiveSeasonal(), Theta(), ExponentialSmoothing()])
+        with self.assertRaises(ValueError):
+            naive.fit(self.seq1)
+
+    def test_fit_univar_ts_with_covariates_for_local_models(self):
+        naive = NaiveEnsembleModel([NaiveDrift(), NaiveSeasonal(), Theta(), ExponentialSmoothing()])
+        with self.assertRaises(ValueError):
+            naive.fit(self.series1, self.series2)
+
```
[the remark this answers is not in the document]
mkos: done.

darts/models/regression_ensemble_model.py
```diff
@@ -58,25 +59,51 @@ def __init__(self,
...
+                                 ) -> Tuple[Sequence[TimeSeries], Sequence[TimeSeries]]:
+        left = [ts[:-n] for ts in ts_sequence]
+        right = [ts[-n:] for ts in ts_sequence]
+        return left, right
+
+    def fit(self,
+            series: Union[TimeSeries, Sequence[TimeSeries]],
+            past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+            future_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+            num_samples: int = 1
```
hrzn, line 72: This should be removed

darts/models/torch_forecasting_model.py
```diff
@@ -1063,6 +1063,9 @@ def _basic_compare_sample(train_sample: Tuple, predict_sample: Tuple):
 
 
 class PastCovariatesTorchModel(TorchForecastingModel, ABC):
+
+    uses_future_covariates = False
```
hrzn, line 1067: Shouldn't this be True?
and is this used?

darts/models/torch_forecasting_model.py
```diff
@@ -1215,6 +1221,9 @@ def _get_batch_prediction(self, n: int, input_batch: Tuple, roll_size: int) -> T
 
 
 class DualCovariatesTorchModel(TorchForecastingModel, ABC):
+
+    uses_past_covariates = False
```
hrzn, line 1225: Same here?

darts/tests/test_regression_ensemble_model.py
```diff
@@ -92,3 +125,84 @@ def test_torch_models_retrain(self):
...
+            ensemble.predict(10, self.seq2, self.cov2)
+
+        def helper_test_models_accuracy(self, model_instance, n, series, past_covariates, min_rmse):
+            # for every model, test whether it predicts the target with a minimum r2 score of `min_rmse`
+            train_f, train_t, test_f, test_t = train_test_split(past_covariates, series, pd.Timestamp("20010101"))
+
+            model_instance.fit(series=train_t, past_covariates=train_f)
+            prediction = model_instance.predict(n=n, past_covariates=past_covariates)
+            current_rmse = rmse(prediction, test_t)
+            print(f"Current RMSE: {current_rmse}")
```
hrzn, line 157: Can be removed

darts/tests/test_regression_ensemble_model.py
```diff
@@ -1,42 +1,75 @@
...
 
 try:
-    from ..models import BlockRNNModel
+    from ..models import RNNModel, BlockRNNModel
     TORCH_AVAILABLE = True
 except ImportError:
     logger.warning("Torch not available. Some tests will be skipped.")
     TORCH_AVAILABLE = False
 
+np.random.seed(42)
```
hrzn, line 26: Try to also add `torch.manual_seed(42)`

hrzn (commented):
This looks a lot better than before :)
We can soon merge it. Just got a couple of minor comments (also a suggestion to make the tests deterministic).

darts/models/ensemble_model.py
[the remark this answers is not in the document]
hrzn: OK, makes sense

darts/models/ensemble_model.py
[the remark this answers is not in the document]
hrzn: No you are right. It was done like that for N=1 already before. I'm just thinking that it's not 100% satisfying because it is abusing the "component" dimension of the TimeSeries IMO. However it is convenient (although works naturally out of the box with RegressionModels, so let's leave it like that (at least for now).

Btw: do you know if it still works if the original series are multivariate?

darts/models/regression_ensemble_model.py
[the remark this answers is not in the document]
hrzn: Yes at the very end you'll need a check like 
```
return results[0] if self._is_single_series else results
```
But I think the way it is now is already OK.

darts/models/regression_ensemble_model.py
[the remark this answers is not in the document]
hrzn: Yes you still need the check at the very end, but it is already better because the logic of calling `predict()` is written only once.

darts/models/ensemble_model.py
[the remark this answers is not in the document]
mkos: > I'm just thinking that it's not 100% satisfying because it is abusing the "component" dimension of the TimeSeries IMO and if you remember, it wasn't obvious for me at the beginning how it works.
> Btw: do you know if it still works if the original series are multivariate?
I never checked, but I don't think it will work tbh, because as you mentioned, we're already 'occupying' `component` dimension for another purpose.

darts/tests/test_regression_ensemble_model.py
mkos: unfortunately, same issue :(

darts/models/torch_forecasting_model.py
mkos: from one of your earlier comments:
> Here's the list:
* `PastCovariatesTorchModel`: only past
* `FutureCovariatesTorchModel`, `DualCovariatesTorchModel` and `DualCovariatesForecastingModel`: only future
* `MixedCovariatesTorchModel` and `SplitCovariatesTorchModel`: both

darts/models/torch_forecasting_model.py
mkos: from one of your earlier comments:
> Here's the list:
> * `PastCovariatesTorchModel`: only past
> * `FutureCovariatesTorchModel`, `DualCovariatesTorchModel` and `DualCovariatesForecastingModel`: only future
> * `MixedCovariatesTorchModel` and `SplitCovariatesTorchModel`: both

it is used [here](https://github.com/unit8co/darts/blob/d7f018904d7e19c510ca414d534322cd5b8fe4ac/darts/models/regression_ensemble_model.py#L119).

darts/models/regression_ensemble_model.py
mkos: `num_samples` only?

darts/models/torch_forecasting_model.py
hrzn: Sorry, I read that too quick, my mistake.

darts/models/torch_forecasting_model.py
hrzn: All good.

darts/models/ensemble_model.py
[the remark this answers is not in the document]
hrzn: We should then at least test then that the target is not multivariate and raise an exception if it is, stating that Ensemble models do not yet support multivariate series.

darts/models/regression_ensemble_model.py
mkos: It is somewhat for ensemble, because you're using `predict()` in the `fit()`, but I get your point. Removing.

hrzn (approved):
Looks good, thanks!

PR #411, opened 2021-07-23
Feature QOL Improvements to Future Covariates and Support for Cyclic Encoding

CompilerLuke:
The API to generate future covariates with datetime_attribute_timeseries was more cumbersome than it needed to be,
as it involved generating a pd.date_range manually.
Moreover cyclical encoding can be preferable to one-hot encoding, as it uses less memory.

### Summary
Modifies datetime_attribute_timeseries and holidays_timeseries by adding an until and add_length parameter. Supports both pd.RangeIndex and pd.DatetimeIndex. The until parameter extends the time axis until the specified timestamp, whereas add_length simply extends the time axis by that amount.
 
Modifies datetime_attribute_timeseries and TimeSeries.add_datetime_attribute by adding a cyclical encoding parameter.


darts/utils/timeseries_generation.py
```diff
@@ -206,10 +206,58 @@ def random_walk_timeseries(length: int = 10,
...
+
+    if not add_length and not until:
+        return time_index
+
+    raise_if(add_length and until, "set only one of add_length and until")
+
+    end = time_index[-1]
+    freq = time_index.freq
+
+    if add_length > 0:
```
pennfranc, line 222: Did you maybe mean to check whether `add_length` is an int here? Because if you check whether it's positive here, the check below will never raise an exception

darts/utils/timeseries_generation.py
```diff
@@ -206,10 +206,58 @@ def random_walk_timeseries(length: int = 10,
...
+        raise_if_not(add_length >= 0, "until, indicating the number of time points to extend into the future by, must be positive")
```
pennfranc, line 223: I guess we don't need this one, since no error will ever be raised due to the if condition before

pennfranc: You could instead add a check before the if condition to make sure no negative value was provided for `add_length`

darts/utils/timeseries_generation.py
```diff
@@ -206,10 +206,58 @@ def random_walk_timeseries(length: int = 10,
...
+        timestamp = pd.Timestamp(until) if datetime_index else until
+
+        raise_if_not(timestamp > end, f"Expected until, {timestamp} to lie past end of time index {end}")
+
+        ahead = timestamp - end
+        raise_if_not((ahead % freq) == pd.Timedelta(0), f"End date must correspond with frequency {freq} of the time axis", logger)
+
+        end = timestamp
+
+    new_time_index = pd.date_range(start= time_index[0], end=end, freq= freq)
```
pennfranc, line 250:
```suggestion
    new_time_index = pd.date_range(start=time_index[0], end=end, freq=freq)
```

darts/utils/timeseries_generation.py
```diff
@@ -278,15 +348,26 @@ def datetime_attribute_timeseries(time_index: Union[pd.DatetimeIndex, TimeSeries
...
+                                                   " with one-hot or cyclical encoding.", logger)
+
+    if one_hot:
         values_df = pd.get_dummies(values)
         # fill missing columns (in case not all values appear in time_index)
         for i in range(1, num_values_dict[attribute] + 1):
             if not (i in values_df.columns):
                 values_df[i] = 0
         values_df = values_df[range(1, num_values_dict[attribute] + 1)]
+    elif cyclic:
```
pennfranc, line 363: I think we might have to give the user a warning if they call the function with `attribute='day'` and `cyclic=True` because the period will be 31 days even though not every month is that long. For all the other attributes that's not an issue.

darts/tests/test_timeseries_generation.py
```diff
@@ -94,3 +94,22 @@ def test_routine(time_index, country_code):
...
                 test_routine(time_index, country_code)
 
+        # test extend time index
+        test_routine(time_index_1, 'US',  add_length= 365)
+        test_routine(time_index_1, 'CH',  until='2016-01-01')
+        test_routine(time_index_1, 'CH',  until='20160101')
+        test_routine(time_index_1, 'AR',  until=pd.Timestamp('2016-01-01'))
+
+        # test overflow
+
```
pennfranc, line 104:
```suggestion
```

darts/utils/timeseries_generation.py
```diff
@@ -256,6 +314,14 @@ def datetime_attribute_timeseries(time_index: Union[pd.DatetimeIndex, TimeSeries
     one_hot
         Boolean value indicating whether to add the specified attribute as a one hot encoding
         (results in more columns).
+    cyclic
```
pennfranc, line 317: Both the one_hot/cyclic pair and the unit/add_length pair are mutually exclusive, i.e. not both of them can be set, and you correctly check for this. Could you also incorporate this requirement into the docstring?

pennfranc:
Looks nice! I just have a few comments. Also, could you update the RNN notebook with this new way of generating future covariates?

darts/tests/test_timeseries_generation.py
```diff
@@ -94,3 +94,22 @@ def test_routine(time_index, country_code):
             for country_code in ['US', 'CH', 'AR']:
                 test_routine(time_index, country_code)
 
+        # test extend time index
+        test_routine(time_index_1, 'US',  add_length= 365)
```
pennfranc, line 98:
```suggestion
        test_routine(time_index_1, 'US',  add_length=365)
```

pennfranc: Ah also, any specific reason for why you put 2 whitespaces between function arguments?

pennfranc (approved):
Looks good to me, @hrzn you also want to take a quick look?

darts/tests/test_timeseries_multivariate.py
```diff
@@ -137,6 +137,24 @@ def test_add_datetime_attribute(self):
...
+        start = times_month[0]
+        end = times_month[-1]
+
+        seriesE = TimeSeries.from_times_and_values(times_month, np.repeat(0.1, len(times_month)))
+        seriesF = seriesE.add_datetime_attribute('day', cyclic=True)
+
+        values_sin = seriesF.values()[:, 1]
+        values_cos = seriesF.values()[:, 2]
+
+        self.assertTrue(np.allclose(np.add(np.square(values_sin), np.square(values_cos)), 1))
```
hrzn, line 151: +1 for sound trigonometry :)

hrzn (approved):
LGTM 👍

PR #414, opened 2021-07-26
Feature/torch batch samples

CompilerLuke:
Previous implementation of the torch_forecasting_model was capable of batching across time-series, but not across samples.

### Summary

Modifies TorchForecastingModel.predict_from_dataset to tile the input series by the sample batch size. Speeds up the typical use-case where you have few time series to predict, but multiple samples.


darts/models/torch_forecasting_model.py
```diff
@@ -616,18 +616,37 @@ def predict_from_dataset(self,
 
                 # repeat prediction procedure for every needed sample
                 batch_predictions = []
-                for i in range(num_samples):
+
+                # if batch_size is not a multiple of input_series, this won't fill the entire batch_size
+                series_length = input_series.shape[0]
+                batch_sample_size = max(batch_size // series_length, 1)
```
pennfranc, line 622: We might have to set `batch_sample_size` equal to 1 explicitly when `num_samples==1`. This is because we could have a `batch_size > 1` to speed up multiple deterministic predictions. But in the last batch, it might be that `series_length < batch_size`. Then we can have a `batch_sample_size` that is bigger than 1, which will lead to tiling, even though we don't want to make a probabilistic forecast

CompilerLuke: I don't believe this occurs, as the loop will check whether sample_length+batch_size > num_samples. So it will never generate more samples than specified. However I added min around batch_sample_size to make it more explicit.

darts/models/torch_forecasting_model.py
```diff
@@ -616,18 +616,43 @@ def predict_from_dataset(self,
...
                     batch_prediction = torch.cat(batch_prediction, dim=1)
+                    out_shape = batch_prediction.shape
+
+                    # drop unnecessary values
                     batch_prediction = batch_prediction[:, :n, :]
-                    batch_prediction = batch_prediction.cpu().detach().numpy()
+
+                    # reshape from 3d tensor (num_series x batch_sample_size, ...)
+                    # into 4d tensor (batch_sample_size, num_series, ...), where dim 0 represents the samples
+                    batch_prediction = batch_prediction.reshape((batch_sample_size, num_series,) + out_shape[1:])
```
pennfranc, line 648: This line is causing some issues now it seems, looking at the tests. I see that you changed the order of the statements in the last commit. Any idea about how to fix this?

darts/models/torch_forecasting_model.py
```diff
@@ -616,18 +616,43 @@ def predict_from_dataset(self,
...
                 # repeat prediction procedure for every needed sample
                 batch_predictions = []
-                for i in range(num_samples):
+
+                # if batch_size is not a multiple of input_series, this won't fill the entire batch_size
+                num_series = input_series.shape[0]
+                batch_sample_size = min(max(batch_size // num_series, 1), num_samples)
+                sample_length = 0
+
+                in_shape = input_series.shape
```
hrzn, line 625: Is this used?

hrzn (commented):
This looks good! It will have to slightly be modified once https://github.com/unit8co/darts/pull/415 is merged, but not much I think.

@CompilerLuke do you have any idea what sort of speedups it produces? It would be interesting to measure the time taken before and after for a few different workloads.

pennfranc:
> This looks good! It will have to slightly be modified once #415 is merged, but not much I think.
> 
> @CompilerLuke do you have any idea what sort of speedups it produces? It would be interesting to measure the time taken before and after for a few different workloads.

Since I'm not sure Lucas will be checking GitHub in the near future, here's a message he sent me:
`The speed up from duplicating timeseries is pretty good, setting the batch_sample_size to 5 decreases probabilistic test execution from 20 to 13s.`

PR #415, opened 2021-07-26
Feat/refactor covariate models

hrzn:
The goal of this PR is to clarify what are past covariates and future covariates in our models supporting covariates.

## Definition:
* Past covariates are covariate time series that are known only in the past at prediction time. Usually things that have to be measure to be known.
* Future covariates are covariate time series that are known in advance (`n` timesteps in advance for a forecast with horizon `n`) at prediction time. Those are things like holidays or weather forecast.

## Why is this important (and challenging)?
It's important to know what series the models are using as past covariates, and what series they are using as future covariates. Not knowing might be confusing when you only specify a `covariates` parameter (e.g. to the `predict()` function). Will the model use future values of this series or will it not? This has to be clear.

To make things more complex, some models might be using only past covariates, but these covariates could, in some cases, be known in advance too. This would allow to obtain forecasts beyond `output_chunk_length`. 

Conversely, some models might be using only *future* values of future covariates, or the history of these future covariates (e.g. imagine a model predicting the flow of a river: it could use weather data of the past few days in addition to weather forecasts of the few next days as inputs for a prediction).

In this PR we address this issue by introducing 5 model types, which define how models use past and/or future covariates. These models will now accept `past_covariates` and/or `future_covariates` parameters, which make it clear what is what.

## Summary of main changes:
* The `fit()` and `predict()` methods of `GlobalForecastingModel` now accept both `past_covariates` and `future_covariates` (instead of simply `covariates`).
* The `TorchForecastingModel` are split into 5 subtypes:
  * `PastCovariatesTorchModel`: using only past covariates - e.g. predicting future river flow using only past-observed rainfalls (in addition to historic flow).
  * `FutureCovariatesTorchModel`: using only future covariates (in the future) - e.g. predicting future river flow using only rainfall forecasts.
  * `DualCovariatesTorchModel`: using only future covariates but both in the past and in the future - e.g. predicting future river flow using both past-observed and forecasted rainfalls.
  * `MixedCovariatesTorchModel`: using both past and future covariates. Future covariates both in past and future. E.g. predicting future river flow using past-observed and forecasted rainfalls, as well as past-observed flow at some earlier point of the river.
  * `SplitCovariatesTorchModel`: using both past and future covariates. Future covariates only in the future. E.g. predicting future river flow using forecasted rainfalls, as well as past-observed flow at some earlier point of the river.
* I have split (and significantly refactored) both the training and the inference datasets to follow these 5 types. So for instance an implementation of `PastCovariatesTrainDataset` contains (past_target, past_covariate, future_target) examples because this is what is needed to train `PastCovariatesTorchModel`.
* The datasets are a bit different than before. The inference datasets return different types than before. Now they extend torch `Dataset`s, which simplifies our codebase.
* I've also made the nomenclature more uniform for our former `ExtendedForecastingModel`s. Those are now `DualCovariatesForecastingModel` and they support only `future_covariates`.
* I have refactored the `predict_from_dataset()` function, extracting things into an abstract `_get_batch_prediction()` function, which is currently only implemented for `PastCovariatesTorchModel`s and `RNNModel` (RNNs need a different logic than block models).
* The covariates will be sliced correctly as long as they contain the required time spans (no need to have same length as target anymore). I think this would make the use of covariates much easier: you can just always provide the whole time series and the library figures out how to slice them for you.
* The models' are dynamically cast to float32 or float64 to follow the dtype of the provided `TimeSeries`. This way training a model on 32-bits data will naturally be faster whereas training on 64-bits data will have potentially higher resolution.

Overall I think it's getting much cleaner than before, but I still have a feeling that types could perhaps be used better... there are lots of types everywhere and perhaps we could use generics in some places (still I would leave that for a future refactoring ;). But at least the user-facing API should be improve.


darts/models/torch_forecasting_model.py
```diff
@@ -582,142 +614,53 @@ def predict_from_dataset(self,
...
         iterator = _build_tqdm_iterator(pred_loader, verbose=verbose)
+
+        self.model.eval()
         with torch.no_grad():
             for batch_tuple in iterator:
 
-                # at this point `input_series` contains both the past target series and past covariates
-                input_series = batch_tuple[0].to(self.device)
-                cov_future = batch_tuple[1] if len(batch_tuple) == 3 else None
+                input_data_tuple, batch_input_series = batch_tuple[:-1], batch_tuple[-1]
```
piaz97, line 644: Here I think we need to move those batches with `.to(self.device)`. Otherwise, we could delegate this responsibility to `self._get_batch_prediction`, but in this last case we would have to remind to move tensors in every implementation of `self._get_batch_prediction` when we are adding a new model.  
```suggestion
                input_data_tuple, batch_input_series = batch_tuple[:-1].to(self.device), batch_tuple[-1].to(self.device)
```

darts/utils/data/sequential_dataset.py
```diff
@@ -33,86 +100,262 @@ def __init__(self,
...
+            If some series turn out to have a length that would allow more than `max_samples_per_ts`, only the
+            most recent `max_samples_per_ts` samples will be considered.
+        """
+        super().__init__()
 
-        # optionally also produce the input covariate
-        input_covariate = None
-        if self.covariates is not None:
-            ts_covariate = self.covariates[ts_idx].values(copy=False)
+        self.ds_past = GenericShiftedDataset(target_series=target_series,
```
pennfranc, line 339: I like this organization of breaking a bigger dataset into 2 smaller ones, seems like a clean approach! But I guess the name `ds_past` isn't pefectly accurate since it includes 'future' information in the form of the future target. Also, it might be a bit arbitrary that the past covariates is included in the dataset with the target, and all other types of covariates require their specialized sub-dataset. But of course I assume that is a product of our existing codebase, and we can reorganize this in later PRs too!

hrzn: @pennfranc I'm not sure to exactly understand what you mean. `ds_past` is named like that simply because it will be the "internal dataset" in charge of serving the past covariates. The past and future target is served both by `ds_past` and `ds_future`, so we make the choice to use the one from `ds_past` and drop those of `ds_future` (naming them `_`). The sole purpose of `ds_future` is to serve `future_covariates`. Behind the scenes the target will be sliced one time too many, but I think it's (hopefully) negligible overhead compared to the code simplicity it brings.
But perhaps I'm missing something in your comment?

darts/models/torch_forecasting_model.py
hrzn: This is now done in the function `TorchForecastingModel._batch_collate_fn()`. For all Numpy arrays returned by the datasets, we stack them into batches, translate to torch Tensors, and finally put on the right device. This function is applied both to training and inference datasets, so I think it's fine like that.

darts/utils/data/sequential_dataset.py
pennfranc: Yeah I think your logic makes sense. I guess the first thought I had when seeing the names was that `ds_past` would supply all past data (both target and covariate), and `ds_future` would provide all future data. What do you think about adding this explanation of the two datasets in the form of comments?

pennfranc (approved):

PR #419, opened 2021-08-03
Feature/probabilistic metric

CompilerLuke:
Current metrics in darts reduce stochastic time series to deterministic time series by taking their median value. 
This ignores the certainty of the time series, which is central to decision-making. 

### Summary

This pull request introduces a new metric `mpgl` short for MEAN PERCENTAGE GAUSSIAN LIKELIHOOD that addresses the above concerns. This new metric fits a Gaussian distribution to stochastic samples of the predicted series and evaluates the likelihood of `actual_series` ± `deviation` occurring in the probabilistic forecast. As such, probabilistic metrics that are too confident in their prediction (small standard deviation), result in a lower likelihood as the prediction band may not contain the ground truth.

Moreover a stoch_reduction parameter has been added to all existing metrics. This will enable the user to select a behaviour other than reducing probabilistic metrics to their median value.


darts/metrics/metrics.py
```diff
@@ -872,9 +933,101 @@ def r2_score(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
...
+                       *,
+                       deviation: float = 1.0,
+                       reduction: Callable[[np.ndarray], float] = np.mean,
+                       inter_reduction: Callable[[np.ndarray], Union[float, np.ndarray]] = lambda x: x,
+                       n_jobs: int = 1,
+                       verbose: bool = False
+):
+    """ MEAN PERCENTAGE GAUSSIAN LIKELIHOOD
+
+    Computes the mean gaussian likelihood of the ground truth series `actual_series` ± deviation,
```
hrzn, line 959: Actually if I understand it well, it computes the mean probability that the ground truth values fall within X +- deviation, where X is the mean of `pred_series`. A few things:
* You should describe it as mean probability (instead of mean likelihood)
* The meaning of `deviation` is a bit weird and left to the users. How about constraining `deviation` to represent a multiple of the standard deviation. This way, for instance for `deviation=2`, we could interpret this metric as "The average probability (over time) that the ground truth lie in the +- 2σ interval defined by the `pred_series`, assuming a Gaussian distribution". `deviation=2` could be a nice default as it corresponds to 95.4% probability.
* You could also optionally avoid using `deviation` (say, if it's None), and in this case compute the mean likelihood (instead of a mean probability). The likelihood has the advantage that it doesn't depend on a (somewhat arbitrary) integration region (given by `deviation`), but has the disadvantage that likelihood values are not bound between 0 and 1 and thus harder to interpret.

darts/metrics/metrics.py
```diff
@@ -872,9 +933,101 @@ def r2_score(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
...
+    isnan_mask = np.logical_or(np.isnan(values), np.any(np.isnan(samples), axis=1))
+    values = np.delete(values, isnan_mask)
+    samples = np.delete(samples, isnan_mask, axis=0)
+
+    # todo vectorize, gaussian_kde cannot be called on all the samples directly
+    n = len(values)
+    mean_probability = 0.0
+    for i in range(n):
+        value = values[i]
+        kde = gaussian_kde(samples[i, :])
```
hrzn, line 1027: I don't think you want to use a kernel estimator here. It probably doesn't do what you want.
Instead, you should probably use the simple normal distribution functions: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.norm.html

Something like:
```
mu, sigma = np.mean(samples[i, :]), np.std(samples[i, :])

# for estimating a probability
proba = cdf(value + deviation, loc=mu, scale=sigma) - cdf(value - deviation, loc=mu, scale=sigma)

# for estimating a density (or likelihood)
likelihood = pdf(value, loc=mu, scale=sigma)

# we should probably leave it as an option to return the log-likelihood instead:
log_likelihood = logpdf(value, loc=mu, scale=sigma)
```

PR #421, opened 2021-08-03
Infer default from frequency

CompilerLuke:
The default for seasonal_periods of 12, in `ExponentialSmoothing` only applies to TimeSeries indexed by month.

### Summary

Leverages the functionality provided by statsmodel to infer seasonal_periods from the frequency of the `TimeSeries`.


darts/models/exponential_smoothing.py
```diff
@@ -75,11 +76,15 @@ def fit(self, series: TimeSeries):
                                            trend=self.trend.value,
                                            damped_trend=self.damped,
                                            seasonal=self.seasonal.value,
-                                           seasonal_periods=self.seasonal_periods)
-
+                                           seasonal_periods= None if self.infer_seasonal_periods else self.seasonal_periods,
+                                           freq=series.freq,
+                                           dates=series.time_index)
```
hrzn, line 81: Will this work if the series is integer-indexed (and its index is a `pandas.Int64Index`)? From the statsmodels doc it looks like this expects rather a DateTimeIndex. I think we should probably condition this inference to cases where the series are datetime-indexed only, and raise an error if the seasonality is not specified and the series is integer-indexed.

darts/models/exponential_smoothing.py
```diff
@@ -75,11 +76,15 @@ def fit(self, series: TimeSeries):
                                            trend=self.trend.value,
                                            damped_trend=self.damped,
                                            seasonal=self.seasonal.value,
-                                           seasonal_periods=self.seasonal_periods)
-
+                                           seasonal_periods= None if self.infer_seasonal_periods else self.seasonal_periods,
+                                           freq=series.freq,
```
hrzn, line 80: From the statsmodels documentation "this is optional if dates are given". Do we need it?

hrzn (commented):
Very cool! Just got one concern regarding integer-indexed series.

PR #424, opened 2021-08-05
Feat/improve doc

hrzn:
Improve documentation

* Remove abstract base classes from documentation
* Copied parameters docstrings from parents classes where needed (torch models)
* Improve Torch models docuemntation
* Create pages for Kalman and Gaussian processes
* Various small improvements here and there

I re-opened the PR because the tests didn't seem to start on #423

### DIFF
--- a/darts/models/gaussian_process_filter.py
+++ b/darts/models/gaussian_process_filter.py
@@ -1,3 +1,8 @@
+"""
+Gaussian Processes
+------------------
+"""
+
 from typing import Optional
 
 import numpy as np
@@ -13,10 +18,12 @@ class GaussianProcessFilter(FilteringModel):
     def __init__(self,
                  kernel: Optional[Kernel] = None,
                  **kwargs):
-        """ GaussianProcessFilter model
-        This model uses the GaussianProcessRegressor of scikit-learn to fit a Gaussian Process to the
-        supplied TimeSeries. This can then be used to obtain samples or the mean values of the
+        """
+        This model uses the `GaussianProcessRegressor` of scikit-learn to fit a Gaussian Process to the
+        supplied TimeSeries. This can then be used to obtain samples from the
         Gaussian Process at the times of the TimeSeries.
+
+        It can for instance be used to fill in missing (NaN) values from a TimeSeries.
 
         Parameters
         ----------
--- a/darts/models/regression_model.py
+++ b/darts/models/regression_model.py
@@ -3,9 +3,11 @@ Regression Model
 ----------------
 
 A `RegressionModel` forecasts future values of a target series based on lagged values of the target values
-and possibly lags of an exoneous series. They can wrap around any regression model having a `fit()`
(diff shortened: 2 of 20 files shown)


darts/models/gaussian_process_filter.py
```diff
@@ -1,3 +1,8 @@
+"""
+Gaussian Processes
+------------------
```
pennfranc, line 3: In some cases you removed these lines. I don't have a preference but I suppose we should be consistent

darts/models/regression_model.py
```diff
@@ -3,9 +3,11 @@
 ----------------
 
 A `RegressionModel` forecasts future values of a target series based on lagged values of the target values
-and possibly lags of an exoneous series. They can wrap around any regression model having a `fit()`
-and `predict()` functions (e.g. scikit-learn regression models), and are using
-`sklearn.linear_model.LinearRegression` by default.
+and possibly lags of a covariate series. They can wrap around any regression model having a `fit()`
```
pennfranc, line 6:
```suggestion
and possibly lags of a covariate series. They can wrap around any regression model having `fit()`
```

darts/models/torch_forecasting_model.py
```diff
@@ -610,29 +628,10 @@ def predict_from_dataset(self,
...
-        All predictions are produced in a recurrent way by taking as input
-        - the previous target value, which will be set to the last known target value for the first prediction,
-          and for all other predictions it will be set to the previous prediction
-        - the previous hidden state
-        - the current covariates (if the model was trained with covariates)
-        As a result, if covariates were used, `n` covariates have to be available into the future.
-
-        If some series in the `input_series_dataset` have more time steps than `input_chunk_length`,
-        only the last `input_chunk_length` time steps will be considered.
+        This method allows for training with a specific `darts.utils.data.InferenceDataset` instance. These datasets
```
pennfranc, line 631:
```suggestion
        This method allows for predicting with a specific `darts.utils.data.InferenceDataset` instance. These datasets
```

darts/models/torch_forecasting_model.py
```diff
@@ -610,29 +628,10 @@ def predict_from_dataset(self,
...
+        implement a PyTorch `Dataset`, and specify how the target and covariates are sliced for inference.
+        In most cases, you'll rather want to call `fit()` instead, which will create an appropriate `InferenceDataset`
```
pennfranc, line 633:
```suggestion
        In most cases, you'll rather want to call `predict()` instead, which will create an appropriate `InferenceDataset`
```

pennfranc (approved):
I think I found a couple of typos and inconsistencies, but other than that it looks really nice, thanks a lot!

darts/models/gaussian_process_filter.py
hrzn: Yes, removing the dotted lines removes the page from the generated doc.

PR #451, opened 2021-08-25
Fix/FFT notebook error

pennfranc:
I saw that there was an error in the FFT notebook because of missing values, which resulted in a NaN prediction. I know moved the missing value interpolation into the FFT model itself so this does not happen anymore.


darts/models/fft.py
```diff
@@ -196,7 +197,6 @@ def __init__(self,
         (controlled by the `nr_freqs_to_keep` argument) and  inverse FFT, combined with the option to detrend
         the data (controlled by the `trend` argument) and to crop the training sequence to full seasonal periods
         (controlled by the `required_matches` argument).
```
hrzn, line 199:
```suggestion
        Note that if the training series contains any NaNs (missing values), these will be filled using `darts.utils.missing_values.fill_missing_values()`.
```

hrzn (commented):
Could you maybe re-run the notebook using another metric than MAPE? The results look bad with MAPE because many values are close to 0.

hrzn (approved):

PR #471, opened 2021-09-08
Remove extra convolutions in TCN

### LINKED ISSUE #470: [BUG] Extra convolutions in TCN
**Describe the bug**
There are extra 1x1 convolutions in the first and the last residual blocks in a case of the fixed number of channels, say input_dim=4, num_filters=4, output_dim=4. There are no need to add such convolutions because the residual summation can be directly performed to the output of the second convolution.

**To Reproduce**
```python
from numpy.random import randn
from darts.models import TCNModel
train_sample = (randn(10, 4), None, randn(10, 4))
model = TCNModel(input_chunk_length=10, output_chunk_length=9, num_layers=2, num_filters=4)
model._create_model(train_sample)
```

**Expected behavior**
There are no extra convolutions

**System**
 - darts version 0.11.0

vpozdnyakov:
Fixes #470. Add the 1x1 convolution in a case of different number of channels between input and output of a residual block

### DIFF
--- a/darts/models/tcn_model.py
+++ b/darts/models/tcn_model.py
@@ -84,7 +84,7 @@ class _ResidualBlock(nn.Module):
         if weight_norm:
             self.conv1, self.conv2 = nn.utils.weight_norm(self.conv1), nn.utils.weight_norm(self.conv2)
 
-        if nr_blocks_below in (0, num_layers - 1) and input_dim != output_dim:
+        if nr_blocks_below in {0, num_layers - 1} and input_dim != output_dim:
             self.conv3 = nn.Conv1d(input_dim, output_dim, 1)
 
     def forward(self, x):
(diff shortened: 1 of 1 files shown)


darts/models/tcn_model.py
```diff
@@ -84,7 +84,7 @@ def __init__(self,
         if weight_norm:
             self.conv1, self.conv2 = nn.utils.weight_norm(self.conv1), nn.utils.weight_norm(self.conv2)
 
-        if nr_blocks_below == 0 or nr_blocks_below == num_layers - 1:
+        if nr_blocks_below in {0, num_layers - 1} and input_dim != output_dim:
```
pennfranc, line 87: Actually do you think we could simplify this to
```suggestion
        if  input_dim != output_dim:
```

darts/models/tcn_model.py
```diff
@@ -103,7 +103,8 @@ def forward(self, x):
         x = self.dropout_fn(x)
 
         # add residual
-        if self.nr_blocks_below in {0, self.num_layers - 1}:
+        if (self.nr_blocks_below in {0, self.num_layers - 1} and
```
pennfranc, line 106: And also here just check, `self.conv1.in_channels != self.conv2.out_channels`?

darts/models/tcn_model.py
vpozdnyakov: Agree! Looks super simple and clear

hrzn (approved):
Thanks!

pennfranc (approved):
Thanks for polishing our TCN!

PR #474, opened 2021-09-12
Feat/priors

hrzn:
* Made all NN models probabilistic (except N-Beats, this one is a bit more complicated and left for future work).
* Added "priors" for likelihoods, based on a KL-divergence regularization of the loss
* Added more distributions
* Re-organised `darts.models`
* Improved documentation of LightGBM model

The most interesting file to review is perhaps `likelihood_models.py `.

Still to do: Some unit tests


darts/utils/likelihood_models.py
```diff
@@ -1,32 +1,150 @@
...
+in which case they are applied to model each component of multivariate series independently.
+Some other distributions (such as ``DirichletLikelihood``) are multivariate,
+in which case they will model all components of multivariate time series jointly.
+
+Univariate likelihoods accept either scalar or array-like values for the optional prior parameters.
+If a scalar is provided, it is used as a prior for all components of the series. If an array-like is provided,
+the i-th value will be used as a prior for the i-th component of the series. Multivariate likelihoods
+require array-like objects when specifying priors.
+
+The target series used for training must always lies within the distribution's support, otherwise
```
pennfranc, line 28:
```suggestion
The target series used for training must always lie within the distribution's support, otherwise
```

darts/utils/likelihood_models.py
```diff
@@ -1,32 +1,150 @@
...
+errors will be raised during training. You can refer to the individual likelihoods' documentation
+to see what is the support. Similarly, the prior parameters also have to lie in some pre-defined domains.
```
pennfranc, line 30:
```suggestion
to see what target values it supports. Similarly, the prior parameters also have to lie in some pre-defined domains.
```?

File: darts/utils/likelihood_models.py [hunk 3]
```diff
@@ -1,32 +1,150 @@
...
 
-class LikelihoodModel(ABC):
+# Some utils for checking parameters' domains
+def _check(param, predicate, param_name, condition_str):
+    if param is None: return
+    if isinstance(param, (collections.Sequence, np.ndarray)):
+        raise_if_not(all(predicate(p) for p in param),
+                     'All provided parameters {} must be {}.'.format(param_name, condition_str))
+    else:
+        raise_if_not(0 < param, 'The parameter {} must be {}.'.format(param_name, condition_str))
```
pennfranc, line 72:
```suggestion
        raise_if_not(predicate(param), 'The parameter {} must be {}.'.format(param_name, condition_str))
```
..unless I'm not understanding something

darts/utils/likelihood_models.py
```diff
@@ -35,64 +153,852 @@ def _sample(self, model_output: torch.Tensor) -> torch.Tensor:
...
-        model_output_means, model_output_vars = self._means_and_vars_from_model_output(model_output)
-        return self.loss(model_output_means.contiguous(), target.contiguous(), model_output_vars.contiguous())
+        super().__init__(prior_strength)
 
-    def _sample(self, model_output: torch.Tensor) -> torch.Tensor:
-        model_output_means, model_output_vars = self._means_and_vars_from_model_output(model_output)
-        return torch.normal(model_output_means, model_output_vars)
+    def _nllloss(self, params_out, target):
+        means_out, sigmas_out = params_out
+        return self.nllloss(means_out.contiguous(), target.contiguous(), sigmas_out.contiguous())
```
pennfranc, line 195: I remember I had to add the `.contigous` for it to work, but is this still the case? because it is only used here.

pennfranc:
Really cool stuff! I guess one of the things that we might still want to add are some test cases involving all likelihoods?

darts/utils/likelihood_models.py
```diff
@@ -1,32 +1,150 @@
...
-        Abstract class for a likelihood model. It contains all the logic to compute the loss
-        and to sample the distribution, given the parameters of the distribution
+        Abstract class for a likelihood model.
         """
-        pass
+        self.prior_strength = prior_strength
 
-    @abstractmethod
-    def _compute_loss(self, model_output: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+    def compute_loss(self, model_output: torch.Tensor, target: torch.Tensor, ):
```
pennfranc, line 90:
```suggestion
    def compute_loss(self, model_output: torch.Tensor, target: torch.Tensor):
```

darts/utils/likelihood_models.py
```diff
@@ -35,64 +153,852 @@ def _sample(self, model_output: torch.Tensor) -> torch.Tensor:
...
-        model_output = self._lambda_from_output(model_output)
-        return self.loss(model_output, target)
+    @property
+    def _prior_params(self):
+        return None
+
+    @staticmethod
+    def _get_r_and_p_from_mu_and_alpha(mu, alpha):
+        # See https://en.wikipedia.org/wiki/Negative_binomial_distribution for the different parametrizations
+        r = 1. / alpha
```
pennfranc, line 297: I think this is causing some numerical issues. `r` can end up having `inf` values here, which then results in NaNs for `p` too. I changed this line to `r = 1. / (alpha + eps)` and get good results with eps=1e-3, but not with other values. Obviously we should find a solution without magic numbers, but I don't have one yet.

darts/utils/likelihood_models.py
hrzn: I'll keep the first edit but not the second :D ("support" is quite the correct technical term I think here).

darts/utils/likelihood_models.py
hrzn: Good spot!

darts/utils/likelihood_models.py
hrzn: Here too I'd rather keep the technical term (since it's also referred to later on).

darts/utils/likelihood_models.py
hrzn: I actually tried to remove it, but it crashes. I didn't investigate it, but it's a bit weird (almost looks like it could be a bug in PyTorch?). In any case let's leave the contiguous() for now.

darts/utils/likelihood_models.py
hrzn: Good one. Actually I think the `eps` trick is fine, that's a standard way to approach it. However we could maybe find a smaller value? How about something like 1e-6?

darts/utils/likelihood_models.py
pennfranc: Ah ok I wasn't aware!

darts/tests/test_probabilistic_models.py
```diff
@@ -77,18 +98,67 @@ def helper_test_probabilistic_forecast_accuracy(self, model_cls, model_kwargs, e
...
+        # test accuracy for decreasing quantiles between 0.3 and 0 (it should ~decrease)
         tested_quantiles = [0.3, 0.2, 0.1, 0.01]
         mae_err = mae_err_median
         for quantile in tested_quantiles:
             new_mae = mae(ts[100:], pred.quantile_timeseries(quantile=quantile))
-            self.assertLess(mae_err, new_mae)
+            self.assertLess(mae_err, new_mae+0.1)
             mae_err = new_mae
+
+    """ More likelihood tests
```
hrzn, line 117: @pennfranc I've added unit tests for all likelihoods here. Could you maybe take a quick look? Thx :)

pennfranc: Looks good!

PR #475, opened 2021-09-13
Feature/metric rho risk

dennisbader:
Addressees DARTS-334.

### Summary

Implemented the rho-risk metric for stochastic predictions following [this paper](https://arxiv.org/abs/1704.04110).

So far the following is supported (only stochastic forecasts allowed, since other metrics should be used for deterministic):
- rho-risk for univariate predictions
- rho-risk for multi-timeseries (*)
- rho-risk for multivariate timeseries (*)

(*) for multi-ts and multivariate ts, the paper from above normalizes the sum of rho-losses from all ts with the sum of Z-values. In my implementation I take the mean due to nature of the multi-ts/multivar. support-wrappers. Should I implement special wrappers for that?

What is missing:
- test cases: since I haven't found a lot of tests for stochastic metrics. Should I add them to darts/tests/test_probabilistic_models.py or darts/tests/test_metrics.py?

### Additional Information

- I slightly changed the _get_values_or_raise() method so that it can return all sample values for stochastic predictions

### DIFF
--- a/darts/metrics/metrics.py
+++ b/darts/metrics/metrics.py
@@ -115,9 +115,28 @@ def multivariate_support(func):
     return wrapper_multivariate_support
 
 
+def _get_values(series: TimeSeries,
+                stochastic_quantile: float = 0.5) -> np.ndarray:
+    """
+    Returns the numpy values of a time series.
+    For stochastic series, return either all sample values with (stochastic_quantile=None) or the quantile sample value
+    with (stochastic_quantile {>=0,<=1})
+    """
+    if series.is_deterministic:
+        series_values = series.univariate_values()
+    else:  # stochastic
+        if stochastic_quantile is None:
+            series_values = series.data_array().values
+        else:
+            raise_if_not(isinstance(stochastic_quantile, float), 'stochastic quantile must be a float')
+            series_values = series.quantile_timeseries(quantile=stochastic_quantile)
+    return series_values
+
+
 def _get_values_or_raise(series_a: TimeSeries,
                          series_b: TimeSeries,
-                         intersect: bool) -> Tuple[np.ndarray, np.ndarray]:
+                         intersect: bool,
+                         stochastic_quantile: Optional[float] = 0.5) -> Tuple[np.ndarray, np.ndarray]:
     """
     Returns the numpy values of two time series. If intersect is true, considers only their time intersection.
     Raises a ValueError if the two time series (or their intersection) do not have the same time index.
@@ -138,11 +157,10 @@ def _get_values_or_raise(series_a: TimeSeries,
                                                                     series_a.time_index, series_b.time_index),
                  logger)
 
-    # TODO: we may want to change this...
-    series_a_det = series_a_common if series_a_common.is_deterministic else series_a_common.quantile_timeseries(quantile=0.5)
-    series_b_det = series_b_common if series_b_common.is_deterministic else series_b_common.quantile_timeseries(quantile=0.5)
-
-    return series_a_det.univariate_values(), series_b_det.univariate_values()
+    # TODO: we may want to change this... -> (Dennis: 13.09.2021) how about this?
+    series_a_det = _get_values(series_a_common, stochastic_quantile=stochastic_quantile)
+    series_b_det = _get_values(series_b_common, stochastic_quantile=stochastic_quantile)
+    return series_a_det, series_b_det
 
 
 def _remove_nan_union(array_a: np.ndarray,
@@ -151,7 +169,6 @@ def _remove_nan_union(array_a: np.ndarray,
     Returns the two inputs arrays where all elements are deleted that have an index that corresponds to
     a NaN value in either of the two input arrays.
     """
-
     isnan_mask = np.logical_or(np.isnan(array_a), np.isnan(array_b))
     return np.delete(array_a, isnan_mask), np.delete(array_b, isnan_mask)
 
@@ -947,13 +964,11 @@ def dtw_metric(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
 
 # rho-risk (quantile risk)
 @multi_ts_support
-# @multivariate_support
+@multivariate_support
 def rho_risk(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
              pred_series: Union[TimeSeries, Sequence[TimeSeries]],
-             rho: Union[float, Sequence[float]],
-             start: Union[pd.Timestamp, int, float] = 1,
-             end: Union[pd.Timestamp, int, float] = 1,
-             # intersect: bool = True,
+             rho: float,
+             intersect: bool = True,
              *,
              reduction: Callable[[np.ndarray], float] = np.mean,
              inter_reduction: Callable[[np.ndarray], Union[float, np.ndarray]] = lambda x: x,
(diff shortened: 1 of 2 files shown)


darts/metrics/metrics.py
```diff
@@ -113,9 +114,28 @@ def wrapper_multivariate_support(*args, **kwargs):
...
+    For stochastic series, return either all sample values with (stochastic_quantile=None) or the quantile sample value
+    with (stochastic_quantile {>=0,<=1})
+    """
+    if series.is_deterministic:
+        series_values = series.univariate_values()
+    else:  # stochastic
+        if stochastic_quantile is None:
+            series_values = series.data_array().values
+        else:
+            raise_if_not(isinstance(stochastic_quantile, float), 'stochastic quantile must be a float')
```
pennfranc, line 130: Not sure we need this, since the function signature specifies that we need a (optional) float. Otherwise we would need checks for all function parameter data types in all other functions too, which is generally not what we do. @hrzn thoghts?

darts/metrics/metrics.py
```diff
@@ -113,9 +114,28 @@ def wrapper_multivariate_support(*args, **kwargs):
     return wrapper_multivariate_support
 
 
+def _get_values(series: TimeSeries,
+                stochastic_quantile: float = 0.5) -> np.ndarray:
```
pennfranc, line 118:
```suggestion
                stochastic_quantile: Optional[float] = 0.5) -> np.ndarray:
```

darts/metrics/metrics.py
```diff
@@ -941,3 +960,100 @@ def dtw_metric(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
...
+    float
+        The rho-risk metric
+    """
+
+    def rho_loss(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
+                 pred_series: Union[TimeSeries, Sequence[TimeSeries]],
+                 rho: float = 0.5,
+                 intersect: bool = True) -> Tuple[float, float]:
+
+        raise_if_not(isinstance(rho, float), 'rho (quantile) must be a float.')
```
pennfranc, line 1037: As above, I'm not sure we need this check

darts/metrics/metrics.py
```diff
@@ -941,3 +960,100 @@ def dtw_metric(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
...
+    verbose
+        Optionally, whether to print operations progress
+
+    Returns
+    -------
+    float
+        The rho-risk metric
+    """
+
+    def rho_loss(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
```
pennfranc, line 1032: Any specific reason for having a function within a function here?

darts/tests/test_metrics.py
```diff
@@ -89,14 +106,26 @@ def helper_test_multiple_ts_duplication_equality(self, metric, **kwargs):
...
+
+        for s1, s2 in test_cases:
+            # univariate
+            non_nan_metric = metric(s1[:9] + 1, s2[:9])
+            nan_s1 = s1.copy()
+            nan_s1._xa.values[-1, :, :] = np.nan
+            nan_metric = metric(nan_s1 + 1, s2)
+            self.assertEqual(non_nan_metric, nan_metric)
+
+            # multivariate
```
pennfranc, line 120:
```suggestion
            # multivariate + multi-TS
```
Very nice extension, thanks!

darts/tests/test_probabilistic_models.py
```diff
@@ -68,27 +67,61 @@ def test_probabilistic_forecast_accuracy(self):
...
+        model = model_cls(**model_kwargs)
+        model.fit(noisy_ts[:100])
+        pred = model.predict(n=100, num_samples=100)
+
+        # test rho-risk (at quantile=0.5) compared to the noiseless ts
+        rrisk_median = rho_risk(ts[100:], pred, rho=0.5)
+        self.assertLess(rrisk_median, err)
+
+        # test risk for increasing quantiles between 0.75 and 1 should increase
+        tested_quantiles = [0.8, 0.9, 0.95, 0.99]  # some test cases actually have minimum risk around 0.75
```
pennfranc, line 114: Why is it that the rho risk is increasing when rho is higher? Sorry if this is obvious, I just briefly looked at the definitions and it's not immediately clear to me.

pennfranc:
Overall this looks really nice! Thanks a lot for also simplifying and generalizing some test cases! I just have a few comments. Also, before merging, it might be nice to integrate this new metric into the example notebooks involving probabilistic models. What do you think?

darts/metrics/metrics.py
dennisbader: I think this makes sense, thanks. Good to know for future PRs.

darts/metrics/metrics.py
dennisbader: Yeah, I thought the same. It was meant for me to make it more clear.
But since it doesn't have any functional reason, I will remove it.

darts/tests/test_probabilistic_models.py
dennisbader: Let's say your predicted stochastic series at quantile/rho=0.5 represents the actual series, then the more you move away from 0.5, the bigger the loss and the higher the risk becomes.
So in these examples with **tested** models and quantiles, the risk increases with each rho in `tested_quantiles`.

If with some other model your predicted series is closest to the actual series at quantile=1., then this test would fail (and also accuracy test `test_probabilistic_forecast_accuracy()`).

I hope this answered your question, if not just let me know :)

dennisbader: Actually, yesterday @pennfranc noticed that the definition of the rho_metric in the originally used paper had a mistake.
From now on we use the rho_metric defined in [this paper](https://papers.nips.cc/paper/2016/hash/03255088ed63354a54e0e5ed957e9008-Abstract.html).

I'll update the pull request comment.

darts/metrics/metrics.py
```diff
@@ -113,9 +114,27 @@ def wrapper_multivariate_support(*args, **kwargs):
...
+    """
+    Returns the numpy values of a time series.
+    For stochastic series, return either all sample values with (stochastic_quantile=None) or the quantile sample value
+    with (stochastic_quantile {>=0,<=1})
+    """
+    if series.is_deterministic:
+        series_values = series.univariate_values()
+    else:  # stochastic
+        if stochastic_quantile is None:
+            series_values = series.data_array().values
```
hrzn, line 128: You could call `series.all_values()` instead. And if you know what you're doing and that you won't be modifying the values, you can make it `series.all_values(copy=False)`, which will be faster.

darts/metrics/metrics.py
```diff
@@ -136,11 +155,10 @@ def _get_values_or_raise(series_a: TimeSeries,
                                                                     series_a.time_index, series_b.time_index),
                  logger)
 
-    # TODO: we may want to change this...
-    series_a_det = series_a_common if series_a_common.is_deterministic else series_a_common.quantile_timeseries(quantile=0.5)
-    series_b_det = series_b_common if series_b_common.is_deterministic else series_b_common.quantile_timeseries(quantile=0.5)
-
-    return series_a_det.univariate_values(), series_b_det.univariate_values()
+    # TODO: we may want to change this... -> (Dennis: 13.09.2021) how about this?
```
hrzn, line 158: Looks good :)
You can change the docstring on line 142: `For stochastic series, return some deterministic quantile value`

darts/metrics/metrics.py
```diff
@@ -941,3 +959,91 @@ def dtw_metric(actual_series: Union[TimeSeries, Sequence[TimeSeries]],
...
+        The rho-risk metric
+    """
+
+    raise_if_not(pred_series.is_stochastic,
+                 'rho (quantile) loss should only be computed for stochastic predicted TimeSeries.')
+
+    z_true, z_hat = _get_values_or_raise(actual_series, pred_series, intersect=intersect, stochastic_quantile=None)
+
+    # adaption of _remove_nan_union(y1, y2): in stochastic case we need to account for different shapes of y1 & y2
+    isnan_mask = np.logical_or(np.isnan(z_true), np.isnan(z_hat).any(axis=2).flatten())
```
hrzn, line 1037: Actually I wonder if we could perhaps move this "remove NaNs" logic into `_get_values_or_raise` function and make it optional there, because it could benefit all metrics potentially!

hrzn (approved):
Looks great! Got a couple of minor suggestions.
Good job @dennisbader 👍

darts/metrics/metrics.py
dennisbader: Nice, very good to know! Thanks 👍

darts/metrics/metrics.py
dennisbader: Good idea, I'll put it into _get_values_or_raise

hrzn (approved):

hrzn:
@dennisbader LGTM - feel free to merge it once you feel confident it's all good. Thx!

PR #477, opened 2021-09-15
Feature/backtest err hist

### LINKED ISSUE #466: [BUG] Accept linewidth and lw interchangeably as the input of the Timeseries plot
**Description**
This is a minor issue, but at the moment it seems that the plot method of Timeseries is passing `lw` to the `matplotlib` library as the argument for the width of the line while an alternative argument is `linewidth` but if one passes this it would give an error as follows:

`
TypeError: Got both 'linewidth' and 'lw', which are aliases of one another
`

A full backtrace of the error could be found here: https://paste.ubuntu.com/p/Cs9wZF8Sz4/

**To Reproduce**
Just pass `linewidth` instead of `lw` to the plot method of Timeseries

**Expected behavior**
It shouldn't matter for the end-user whether to pass lw or linewidth

 - Python version: Python 3.8.5
 - darts version: '0.9.1'

dennisbader:
- adds method `plot_hist()` to `darts/utils/statistics.py` that plots histogram of input data -> addressing [DARTS-226](https://unit8.atlassian.net/browse/DARTS-226)
- fixes #466 -> now either `lw` or `linewidth` can be passed to `TimeSeries.plot()`


darts/utils/statistics.py
```diff
@@ -305,9 +305,54 @@ def plot_acf(ts: TimeSeries,
...
+    ax
+        Optionally, an axis object to plot the histogram on.
+    """
+
+    raise_if_not(any([isinstance(data, x) for x in [TimeSeries, list, np.ndarray]]),
+                 'input data must be a TimeSeries instance or an array-like')
+
+    if isinstance(data, TimeSeries):
+        data._assert_univariate()
+        data._assert_deterministic()
```
hrzn, line 338: If the series is stochastic you could imagine getting something like `values = data.all_data().flatten()`.

hrzn (commented):
This looks good. It might be nice to add an example in the intro notebook perhaps?
Btw: if you don't change the notebook (but only re-ran it), could you keep it out of the commits? Because otherwise notebooks will have a huge git history unnecessarily.

darts/utils/statistics.py
```diff
@@ -305,9 +305,53 @@ def plot_acf(ts: TimeSeries,
...
+    ax
+        Optionally, an axis object to plot the histogram on.
+    """
+
+    raise_if_not(any([isinstance(data, x) for x in [TimeSeries, list, np.ndarray]]),
+                 'input data must be a TimeSeries instance or an array-like')
+
+    if isinstance(data, TimeSeries):
+        data._assert_univariate()
+        values = data.univariate_values() if data.is_deterministic else data.all_values().flatten()
```
hrzn, line 338: I think simply having `values = data.all_values().flatten()` should work in all cases.

Actually, I'm thinking that you could remove the requirement for the series to be univariate, and plot several histograms (maybe on the same figure) with different colors/labels for multivariate series. One histogram per component - and maybe only plot the top-5 or something like this if the series contains too many components (you can raise a warning if this happens).

dennisbader: I updated it now with https://github.com/unit8co/darts/pull/477/commits/51da4dee1f1fbb0d4042de36776f21583b772c93 that if the TimeSeries contains more than one component, it plots up to four histograms of the first four components.

If `data` is a list or np.ndarray I will simply flatten the values as I think that is the major use case.

hrzn (approved):
👍

PR #479, opened 2021-09-16
Timeseries utilities

mkos:
Utilities in `TimeSeries`:
- **New**
    - `to_csv()` - only for deterministic timeseries
    - `to_pickle()` - also for stochastic timeseries
    - `from_pickle()`
    - `head()` - by default first 5 rows over time series
    - `tail()` - by default last 5 rows over time series
    - `concatenate()`
- **Updated**
    - `from_csv()`, `from_dataframe()` - if there is a column named `time` it will be automatically parsed as timeseries and used as an index for time


dennisbader (commented):
Looks good, thanks a lot! Just got a couple of minor issues.
The new utils will definitely be very useful!

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        if ((axis == 'time' and not (component_axis_equal and sample_axis_equal)) or
+            (axis == 'component' and not ((time_axis_equal or ignore_time_axes) and sample_axis_equal)) or
+            (axis == 'sample' and not ((time_axis_equal or ignore_time_axes) and component_axis_equal))):
+
+            raise_log(AttributeError('Remaining (non-concatenating) axes need to be equal.'))
+
+        da_sequence = [ts.data_array() for ts in timeserie_sequence]
+
+        if axis == DIMS[0]: # time
+            if time_axis_equal:
```
hrzn, line 1051: I don't really think you need to explicitly check this case here, because it will later be assessed that `consecutive_time_axes == True`, and `time_axis_equal` is just a sub case of time axes not being consecutive but I don't think we need to check for it and adapt the error message in this case.

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                if ignore_time_axes:
+                    da_concat=da_sequence[0]
+                else:
+                    raise_log(AttributeError("All concatenated have the same time axis. "
+                                             "If this is not an error, use `ignore_time_axis=True`."))
```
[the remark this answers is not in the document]
hrzn: I think an error makes sense here, because if the user is calling concatenate() and we return a TimeSeries without raising an error, it might lead to misunderstandings and user expecting the result to be a concatenated series (whereas it wouldn't be if we just return the first).

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                consecutive_time_axes = True
+                for i in range(1, len(timeserie_sequence)):
+                    if timeserie_sequence[i - 1].end_time() + pd.Timedelta(timeserie_sequence[0].freq) != \
+                            timeserie_sequence[i].start_time():
+                        consecutive_time_axes = False
+                        break
+
+                if consecutive_time_axes:
+                    da_concat = xr.concat(da_sequence, dim=axis)
+                else:
```
hrzn, line 1067: Here you should not raise if `ignore_time_axis == True`. Basically:

* Either series are consecutive (in which case you concat them directly),
* Or `ignore_time_axis == True` (in which case you concat their values and then resets the resulting time index so that the beginning matches the beginning of the first series).

If none of these two conditions are satisfied, raise an error.

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        start_date_equal = len(set([ts.start_time() for ts in timeserie_sequence])) == 1
+        end_date_equal = len(set([ts.end_time() for ts in timeserie_sequence])) == 1
+        freq_equal = len(set([ts.freq_str for ts in timeserie_sequence])) == 1
+        time_axis_equal = start_date_equal and end_date_equal and freq_equal
+        component_axis_equal = len(set([ts.width for ts in timeserie_sequence])) == 1
+        sample_axis_equal = len(set([ts.n_samples for ts in timeserie_sequence])) == 1
+
+        if ((axis == 'time' and not (component_axis_equal and sample_axis_equal)) or
+            (axis == 'component' and not ((time_axis_equal or ignore_time_axes) and sample_axis_equal)) or
+            (axis == 'sample' and not ((time_axis_equal or ignore_time_axes) and component_axis_equal))):
```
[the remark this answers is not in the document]
hrzn: Yes it's important, there are cases where it could fail otherwise.

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                else:
+                    raise_log(AttributeError("When concatenating over time axis, all series need to be subsequent, i.e."
+                                             " end_time of the previous one should be equal to start_time - 1 of the "
+                                             "following timeserie."
+                                         ))
+
+        else: # component or sample
+            if len(set([ts.shape[0] for ts in da_sequence])) != 1:
+                raise AttributeError("All concatenating time series need to have the same time axis or at least"
+                                     " be of the same length.")
```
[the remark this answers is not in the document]
hrzn: I think to avoid confusion we should either have
* `time_axis_equal == True`, or
* `ignore_time_axes == True` and `len(set([ts.shape[0] for ts in da_sequence])) == 1`.

I wouldn't silently concatenate series that do not have the same time axes even if they have the same length. This is basically what is implemented here, except that the check `len(set([ts.shape[0] for ts in da_sequence])) == 1` could be move and not be done preemptively, but performed only if `time_axis_equal == False`.

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                    ts1_time_coord = da_sequence[0].coords[DIMS[0]]
+                    for i in range(1, len(da_sequence)):
+                        da_sequence[i] = da_sequence[i].assign_coords(time=ts1_time_coord)
+
+                    da_concat = xr.concat(da_sequence, dim=axis)
+                else:
+                    raise_log(AttributeError("All concatenating time series need to have the same time axis."
+                                             " You can override this error by setting `ignore_time_axes=True`."))
+
+        return TimeSeries.from_xarray(da_concat, fill_missing_dates=True, freq=timeserie_sequence[0].freq_str)
```
hrzn, line 1090: I don't think you need to call the factory method here, probably the constructor is enough. Furthermore you for sure shouldn't set `fill_missing_dates=True` because this is expensive to compute and should not be needed at this stage. Also no need to set the freq if the time axis of the DataArray contains the correct freq already.

darts/timeseries.py
```diff
@@ -1725,6 +1903,55 @@ def to_json(self) -> str:
         """
         return self.pd_dataframe().to_json(orient='split', date_format='iso')
 
+    def to_csv(self,
+               path_or_buf: Optional[Union[str, TextIO]],
+               sep: str = ",",
```
hrzn, line 1908: Couldn't we just leverage Pandas' `**kwargs` here and not list/maintain them on our side?

darts/utils/timeseries_generation.py
```diff
@@ -391,5 +391,5 @@ def datetime_attribute_timeseries(time_index: Union[pd.DatetimeIndex, TimeSeries
     if one_hot:
         values_df.columns = [attribute + '_' + str(column_name) for column_name in values_df.columns]
 
-    return TimeSeries.from_dataframe(values_df)
+    return TimeSeries.from_dataframe(values_df, time_col=None)
```
hrzn, line 394: This change is not needed once the default is restored to `None`.

hrzn (changes requested):

darts/timeseries.py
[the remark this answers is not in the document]
hrzn: Actually for the name of the time axis you should use `TimeSeries._time_dim` because we allow this one to have a different name.

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+
+            Returns
+            -------
+            TimeSeries
+                concatenated timeseries
+        """
+
+        axis = TimeSeries._get_str_axis(axis)
+        if len(timeserie_sequence) < 2:
+            return timeserie_sequence
```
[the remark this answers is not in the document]
mkos: good catch.

darts/timeseries.py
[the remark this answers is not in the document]
mkos: I can't use `TimeSeries._time_dim` because it's `@staticmethod`, and since you have suggested moving it out from `TimeSeries` class, I won't be able to use it either. Unless I am missing something, @hrzn? For now I used suggestion from @dennisbader.

darts/timeseries.py
mkos: Not exactly: in the former case, `ignore_time_axes` can be used to override it and return only the first timeseries. In the latter case, `ignore_time_axes` doesn't do anything (and maybe it shouldn't be this way). Should I change the behavior to one case, where `ignore_time_axes` returns first series only, whether they perfectly overlap or not?

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+            if time_axis_equal:
+                if ignore_time_axes:
+                    da_concat=da_sequence[0]
+                else:
+                    raise_log(AttributeError("All concatenated have the same time axis. "
+                                             "If this is not an error, use `ignore_time_axis=True`."))
+            else:
+                consecutive_time_axes = True
+                for i in range(1, len(timeserie_sequence)):
+                    if timeserie_sequence[i - 1].end_time() + pd.Timedelta(timeserie_sequence[0].freq) != \
```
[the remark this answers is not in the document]
mkos: I removed `pd.Timedelta` per your comment, but I had to leave `timeserie_sequence[0]` because the `timeserie_sequence` is a simple list and does not have `freq` parameter.

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                Three dimensional array of (time, component, samples) where "samples" dimension has been cut off after
+                # of ``samples`` samples.
+        """
+        # test how this will work for univariate and multivariate cases
+        axis = TimeSeries._get_str_axis(axis)
+
+        try:
+            return TimeSeries(self._xa[{axis: range(samples)}])
+        except IndexError:
+            return self
```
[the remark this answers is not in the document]
mkos: I did the suggested change, but I don't really understand why avoid try/except? Even official python documentation says that try/except is the preferred way: [EAFP](https://docs.python.org/3/glossary.html#term-eafp)

darts/timeseries.py
```diff
@@ -911,6 +935,160 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                    ignore_time_axes: Optional[bool] = False):
+        """Concatenates other timeseries to the current one along given axis.
+
+            Note: when concatenating along the ``time`` dimension, first concatenated timeserie marks the start date of
+            concatenated series and remaining series will have their date indices overwritten.
+
+            Parameters
+            ----------
+            timeserie_sequence : sequence of TimeSeries
+                another timeseries to concatenate to this one
```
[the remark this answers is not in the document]
mkos: yes, the idea was different at the beginning and changed along the way, but I didn't update docstring - hence the confusion. I think this is good idea, definitely a little less of typing.

darts/timeseries.py
```diff
@@ -308,7 +309,25 @@ def from_dataframe(df: pd.DataFrame,
 
         # get time index
         if time_col:
```
dennisbader, line 311: Btw, if we rule out `if time_col is None:` and `if time_col not in df.columns:` we could reduce the indentation level quite a bit and make it easier to read:

```suggestion
        if time_col is None:
            raise_if_not(isinstance(df.index, pd.Int64Index) or isinstance(df.index, pd.DatetimeIndex),
                         'If time_col is not specified, the DataFrame must be indexed either with'
                         'a DatetimeIndex, or with a Int64Index (incl. RangeIndex).', logger)
    
        if time_col not in df.columns:
            raise_log(AttributeError('time_col=\'{}\' is not present.'.format(time_col)))
    
        if np.issubdtype(df[time_col].dtype, object):
            ...
```

darts/timeseries.py
```diff
@@ -904,6 +945,96 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+            axis along which we intend to display records
+
+        Returns
+        -------
+        TimeSeries
+            Three dimensional array of (time, component, samples) where "samples" dimension has been cut off except
+            # of ``samples`` from the bottom. [Default: 5]
+        """
+
+        axis = TimeSeries._get_str_axis(axis)
```
dennisbader, line 993:
```suggestion
        axis = TimeSeries._get_str_axis(axis)
        
```

darts/timeseries.py
```diff
@@ -904,6 +945,96 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                    axis: Optional[Union[str, int]] = 0,
+                    ignore_time_axes: Optional[bool] = False):
+        """Concatenates another timeseries to the current one along given axis.
+
+            Note: when concatenating along the ``time`` dimension, first concatenated timeserie marks the start date of
+            the resulting series, and the remaining series will have their date indices overwritten.
+
+            Parameters
+            ----------
+            other : sequence of TimeSeries
```
dennisbader, line 1024:
```suggestion
            other : TimeSeries
```

darts/timeseries.py
```diff
@@ -904,6 +945,96 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        return axis
+
+
+    def concatenate(self,
+                    other: 'TimeSeries',
+                    axis: Optional[Union[str, int]] = 0,
+                    ignore_time_axes: Optional[bool] = False):
+        """Concatenates another timeseries to the current one along given axis.
+
+            Note: when concatenating along the ``time`` dimension, first concatenated timeserie marks the start date of
```
dennisbader, line 1019:
```suggestion
            Note: when concatenating along the ``time`` dimension, timeseries `self` marks the start date of
```

darts/timeseries.py
```diff
@@ -2270,3 +2435,97 @@ def _set_freq_in_xa(xa_: xr.DataArray):
...
 
         raise_log(IndexError("The type of your index was not matched."), logger)
+
+
+def concatenate(timeserie_sequence: Sequence['TimeSeries'],
+                axis: Optional[Union[str, int]] = 0,
+                ignore_time_axes: Optional[bool] = False):
+    """Concatenates multiple timeseries along given axis.
+
+    Note: when concatenating along the ``time`` dimension, first concatenated timeserie marks the start date of
```
dennisbader, line 2445:
```suggestion
    Note: when concatenating along the ``time`` dimension, first timeseries of `timeseries_sequence` marks the start date of
```

darts/timeseries.py
```diff
@@ -2270,3 +2435,97 @@ def _set_freq_in_xa(xa_: xr.DataArray):
                 return TimeSeries(xa_)
 
         raise_log(IndexError("The type of your index was not matched."), logger)
+
+
+def concatenate(timeserie_sequence: Sequence['TimeSeries'],
+                axis: Optional[Union[str, int]] = 0,
+                ignore_time_axes: Optional[bool] = False):
```
dennisbader, line 2442:
```suggestion
                axis: Union[str, int] = 0,
                ignore_time_axes: bool = False):
```

darts/timeseries.py
```diff
@@ -904,6 +945,96 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        else:
+            raise_log(AttributeError(axis_error_string))
+
+        return axis
+
+
+    def concatenate(self,
+                    other: 'TimeSeries',
+                    axis: Optional[Union[str, int]] = 0,
+                    ignore_time_axes: Optional[bool] = False):
```
dennisbader, line 1016:
```suggestion
                    ignore_time_axes: Optional[bool] = False) -> 'TimeSeries':
```

dennisbader (approved):
Very nice now, thanks a lot! Also cool how concatenate works now!
I just added a few minor doc suggestions and one place to reduce indentation level.

darts/timeseries.py
```diff
@@ -904,6 +945,96 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+                axis along which timeseries will be concatenated. ['time', 'component' or 'sample'; Default: 'time']
+            ignore_time_axes : bool, default False
+                Ignore errors when time axis varies for some timeseries. Note that this may yield unexpected results
+
+            Returns
+            -------
+            TimeSeries
+                concatenated timeseries
+        """
+        return concatenate(timeserie_sequence=[self, other], axis=axis, ignore_time_axes=ignore_time_axes)
```
dennisbader, line 1036: actually, since the new concatenate supports sequences of any length we could also concatenate a sequence of timeseries with `self` with something like this (and docs update):

```suggestion
        return concatenate(timeserie_sequence=[self] + other if isinstance(other, list) else [other], axis=axis, ignore_time_axes=ignore_time_axes)
```

darts/timeseries.py
mkos: Note the different error strings. Reduce level → reduce verbosity.

darts/timeseries.py
mkos: I think this is a bit overcomplicated.

darts/timeseries.py
mkos: I don't understand why have you removed `Optional`? Since we have defaults, this indeed is optional. Unless I'm not seeing something.

darts/timeseries.py
mkos: I think the original as clear as yours.

darts/timeseries.py
dennisbader: Optional means it can either be `None` or a `bool`. It means the same as `Union[bool, None] = False`

darts/timeseries.py
dennisbader: Could you please explain a bit more what that means (out of curiosity)? That the error message in the order from before are more clear/verbose?

darts/timeseries.py
```diff
@@ -308,7 +309,25 @@ def from_dataframe(df: pd.DataFrame,
...
+                if np.issubdtype(df[time_col].dtype, object):
+                    try:
+                        time_index = pd.Int64Index(df[time_col].astype(int))
+                    except ValueError:
+                        try:
+                            time_index = pd.DatetimeIndex(df[time_col])
+                        except ValueError:
+                            raise_log(
+                                AttributeError("'time_col' is of 'object' dtype but doesn't contain valid timestamps"))
+                elif np.issubdtype(df[time_col].dtype, np.number):
```
hrzn, line 322: Would this work? If yes, I think it's better
```suggestion
                elif np.issubdtype(df[time_col].dtype, np.integer):
```

darts/timeseries.py
```diff
@@ -904,6 +945,97 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
         else:
             return self._xa[:, 0, sample].values
 
+    def head(self,
+             size: Optional[int] = 5,
+             axis: Optional[Union[int, str]] = 'time') -> 'TimeSeries':
```
hrzn, line 950:
```suggestion
             axis: Optional[Union[int, str]] = 0) -> 'TimeSeries':
```

darts/timeseries.py
```diff
@@ -904,6 +945,97 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        ----------
+        size : int, default 5
+            number of samples to retain
+        axis : str or int, optional, default 'time'
+            axis along which we intend to display records
+
+        Returns
+        -------
+        TimeSeries
+            Three dimensional array of (time, component, samples) where "samples" dimension has been cut off after
```
hrzn, line 964:
```suggestion
            Three dimensional array of (time, component, samples) where "axis" dimension has been cut off after
```

darts/timeseries.py
```diff
@@ -904,6 +945,97 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        """
+
+        axis = TimeSeries._get_str_axis(axis)
+
+        display_n = range(min(size, self._xa.sizes[axis]))
+        return TimeSeries(self._xa[{axis: display_n}])
+
+    def tail(self,
+             size: Optional[int] = 5,
+             axis: Optional[Union[int, str]] = 'time') -> 'TimeSeries':
```
hrzn, line 975:
```suggestion
             axis: Optional[Union[int, str]] = 0) -> 'TimeSeries':
```

darts/timeseries.py
```diff
@@ -904,6 +945,97 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+        """
+        Return last n samples from TimeSeries.
+
+        Parameters
+        ----------
+        size : int, default: 5
+            number of samples to retain
+        axis : str or int, optional, default: 'time'
```
hrzn, line 983:
```suggestion
        axis : str or int, optional, default: 0 (time dimension)
```

darts/timeseries.py
```diff
@@ -904,6 +945,97 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+            axis along which we intend to display records
+
+        Returns
+        -------
+        TimeSeries
+            Three dimensional array of (time, component, samples) where "samples" dimension has been cut off except
```
hrzn, line 989:
```suggestion
            Three dimensional array of (time, component, samples) where "axis" dimension has been cut off except
```

darts/timeseries.py
```diff
@@ -904,6 +945,97 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
...
+
+            Note: when concatenating along the ``time`` dimension, timeseries `self` marks the start date of
+            the resulting series, and the remaining series will have their date indices overwritten.
+
+            Parameters
+            ----------
+            other : TimeSeries
+                another timeseries to concatenate to this one
+            axis : str or int
+                axis along which timeseries will be concatenated. ['time', 'component' or 'sample'; Default: 'time']
```
hrzn, line 1028:
```suggestion
                axis along which timeseries will be concatenated. ['time', 'component' or 'sample'; Default: 0 (time)]
```

darts/timeseries.py
```diff
@@ -2270,3 +2436,97 @@ def _set_freq_in_xa(xa_: xr.DataArray):
...
+
+        da_concat = xr.concat(da_sequence, dim=axis)
+
+        if not consecutive_time_axes:
+            if ignore_time_axes:
+                tindex = pd.date_range(timeserie_sequence[0].start_time(),
+                                       freq=timeserie_sequence[0].freq,
+                                       periods=da_concat.sizes[axis]
+                                       )
+                da_concat = da_concat.assign_coords({DIMS[0]:tindex})
```
hrzn, line 2501:
```suggestion
                da_concat = da_concat.assign_coords({DIMS[0]: tindex})
```

hrzn (approved):
Just a couple of minor suggestions, then we can merge. Well done!

darts/timeseries.py
mkos: it should according to this: https://numpy.org/doc/stable/reference/arrays.scalars.html#arrays-scalars.

PR #483, opened 2021-09-20
Feat/perf improvements

hrzn:
* Some performance improvements in `TimeSeries` factory methods (removed a couple of unnecessary copying operations)

UPDATE: We actually don't need the `sort` parameter, we can dynamically detect at creation time if the series needs to be sorted; avoid (expensive) sorting if the time axis is already monodically increasing.

So we have the guarantee that ALL TimeSeries are always sorted AND we always get the performance gain of the figure below.

image


darts/timeseries.py
```diff
@@ -445,33 +467,29 @@ def from_values(values: np.ndarray,
...
         time_index = pd.RangeIndex(0, len(values), 1)
-
         values_ = np.reshape(values, (len(values), 1)) if len(values.shape) == 1 else values
+
         return TimeSeries.from_times_and_values(times=time_index,
                                                 values=values_,
-                                                fill_missing_dates=fill_missing_dates,
-                                                freq=freq,
-                                                columns=columns)
+                                                fill_missing_dates=False,
```
dumjax, line 486: What is the role of fill_missing_date in from_values function if it is set to false here?

darts/timeseries.py
```diff
@@ -2119,51 +2123,53 @@ def _set_freq_in_xa(xa_: xr.DataArray):
             # TODO: unit-test this
             _set_freq_in_xa(xa_)
 
-            return TimeSeries(xa_)
+            return TimeSeries(xa_, sort=True)  # note: we keep the sorting here, in case provided index is not sorted
         elif isinstance(key, pd.Int64Index):
```
dumjax, line 2127: naive question: why not proposing sort option in get item instead?

darts/timeseries.py
```diff
@@ -1467,30 +1493,8 @@ def append_values(self, values: np.ndarray) -> 'TimeSeries':
...
-        values
-            An array containing the values to replace (optional).
-
-        Returns
-        -------
-        TimeSeries
-            A new TimeSeries with updated values.
-        """
-        # TODO: I don't think this is needed... probably better to just create a new TimeSeries
-        raise NotImplementedError('TimeSeries.update() is not supported anymore.')
```
dumjax, line 1493: I know update is not a full new Timeseries, but sometimes it is easier to change values like the .loc[index, col] = new_value of pandas. What is the alternative to change just a small part of a timeseries? to extract the values via pd_daframe, change the values and create a new one?

darts/models/forecasting_model.py
```diff
@@ -151,11 +151,15 @@ def _generate_new_dates(self,
...
-            return pd.DatetimeIndex(new_dates, freq=input_series.freq_str)
+            # time_index = input_series.time_index
+            # new_dates = [
+            #     (time_index[-1] + (i * input_series.freq)) for i in range(1, n + 1)
+            # ]
+            # return pd.DatetimeIndex(new_dates, freq=input_series.freq_str)
+            start_time = input_series.end_time() + input_series.freq
+            return pd.date_range(start=start_time,
+                                 end=start_time + (n-1) * input_series.freq,
+                                 freq=input_series.freq)
```
dennisbader, line 162:
```suggestion
            return pd.date_range(start=start_time,
                                 periods=n,
                                 freq=input_series.freq)
```
Maybe using periods here could make it even simpler

darts/timeseries.py
```diff
@@ -111,7 +112,8 @@ def __init__(self, xa: xr.DataArray):
     @staticmethod
     def from_xarray(xa: xr.DataArray,
                     fill_missing_dates: Optional[bool] = False,
-                    freq: Optional[str] = None) -> 'TimeSeries':
+                    freq: Optional[str] = None,
+                    sort: Optional[bool] = True) -> 'TimeSeries':
```
dennisbader, line 116: should `sort` here and later be `Optional` or just `bool`?

darts/timeseries.py
dennisbader: I would be interested in that as well. Just out of curiosity: would we have to implement this through `__getitem__` and `__setitem__`?

dennisbader (commented):
Very cool update with nice performance boost 👍

darts/timeseries.py
hrzn: I removed it from `from_values()` because it doesn't make sense to have it here (we are creating the time axis here anyway, so there's no "missing value").

darts/timeseries.py
hrzn: I'm very rigid-minded about the immutability as you know :D
The "best" way today (not fully satisfying I agree) is probably to get the underlying DataArray (calling `data_array()`), edit the values as you want, and then create a new `TimeSeries` from there.
I don't want to provide functions that could mutate the series. However, we could have a function like `with_values_updated()` or something like this, which returns a _new_ TimeSeries with changes in the desired places.

PR #485, opened 2021-09-23
fix wrong column name in add_datetime_attribute

IncubatorShokuhou:
### Summary

Now in darts, when use `Timeseries.add_datetime_attribute()` with `one_hot = False` and `cyclic = False`, the component name is always `Month`, no matter what attribute you are using. This is caused by the mising column name of `values_df` in `datetime_attribute_timeseries` in `darts/utils/timeseries_generation.py`. This PR fixes it.


darts/utils/timeseries_generation.py
```diff
@@ -385,7 +385,8 @@ def datetime_attribute_timeseries(time_index: Union[pd.DatetimeIndex, TimeSeries
             attribute+"_cos": np.cos(freq * values)
         })
     else:
-        values_df = pd.DataFrame(values)
+        values_df = pd.DataFrame({attribute: values})
```
hrzn, line 388: Good one, thanks! I suspect however we may have the same issue for the `one_hot` case (line 368), could you maybe fix this one as well while you're at it? Thanks!

IncubatorShokuhou: @hrzn Seems that `one_hot` case does not have this issure because the column name is set in line 392-393, which is introduced in PR #246

IncubatorShokuhou:
@hrzn I pushed a new commit, in which I move the naming of the columns in `one_hot` case to the if branch in the front to make it more readable, add more details in comments and logging, and add `microsecond` and `nanosecond` in `num_values_dict`.

hrzn (approved):
Looks good - thanks @IncubatorShokuhou !

hrzn:
The test issues are due to sklearn 1.0 that has just been released and causes issues with pmdarima. https://github.com/unit8co/darts/pull/489 should fix it.

Issue #224, opened 2020-10-26
Prophet return every prediction value


fawolfmann:
**Is your feature request related to a current problem? Please describe.**
In order to create an outlier detection with Prophet, i need the full dataframe that's return Prophet

**Describe proposed solution**
Remove the hardcoded `["yhat"]` from `Prophet.predict` add a variable asking to return just `yhat` or all the predictions: `'yhat_lower', 'yhat_upper', etc..`
https://unit8co.github.io/darts/_modules/darts/models/prophet.html#Prophet.predict

Thanks very much! very useful framework

LeoTafti:
Hi @fawolfmann,

Thanks for reporting.

I think that the solution you propose could indeed be a nice small tweak to the `predict()` method of `Prophet`. It would be nice to keep the current behavior as default though, so maybe adding something like `simple_prediction:bool = True` to the `Prophet.predict()` method could do the trick.

I'll add this to our backlog. In the meantime if you have a little time, feel free to [contribute](https://github.com/unit8co/darts#contribute)!

fawolfmann:
Yes, will code it!

enesgur:
Hello @LeoTafti,

Thanks for the library. It's very useful. Is there any update about this issue?

dennisbader:
Hi @enesgur and thanks for writing and your PR as well.
We are currently working on expanding support for Prophet including this feature.
So stay tuned!

enesgur:
Hi @dennisbader.

You saved me from pure Prophet codes. :) Thanks for your development.

PR #493, opened 2021-09-27
Added raw options for Prophet 

enesgur:
**Updated**
* Added `raw` option for `prophet.predict()`.


dennisbader (commented):
Hi @enesgur and thanks for the PR!
As mentioned in #224, we are currently working on expanding Prophet including this feature!

PR #513, opened 2021-10-20
Feature/tft darts

dennisbader:
## Temporal Fusion Transformer (TFT)

From paper https://arxiv.org/pdf/1912.09363.pdf. Internal model architecture uses a majority of [PyTorch Forecasting's TFT implementation](https://pytorch-forecasting.readthedocs.io/en/latest/api/pytorch_forecasting.models.temporal_fusion_transformer.TemporalFusionTransformer.html#pytorch_forecasting.models.temporal_fusion_transformer.TemporalFusionTransformer) (PF).

### Model info:

- TFT is subclass of `TorchParametricProbabilisticForecastingModel` and `MixedCovariatesTorchModel`
- TFT uses the `MixedCovariatesSequentialDataset` loaders

### Support
- mixed covariates: past and/or future covariates
- univariate forecasting 
- multivariate forecasting
- multi-TS forecasting
- probabilistic forecasting
- multi-losses (`QuantileLoss` from `darts.models.forecasting.tft_submodels.py`) or regularx losses (such as `MSELoss`)

### Other Information
- TFT requires a future input. Right now, when there are no future_covariates passed to `fit()`, a placeholder zero tensor of length `output_chunk_length` is created.
As discussed in last synch, we could implement automatic cyclic encoding of the timeseries if there are no future_covariates?
- Interpretable output not yet accessible in this PR (such as attention information). I guess that `predict()` should only return the predicted target values? We should think of a smart way to give access to the interpretable parameters.
- Static covariates are currently not supported in this PR. The code for static covariates integration by PF was mostly kept as is, as we might want to add static covariate support to Darts in the future as well? 
- Categorical embdeding not supported in this PR. The torch submodules were kept in case we want to add this in the future. All values are treated as reals currently.


hrzn (commented):
As discussed @dennisbader , it'd be better to import the code from PyTorch forecasting directly when/where possible.

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+        # for post lstm static enrichment
+        self.static_context_enrichment = _GatedResidualNetwork(
+            input_size=self.hidden_size,
+            hidden_size=self.hidden_size,
+            output_size=self.hidden_size,
+            dropout=self.dropout
+        )
+
+        # lstm encoder (history) and decoder (future) for local processing
+        self.lstm_encoder = _LSTM(
```
hrzn, line 176: Do we need PTF's LSTM here? (Compared to PyTorch original LSTM).
As I understand the different is that PTF "handles 0-length sequences",  but I guess it's never the case we'll get a 0-length sequence as long as our input and output chunks are strictly positive?

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+
+        embeddings_varying_decoder = {name: input_vectors_future[name] for name in self.decoder_variables}
+        embeddings_varying_decoder, decoder_sparse_weights = self.decoder_vsn(
+            x=embeddings_varying_decoder,
+            context=static_context_expanded[:, encoder_length:],
+        )
+
+        # LSTM
+        # calculate initial state
+        input_hidden = self.static_context_initial_hidden_lstm(static_embedding).expand(
```
hrzn, line 393: I guess this initial GRN pass on the LSTM initial state might make sense, but i believe it's not in the TFT paper?

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+        # static enrichment
+        static_context_enriched = self.static_context_enrichment(static_embedding)
+        attn_input = self.static_enrichment_grn(
+            x=lstm_out,
+            context=self.expand_static_context(context=static_context_enriched, timesteps=timesteps)
+        )
+
+        # multi-head attention
+        attn_out, attn_out_weights = self.multihead_attn(
+            q=attn_input[:, encoder_length:],  # query only for predictions
```
hrzn, line 431: Out of curiosity (but no need to do it for this PR), I'd be interested to see how it performs without taking future values only for queries here.

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+            k=attn_input,
+            v=attn_input,
+            mask=self.get_attention_mask(
```
hrzn, line 434: I guess often this mask will be the the same (when operating on the full input+output chunks), so it might make sense to cache the "full" version to avoid re-generating and moving to device it every time.
It's nitpicking though and I doubt it'd have a big performance impact so we can leave it like that for now :)

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+                decoder_length=timesteps - encoder_length,
+                device=past_target.device
+            ),
+        )
+
+        # skip connection over attention
+        attn_out = self.post_attn_gan(x=attn_out, skip=attn_input[:, encoder_length:])
+
+        # position-wise feed-forward
+        out = self.positionwise_feedforward_grn(x=attn_out, context=None)
```
hrzn, line 445: Do you know if this is in the paper?

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+            hidden state size of the TFT. It is the main hyper-parameter and common across the internal TFT architecture.
+        lstm_layers : int
+            number of layers for the Long Short Term Memory (LSTM) Encoder and Decoder (2 is a good default).
+        num_attention_heads : int
+            number of attention heads (4 is a good default)
+        dropout : float
+            Fraction of neurons afected by Dropout.
+        hidden_continuous_size : int
+            default for hidden size for processing continuous variables
+        add_cyclic_encoder : optional str
```
hrzn, line 530: Could we find a way to make this one work for integer-indexed time series?
I think in such cases, we could "simply" generate some sine-based encodings of the positions; similar to what is done in the original transformer paper. I guess we could perhaps simply change the `datetime_attribute_timeseries()` to handle those integer-series cases too. Maybe dynamically ignoring the `attribute` argument when `time_index` is a `pd.Int64Index`.

If that is too complex, it can be left out for a future PR :) But it'd be nice to have at some point.

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,915 @@
...
+            future_covariates = self._add_cyclic_encoder(target, future_covariates=future_covariates)
+
+        return MixedCovariatesSequentialDataset(target_series=target,
+                                                past_covariates=past_covariates,
+                                                future_covariates=future_covariates,
+                                                input_chunk_length=self.input_chunk_length,
+                                                output_chunk_length=self.output_chunk_length,
+                                                max_samples_per_ts=self.max_sample_per_ts)
+
+    def _add_cyclic_encoder(self,
```
hrzn, line 722: Very nice!

hrzn (commented):
This looks great - very good job @dennisbader !

My "main" remark would be that it'd be nice to extend the cyclic encoding to work on integer-indexed series too (see comment).
Also it'd be nice to add an example notebook just for the TFT, as it is a little bit different from other models (cyclic encoding + probabilistic forecasting by default).

Let me know what you think should/should-not be part of this PR, then we can merge.

darts/models/forecasting/tft_model.py
dennisbader: good catch!

darts/models/forecasting/tft_model.py
dennisbader: Yes, see arrow below

Screenshot 2021-11-05 at 08 42 47

darts/models/forecasting/tft_model.py
dennisbader: The GRN comes from the static covariate encoders (you can see them in the bottom right of the image below in the other comment). It's not obvious from the image but when you read through the text:

From TFT paper:
image
and for the encoder LSTM:
image

darts/models/forecasting/tft_model.py
```diff
@@ -0,0 +1,836 @@
...
+                                                max_samples_per_ts=self.max_sample_per_ts)
+
+    def _verify_train_dataset_type(self, train_dataset: TrainingDataset):
+        raise_if_not(isinstance(train_dataset, MixedCovariatesTrainingDataset),
+                     'TFTModel requires a training dataset of type MixedCovariatesTrainingDataset.')
+
+    def _produce_train_output(self, input_batch: Tuple):
+        return self.model(input_batch)
+
+    def predict(self, n, *args, **kwargs):
```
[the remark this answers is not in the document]
dennisbader: If we leave `n`, and `n < self.output_chunk_length`, the inference dataset only loads the next `n` future covariates. 
So to get a complete output chunk for prediction, we would have to step back 
`-(self.output_chunk_length - n)` and then do prediction for `n_new = self.output_chunk_length`.
This means that we ignore the last values from `past_target`.

To avoid this, I intercepted calls to `predict()` so that we load a minimum of `self.output_chunk_length` points from future data.

darts/models/forecasting/tft_model.py
dennisbader: Included it :)

darts/models/forecasting/tft_model.py
dennisbader: as discussed with @hrzn, we should do this in another PR. Opened ticket [DARTS-499](https://unit8.atlassian.net/browse/DARTS-499)

darts/models/forecasting/tft_model.py
hrzn: Did you push your changes? I don't see it.

dennisbader: Now I did

hrzn (approved):
Very nice one 💯

PR #520, opened 2021-10-22
u8darts installation instructions fixed

pohndorff:
### Summary

Installation instruction using PyPI for the core + X with u8darts had quotes in the command. These are causing errors when copy-pasted. Removing the quotes solves the issue and the package with additional dependencies can be installed.


tomasvanpottelbergh:
@pohndorff could you test with double quotes? That might be a universal solution.

pohndorff:
@tomasvanpottelbergh Thank you, double quotes do work in the command and powershell terminal (Windows), i.e. `pip install "u8darts[pmdarima]"`.

hrzn (approved):

hrzn:
Perfect, thank you @pohndorff

PR #522, opened 2021-10-25
Feature/randomsearch

### LINKED ISSUE #514: Random Search
Grid searches with multiple hyperparameters gets computationally expensive very quickly as you expand the range of possible hyperparameters. My current data set takes approx 7.5 hrs for each iteration of the gridsearch, so a grid search becomes overwhelmingly expensive very quickly. Implementing a random search can search through a random subset of a parameter grid more quickly to find optimal parameters. 

I have written a simple ammendment to the GridSearch function that introduces a new optional parameter, n_samples, which defines the number of parameter combinations to test. I've made it take either an integer, which defines an absolute number of samples, or a float which takes a percentage. After the parameter cross product is calculated, it randomly takes that number of samples from the list and the iterator is created from them instead of the full cartesian product list. 

Is this a worthwhile addition to the code base? I can add it in if so.

sfrischkorn:
Fixes #514.

### Summary

Added an extra parameter, num_samples, to grid search that will take that number of parameter combinations randomly, either as an absolute number if an integer, or a percentage if a float. The sampling uses random.sample and is implemented in a private method to allow it to be tested.

Tests have been implemented and the new code is all covered.
This is the first time I've done a pull request, hopefully I've done it right :)


hrzn (approved):
Thanks! We'll merge once we have fixed the issues with documentation.

Issue #510, opened 2021-10-12
Train a model, but don't save model artifacts


jonhilgart22:
**Is your feature request related to a current problem? Please describe.**
A clear and concise description of what the problem is.

I'd like to train a Nbeats and TCN model but not save the model artifacts. I'm running this code on AWS Lambda which has size restrictions for these model files.

**Describe proposed solution**
A clear and concise description of what the library should provide to solve missing functionality.

Can we pass in a param here https://github.com/unit8co/darts/blob/484a184a97be37e64ed62f6fa6fad2bdef898adf/darts/models/forecasting/torch_forecasting_model.py#L859 indicating whether save=True or save=False ?

**Describe potential alternatives**
A clear and concise description of any alternative solutions or existing features that might solve it.

**Additional context**

dennisbader:
Hi @jonhilgart22 and thanks for pointing this out.

Currently, the models are saved automatically without the option to bypass it.

We added this feature to our to-do list!

hrzn:
Since v0.13.0, checkpoints are not saved by default anymore (you can still save them by passing `save_checkpoints=True` to the model).

Issue #504, opened 2021-10-05
How NBEATS stops writing numerous checkpoints files


lyn-bupt:
Whenever I use NBEATS, hundreds of GB of checkpoint files are created. I also tried setting `work_dir=None` in the `NBEATSModel()`, but this will report an error.

microprediction:
Oh my goodness I just discovered the same issue. Is there a workaround?

PR #523, opened 2021-10-25
Feature/torchmodel saving

dennisbader:
Addresses #510, #504.

### Summary
General updates to `TorchForecastingModel`'s save and load methods.
- automatic saving of untrained model and checkpoints can by bypassed by creating model with `save_checkpoints=False`
- model can be manually saved with `YourTorchModel.save_model(file_path)` (file_path pointing to the .pth.tar file).
- model can be manually loaded with `YourTorchModel.load_model(file_path)` or the original method `YourTorchModel.load_from_checkpoint()`

#### Other Information

- added unit tests to compare manual save/load with original automatic save/load


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -95,7 +95,8 @@ def __init__(self,
                  log_tensorboard: bool = False,
                  nr_epochs_val_period: int = 10,
                  torch_device_str: Optional[str] = None,
-                 force_reset=False):
+                 force_reset=False,
+                 save_checkpoints=True):
```
hrzn, line 99: I'm actually thinking that we should perhaps default this one to `False`, wdyt? I think it can cause issues for some users because it's happening silently, can generate files quite large, and even though we keep only the last 5, each model name is (by default) unique so each model instance will generate 5 files by default.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -924,22 +981,29 @@ def _prepare_tensorboard_writer(self):
...
             The name of the model (used to retrieve the checkpoints folder's name).
         work_dir
             Working directory (containing the checkpoints folder). Defaults to current working directory.
-        filename
+        file_name
             The name of the checkpoint file. If not specified, use the most recent one.
         best
-            If set, will retrieve the best model (according to validation loss) instead of the most recent one.
+            If set, will retrieve the best model (according to validation loss) instead of the most recent one. Only
+            works when `file_name` is not given.
```
hrzn, line 1006:
```suggestion
            is ignored when `file_name` is given.
```
```suggestion
            works when `file_name` is not given.
```

hrzn (approved):
Nice enhancement! 
I'd just like to propose that we set `save_checkpoints=False` by default.

darts/models/forecasting/torch_forecasting_model.py
dennisbader: agreed!

hrzn:
```suggestion
                 save_checkpoints=False):
```

hrzn (approved):

darts/models/forecasting/forecasting_model.py
```diff
@@ -701,6 +733,9 @@ def _sample_params(model_class, params, n_random_samples):
                          "If supplied as a float, n_random_samples must be greater than 0.0 and less than 1.0.")
             return sample(params, int(n_random_samples * len(params)))
 
+    def untrained_model(self):
```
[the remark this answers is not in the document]
dennisbader: New approach uses a meta-class `MetaModel` on parent class `ForecastingModel`.
With `MetaModel`, model creation parameters can be extracted and stored.
This approach gets rid of public issue, the need to pass locals() and avoids a lot of repetition

hrzn (approved):
Nice improvement 👍

PR #525, opened 2021-10-29
docs/covariates

dennisbader:
# Covariates tutorial

Added a tutorial on covariates in Darts to clarify the topic for newcomers to the library.
This is to my understanding how covariates work/are used in Darts. So feel free to correct me if I misinterpreted something or if there is anything to add!

## Overview
Section 1 and section 2 cover the most important points:

- What are covariates in general and in Darts
- How to find out which type of covariate your variable is
- Intro to Darts' `XCovariatesModel` and which covariate types they support
- Covariate support for each of Darts' forecasting models
- How to use covariates in Darts' Standard and Torch Forecasting Models

Section 3 gives an in-depth guide of how covariates are used in Darts' Torch Forecasting Models and which requirements they have to fulfil.

## Additional info
what is not covered:
- covariate length requirements for regression models (maybe we can go into that in the regression model tutorial that we planned?
- covariate requirements for torch forecasting models with shifted datasets (the ones that use training length instead of output_chunk_length)


dennisbader:
closed as old version of #691

PR #526, opened 2021-10-29
Feat/quantile loss as likelihood

hrzn:
Transformed the `QuantileLoss` into a new `QuantileRegression` class, which makes it possible to use with other probabilistic models as well. TFT is also using this now by default.

The sampling is made by sampling the quantile values directly, with a "trick" relying on the presence of the median in the quantiles in order to avoid over-sampling any of the smallest or highest quantiles.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ Some of the models depend on `prophet` and `torch`, which have non-Python depend
 A Conda environment is thus recommended because it will handle all of those in one go.
 
 ### From conda-forge
-Currently only the x86_64 architecture with Python 3.7 or 3.8 
+Currently only the x86_64 architecture with Python 3.7-3.9
 is fully supported with conda; consider using PyPI if you are running into troubles.
 
 To create a conda environment for Python 3.7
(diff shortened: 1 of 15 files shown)


dennisbader (approved):
Nice one! I I like how we can use quantile regression now for probabilistic forecasting as well.
I changed one or two minor things, which had to be adjusted now that QuantileRegression is a likelihood model and not a regular loss anymore (TFT forward output shape).

And a last point about how many quantiles we should use per deafult in QuantileRegression:
I assumed that the number of quantiles would not have an a large effect (at least) on the median quantile, but apparently it does (same results in original TFT branch with QuantileLoss).
I'm not sure if this is how it's supposed to work. What do you think?
- first plot: quantiles=[0.02, 0.1, 0.25, 0.5, 0.75, 0.9, 0.98]
- second plot: quantiles=[0.01, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.4, 0.5, 0.6, 0.7, 0.75, 0.8, 0.85, 0.9, 0.95, 0.99]

tft_5_quantiles
:
tft_17_quantiles

PR #532, opened 2021-11-01
Fix/regression model multivariate

### LINKED ISSUE #508: [BUG] RegressionModel can't predict
**Bug description**
LightGBMModel, RandomForest and I believe any other regression model is not able to forecast multivariate (11 target values) time series.
The model.fit() is successful, but when I try model.predict() on test series or even the same series it was trained, I get ValueError: conflicting sizes for dimension 'component': length 1 on the data but length 11 on coordinate 'component'

**To Reproduce**
`train = [TimeSeries.from_dataframe(df_series.asfreq('MS'), fill_missing_dates=True) for df_series in df_set]`
`train = [fill_missing_values(series) for series in train]`
`model = LightGBMModel(lags=12, random_state=67)`
`model.fit(train)`
`preds = model.predict(n=1, series=series)`

**System:**
 - Python version: 3.7.11
 - darts version 0.12.0

**Additional context**
When I use the same data preprocessing and training pipeline with neural network models (NBEATSModel, TCNModel) there are no such bugs.

brunnedu:
Addresses Fixes #508, DARTS-466

### Summary

- Fixed preprocessing of RegressionModel to work with multivariate timeseries
- Added MultiOutputRegressor wrapper to RegressionModel to handle multi-output regression for multivariate timeseries
- Fixed old unit tests that relied on univariate timeseries assumption & added new unit tests for multivariate timeseries
- Renamed file containing LinearRegressionModel from lgbm_model.py to linear_regression_model.py and modified __init__.py to import from linear_regression_model

### Other Information

I haven't written a lot of additional unit tests. If you have any idea what else we should test for please let me know and I would add tests for those cases too.


hrzn (approved):
LGTM, good job!

PR #533, opened 2021-11-02
Docs/cov update

hrzn:
## Changes I made
* Add a first "TL;DR" section at the beginning (most users won't have time to read everything, we should extract the gist for them)
* Various modifications / fixes
* Changed the example with ice-creams (in Sections 1 and 2), in order to use *measured* traffic in a neighbouring store and *predicted* temperature. I think it might be confusing to have to distinguish between *measured* temperature and *estimated* temperature; I thought it would be better to simply use different quantities (traffic in neighbourhood for past covariates instead of temperature).
* I tried to trim down a bit Section 3, which (IMO) had a lot of repetitive elements.

## Overall review of the document
Thank you for the initiative, and for giving a stab at this topic, which is pretty tricky to explain :) It's a very nice start, but I think it could still be improved, along the following lines:
* Try to make it more crisp. Minimize the content that people will have to go over. Right now it is a bit overwhelming and somewhat inflates the impression of complexity of the topic. I tried to trim down Section 3 a bit already, but I think more could be done.
* There are quite a few repetitions (for instance, it mentions and explains the case with auto-regressive predictions for the case `n > output_chunk_length` many times). I think here too it could be more crips and avoid repetitions. Minimizing the number of sentences, sections, etc, and making them very direct and to the point.
* You try to introduce quite a few notions which are a bit separate (IMO) from the topic - like how the loss is computed in neural nets. Again - there are opportunities for simplification I think.

To re-works (shuffle a bit and simplify) the document, here's a proposed (draft) outline:

* Section 1 - A high-level "theoretical" intro to the topic.
  * Definitions of past/future covariates
  * Simple explanation that sometimes past values of future covs are needed, and sometimes future values of past covs are needed.
  * Table showing the models and what they support need
  * List of required *time spans* (!= lengths) of past & future covariates
  
* Section 2 - Example of how to train a model with past and future covariates, with code snippets. You can reuse my "river flow" example from here (https://medium.com/unit8-machine-learning-publication/time-series-forecasting-using-past-and-future-external-data-with-darts-1f0539585993) if you want

* Section 3 - How ML models work under the hood
  * Datasets and slicing
  * "step-by-step" example (like what you already have)

### DIFF
--- a/COVARIATES.md
+++ b/COVARIATES.md
@@ -203,14 +203,15 @@ Model | Standard <br /> Forecasting Model | Torch <br /> Forecasting Model | `Pa
 `NBEATSModel` |  | 🟢 | ✅ |  |  |  |  |
 `TCNModel` |  | 🟢 | ✅ |  |  |  |  |
 `TransformerModel` |  | 🟢 | ✅ |  |  |  |  |
-`TFT` |  | 🟢 |  |  |  | ✅ |  |
+`TFT` (WIP) |  | 🟢 |  |  |  | ✅ |  |
 
 **Table 2: Darts' forecasting models and their `{X}CovariatesModel`**
 
 `*` models that do not support covariates
 
 `**` `RegressionModel` including `RandomForest`, `LinearRegressionModel` and `LightGBMModel`. `RegressionModel` is a
-special kind of model which uses lags on covariates and past targets to do predictions.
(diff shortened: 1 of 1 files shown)


dennisbader:
closed as old version of #691

PR #534, opened 2021-11-03
Fix rendering of lists in examples/*

sotte:
In the example notebooks newlines are added before list such they render properly when build with sphinx.

### Summary
The markdown that pandoc uses requires new lines before list items. I just added the missing newlines.

Before:
img_journal_1

After:
img_journal_2

How to reproduce:
```
./gradlew buildDocs
# Then look at an example
xdg-open docs/build/html/examples/02-multi-time-series-and-covariates.html
```


brunnedu:
Hi @sotte,

Thanks a lot for pointing out this mistake!
Since I've seen this mistake in some other places I will soon push some additional corrections to this branch and will then merge it into our master branch.

hrzn (approved):
Very good one, actually it was bugging me for a while :D thanks a lot @sotte !

sotte:
@brunnedu I actually thought I caught all of the cases, but that is obviously wrong. Code reviews are good ;)

Thanks so much and keep up the great work!


(Side note:
With pandoc one can also specify the markdown variant. It might make sense to switch to a version that does not require newlines before lists. But of cause this change might have some other unwanted ramifications.)

brunnedu:
@sotte 
Thank you for quickly helping us out when you noticed a problem with our formatting!
Thanks for the tip, as we are currently revising some things regarding our docs, this could be very useful.

PR #535, opened 2021-11-04
Feat/autoregressive timeseries generation

brunnedu:
### Summary

- Added autoregressive_timeseries function to generate autoregressive timeseries with specific coefficients.
- Added unit tests for autoregressive_timeseries function.

### Other Information

Wrote this function for myself to sanity check some coefficients of the LinearRegressionModel and then thought why not add it to the timeseries_generation module.

Feel free to add suggestions on additional hyperparameters or how I could make the explanation of its functionality clearer in the docs.


darts/utils/timeseries_generation.py
```diff
@@ -308,6 +308,65 @@ def random_walk_timeseries(mean: float = 0.,
     return TimeSeries.from_times_and_values(index, values, freq=freq, columns=pd.Index([column_name]))
 
 
+def autoregressive_timeseries(coef: Sequence[float],
```
hrzn, line 311: How about `coefs` ?

darts/utils/timeseries_generation.py
```diff
@@ -308,6 +308,65 @@ def random_walk_timeseries(mean: float = 0.,
...
+                              length: Optional[int] = None,
+                              freq: str = 'D',
+                              column_name: Optional[str] = 'autoregressive') -> TimeSeries:
+    """
+    Creates a univariate, autoregressive TimeSeries whose values are calculated using specified coefficients `coef` and
+    starting values `start_values`.
+
+    Parameters
+    ----------
+    coef : list of float
```
hrzn, line 324:
```suggestion
    coef : Sequence of float
```

darts/utils/timeseries_generation.py
```diff
@@ -308,6 +308,65 @@ def random_walk_timeseries(mean: float = 0.,
...
+        The autoregressive coefficients used for calculating the next time step.
+        series[t] = coef[-1] * series[t-1] + coef[-2] * series[t-2] + ... + coef[0] * series[t-len(coef)]
+    start_values : list of float, default: np.ones(len(coef))
```
hrzn, line 327:
```suggestion
    start_values : Sequence of float, default: np.ones(len(coef))
```

darts/utils/timeseries_generation.py
```diff
@@ -308,6 +308,65 @@ def random_walk_timeseries(mean: float = 0.,
...
+        The starting values used for calculating the first few values for which no lags exist yet.
+        series[0] = coef[-1] * starting_values[-1] + coef[-2] * starting_values[-2] + ... + coef[0] * starting_values[0]
+    start : pd.Timestamp or int, optional
+        The start of the returned TimeSeries' index. If a pandas Timestamp is passed, the TimeSeries will have a pandas
+        DatetimeIndex. If an integer is passed, the TimeSeries will have a pandas Int64Index index. Works only with
+        either `length` or `end`.
+    end : pd.Timestamp or int, default: pd.Timestamp('2000-01-01')
```
hrzn, line 334: I think the default is on `start`, not `end`.

hrzn (approved):
Nice 👍

darts/utils/timeseries_generation.py
brunnedu: I also don't like 'list' but it is what the numpydoc styleguide suggests [numpydoc - styleguide - parameters](https://numpydoc.readthedocs.io/en/latest/format.html#:~:text=4.-,Parameters,-%C2%B6)
Should I still change it to 'Sequence' ?

darts/utils/timeseries_generation.py
brunnedu: Thought about naming it `coefs` but then found that `coefs` is way less popular than `coef` as a parameter name. Also sklearn uses `coef_` to get the model coefficients. But yes coefs would make it clearer that we can take multiple coefficients...

hrzn: OK, we can leave `coef` then.

darts/utils/timeseries_generation.py
hrzn: I didn't know that! Let's leave `list` then.

Issue #515, opened 2021-10-21
Easier inheritance of TimeSeries object


rbagd:
I was looking to extend `TimeSeries` object to make it easier to convert it to other not yet implemented and internal formats. Usually, I'd do something like

```python
from darts import TimeSeries as DartsTimeSeries

class TimeSeries(DartsTimeSeries):
    def to_my_custom_format(self):
        pass

```

The problem with this approach is that all factory methods in `TimeSeries` are static and are not aware of the underlying class, which means that

```python
from my_module import TimeSeries
TimeSeries.from_dataframe(df, "time", "value")   # -> darts.timeseries.TimeSeries
```
still returns `darts.timeseries.TimeSeries` object without `to_my_custom_format()` method. Is there a reason why factory methods are implemented as `@staticmethod` and not `@classmethod`? The latter would be more friendly from inheritance point of view.

hrzn:
Yes, this is a good point. There's no good reason and we'll make them classmethods.

hrzn:
@rbagd Could you have a quick look at this PR and tell me if that would solve your issues (or if I have forgotten corner cases?)
https://github.com/unit8co/darts/pull/536

rbagd:
Thanks to @hrzn, factory methods got converted to classmethod, so this can be closed.

GuillaumeDMMarion:
I think there is still another case for making inheritance easier that is (AFAIK) not yet solved by #536. 

In some methods of the TimeSeries class, instantiation is done by explicitly calling TimeSeries. Some examples of methods where this is done: head, tail, but also \_\_abs__ etc. I illustrate how this impedes easier inheritance below:

```python
import pandas as pd
from darts.timeseries import TimeSeries # from feature/classmethods branch

class Child(TimeSeries):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        
df = pd.DataFrame({'time':pd.date_range('20210101', periods=10), 'value': range(10)})
ts = Child.from_dataframe(df, time_col='time')
print(type(ts), type(ts.head()))
```

The above code will result in: 
<class '__main__.Child'> <class '__main__.TimeSeries'>

If one would change the current ```TimeSeries``` call in the head method to ```self.__class__``` the result then becomes:
<class '__main__.Child'> <class '__main__.Child'>

Thus retaining any methods defined in Child.

hrzn:
> I think there is still another case for making inheritance easier that is (AFAIK) not yet solved by #536.
> 
[...]

PR #536, opened 2021-11-07
factory methods as classmethod instead of staticmethod

hrzn:
Solves https://github.com/unit8co/darts/issues/515


dennisbader (approved):
Nice, looks good to me 👍

PR #545, opened 2021-11-11
Feat/pytorch perf improvements

hrzn:
* Add `num_loader_workers` parameters to `fit()` and `predict()` methods of Torch models, in order to control the `num_workers` of PyTorch `DataLoader`s.
* New method `TimeSeries.astype()` to easily cast `np.float64` <--> `np.float32`.
* Add `dtype` as an option to the time series generation modules
* Change the TFT method generating cyclic encodings to use the right dtype.
* Added a small performance guide in a new `guides/` folder. (@dennisbader maybe we can put your covariates guide in there too?)

Another change, not related to performance:
* Replace the construct and factory calls of `TimeSeries` with `self.__class__` to make inheritance easier


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -580,6 +594,11 @@ def predict(self,
         num_samples
             Number of times a prediction is sampled from a probabilistic model. Should be left set to 1
             for deterministic models.
+        num_loader_workers
+            Optionally, an integer specifying the ``num_workers`` to use in PyTorch ``DataLoader`` instances,
+            both for the training and validation loaders (if any).
```
dennisbader, line 599:
```suggestion
            for the inference/prediction dataset loaders (if any).
```

darts/timeseries.py
```diff
@@ -829,6 +833,22 @@ def quantiles_df(self, quantiles: Tuple[float] = (0.1, 0.5, 0.9)) -> pd.DataFram
...
+        dtype
+            A NumPy dtype (np.float32 or np.float64)
+
+        Returns
+        -------
+        TimeSeries
+            A TimeSeries having the desired dtype.
+        """
+        return self.__class__(self._xa.astype(dtype))
+
```
dennisbader, line 851: Nice one, I always wondered how we'd best convert dtypes for our torch models! 
Could we image adding option argument `dtype` to all of the `TimeSeries.from_*()` methods?
That way, users could avoid having to regenerate the time series to convert the values?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -245,14 +245,15 @@ def _init_model(self) -> None:
         """
 
         # the tensors have shape (chunk_length, nr_dimensions)
-        model = self._create_model(self.train_sample)
+        self.model = self._create_model(self.train_sample)
```
dennisbader, line 248: Cool!

guides/performance.md
```diff
@@ -0,0 +1,91 @@
+# Darts Performance Recommendations
```
dennisbader, line 1: Very nice!

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -493,7 +506,7 @@ def fit_from_dataset(self,
         train_loader = DataLoader(train_dataset,
                                   batch_size=self.batch_size,
                                   shuffle=True,
-                                  num_workers=0,
+                                  num_workers=num_loader_workers,
```
dennisbader, line 509: Do you also get warnings for every epoch?
Warning:
[W ParallelNative.cpp:214] Warning: Cannot set number of intraop threads after parallel work has started or after set_num_threads call when using native parallel backend (function set_num_threads)

image

to reproduce (also checked with other test cases):
```

x = linear_timeseries(length=100, freq='h')
model = TFTModel(input_chunk_length=10, output_chunk_length=10, add_cyclic_encoder='hour')
model.fit(x, num_loader_workers=2, verbose=True)
```
version:
python==3.7.11
torch==1.10.0
latest dev-all.txt requirements

dennisbader (commented):
very nice update! just a couple of minor comments and an warning I get when fitting with `num_loader_workers > 0`

darts/timeseries.py
hrzn: Yes we could but not sure yet if it would be really needed so I'd rather leave that to a future PR - it's not too bad now:
`my_series = TimeSeries.from_csv(fname).astype(np.float32)` -vs-
`my_series = TimeSeries.from_csv(fname, dtype=np.float32)`

guides/performance.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Data in Memory and I/O bottlenecks
+It's helpful to load all your `TimeSeries` in memory upfront if you can.
+Darts offers the possibility to train models on any `Sequence[TimeSeries]`,
+which means that for big datasets, you can write your own `Sequence` implementation, and read the time series lazily from disk. This will typically incur a high I/O cost, though. So when training on multiple series, first try to build a simple `List[TimeSeries]` upfront, and see if it holds in the computer memory.
+
+-------------
+
+## Example Benchmark
+As an example, we show here the time required to train one epoch on the energy dataset (`darts.datasets.EnergyDataset`), which consists of one multivariate series that is 28050 timesteps long and has 28 dimensions.
```
hrzn, line 40:
```suggestion
As an example, we show here the time required to train one epoch on the first 80% of the energy dataset (`darts.datasets.EnergyDataset`), which consists of one multivariate series that is 28050 timesteps long and has 28 dimensions.
```

darts/models/forecasting/torch_forecasting_model.py
hrzn: Mmmh no I don't get it. It might be an issue with OS/Python/PyTorch (not sure)
https://stackoverflow.com/questions/64772335/pytorch-w-parallelnative-cpp206
In any case, it won't show by default as `num_loader_workers=0` by default, so we can leave it and hope it'll solve itself in the future...

PR #547, opened 2021-11-12
Feat/tft relative index

dennisbader:
### Summary

- relative time index can now be generated without having to specify any future covariates
- @hrzn I think this is the closest we get to a "historic"-only TFT as:
  - it creates the required decoder part if no future covariates are given
  - relative index is identical for all samples since input_ - and output_chunk_length stay the same.
  - if the relative index doesn't generate any additional value, the TFT should be able to ignore it through its Variable Selection Networks.
  - (+) relative index can indeed be helpful! It was most obvious when input and output chunk length were a multiple of a seasonal period.

### DIFF
--- a/darts/models/forecasting/tft_model.py
+++ b/darts/models/forecasting/tft_model.py
@@ -84,7 +84,8 @@ class _TFTModule(nn.Module):
             Fraction of neurons afected by Dropout.
         add_relative_index : bool
             Whether to add positional values to future covariates. Defaults to `False`.
-            This gives a value to the position of each step from input and output chunk relative to the prediction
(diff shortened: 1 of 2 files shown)


hrzn (approved):
@dennisbader Thanks for this. As discussed, I'm surprised that it can actually be helpful in some cases, because, like you say, it adds a constant component to each sample. I would expect that adding something constant that doesn't depend on the sample shouldn't help models on average. It might be worth doing some experiments to try (dis)proving this.

Nevertheless we can leave it for now, it doesn't hurt for sure. I still think that some global (non-relative) encodings might greatly help all future covariates models though. Anything encoding a global aspect of the time axes of the series (like the year, or the integer index). We could add it at some later point.

PR #548, opened 2021-11-12
A probabilistic version of NBEATS

hrzn:
This is a probabilistic version of N-BEATS. N-BEATS is more tricky than the other models to make it probabilistic, because of the way the forecasts are built internally.

The way I've done it is by changing the dimensionality of the tensors returned by the blocks, from `(target_length,)` to `(target_length, nr_params)`, and then reshaping everything correctly to have the parameters at the right place in the output tensors. It requires care, because the time dimension may already contain interleaved dimensions of multivariate series (as N-BEATS is all univariate internally), but I think this is a way to do it that respects the architectural benefits of N-BEATS. There might still be better ways to do it, though.

I've tested with multivariate series, and also with/without generic architecture (see an example with a QuantileRegression Likelihood below).

* Still to do: write a couple of unit tests

image

### DIFF
--- a/darts/models/forecasting/nbeats.py
+++ b/darts/models/forecasting/nbeats.py
@@ -197,7 +197,6 @@ class _Stack(nn.Module):
             The number of parameters of the likelihood (or 1 if no likelihood is used)
         expansion_coefficient_dim
             The dimensionality of the waveform generator parameters, also known as expansion coefficients.
-            Only used if `generic_architecture` is set to `True`.
         input_chunk_length
             The length of the input sequence fed to the model.
         target_length
--- a/darts/models/forecasting/nbeats.py
+++ b/darts/models/forecasting/nbeats.py
@@ -31,46 +31,34 @@ class _TrendGenerator(nn.Module):
 
     def __init__(self,
                  expansion_coefficient_dim,
-                 target_length,
-                 nr_params):
+                 target_length):
         super(_TrendGenerator, self).__init__()
 
-        # coefs is of size (expansion_coefficient_dim, target_length)
-        coefs = torch.stack([(torch.arange(target_length) / target_length)**i for i in range(expansion_coefficient_dim)], 
+        # basis is of size (expansion_coefficient_dim, target_length)
+        basis = torch.stack([(torch.arange(target_length) / target_length)**i for i in range(expansion_coefficient_dim)], 
                             dim=1).T
 
-        # to make it work seamlessly with likelihood parameters, 
-        # we tile it to size (expansion_coefficient_dim, target_length)
-        if nr_params is not None:
-            coefs = torch.tile(coefs, dims=(1, 1, nr_params))
-        
-        self.T = nn.Parameter(coefs, False)
+        self.basis = nn.Parameter(basis, requires_grad=False)
 
 
     def forward(self, x):
-        return torch.matmul(x, self.T)
+        return torch.matmul(x, self.basis)
 
 
 class _SeasonalityGenerator(nn.Module):
 
     def __init__(self,
-                 target_length,
-                 nr_params):
+                 target_length):
         super(_SeasonalityGenerator, self).__init__()
         half_minus_one = int(target_length / 2 - 1)
         cos_vectors = [torch.cos(torch.arange(target_length) * 2 * np.pi * i) for i in range(1, half_minus_one + 1)]
         sin_vectors = [torch.sin(torch.arange(target_length) * 2 * np.pi * i) for i in range(1, half_minus_one + 1)]
-        coefs = torch.stack([torch.ones(target_length)] + cos_vectors + sin_vectors, dim=1).T
+        basis = torch.stack([torch.ones(target_length)] + cos_vectors + sin_vectors, dim=1).T
 
-        # to make it work seamlessly with likelihood parameters, 
-        # we tile it to size (expansion_coefficient_dim, target_length)
-        if nr_params is not None:
-            coefs = torch.tile(coefs, dims=(1, 1, nr_params))
-
-        self.S = nn.Parameter(coefs, False)
+        self.basis = nn.Parameter(basis, requires_grad=False)
 
     def forward(self, x):
-        return torch.matmul(x, self.S)
+        return torch.matmul(x, self.basis)
 
 
 class _Block(nn.Module):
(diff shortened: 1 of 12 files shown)


dennisbader (commented):
Very nice, now all of the TorchModels support probabilistic forecasts!
I added a comment on the way the model output is reshaped at the end of `forward()`.
Apart from that, everything looks good to me!

darts/models/forecasting/nbeats.py
```diff
@@ -346,18 +407,18 @@ def forward(self, x):
...
-        # the covariates are by construction added as extra time series on the right side. So we need to get rid of this
-        # right output
-        y = y[:, :, :self.output_dim]
+        # In multivariate case, we get a result x1, y1, z1, x2, y2, z2, ... we want to reshape to original format
+        # In addition, we reshape y in order to have consecutive "blocks" containing the different parameters along the
+        # last dimension. We also get rid of the covariates and keep only the target dimensions.
+        # The covariates are by construction added as extra time series on the right side. So we need to get rid of this
+        # right output (keeping only :self.output_dim).
+        y = torch.cat([y[:,:,i].reshape(y.shape[0], self.output_chunk_length, self.input_dim)[:, :, :self.output_dim]
+                       for i in range(self.nr_likelihood_params)], dim=2)
```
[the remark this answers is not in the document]
hrzn: Mmmh I think you're right, but we may have an inconsistency here:
* `QuantileRegression` requires inputs (i.e. model outputs) to be of shape `(batch_size, n_timesteps, n_components, n_quantiles)` (it does the reshaping into this shape when sampling and computing the loss),
* The other `Likelihood` objects having several parameters assume a shape of `(batch_size, n_timesteps, n_components * n_params)` and they take consecutive blocks of the last dimensions as the parameters. For instance in the Gaussian case:
```
mu = model_output[:, :, :output_size // 2]
sigma = self.softplus(model_output[:, :, output_size // 2:])
```
I'm not sure that by just reshaping we get the right "blocks" in the right place, we might have to change the way either Likelihoods or QuantileRegressions take their parameters from the model outputs... (but I still have to think about it :D). Another solution (maybe quite clean) would be to force all models outputs to be of shape `(batch_size, n_timesteps, n_components, n_params)` rather than "abusing" the last dimension.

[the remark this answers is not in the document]
dennisbader: I agree!
I vote for `(batch_size, n_timesteps, n_components, n_params)` as it's quite clean and similar to the shape we have for TimeSeries with samples. 
For this, we would have to unsqueeze `y_true` in `dim=-1` to compute the losses correctly for `MSELoss()` etc. as well (I believe currently `y_true` comes as `(batch_size, n_timesteps, n_components)` right now)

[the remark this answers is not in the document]
hrzn: OK, I'll give it a try later.

darts/models/forecasting/nbeats.py
```diff
@@ -30,48 +31,72 @@ class _TrendGenerator(nn.Module):
...
-        self.T = nn.Parameter(
-            torch.stack([(torch.arange(output_dim) / output_dim)**i for i in range(expansion_coefficient_dim)], 1),
-            False)
+
+        # coefs is of size (expansion_coefficient_dim, target_length)
+        coefs = torch.stack([(torch.arange(target_length) / target_length)**i for i in range(expansion_coefficient_dim)], 
+                            dim=1).T
+
+        # to make it work seamlessly with likelihood parameters, 
+        # we tile it to size (expansion_coefficient_dim, target_length)
```
pennfranc, line 43:
```suggestion
        # we tile it to size (expansion_coefficient_dim, target_length, nr_params)
```?

File: darts/models/forecasting/nbeats.py [hunk 3]
```diff
@@ -30,48 +31,72 @@ class _TrendGenerator(nn.Module):
...
-                 output_dim):
+                 target_length,
+                 nr_params):
         super(_TrendGenerator, self).__init__()
-        self.T = nn.Parameter(
-            torch.stack([(torch.arange(output_dim) / output_dim)**i for i in range(expansion_coefficient_dim)], 1),
-            False)
+
+        # coefs is of size (expansion_coefficient_dim, target_length)
+        coefs = torch.stack([(torch.arange(target_length) / target_length)**i for i in range(expansion_coefficient_dim)],
```
pennfranc, line 39: Nitpicking, but `coefs` might not be the best name here because the paper refers to the values that are multiplied with this matrix as 'expansion coefficients' already, so this might be confusing. Maybe we could call it `basis`... what do you think?

darts/models/forecasting/nbeats.py
```diff
@@ -30,48 +31,60 @@ class _TrendGenerator(nn.Module):
...
         ----------
         num_layers
             The number of fully connected layers preceding the final forking layers.
         layer_width
             The number of neurons that make up each fully connected layer.
+        nr_params
+            The number of parameters of the likelihood (or 1 if no likelihood is used)
         expansion_coefficient_dim
             The dimensionality of the waveform generator parameters, also known as expansion coefficients.
             Only used if `generic_architecture` is set to `True`.
```
pennfranc, line 90: I just realized this is not true, because the trend module also needs the `expansion_coefficient_dim`, which it uses as degree for the polynomial basis.

```suggestion
            Used in the generic architecture and the trend module of the interpretable architecture, where it determines the degree of the polynomial basis.
```

pennfranc: I guess you might want to split this into 2 lines or something..

darts/models/forecasting/nbeats.py
```diff
@@ -166,6 +191,8 @@ def __init__(self,
             The number of fully connected layers preceding the final forking layers in each block.
         layer_width
             The number of neurons that make up each fully connected layer in each block.
+        nr_params
+            The number of parameters of the likelihood (or 1 if no likelihood is used)
         expansion_coefficient_dim
             The dimensionality of the waveform generator parameters, also known as expansion coefficients.
             Only used if `generic_architecture` is set to `True`.
```
pennfranc, line 198: This also has to be updated.

pennfranc: But in the `_NBEATSModule` and `NBEATSModel` classes the description does not have to be changed, because those still accept a separate parameter for the polynomial degree.

darts/models/forecasting/nbeats.py
```diff
@@ -108,13 +123,14 @@ def __init__(self,
...
+        # Fully connected layer producing forecast/backcast expansion coeffcients (waveform generator parameters).
+        # The coefficients are emitted for each parameter of the likelihood.
         if g_type == _GType.SEASONALITY:
             self.backcast_linear_layer = nn.Linear(layer_width, 2 * int(input_chunk_length / 2 - 1) + 1)
-            self.forecast_linear_layer = nn.Linear(layer_width, 2 * int(target_length / 2 - 1) + 1)
+            self.forecast_linear_layer = nn.Linear(layer_width, nr_params * (2 * int(target_length / 2 - 1) + 1))
         else:
             self.backcast_linear_layer = nn.Linear(layer_width, expansion_coefficient_dim)
-            self.forecast_linear_layer = nn.Linear(layer_width, expansion_coefficient_dim)
+            self.forecast_linear_layer = nn.Linear(layer_width, nr_params * expansion_coefficient_dim)
```
pennfranc, line 133: I think this is good now, we also have a nice consistency in how we create multiple likelihood parameters between generic and interpretable architectures.

darts/models/forecasting/nbeats.py
```diff
@@ -30,48 +31,60 @@ class _TrendGenerator(nn.Module):
...
         super(_SeasonalityGenerator, self).__init__()
-        half_minus_one = int(output_dim / 2 - 1)
-        cos_vectors = [torch.cos(torch.arange(output_dim) * 2 * np.pi * i) for i in range(1, half_minus_one + 1)]
-        sin_vectors = [torch.sin(torch.arange(output_dim) * 2 * np.pi * i) for i in range(1, half_minus_one + 1)]
-        self.S = nn.Parameter(torch.stack([torch.ones(output_dim)] + cos_vectors + sin_vectors, 1),
-                              False)
+        half_minus_one = int(target_length / 2 - 1)
+        cos_vectors = [torch.cos(torch.arange(target_length) * 2 * np.pi * i) for i in range(1, half_minus_one + 1)]
+        sin_vectors = [torch.sin(torch.arange(target_length) * 2 * np.pi * i) for i in range(1, half_minus_one + 1)]
+        basis = torch.stack([torch.ones(target_length)] + cos_vectors + sin_vectors, dim=1).T
```
pennfranc, line 56: could we add some comment here like 'basis is of size `(2 * int(target_length / 2 - 1) + 1, target_length)`? Similarly to the comment in the `_TrendGenerator` class.

darts/models/forecasting/nbeats.py
hrzn: Actually I think this parameter is used only when `generic_architecture` is `True`. When it is `False, we are using `trend_polynomial_degree + 1` (line 354) for the trend (and `2 * int(input_chunk_length / 2 - 1) + 1` for the seasonality).

pennfranc: Yeah that's right but at line 354 we are passing `trend_polynomial_degree + 1` as `expansion_coefficient_dim` parameter to the stack module. The stack module does not accept a separate `trend_polynomial_degree` parameter. So I the description is correct for the `_NBEATSModule` and `NBEATSModel` classes, but from `_Stack` downward it is not correct anymore because the trend module also uses `expansion_coefficient_dim` to determine the basis dimension. Does that make sense?

hrzn: Right, makes sense 👍

PR #554, opened 2021-11-17
Feat/loader improvement

dennisbader:
### Summary

- improved performance of `GenericShiftedDataset` loader through:
  - memorizing indexes of first observed sample for each timeseries (ts_idx) allows to calculate all future sample indexes by simple position difference-adjustments 
  - supports all index types with the same logic
  - covariates that are >> longer than the target series now have no more negative impact on computational time
- change from indexes that started at the end to absolute indexes 
- minor adjustment of error messages in inference dataset loaders.

### Other info:
- tested datetime and integer indexed series
- checked whether new approach produces the exact same indexes as before
- compared original approach with new approach

## Comparison
The graphs below show the average time required for loading relative to the training time (t_load / t_train) in one epoch with the original and new approach.
In all cases, the new approach was faster.

Air Pass. - air passengers | Electr. - electricity demand

comp_1
**Graph 1**: 
Left: covariates with matching time index as target series
Right: Lazy Loading for airpassengers with matching and longer covariate series (1000 time steps)

comp_2
**Graph 2**: 
Left: same as Graph 1 left but on the same y-axis as Graph 2 right.
Right: Single and Multi TimeSeries training with covariates that are >> longer than the target series (1000 covariate steps vs ~200 target steps)

## Update 22/11/2021

- `HorizonBasedDataset` now leverages of `_memory_indexer()` as well.
- added `_memory_indexer()` to base class `TrainingDataset` so that any future dataset classes can use it as well.


darts/utils/data/shifted_dataset.py
```diff
@@ -459,59 +463,123 @@ def __getitem__(self, idx) -> Tuple[np.ndarray, Optional[np.ndarray], np.ndarray
...
+                # extract actual index value (respects datetime- and integer-based indexes; also from non-zero start)
+                start_time = ts_target.time_index[start]
+                end_time = ts_target.time_index[end - 1]
+
+                raise_if_not(start_time in ts_covariate.time_index and end_time in ts_covariate.time_index,
+                             f'Missing covariates; could not find {"future" if self.shift_covariates else "past"} '
+                             f'covariates in index value range: {start_time} - {end_time}.')
+
+                # extract the index position (index) from index value
+                cov_start = ts_covariate.time_index.get_loc(start_time)
```
hrzn, line 559: This is much cleaner :)

darts/utils/data/shifted_dataset.py
```diff
@@ -459,59 +463,123 @@ def __getitem__(self, idx) -> Tuple[np.ndarray, Optional[np.ndarray], np.ndarray
...
+                cov_end = ts_covariate.time_index.get_loc(end_time) + 1
+
+            # store position of initial sample and all relevant sub set indices
+            self.index_memory[ts_idx] = {
+                'end_of_output_idx': end_of_output_idx,
+                'past_target': (past_start, past_end),
+                'future_target': (future_start, future_end),
+                'covariate': (cov_start, cov_end),
+            }
+        else:
```
hrzn, line 569: Do you know (roughly) what is the fraction of the gain coming from this caching?
Is this avoiding some expensive O(N) `get_loc()` calls?

hrzn (commented):
That looks good to me, nice one :)
One point: It would be nice if this could be extended to `HorizonBasedDataset` too, WDYT?

darts/utils/data/shifted_dataset.py
dennisbader: Yes, I did it experimentally, see below.

- n_calls is the number of times we go through all indexes of `__getitem()__`  from a GenericShiftedDataset (n_calls = 3 would be the equivalent of going through 1000 epochs)
- `time` is the time spent for computing `memory_indexer()`

cach_vs_no_cache

darts/utils/data/shifted_dataset.py
```diff
@@ -12,9 +12,11 @@
                                DualCovariatesTrainingDataset,
                                MixedCovariatesTrainingDataset,
                                SplitCovariatesTrainingDataset)
-from .utils import _get_matching_index
```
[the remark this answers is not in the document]
dennisbader: Yes, would be nice. I will look into it

darts/utils/data/training_dataset.py
```diff
@@ -63,6 +70,102 @@ def __len__(self) -> int:
...
 
+    def _memory_indexer(self,
+                        ts_idx: int,
+                        ts_target: TimeSeries,
+                        shift: int,
+                        input_chunk_length: int,
+                        output_chunk_length: int,
+                        end_of_output_idx: int,
+                        ts_covariate: TimeSeries,
+                        cov_type: Optional[bool] = None) -> SampleIndexType:
```
hrzn, line 81: [Nitpicking] It seems the type hint is not good here. Also, I think you could use an Enum instead of two strings.

hrzn (approved):
Nice one 👍

darts/utils/data/training_dataset.py
dennisbader: True!

PR #555, opened 2021-11-19
fix end_ts+1

dumjax:
Fix to respect the definition of the method slice of TimeSeries. 

### Summary

TimeSeries slice method explicitly states:
Returns a new TimeSeries, starting later than `start_ts` and ending before `end_ts`, inclusive on both ends.
Nevertheless, in the case start_ts and end_ts are int, the range of index was defined by pd.RangeIndex(start_ts, end_ts, step=1) which means that end side was excluded. We just have to replace by end_ts+1.


hrzn:
Could we maybe add a test case to capture this case in the future? That might turn out useful :D

tomasvanpottelbergh:
Should the slicing method actually include the end? In my experience, you often want to exclude the end (slicing from midnight till midnight to cover a full day for example). Or do you think the current definition is more useful?

dumjax:
> 

I had the same question cause intuitively I would have said that we need to follow python way (excluding last element). But it was defined since the beginning like this and maybe there are some other parts which depends on this.

hrzn:
I have made some tests and it's OK to make the interval open on the right. I just had to adapt a couple of unit tests, but nothing major was relying on it (in fact `slice()` should probably be deprecated - we prefer slicing with `[]` instead). I've updated the PR.

hrzn (approved):

darts/timeseries.py
```diff
@@ -1278,11 +1278,11 @@ def slice(self, start_ts: Union[pd.Timestamp, int], end_ts: Union[pd.Timestamp,
         if isinstance(start_ts, pd.Timestamp):
             raise_if_not(self._has_datetime_index, 'Timestamps have been provided to slice(), but the series is '
                                                    'indexed using an integer-based Int64Index.', logger)
-            idx = pd.DatetimeIndex(filter(lambda t: start_ts <= t <= end_ts, self._time_index))
+            idx = pd.DatetimeIndex(filter(lambda t: start_ts <= t < end_ts, self._time_index))
         else:
             raise_if(self._has_datetime_index, 'start and end times have been provided as integers to slice(), but '
                                                'the series is indexed with a DatetimeIndex.', logger)
-            idx = pd.RangeIndex(start_ts, end_ts, step=1)
+            idx = pd.RangeIndex(start_ts, end_ts+1, step=1)
```
tomasvanpottelbergh, line 1285: Shouldn't this change be reverted and a test added?

tomasvanpottelbergh (changes requested):

hrzn:
As discussed recently with @dumjax - we'll close it. Having exclusive bounds on ints and inclusive on datetimes matches the Pandas way to index (and our `[]` operator too), so we'll leave it as is for now. We could revisit it at some point, because actually when ints represent time steps (and not indexes), one could make the case that all bounds should be inclusive.

PR #565, opened 2021-11-24
Feat/add encoders

dennisbader:
### Summary

- added encoder base classes for new general encoders
  - has an `encode_train()` method for generating encodings on the train dataset
  - has an `encode_inference()` method for generating encodings on the inference dataset

- added index generator classes for generating indexes for past or future covariate encoders
  - past covariate encoders rely on `PastCovariateIndexGenerator`, 
  future covariate encoders on `FutureCovariateIndexGenerator`
- added `SingleEncoder` base class (inherits from encoder) which can be used to implement new single encoders such as cyclic encoder, datetime attribute encoder, ...
  - must have an `encode()` which implements the logic of the encoder
- added `SequentialEncoder` class (inherits from encoder) which automatically instantiates single encoder objects from parameters `model_kwargs` used when creating a `TorchForecastingModel`.
  - for generating encoded past and/or future covariates all under one hood
  - you can read below if you want to know more about the kwargs for the SequetialEncoder
  - `encode_train()` and `encode_inference()` launches all single encoders for past and future covariates in sequence
- refactored inference datasets
  - new `GenericShiftedInferenceDataset` now combines logic of Past and Future Inference datasets

### What works
- **update**: supported by all TorchForecastingModels
- I also added an encoder `DatetimeAttributeEncoder` which could be used instead of `datetime_attribute_series` from preprocessing (for this we would need transformers (Scaler) in SequentialEncoder).
- with the encoders users wouldn't have to worry about how many covariates need to be passed to training and prediction. 
- new encoders can be easily added
- **update:** we decided against distinguishing between lazy and eager loading which reduces complexity a lot. A warning is now raised when lazy loading is suspected and `add_encoders` are used.
- **update:** added absolute and relative integer index encoding for both datetime and integer-indexed data.
  - 'absolute' gives the integer index position: from index 0 where train target series starts
  - 'relative' gives the integer index position relative to the prediction point: index == 0 where train target series ends
  - moved position of `initialize_encoders()` so that the same encoder objects can be used for train, validation and inference
- **update:** Transformers are now supported. Respects training, validation and inference datasets, non-transformation of certain covariates/encodings (user-defined, cyclic encoders). Fitted on training dataset.

## SequentialEncoder **(updated)**
        SequentialEncoder automatically creates encoder objects from parameter `add_encoders` used when creating a
        `TorchForecastingModel`.

        *   Only kwarg `add_encoders` of type `Optional[Dict]` will be used to extract the encoders.
            For example: `model = MyModel(..., add_encoders={...}, ...)`

        The `add_encoders` dict must follow this convention:
            `{encoder keyword: {temporal keyword: List[attributes]}, ..., transformer keyword: transformer object}`
        Supported encoder keywords:
            `'cyclic'` for cyclic temporal encoder. See the docs
            :meth:`CyclicTemporalEncoder <darts.utils.data.encoders.CyclicTemporalEncoder>`;
            `'datetime_attribute'` for adding scalar information of pd.DatetimeIndex attribute. See the docs
            :meth:`DatetimeAttributeEncoder <darts.utils.data.encoders.DatetimeAttributeEncoder>`
            `'position'` for integer index position encoder. See the docs
            :meth:`IntegerIndexEncoder <darts.utils.data.encoders.IntegerIndexEncoder>`;
            `'custom'` for encoding index with custom callables (functions). See the docs
            :meth:`CallableIndexEncoder <darts.utils.data.encoders.CallableIndexEncoder>`;
        Supported temporal keywords:
            'past' for adding encoding as past covariates
            'future' for adding encoding as future covariates
        Supported attributes:
            for attributes read the referred docs for the corresponding encoder from above
        Supported transformers:
            a transformer can be added with transformer keyword 'transformer'. The transformer object must be an
            instance of Darts' :meth:`FittableDataTransformer
            <darts.dataprocessing.transformers.fittable_data_transformer.FittableDataTransformer>` such as Scaler() or
            BoxCox(). The transformers will be fitted on the training dataset when calling calling `model.fit()`.
            The training, validation and inference datasets are then transformed equally.

        An example of a valid `add_encoders` dict for hourly data:
            add_encoders={
                'cyclic': {'future': ['month']},
                'datetime_attribute': {'past': ['hour'], 'future': ['year', 'dayofweek']},
                'position': {'past': ['absolute'], 'future': ['relative']},
                'custom': {'past': [lambda index: (index.year - 1950) / 50]},
                'transformer': Scaler()
            }

        Tuples of `(encoder_id, attribute)` are extracted from `add_encoders` to instantiate the `SingleEncoder`
        objects:
        * The `encoder_id` is extracted as follows:
            str(encoder_kw) + str(temporal_kw) -> 'cyclic' + 'past' -> `encoder_id` = 'cyclic_past'
            The `encoder_id` is used to map the parameters with the corresponding `SingleEncoder` objects.
        * The `attribute` is extracted from the values given by values under `temporal_kw`
            `attribute` = 'month'
            ...
            The `attribute` tells the `SingleEncoder` which attribute of the index to encode

        New encoders can be added by appending them to the mapping property `SequentialEncoder.encoder_map()`


hrzn:
Just a random idea about scaling: I suspect it will mostly be needed for Datetime attributes (e.g. years). Cyclic encodings should be more or less fine. To address this for Datetime attributes, we could perhaps allow users to provide functions in addition to datetime attributes. Then in the function users could be free to do some (rough) scaling. For instance:
```
add_encoders = {'datetime_attribute': {'future': [lambda dt: (dt.year - 1950) / 50]}}
```
Do you think this might make sense?

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,570 @@
...
+                 output_chunk_length: int,
+                 shift: int,
+                 takes_past_covariates: bool = False,
+                 takes_future_covariates: bool = False) -> None:
+
+        """
+        SequenceEncoder automatically creates encoder objects from parameters `model_kwargs` used when creating a
+        `TorchForecastingModel`.
+
+        *   Only kwarg `add_encoders` of type `Optional[Dict]` will be used to extract the encoders.
```
hrzn, line 250: Was there a reason why taking all the model's kwargs and not just the `add_encoders` parameter here?

darts/models/forecasting/tft_model.py
```diff
@@ -559,15 +560,22 @@ def __init__(self,
...
-            For more information, check out :meth:`datetime_attribute_timeseries()
-            <darts.utils.timeseries_generation.datetime_attribute_timeseries>`
+        add_encoders : optional Dict
+            If other than None, add index encoding/s to the past and/or future covariates for training and prediction.
+            The dict must follow this convention:
+                `{encoder keyword: {temporal keyword: List[attributes]}}`
+            Supported encoder keywords:
+                `cyclic` for cyclic temporal encoder. See the docs :meth:`CyclicTemporalEncoder
+                <darts.utils.data.encoders.CyclicTemporalEncoder>`;
+                `datetime_attribute` for adding scalar information of pd.DatetimeIndex attribute. See the docs
```
hrzn, line 570: Question: is there a way to add the "naive" index values for integer-indexed series?

darts/utils/data/horizon_based_dataset.py
```diff
@@ -107,7 +117,7 @@ def __getitem__(self, idx: int) -> Tuple[np.ndarray, Optional[np.ndarray], np.nd
 
         # optionally, load covariates
         ts_covariate = self.covariates[ts_idx] if self.covariates is not None else None
-        cov_type = CovariateType.NONE if self.covariates is None else CovariateType.PAST
+        main_cov_type = CovariateType.NONE if self.covariates is None else CovariateType.PAST
```
hrzn, line 120: Shouldn't you re-use `self.covariate_type` (and set it to None in init() if covariates are None)?

darts/utils/data/training_dataset.py
```diff
@@ -171,6 +167,41 @@ def _memory_indexer(self,
 
         return past_start, past_end, future_start, future_end, cov_start, cov_end
 
+    def _generate_covariates(self,
```
hrzn, line 170: It might be a bit clearer to renamed into sth like `_add_encodings_to_covariates()` ?

darts/utils/data/training_dataset.py
```diff
@@ -171,6 +167,41 @@ def _memory_indexer(self,
...
+                                                           past_covariate=covariate,
+                                                           future_covariate=None,
+                                                           encode_future=False)
+        elif cov_type is CovariateType.HISTORIC_FUTURE:
+            _, covariate = self.lazy_encoders.encode_train(target=target,
+                                                           past_covariate=None,
+                                                           future_covariate=covariate,
+                                                           encode_past=False)
+        elif cov_type is CovariateType.FUTURE:
+            _, covariate = self.lazy_encoders.encode_inference(n=self.lazy_encoders.output_chunk_length,
```
hrzn, line 195: I find this naming `encode_inference` not ideal, as it is also called for producing training samples.
Perhaps `encode_train` could be renamed into `encode_past` and `encode_inference` into `encode_future` ?

hrzn: `n` is supposed to be the forecast horizon, but here it's set to the `output_chunk_length`. Is this intended?

darts/utils/data/training_dataset.py
```diff
@@ -171,6 +167,41 @@ def _memory_indexer(self,
...
+                                                               target=target,
+                                                               past_covariate=None,
+                                                               future_covariate=covariate,
+                                                               encode_past=False)
+            if len(covariate[0]) != self.lazy_encoders.output_chunk_length:
```
hrzn, line 200: A comment would be welcome here.

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,570 @@
...
+
+        If `n` is `None` it is a prediction and method `encoder.encode_inference()` is called.
+        Otherwise, it is a training case and `encoder.encode_train()` is called.
+        """
+
+        encode_method = 'encode_train' if n is None else 'encode_inference'
+
+        encoded_sequence = []
+        if covariate is None:
+            covariate = [None] * len(target)
```
hrzn, line 429: If I understand well, these lists have size > 1 only in the "eager" case, right?

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,570 @@
...
+        else:
+            covariate = [covariate] if isinstance(covariate, TimeSeries) else covariate
+
+        for ts, pc in zip(target, covariate):
+            encoded = concatenate([getattr(enc, encode_method)(target=ts,
```
hrzn, line 434: Nice, a use of `darts.concatenate` 😍

hrzn (commented):
I'm not completely done reviewing, but it looks really good! I really like the fact users could have the ability to specify simple encodings/attributes as covariates when creating the models without having to do this manually.

My main remark for the time being would be that I'd try to avoid to treat the non-lazy case separately, and treat everything in the datasets (in the lazy fashion). This would keep things simple, and we can perhaps optimise later if that's really needed.

It'd also be nice if this could be extend to other models (but it can wait for a future PR - this one is already huge!)

darts/utils/data/encoder_base.py
```diff
@@ -0,0 +1,292 @@
...
+                              covariate: Optional[TimeSeries] = None) -> SupportedIndexes:
+        """For training (when `n` is `None`) we can simply use the future covariates (if available) or target as
+        reference to extract the time index.
+        """
+
+        # TODO -> we can probably summarize this in CovariateIndexGenerator as I think it's the same for
+        #  future & past covs
+        super(FutureCovariateIndexGenerator, self).generate_train_series(target, covariate)
+
+        return covariate.time_index if covariate is not None else target.time_index
```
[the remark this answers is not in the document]
dennisbader: I think this is correct for training.

For the training dataset, the last `output_chunk_length` points of the target series are never used as a "past/input" chunk because we need to evaluate the loss on the future chunk.
So the future covariates can end at the same index as the target.

But yeah maybe I missed something, WDYT?

darts/models/forecasting/tft_model.py
dennisbader: Yes, there is

darts/utils/data/training_dataset.py
dennisbader: True! +1

darts/utils/data/encoder_base.py
```diff
@@ -0,0 +1,292 @@
...
+        if covariate is not None:
+            return covariate.time_index
+        else:
+            return _generate_index(start=target.end_time() - target.freq * (self.input_chunk_length - 1),
+                                   length=self.input_chunk_length + max(0, n - self.output_chunk_length),
+                                   freq=target.freq)
+
+
+class FutureCovariateIndexGenerator(CovariateIndexGenerator):
+    """generates index for future covariate."""
```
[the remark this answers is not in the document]
dennisbader: same answer as before

darts/utils/data/encoders.py
dennisbader: True, we can directly use the `add_encoders` parameter.
I did it in the beginning to be consistent with the docs for the Model itself and the SequenceEncoder

darts/utils/data/encoder_base.py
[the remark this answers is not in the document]
dennisbader: Maybe it would be smarter though to include the next `output_chunk_length` points. Then we wouldn't have to call `encode_inference()` in the TrainingDatasets. Then we just have `output_chunk_length` unused future covariates when adding future covariates in the "eager/non-lazy" way.

darts/utils/data/training_dataset.py
dennisbader: Yes, here it is intended. We want the future part of future covariates on the output chunk of length `output_chunk_length` for the current training sample.

darts/utils/data/encoder_base.py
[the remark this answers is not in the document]
dennisbader: since we don't distinguish between lazy and eager case anymore, this can stay as it is

darts/utils/data/training_dataset.py
dennisbader: as discussed, we can now leave it as it is

darts/models/forecasting/tft_model.py
hrzn: How would this work?

dennisbader: Implementing it right now :)

darts/utils/data/encoders.py
dennisbader: yes :)

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,891 @@
...
+        super(FutureCallableIndexEncoder, self).__init__(
+            index_generator=FutureCovariateIndexGenerator(input_chunk_length, output_chunk_length),
+            attribute=attribute
+        )
+
+
+class SequenceEncoder(Encoder):
+    """A `SequenceEncoder` object can store and control multiple past and future covariate encoders at once.
+    It provides the same functionality as single encoders (`encode_train()` and `encode_inference()`).
+    Sequence encoders can be used with Darts' darts `Datasets`.
```
hrzn, line 438: Is this still relevant?

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,891 @@
...
+                `attribute = lambda index: (index - 1950) / 50]}`
+        """
+
+        super(FutureCallableIndexEncoder, self).__init__(
+            index_generator=FutureCovariateIndexGenerator(input_chunk_length, output_chunk_length),
+            attribute=attribute
+        )
+
+
+class SequenceEncoder(Encoder):
```
hrzn, line 435: The name `SequenceEncoder` might be slightly misleading as it could be understood as an encoder for a sequence. How about e.g. `EncoderSequence` ?

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,891 @@
...
+        -------
+        Tuple[past_covariate, future_covariate]
+            The past_covariate and/or future_covariate for training including the encodings.
+            If input {x}_covariate is None and no {x}_encoders are given, will return `None`
+            for the {x}_covariate.
+        Raises
+        ------
+        Warning
+            If model was created with `add_encoders` and there is suspicion of lazy loading.
+            The encodings/covariates are generated pre-train for all individual targets and
```
hrzn, line 566:
```suggestion
            The encodings/covariates are generated eagerly before starting training for all individual targets and
```

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,891 @@
...
+            loaded into memory. Depending on the size of target data, this can create memory
+            issues. In case this applies, consider setting `add_encoders=None` at model
+            creation
```
hrzn, line 569:
```suggestion
            creation and build your encodings covariates manually for lazy loading.
```

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,891 @@
...
+            self._future_transformer = SequenceEncoderTransformer(copy.deepcopy(transformer), transform_future_mask)
+
+    def _process_input_encoders(self, params: Dict) -> Tuple[List, List]:
+        """Processes input and returns two lists of tuples `(encoder_id, attribute)` from relevant encoder
+        parameters at model creation.
+
+        Parameters
+        ----------
+        params
+            A dict of type Optional[Dict] from parameter `add_encoders` used at model creation.
```
hrzn, line 776: Not sure you need to repeat the full docstring in those private methods.

darts/utils/data/encoders.py
```diff
@@ -0,0 +1,891 @@
...
+ENCODER_KEYS = ['cyclic', 'datetime_attribute', 'position', 'custom']
+FUTURE = 'future'
+PAST = 'past'
+VALID_TIME_PARAMS = [FUTURE, PAST]
+VALID_ENCODER_DTYPES = (str, Sequence)
+
+TRANSFORMER_KEYS = ['transformer']
+VALID_TRANSFORMER_DTYPES = FittableDataTransformer
+
+DIMS = ('time', 'component', 'sample')
```
hrzn, line 41: Would be safer to import this one from `darts.timeseries`

darts/utils/data/inference_dataset.py
```diff
@@ -3,14 +3,19 @@
...
+from typing import Union, Sequence, Optional, Tuple
 
 from ...timeseries import TimeSeries
 from ...logging import raise_if_not
-from .utils import _get_matching_index
+from .utils import CovariateType
+from .encoder_base import (PastCovariateIndexGenerator,
+                           FutureCovariateIndexGenerator)
+
+SampleIndexType = Tuple[int, int, int, int, int, int]
```
hrzn, line 18: What is this used for?

darts/utils/data/inference_dataset.py
```diff
@@ -3,14 +3,19 @@
...
+
+from abc import ABC, abstractmethod
 from torch.utils.data import Dataset
+from typing import Union, Sequence, Optional, Tuple
 
 from ...timeseries import TimeSeries
 from ...logging import raise_if_not
-from .utils import _get_matching_index
+from .utils import CovariateType
+from .encoder_base import (PastCovariateIndexGenerator,
```
hrzn, line 15: Is this needed?

hrzn (commented):
This is really great work @dennisbader - looks very good!
Also big kudos for the code quality and the good unit tests 👍 

I have a few minor comments (nothing important). I think we can merge it soon :) 

My main comment, though, would be that it would be nice to already extend `add_encoders` to all Torch models, WDYT?

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -178,28 +180,49 @@ def transform(self,
...
-
             The output is a 2-D matrix where each column corresponds to a component (dimension)
             of the series, and the columns' values are the flattened values over all samples
+
+            If `component_mask` is given, extract only columns were `component_mask` is True.
         """
+
         vals = series.all_values(copy=False)
-        return np.stack([vals[:, i, :].reshape(-1) for i in range(series.width)], axis=1)
+        if component_mask is None:
```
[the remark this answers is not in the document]
dennisbader: looks much better!

darts/utils/data/encoder_base.py
```diff
@@ -0,0 +1,382 @@
...
+    @abstractmethod
+    def __init__(self):
+        self.attribute = None
+        self.dtype = np.float64
+
+    @abstractmethod
+    def encode_train(self,
+                     target: TimeSeries,
+                     covariate: Optional[TimeSeries] = None,
+                     merge_covariate: bool = True,
```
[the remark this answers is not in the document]
dennisbader: yes, the SequenceEncoder sets merge_covariate to False when it calls the SingleEncoders so that merging the encodings with optional user-supplied covariates is only applied once when all encodings were generated.

darts/utils/data/encoders.py
dennisbader: No, thanks :)

darts/utils/data/encoders.py
dennisbader: What do you think about `SequentialEncoder`? Would be consistent with the style I used for the other classes  `{X}Encoder` like `SingleEncoder`

hrzn: Sounds good!

darts/models/forecasting/tft_model.py
```diff
@@ -533,41 +533,48 @@ def __init__(self,
...
+            `add_encoders`. An example showing some of `add_encoders` features:
+
+                add_encoders={
+                    'cyclic': {'future': ['month']},
+                    'datetime_attribute': {'past': ['hour'], 'future': ['year', 'dayofweek']},
+                    'position': {'past': ['absolute'], 'future': ['relative']},
+                    'custom': {'past': [lambda index: (index.year - 1950) / 50]}
+                    'transformer': Scaler()
+
         add_relative_index : bool
```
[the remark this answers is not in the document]
dennisbader: Hmm, I'm not sure I would add it to `add_encoders` as part of the `SequentialEncoder`. 
This relative index stays constant over all samples and gives the index relative to the sample prediction point.
Whereas the others are computed for the entire train, validation and inference time spans.

I think instead of adding it to the `add_encoders`, we can simply leverage the `FutureIntegerIndexEncoder` in `TFTModel` to compute this `relative_index`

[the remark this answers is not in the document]
hrzn: OK sounds good

darts/tests/test_encoders.py
```diff
@@ -0,0 +1,501 @@
...
+                                                                   past_covariate=self.covariate_multi,
+                                                                   future_covariate=self.covariate_multi)
+
+        # encoded multi TS covariates should have same number as input covariates
+        self.assertEqual(len(past_covs_train), 2)
+        self.assertEqual(len(future_covs_train), 2)
+
+        # each attribute (i.e., 'month', ...) generates 2 output variables (+ 1 covariate from input covariates)
+        self.assertEqual(past_covs_train[0].n_components, 3)
+        self.assertEqual(future_covs_train[0].n_components, 3)
```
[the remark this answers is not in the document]
dennisbader: This is due to cyclic encoders generating 2 components: a sine and a cosine wave :). When generating 2 monthly cyclic encoders we actually generate 4 components . And with 1 monthly cyclic encoder only 2 components.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -436,15 +441,27 @@ def fit(self,
...
+            if self.encoders.encoding_available:
+                val_past_covariates, val_future_covariates = \
+                    self.encoders.encode_train(target=val_series,
+                                               past_covariate=val_past_covariates,
+                                               future_covariate=val_future_covariates)
+
             val_dataset = self._build_train_dataset(target=val_series, 
                                                     past_covariates=val_past_covariates, 
-                                                    future_covariates=future_covariates,
+                                                    future_covariates=val_future_covariates,
```
hrzn, line 464: Oh! That was a bug?

hrzn (approved):
Super nice 👍  great job!

darts/models/forecasting/torch_forecasting_model.py
dennisbader: I believe so, right?

PR #574, opened 2021-11-26
Feat/adapt scalers to stochastic

hrzn:
Extend the scalers and BoxCox transformers to work with stochastic series.
The scaling is always done per-component, jointly over all samples of the same component.

I have also added a new method `with_values()` to `TimeSeries`, to swap the array of values of a TimeSeries with another one.


darts/dataprocessing/transformers/boxcox.py
```diff
@@ -109,21 +134,16 @@ def ts_fit(series: TimeSeries,
...
+        return series.with_values(BoxCox._reshape_out(transformed_vals, series.width, series.n_samples))
 
-        def _boxcox_wrapper(col):
-            idx = series.pd_dataframe(copy=False).columns.get_loc(col.name)  # get index from col name
-            return boxcox(col, lmbda[idx])
-
-        return TimeSeries.from_dataframe(series.pd_dataframe(copy=False).apply(_boxcox_wrapper))
 
     @staticmethod
     def ts_inverse_transform(series: TimeSeries, lmbda: Union[Sequence[float], pd.core.series.Series]) -> TimeSeries:
```
dennisbader, line 143: Out of interest, do you know what the benefit is of having these methods defined as static methods in the parent classes?
If they weren't we could access the class methods (like `BoxCox._reshape_in()` more elegantly.

darts/tests/dataprocessing/transformers/test_boxcox.py
```diff
@@ -81,3 +82,15 @@ def test_boxcox_multiple_calls_to_fit(self):
         lambda2 = deepcopy(box_cox._fitted_params)[0].tolist()
 
         self.assertNotEqual(lambda1, lambda2, "Lambdas should change when the transformer is retrained")
+
+    def test_multivariate_stochastic_series(self):
```
dennisbader, line 86: Is this stochastic? I believe `n_samples==1` here

darts/tests/dataprocessing/transformers/test_data_transformer.py
```diff
@@ -72,3 +72,17 @@ def test_multi_ts_scaling(self):
...
+        scaler = Scaler(MinMaxScaler())
+        vals = np.random.rand(10,5,50)
+        s = TimeSeries.from_values(vals)
+        ss = scaler.fit_transform(s)
+        ssi = scaler.inverse_transform(ss)
+
+        # Test inverse transform
+        np.testing.assert_allclose(s.all_values(), ssi.all_values())
+
+        # Test that the transform is done per component (i.e max value over each component should be 1)
```
dennisbader, line 86: could we also test for min() almost equal to 0, to test both sides, wdyt?

darts/dataprocessing/transformers/boxcox.py
```diff
@@ -88,15 +91,37 @@ def _transform_iterator(self, series: Sequence[TimeSeries]) -> Iterator[Tuple]:
     def _inverse_transform_iterator(self, series: Sequence[TimeSeries]) -> Iterator[Tuple]:
         return zip(series, self._fitted_params)
 
+    @staticmethod
+    def _reshape_in(series: TimeSeries) -> np.ndarray:
```
dennisbader, line 95: I imagine that methods `_reshape_in/_out` can be useful for all transformers (as redefined in Scaler). Could these be part of `BaseDataTransformer`?

hrzn: Indeed, absolutely!

darts/dataprocessing/transformers/boxcox.py
hrzn: I'm not exactly sure actually, I think the idea was maybe to have a more functional way to write these methods (not having to access the object's state), but I agree it might be cleaner to at some point remove this. I'd leave this for later though.

PR #682, opened 2021-12-09
Fix/example notebooks

dennisbader:
### Summary

- added TFTModel and DTW to the docs examples page (they were not appearing before)
- updated TFTModel example with new add_encoders


tomasvanpottelbergh:
While you're touching the examples, could you also add all examples to the lists in `.github/workflows`? Even better if you can find an easy way to automatically check all examples in the folder :)

dennisbader:
As I'm new to github actions: I assume they have to be added to the lines below only in `develop.yml` and `merge.yml` (not `release.yml`)?
```
check-examples:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        example-name: [04-FFT-examples.ipynb, 05-RNN-examples.ipynb, 01-darts-intro.ipynb, 03-data-processing.ipynb, 02-multi-time-series-and-covariates.ipynb]
```

tomasvanpottelbergh:
> 

Exactly. We could remove them from `develop.yml`, so they only get run after merging the PR. That's what we do for the tests for other flavors (which are currently failing actually).

dennisbader:
Yes, I totally missed adding the TORCH check for the add_encoders PR.
I already upladed a PR that should fix it

docs/source/examples.rst
```diff
@@ -108,6 +108,16 @@ You will find below an example Jupyter notebook showcasing the usage of the Deep
 
    examples/10-DeepTCN-examples.ipynb
 
+Temporal Fusion Transformer (TFT) Model
+===================
+
+You will find below an example Jupyter notebook showcasing the usage of the TFT model
```
hrzn, line 114: Actually could we make those lines less verbose? No need to say each time "You will find below an example ..."

We could just say
```
TFT model example notebook:
```

And similarly for all notebooks.

hrzn:
+1 for adding all tests in `merge.yml` (and keep most out of `develop.yml`).

dennisbader:
@tomasvanpottelbergh I researched a bit and so far the best I've found on dynamically building the matrix would be with `fromJson()` to create the matrix from a json ([see here](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson)). That would leave the question to how we write the json when pushing/merging. Do you have any ideas on that?

hrzn (approved):
👍

hrzn:
Anything missing before we can merge this one?

Issue #684, opened 2021-12-09
[Question]


davidrpugh:
Thanks for the awesome library!

I am working with asset price data and I would like to perform log transform as part of a preprocessing step to insure that the model never predicts negative prices. I tried what I thought was the obvious thing to do.
```python
X = darts.TimeSeries.from_series(asset["price"])
preprocessor = dataprocessing.transformers.InvertibleMapper(fn=np.log, inverse_fn=np.exp)
Z = preprocessor.transform(X)
model = models.ARIMA()
model.fit(Z)
```
So far so good. I them make my predictions. I am using the model probabilistic forecasting to I am going to generate `N` forecasts.
```python
predictions = model.predict(T, num_samples=N)
```
I then need to apply the inverse transform on my predictions to get back to the original price forecasts.
```python
forecasted_prices = preprocessor.inverse_transform(predictions)
```
However this last call generates and error.
```python
[2021-12-09 18:02:35,340] ERROR | main_logger | AssertionError: The pd_dataframe() method can only return DataFrames of deterministic time series, and this series is not deterministic (it contains several samples). Consider calling quantile_df() instead.
[2021-12-09 18:02:35,340] ERROR | main_logger | AssertionError: The pd_dataframe() method can only return DataFrames of deterministic time series, and this series is not deterministic (it contains several samples). Consider calling quantile_df() instead.
[2021-12-09 18:02:35,340] ERROR | main_logger | AssertionError: The pd_dataframe() method can only return DataFrames of deterministic time series, and this series is not deterministic (it contains several samples). Consider calling quantile_df() instead.
[2021-12-09 18:02:35,340] ERROR | main_logger | AssertionError: The pd_dataframe() method can only return DataFrames of deterministic time series, and this series is not deterministic (it contains several samples). Consider calling quantile_df() instead.
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
/var/folders/69/7vb352bd765cb9jwg51q78n40000gp/T/ipykernel_13515/4062002826.py in <module>
      1 val.plot()
----> 2 preprocessor.inverse_transform(predictions)
      3 #plt.yscale("log")

~/Research/crypto-research/env/lib/python3.9/site-packages/darts/dataprocessing/transformers/mappers.py in inverse_transform(self, series, *args, **kwargs)
    115                           *args, **kwargs) -> Union[TimeSeries, List[TimeSeries]]:
    116         # adding the inverse_fn param
[...]

PR #688, opened 2021-12-09
Feat/extend map to stochastic

hrzn:
`TimeSeries.map()` was not working on stochastic series; this solves the issue. It also slightly changes the API of the method. Now it expects functions that can act on NumPy arrays (rather than on isolated scalars). Most elementwise lambdas (such as `lambda x: x**2`) should keep working though.

It also makes `Mapper`s work properly (incl with inverse transforms on stochastic forecasts).

Fixes: https://github.com/unit8co/darts/issues/684 https://github.com/unit8co/darts/issues/687

### DIFF
--- a/darts/tests/test_encoders.py
+++ b/darts/tests/test_encoders.py
@@ -79,7 +79,6 @@ class EncoderTestCase(DartsBaseTestClass):
 
     @unittest.skipUnless(TORCH_AVAILABLE, 'Torch not available. SequentialEncoder tests with models will be skipped.')
     def test_sequence_encoder_from_model_params(self):
-        print('was called')
         """test if sequence encoder is initialized properly from model params"""
         # valid encoder model parameters are ('past', 'future') for the main key and datetime attribute for sub keys
         valid_encoder_args = {
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -1852,13 +1852,18 @@ class TimeSeries:
             
         elif num_args == 2:  # map function uses timestamp f(timestamp, x)
 
-            # Poor man's solution - we iterate over all timestamps 
-            # (not sure there's a better solution here)
-            ars = []
-            for ts in self.time_index:
-                vals = self._xa.sel({self._time_dim: ts})
-                ars.append(np.expand_dims(fn(ts, vals), 0))
-            new_xa.values = np.vstack(ars)
+            # go over shortest amount of iterations, either over time steps or components and samples
+            if self.n_timesteps <= self.n_components * self.n_samples:
+                new_vals = np.vstack([
+                    np.expand_dims(fn(self.time_index[i], self._xa[i, :, :]), axis=0) for i in range(self.n_timesteps)
+                ])
+            else:
+                new_vals = np.stack([
+                    np.column_stack([fn(self.time_index, self._xa[:, i, j]) for j in range(self.n_samples)])
+                    for i in range(self.n_components)
+                ], axis=1)
+            new_xa.values = new_vals
+            
         else:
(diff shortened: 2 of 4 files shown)


darts/timeseries.py
```diff
@@ -1807,18 +1807,21 @@ def map(self,
...
         ----------
         fn
-            Either a function which takes a value and returns a value ie. `f(x) = y`
-            Or a function which takes a value and its timestamp and returns a value ie. `f(timestamp, x) = y`
-            The type of `timestamp` is either `pd.Timestamp` (if the series is indexed with a DatetimeIndex),
+            Either a function which takes a NumPy array and returns a NumPy array of same shape; 
+            e.g., `lambda x: x ** 2`, `lambda x: x / x.shape[0]` or `np.log`.
+            It can also be a function which takes a timestamp and array, and returns a new array of same shape;
+            e.g., `lambda ts, x: x / ts.days_in_month`.
+            The type of `ts` is either `pd.Timestamp` (if the series is indexed with a DatetimeIndex),
```
dennisbader, line 1824: Why again do we expect to have a `pd.Timestamp` or `int` instead of a `pd.DatetimeIndex` or `(pd.Int64Index, pd.RangeIndex)`?

`lambda ts, x: x / ts.days_in_month` can also be applied on the entire`time_index`. ->
`lambda time_index, x: x / time_index.days_in_month`

Similar to darts.utils.data.encoders.CallableIndexEncoder


Could this avoid having to iterate over every time stamp anymore as you mentioned below?

Maybe I'm missing something :)

darts/tests/test_encoders.py
```diff
@@ -79,7 +79,6 @@ class EncoderTestCase(DartsBaseTestClass):
 
     @unittest.skipUnless(TORCH_AVAILABLE, 'Torch not available. SequentialEncoder tests with models will be skipped.')
     def test_sequence_encoder_from_model_params(self):
-        print('was called')
```
dennisbader, line 82: oh no, thanks!

dennisbader (commented):
Very nice that now this also works with stochastic time series!
My only comment would be to think about expecting a time_index instead of a time stamp as `ts` in`TimeSeries.map()`

darts/timeseries.py
hrzn: You are right it would be nicer implementation-wise, however the signature of the resulting function is not as nice.
For instance instead of writing
```
lambda ts, vals: vals/ts.days_in_month
```

We would have to write:
```
lambda idx, vals: (vals.T/idx.days_in_month.values).T
```

More efficient for sure, but not that nice, no?

dennisbader: I see what you mean. What do you think about finding the shortest path to compute the result? Something like this

```
            # go over shortest amount of iterations, either over time steps or components and samples
            if self.n_timesteps <= self.n_components * self.n_samples:
                new_vals = np.vstack([
                    np.expand_dims(fn(self.time_index[i], self._xa[i, :, :]), axis=0) for i in range(self.n_timesteps)
                ])
            else:
                new_vals = np.stack([
                    np.column_stack([fn(self.time_index, self._xa[:, i, j]) for j in range(self.n_samples)])
                    for i in range(self.n_components)
                ], axis=1)
            new_xa.values = new_vals
```
this could reduce cost by quite a bit (see below). I also assume that in most cases during training `n_timesteps >> n_components * n_samples` and for prediction `n_timestamps << n_components * n_samples`. This approach could leverage the fast route for both. WDYT?

image

hrzn: OK nice, I like it :) thanks!

dennisbader (approved):
Very nice, thanks!

PR #691, opened 2021-12-14
Docs/cov update2

dennisbader:
## Summary

- continuation of @hrzn 's suggestions from #533.
- split everything into two guides:
  - general covariates in darts: basically tells you all you need to know to get you started with covariates in Darts
  - in depth look at TorchForecastingModel with covariates. I believe this could come in handy as an explanation on how TFMs work (we received some questions on this already)
- moved files into ./guides
## General info:
- added tl;drs
- adapted a large amount of sections -> tried to make it more crips, better structured, ...
- changed TFM example to use more distinct past/future covariates and MixedCovariatesModel


guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
+# In-depth look at Covariates in Torch Forecasting Models
+This document was written for darts version 0.15.0.
+
+We assume that you already know about covariates in Darts. If you're new to the topic we recommend you to read darts/guides/covariates.md first.
```
hrzn, line 4: Here too, would be good to put actual link.

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Summary - TL;DR
+In Darts, **Torch Forecasting Models (TFMs)** are broadly speaking "machine learning based" models, which denote PyTorch-based (deep learning) models.
+
+TFMs train and predict on fixed-length chunks (sub-samples) extracted from your input data. Each chunk contains an input chunk - resembling the sample's past - and an output chunk - the sample's future. The length of these chunks has to be specified at model creation with parameters `input_chunk_length` and `output_chunk_length`.
```
hrzn, line 9: here too, did you mean "containing" instead of "resembling" ?

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+All TFMs can be trained on single or multiple `target` series and, depending on their covariate support, `past_covariates` and / or `future_covariates`.
+
+```
+# fit the model on a single target series with optional past and / or future covariates
+model.fit(target, 
+          past_covariates=past_covariates, 
+          future_covariates=future_covariates)
+```
+
+You can produce forecasts for any input `target` TimeSeries (also unobserved targets) as long as it contains `input_chunk_length` time steps.
```
hrzn, line 27:
```suggestion
You can produce forecasts for any input `target` TimeSeries or for several targets given as a sequence of TimeSeries. This will also work on series that have not been seen during training, but the series must contain `input_chunk_length` time steps.
```

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+- time span requirements for target and covariate series
+
+[Section 2](#2-in-depth-look-at-how-past-and-future-covariates-are-used-in-a-torch-forecasting-model) gives
+an in-depth guide of how covariates are used in Darts' PyTorch-based Forecasting Models.
+
+## 1. Torch Forecasting Models
+## 1.1. Introduction
+TFMs are broadly speaking "machine learning based" models, which denote PyTorch-based (deep learning) models.
+
+TFMs train and predict on fixed-length chunks (sub-samples) of the `target` and `*_covariates` series (if supported). `Target` is the series for which we want to predict the future, `*_covariates` are the past and / or future covariates (explanatory variables) that can help models improve forecasts.
```
hrzn, line 53: This part is somewhat a repetition of much of what's written in the TL;DR. Perhaps you could rename the TL;DR in "Introduction" and cut this part (and assume everyone will read the intro first)

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+Each chunk contains an input chunk - resembling the sample's past - and an output chunk - the sample's future. The sample's prediction point lies at the end of the input chunk. The length of these chunks has to be specified at model creation with parameters `input_chunk_length` and `output_chunk_length`.
+
+```
+# model with chunks looking 7 time steps back (past) and 1 time step ahead (future)
+model = SomeTorchForecastingModel(input_chunk_length=7, output_chunk_length=1)
+```
+
+Depending on their capabilities, TFMs accept past and / or future covariates as optional parameters `past_covariates` and `future_covariates` in `fit()` and `predict()`. You can find the covariate support for each TFM in [section 1.2.](#12-torch-forecasting-model-covariates-support). 
+
+***The same covariates can be used for training and prediction if they have the required time spans (explained in [section 1.3.](#13-required-target-time-spans-for-training-validation-and-prediction))***
```
hrzn, line 64: Here too this is somewhat of a repetition from the higher-level document, which I would assume they would have read first (as you link to it at the top, saying it's assumed the read it).

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+
+So during training, the torch models will go through the training data in sequences (see Figure 3). Using information from the **input chunk** and **output chunk**, the model predicts the future target on the output chunk. The training loss is evaluated between the predicted future target and the actual target value on the output chunk. The model trains itself by minimizing the loss over all sequences.
+
+![figure5](./images/covariates/seq_covs_1.png)
+
+**Figure 3: Prediction and loss evaluation in a single sequence**
+
+After having completed computations on the first sequence, the model moves to the next sequence, which, in the
+case of sequential models, starts one time step later than the previous sequence. This process is repeated until all 365 days were covered (see Figure 4).
```
hrzn, line 206:
```suggestion
case of sequential datasets, starts one time step later than the previous sequence. This process is repeated until all 365 days were covered (see Figure 4).
```

hrzn: Actually this is not quite accurate, because loading is randomized, so the model will (in general) not see the sub-slices sequentially, but in a random order.

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+# load the model state that performed best on validation set 
+best_model = model.load_from_checkpoint(model_name='MyModel', best=True)
+```
+
+You can also save or load manually:
+```
+model.save_model(model_path)
+loaded_model = model.load_model(model_path)
+```
+
```
hrzn, line 254:
```suggestion
/!\ Warning /!\ At this stage of Darts development, we are not (yet) ensuring backward compatibility, so it might not always be possible to load a model saved by an older version of the library.
```

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
...
+- If `n <= output_chunk_length`: we can predict `n` in one go (using one "internal model call")
+    - in our example: predict the next day's ice-cream sales (`n = 1`)
+- If `n > output_chunk_length`: we must predict `n` by calling the internal model multiple times. Each call outputs `output_chunk_length` prediction points. We go through as many calls as needed until we get to the final `n` prediction points, in an auto-regressive fashion.
+    - in our example: predict ice-cream sales for the next 3 days at once (`n = 3`)
+
+  To do this we have to supply additional `past_covariates` for the next `n - output_chunk_length = 2` time steps (days) after the end of our 365 days training data. Unfortunately, we do not have measured `temperture` for the future. But let's assume we have access to temperature forecasts for the next 2 days. We can just append them to `temperature` and the prediction will work!
+  
+  ```
+  temperature = temperature.concat(temperature_forecast)
```
hrzn, line 272: Not sure we have a method `TimeSeries.concat()`. But you can do something like
```
temperature = temperature.concatenate(temperature_forecast, axis=0)
```

guides/covariates_tfms.md
```diff
@@ -0,0 +1,289 @@
+# In-depth look at Covariates in Torch Forecasting Models
```
hrzn, line 1: How about renaming the document to `torch_forecasting_models.md` ?
You are actually covering more than just covariates (which is great :) )

hrzn (commented):
This is great!!
Just have a few minor comments :)

guides/covariates.md
```diff
@@ -0,0 +1,241 @@
...
+              past_covariates=past_covariates_2,
+              # optional future_covariates,
+              )
+```
+
+If you train a model using `past_covariates`, you'll have to provide these `past_covariates` also at prediction time to `predict()`. This applies to `future_covariates` too, with a nuance that `future_covariates` have to extend far enough into the future at prediction time (all the way to the forecast horizon `n`). This can be seen in the graph below. `past_covariates` needs to include the same time steps as `target`, and `future_covariates` must include the same time span plus additional `n` forecast horizon time steps. 
+
+***You can use the same `*_covariates` for both training and prediction, given that they contain the required time spans.***
+
+![figure0](./images/covariates/top_level.png)
```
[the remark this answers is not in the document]
dennisbader: When looking at the document from the branch itself on github it seems to work

guides/covariates_tfms.md
dennisbader: True

hrzn (approved):
Thanks!

hrzn:
I guess we can close the other PR as it's not needed anymore.

PR #695, opened 2021-12-16
granger causality init

dumjax:
New feature

### Summary

Granger causality test between 2 univariate time series. Wrapper of the statsmodels granger causality function.

### Other Information

We probably want also to add later a stationarity test (it seems some non stationarity time series can still be ok for granger causality test see https://davegiles.blogspot.com/2011/04/testing-for-granger-causality.html, but at least a warning), hence probably add some more statistic utils functions coming from statsmodels, like adfuller and kpss.


darts/utils/statistics.py
```diff
@@ -255,6 +255,56 @@ def remove_trend(ts: TimeSeries,
...
+    maxlag
+        If an integer, computes the test for all lags up to maxlag. 
+        If an iterable, computes the tests only for the lags in maxlag.
+    addconst
+        Include a constant in the model.
+    verbose
+        Print results.
+    Returns
+    -------
+    Dict
```
hrzn, line 286: Could we maybe link to statsmodels doc here - as we return exactly what they return? (We can also leave the current explanations in our docstring too).

darts/utils/statistics.py
```diff
@@ -255,6 +255,56 @@ def remove_trend(ts: TimeSeries,
...
+        All test results, dictionary keys are the number of lags. For each lag the values are a tuple, 
+        with the first element a dictionary with test statistic, pvalues, degrees of freedom, the second element are 
+        the OLS estimation results for the restricted model, the unrestricted model and the restriction (contrast) 
+        matrix for the parameter f_test.
+    """
+
+    ts_cause._assert_univariate()
+    ts_effect._assert_univariate()
+
+    raise_if(not ts_cause.has_same_time_as(ts_effect),
```
hrzn, line 296: You could consider `slice_intersect()`ing the series for the user and then checking if they have the same time, which could be more user friendly I guess.

darts/utils/statistics.py
```diff
@@ -255,6 +255,56 @@ def remove_trend(ts: TimeSeries,
...
+            'ts_cause and ts_effect time series have different time index.')
+
+
+
+    return grangercausalitytests(
+        np.concatenate((ts_effect.values(), ts_cause.values()), axis=1),
```
hrzn, line 302: I think you can probably call `values(copy=False)` here, which will be slightly more efficient.

darts/utils/statistics.py
dumjax: Don't they have the same time anyway if we do `slice_intersect()`? what do you mean by "they" , is it the slice_intersect result compared to the 2 input time series and then a warning (printing the differences of time)?

darts/utils/statistics.py
dumjax: I slightly changed the explanation, but the parameter doc is the same. In case of link to statsmodels doc, just a link is enough in introduction?

darts/utils/statistics.py
```diff
@@ -255,6 +255,196 @@ def remove_trend(ts: TimeSeries,
...
+
+    Returns
+    -------
+    Boolean
+        If ts is stationary or not.
+    """
+    adf_res = stationarity_test_adf(ts)
+    kpss_res = stationarity_test_kpss(ts)
+
+    return (adf_res[1]<p_value_threshold_adfuller) & (kpss_res[1]>p_value_threshold_kpss)
```
hrzn, line 290:
```suggestion
    return (adf_res[1]<p_value_threshold_adfuller) and (kpss_res[1]>p_value_threshold_kpss)
```

darts/utils/statistics.py
hrzn: Yes, i think it's enough.

hrzn (approved):
Looks good 👍

PR #700, opened 2021-12-19
Feat/improve tests

hrzn:
* Use absolute imports in tests
* Reorganise test files into submodules matching that of the library
* Use `pytest` to run the tests (it provides a nice warnings summary instead of printing all warnings, and good functions timing out of the box)
* Use `pytest-cov` to compute coverage, and compute coverage in one go (currently our tests are always run twice!)
* Slightly speed up some of the tests, and remove a few warnings
* Started to use `black` locally (and formatting applied on test files already) *99% of the diffs will come from that :D*


dennisbader (approved):
Very nice, thanks!

Issue #707, opened 2021-12-27
Error in RNNModel - LSTM after specifying lr_scheduler_cls


shashankdeshpande:
Getting error while fitting RNNModel-LSTM if we specify lr_scheduler_cls

**To Reproduce**
Code snippet
```python
lstm_model = RNNModel(
	model='LSTM',
	optimizer_kwargs={'lr': 1e-3},
	lr_scheduler_cls=torch.optim.lr_scheduler.ReduceLROnPlateau,
	...
	)
```
Error Screenshot
![Screenshot (35)](https://user-images.githubusercontent.com/47742503/147477737-42380f03-5a63-41b5-b57d-286616674786.png)

**System**
 - Python version: 3.8.12
 - darts version: 0.14.0
 - torch version: 1.10.1+cpu

pennfranc:
Thanks for reporting! I added support for `ReduceLROnPlateau` in #711, at least using the training loss as a metric.

PR #711, opened 2021-12-28
Fix/support for ReduceLROnPlateau scheduler

pennfranc:
Addresses #707. This only adds limited support for `ReduceLROnPlateau` in the sense that it only takes the training loss as `metrics` argument to its `step()` method. To add, say, validation loss we would have to change a few more things since the lr_scheduler is called at every epoch of the training loop, whereas new validation losses are not necessarily computed everytime. I guess it's not worth spending too much time on this considering the new PTL refactoring.


hrzn (approved):
Thanks!

PR #725, opened 2022-01-08
fixed timedelta issue in longest_contiguous_slice

### LINKED ISSUE #716: [BUG] `longest_contiguous_slice` throws TypeError if `max_gap_size` is greater or equal than `'gap_size'`
**Describe the bug**
I am trying to use the `longest_contiguous_slice` method on a `TimeSeries` object with a `freq` of `'W-MON'` but I get this error.
`TypeError: Cannot add Timedelta to Week`

**To Reproduce**
```
# Simulating random data
idx = pd.date_range(start='2018-01-01', periods=100, freq='W-MON')
values = np.random.uniform(low=1, high=300, size=(100))
values.ravel()[np.random.choice(values.size, 40, replace=False)] = np.nan
s = pd.Series(values, index=idx)
# Creating TimeSeries Object
ts = TimeSeries.from_series(s)
# Returning the largest TimeSeries slice that contains no gaps larger than 2
ts.longest_contiguous_slice(max_gap_size=2)
```

**Expected behavior**
Returning the largest TimeSeries slice that contains no gaps larger than 2

 - Python version: [e.g. 3.9.5]
 - darts version [e.g. 0.15.0]

**Additional context**
I noticed that the method would work if I set the `max_gap_size` to a number greater than or equal to 
`ts.gaps()['gap_size'].max()`

dennisbader:
Fixes #716

### Summary

- fixed timedelta issue in `TimeSeries.longest_contiguous_slice()`


darts/timeseries.py
```diff
@@ -1468,7 +1468,7 @@ def longest_contiguous_slice(self, max_gap_size: int = 0) -> 'TimeSeries':
         max_slice_start = None
         max_slice_end = None
         for index, row in relevant_gaps.iterrows():
-            size = row['gap_start'] - curr_slice_start - self._freq
```
hrzn, line 1471: Does changing the order of the operands make a difference here?

dennisbader: Yes, 
- `row['gap_start'] - curr_slice_start`  gives a time delta in days from which we cannot subtract the frequency.
- `row['gap_start'] - self._freq` gives a time stamp from which we can subtract `curr_slice_start`

tomasvanpottelbergh: Could we change our API to avoid relying on the order of the operands? Because this seems a bit fragile... I would expect a subtraction to always return the same type.

dennisbader: This is how pandas handles the subtraction
- `pd.Timestamp - pd.Timestamp -> pd.TimeDelta`
- `pd.Timestamp - pandas DateOffset (freq) -> pd.Timestamp`

As TimeDelta doesn't carry information about the exact dates, we cannot subtract **irregular** frequencies (such as month start 'MS', etc.) which is why the original operand order fails. 

For **regular** frequencies (such as 'D') the order doesn't matter for the results.

Maybe we could make it more obvious with:

```
curr_slice_end = row['gap_start'] - self._freq
size = curr_slice_end - curr_slice_start
```
?

hrzn: +1 with your solution (and adding a comment) @dennisbader

hrzn (approved):

Issue #724, opened 2022-01-06
[BUG] TypeError: __deepcopy__() takes 1 positional argument but 2 were given when using darts.models.forecasting.regression_ensemble_model.RegressionEnsembleModel


omerfguzel:
**Describe the bug**
I'm trying to use Ensemble model but I got this error. Any suggestions?

**System:**
 - Python version: 3.7
 - darts version 0.15.0

**Additional context**
full error line: TypeError: __deepcopy__() takes 1 positional argument but 2 were given

dennisbader:
Hi, could you please send a code snippet so we can reproduce the issue? Thanks.

omerfguzel:
Solved and here is how it is solved:
https://github.com/h3ik0th/Darts_ensembleFC/issues/1#issuecomment-1007209786

dennisbader:
When creating a model object all parameters are saved.
Deepcopying fitted/trained `forecasting_models` with EnsembleModels fails.

You can reset your models with `models_untrained = [m.untrained_model() for m in models]`

PR #728, opened 2022-01-09
added check for untrained models for copying of model parameters

dennisbader:
Addresses #724.

### Summary

- instantiating ensemble models with fitted/trained forecasting model resulted in deepcopy error. 
- added check for untrained models with explanatory error message.


hrzn (approved):
Thanks!

PR #732, opened 2022-01-11
Feat/set up black

tomasvanpottelbergh:
Closes #585.

- Adds `black` to dev requirements
- Adds instructions to CONTRIBUTING.md
- Adds pre-commit hook to format automatically using Black
- Adds Black linting to CI

TO DO:
- [x] Add `.git-blame-ignore-revs` file for commit applying black to codebase (see #733)

PR #733, opened 2022-01-11
Feat/format using black

tomasvanpottelbergh:
Applies Black to entire codebase + fixes some minor formatting issues.

This is done as a separate PR so that the resulting squash-merge commit can be added to the `.git-blame-ignore-revs` file in #732.


hrzn (approved):
Looking good ™️

Issue #606, opened 2021-12-01
Vectorize dataset creation in `RegressionModel`s


hrzn:
Goal: make it faster to create the training dataset on long series.

This can be done only on a subset of cases, where the target is a “simpler” function of covariates (wont work for arbitrary lags combinations).

See how it’s done with LGBM.

PR #736, opened 2022-01-12
Fix/optimize regression model fit

brunnedu:
Addresses #606 

### Summary

- Vectorized computation of X and y matrices during fitting of RegressionModel by replacing the _create_lagged_data function
- Adjusted unit tests of regression models and regression ensemble model

### Other Information

- RegressionModel now imports pandas which is used for matching time indices of target, past covariate and future covariate series
- If no samples can be built, the error message is now less descriptive than the one in MixedCovariatesSequentialDataset
- Significant speedup over previous preprocessing method (up to 400x)
- Works for arbitrary lag combinations (no need to be contiguous)


darts/models/forecasting/regression_model.py
```diff
@@ -282,65 +282,94 @@ def _get_last_prediction_time(self, series, forecast_horizon, overlap_end):
...
-            past_covariates=past_covariates,
-            future_covariates=future_covariates,
-            input_chunk_length=max(0, -self.min_lag),
-            output_chunk_length=max(
-                1, self.max_lag + 1
-            ),  # max lag + 1 since we need to access that index
-            max_samples_per_ts=max_samples_per_ts,
-        )
+            # combine lags
+            df_X = pd.concat(df_X, axis=1)
```
hrzn, line 352: This is neat 👌

hrzn (approved):
Very nice 💯

darts/models/forecasting/regression_model.py
brunnedu: Fortunately, I could let pd.concat do the task of aligning the different series with each other 😄

PR #741, opened 2022-01-15
Feat/return estimator

hrzn:
Make `fit()` method of models return `self` to allow chained calls.


brunnedu (approved):
Nice update! Checked all models and think you didn't miss any fit method 👍

tomasvanpottelbergh:
Nice, this will be really useful for pipelines! It would be nice to use it in the tests and example for `KalmanFilter` as well.

tomasvanpottelbergh (approved):
Nice, this will be really useful for pipelines! It would be nice to use it in the tests and example for `KalmanFilter` as well.

Issue #661, opened 2021-12-02
[New model] Forecasting from Kalman Filter


hrzn:
Should leverage the Kalman Filter filtering model we already have.

tomasvanpottelbergh:
Duplicate of #634

PR #743, opened 2022-01-15
Feat/kalman forecaster

hrzn:
Fixes: #661 #735 

* Introduces `KalmanForecaster`, a new forecasting model based on `KalmanFilter`
* Improved `KalmanFilter` to handle missing values
* Improved `KalmanFilter` to return a series with columns names matching that of the original series
* Improved `ForecastingModel`s to return predicted series with a time axis having the same name as the original series being forecasted
* Added new unit tests for `KalmanForecaster`
* Other small improvements.


darts/models/forecasting/regression_model.py
```diff
@@ -335,7 +338,7 @@ def _create_lagged_data(
             ]
 
             df_X = []
-            df_y = target_ts.pd_dataframe()
+            df_y = target_ts.pd_dataframe(copy=False)
```
brunnedu, line 341: Good modification, should've thought of that 👍

brunnedu (commented):
Looks good to me, but @tomasvanpottelbergh could probably give better feedback 😄

darts/models/filtering/kalman_filter.py
```diff
@@ -197,48 +204,50 @@ def filter(
...
+            sampled_outputs = np.zeros((len(y_values), self.dim_y, num_samples))
 
+        mean_vec = None  # store previous mean
         for i in range(len(y_values)):
             y = y_values[i, :].reshape(-1, 1)
+
+            if np.isnan(y).any():
+                raise_if(i == 0, "The first value of the series must not be NaN.")
+                # if an observation is missing, take the previous mean as an observation
+                y = mean_vec.reshape(-1, 1)
```
tomasvanpottelbergh, line 232: I'm not sure if this is the best way of handling missing values here, especially if used for forecasting. I think this might result in an overconfident probabilistic output. I will have to take a closer look, but maybe we should use `nfoursid`'s `extrapolate()` method instead or add something similar for this purpose?

darts/models/filtering/kalman_filter.py
```diff
@@ -102,12 +107,12 @@ def fit(
 
         # TODO: Handle multiple timeseries. Needs reimplementation of NFourSID?
         self.dim_y = series.width
-        outputs = series.pd_dataframe()
+        outputs = series.pd_dataframe(copy=False)
```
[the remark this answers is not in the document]
hrzn: I tested the column renaming and it is fine - I think the DF contains a copy of the series' columns anyway.

darts/models/filtering/kalman_filter.py
hrzn: You're right, this is not quite correct. I guess we should run the state space system one step instead. I have [opened an issue with nfoursid](https://github.com/spmvg/nfoursid/issues/3) to ask what might be the best way to do it.

hrzn:
## Examples

Forecasting:
image

Missing values imputation:
image

darts/models/filtering/kalman_filter.py
hrzn: @tomasvanpottelbergh I have now changed it to take a step of the Kalman filter (which takes a step of the underlying state space model), but not taking the measurements into account in the update.
To this end I had to overwrite nfoursid `step()` function with a custom-made `step_accepting_missing_obs()` function. As this is not very future-proof, I've also fixed nfoursid's version in the requirements for the time being.

For more context about this choice, you can checkout https://github.com/spmvg/nfoursid/issues/3

darts/models/filtering/kalman_filter.py
```diff
@@ -204,48 +280,50 @@ def filter(
         y_values = series.values(copy=False)
         if self._expect_covariates:
             u_values = covariates.values(copy=False)
+
+            # set control signal to 0 if it contains NaNs:
+            # u_values = np.nan_to_num(u_values, copy=True, nan=0.0)
```
tomasvanpottelbergh, line 285: This should probably be uncommented? Also, it would be good to add tests for missing values both in observations and covariates

darts/models/filtering/kalman_filter.py
```diff
@@ -10,13 +10,86 @@
...
+    x_pred = (
+        self.x_predicteds[-1]
+        if self.x_predicteds
+        else np.zeros((self.state_space.x_dim, 1))
+    )
+    p_pred = (
+        self.p_predicteds[-1] if self.p_predicteds else np.eye(self.state_space.x_dim)
+    )
+
+    if y is not None:
```
tomasvanpottelbergh, line 47: Nice, I think this makes sense! It's equivalent to setting the gains to zero, right? Maybe doing that explicitly is clearer? This is consistent with https://github.com/pykalman/pykalman/blob/master/pykalman/standard.py#L289.

hrzn: Indeed I think this is equivalent, but the reason I didn't do it is computational:
* If we set K=0, we need the operand to multiply it with, which in turns requires `y` (and would crash if `y` is NaN)
* Even if I treated this above case, we would still need the rest of the operand, which requires computing some matrix multiplications for nothing.

Maybe I'll add a comment to the code to make this clear.

darts/models/filtering/kalman_filter.py
```diff
@@ -204,48 +208,50 @@ def filter(
...
-            sampled_states = np.zeros((len(y_values), self.dim_y, num_samples))
+            sampled_outputs = np.zeros((len(y_values), self.dim_y, num_samples))
 
+        # mean_vec = None  # store previous mean
         for i in range(len(y_values)):
             y = y_values[i, :].reshape(-1, 1)
             u = u_values[i, :].reshape(-1, 1)
+
+            if np.isnan(y).any():
+                raise_if(i == 0, "The first value of the series must not be NaN.")
```
tomasvanpottelbergh, line 234: I think this is not necessary anymore?

darts/models/filtering/kalman_filter.py
```diff
@@ -204,48 +208,50 @@ def filter(
...
                 (
                     len(y_values),
                     self.dim_y,
                 )
             )
         else:
-            sampled_states = np.zeros((len(y_values), self.dim_y, num_samples))
+            sampled_outputs = np.zeros((len(y_values), self.dim_y, num_samples))
 
+        # mean_vec = None  # store previous mean
```
tomasvanpottelbergh, line 228: Can be removed?

darts/models/filtering/kalman_filter.py
hrzn: Indeed, thanks.

tomasvanpottelbergh (approved):
Great improvement and addition, thanks! :) LGTM!

PR #752, opened 2022-01-24
Remove broken type annotation

hrzn:


brunnedu (approved):
Seems to be the only place we used it. Thanks for the quick fix!

PR #760, opened 2022-01-27
Fix/added unique ids to model_name

### LINKED ISSUE #759: [BUG] tensorboard does not work with gridsearch when model_name is provided
**Describe the bug**
providing a model name when using gridsearch puts all files under the same folder `./darts/runs/MODEL_NAME`
this breaks tensorboard 

**To Reproduce**
```
NBEATS_params = {
        "log_tensorboard": [True],
        "model_name": ["NBEATS_gs"],
        "layer_widths": [256, 512, 1024],
}

best_model, dict_params = NBEATSModel.gridsearch(NBEATS_params, train_air, val_series=val_air, metric=mape, n_jobs=3, verbose=True)
```

**Expected behavior**
each run should have its own dir

 - Python version: 3.7
 - darts version 0.16.0

**Additional context**

gdevos010:
Fixes #759 

### Summary

added date and pid to name


darts/models/forecasting/forecasting_model.py
```diff
@@ -681,6 +683,12 @@ def _evaluate_combination(param_combination):
             param_combination_dict = dict(
                 list(zip(parameters.keys(), param_combination))
             )
+            if param_combination_dict.get("model_name", None):
+                current_time = time.strftime("%Y-%m-%d_%H:%M:%S", time.localtime())
+                param_combination_dict[
+                    "model_name"
+                ] = f"{current_time}_{param_combination_dict['model_name']}_{os.getpid()}"
```
hrzn, line 690: How about using something that allows mapping back to the parameters combination being used? Maybe instead of the current time, using the index of the currently evaluated parameters within `params_cross_product` would make this easier? Also do we need the process PID?

gdevos010: I followed the current naming convention.

https://github.com/unit8co/darts/blob/96125f839c115e49d44699ab7d911e8413d4d0b6/darts/models/forecasting/torch_forecasting_model.py#L180-L182

The naming was discussed in #346

The `params_cross_product` is not provided to the user, so it's unclear how useful this would be. The name also needs to be unique across consecutive runs; otherwise, multiple events files will be in the same folder again. So a simple index would not surface. We could remove the PID and just use the current time.

gdevos010: The hyperparameters could be saved to tensorboard to map them back quickly

hrzn: OK, so let's leave it as you propose, but without the PID, OK?
Also, could you unit-test this? Thanks.

gdevos010: I updated the naming convention. I'm not sure its worth the time to unit test this since hyperparameter tuning will completely change with pytorch-lightning. I could try to do it, but work has really picked up this week.

hrzn (approved):
Thanks!

Issue #618, opened 2021-12-01
Handle user-defined `output_chunk_length` in RegressionModels


hrzn:
Right now our regression models always predict 1 time step in advance auto-regressively. We should update this to the more general case of predicting `output_chunk_length` at a time.
This could increase model quality in some cases, and allow models to exploit past covariates to predict more time steps in advance.

PR #761, opened 2022-01-31
Feat/output chunk length regression model

brunnedu:
Addresses #618 

### Summary

RegressionModels:
- Implemented `output_chunk_length` parameter for all RegressionModels allowing for much more flexible prediction with less covariate data
- deleted now unused functions (`_get_prediction_data`, `_shift_matrices`, `_update_min_max`)
- turned `lags` into a dictionary for simpler code
- updated check of covariate input dimensions
- updated checks for enough covariate data (now the requirements are minimal and can differ between covariate types)
- only wraps model with `MultiOutputRegressor` if it is necessary and model doesn't support multi-output regression natively
- doesn't rely on torch datasets anymore

RegressionModel Tests:
- Deleted now unused tests (`test_models_denoising_multi_input`, `test_models_denoising`, `test_shift_matrices`)
- Added new tests (`test_models_accuracy_*`, `test_multioutput_wrapper`)
- Added checks for `output_chunk_length=1` and `output_chunk_length=5` in some already existing tests
- Updated `test_not_enough_covariates` to check if minimal covariates are required

### Other Information

- changed the `__str__` of `LinearRegressionModel`, `RandomForest`, `LightGBMModel`
- should we consider deleting the `LinearRegressionModel`, since it behaves the same as `RegressionModel(model=None)`?


darts/models/forecasting/regression_model.py
```diff
@@ -428,105 +352,61 @@ def fit(
...
+            # and model isn't wrapped already
+            if not isinstance(self.model, MultiOutputRegressor):
+                # check whether model supports multi-target regression natively
+                if not (
+                    callable(getattr(self.model, "_get_tags", None))
+                    and isinstance(self.model._get_tags(), dict)
+                    and self.model._get_tags().get("multioutput")
+                ):
+                    # if not, wrap model with MultiOutputRegressor
+                    self.model = MultiOutputRegressor(self.model, n_jobs=1)
```
hrzn, line 402: I wonder there could be some cases where `n_jobs > 1` might provide significant gains...

darts/models/forecasting/regression_model.py
```diff
@@ -582,123 +463,114 @@ def predict(
...
-            target_series=series,
-            past_covariates=past_covariates,
-            future_covariates=future_covariates,
-            n=max(
-                n + self.max_lag, n
-            ),  # required for retrieving enough future covariates
-            input_chunk_length=max(0, -self.min_lag),
-            output_chunk_length=1,
-        )
+                    if cov.has_datetime_index:
```
hrzn, line 519: Ouch, it hurts that we have to differentiate the two cases here 😱

hrzn (approved):
This is really awesome work @brunnedu, thanks!
You are bringing our RegressionModels to the next step! 🚀

darts/models/forecasting/regression_model.py
brunnedu: yes, maybe it would be nice to have a slice_inclusive in TimeSeries that always includes the start and stop for both datetime and integer indices? 🤔

darts/models/forecasting/regression_model.py
brunnedu: right, I will test it out 👍

darts/models/forecasting/regression_model.py
hrzn: Yes maybe. I think our slicing is at least consistent with Pandas, but it's not the first time we find this a bit frustrating.

PR #762, opened 2022-01-31
Add workaround for broken pip-tools

tomasvanpottelbergh:
Fixes broken builds due to https://github.com/jazzband/pip-tools/issues/1558.


brunnedu (approved):
Thanks for the quick fix! 👍

PR #769, opened 2022-02-03
fix: bad file descriptor

### LINKED ISSUE #768: [BUG] bad file descriptor error when using Prophet with Dask
**Describe the bug**
When using Prophet, which uses a context to suppress output from Prophet logging, an error `bad file descriptor` may happen if running in parallel with Dask for a long time (about 15 minutes).

**To Reproduce**
- Create a Dask cluster
- Prepare a time series source with multiple time series to forecast
- Load the big time series file with Dask dataframe
- Groupby the division of the series, and apply forecasting to each division; for example, if we have columns `[store, product, time, sales]`, and we want to forecast for each product in a store, we would do `data.groupby(['store', 'product']).apply(forecast, meta=...)`
- Wait for the error to occur

**Expected behavior**
The forecast processes on Dask should finish without error, or more specifically, without error not relating for forecasting.

 - Python version: [e.g. 3.8]
 - darts version [e.g. 0.14.0]

**Additional context**
 - System: Macbook Pro M1 - Monterey

nlhkh:
Fixes #768 .

### Summary

When using Prophet under a multi-threading environment such as Dask, the error `[Errno 9] Bad File Descriptor` could happen. I can consistently reproduce this error while running forecasts with Dask. This implementation seems to fix the issue.

I do not fully understand why this happens, because only when running Dask for a long time would this happen. My implementation is semantically equivalent to the original, but I guess avoiding to open `devnull` too many times can help.

### Other Information

I am running forecast using Prophet for about 500_000 time series, all in 1 big Dask data frames (split into 91 smaller Pandas data frames).
If I don't use Dask, and instead use *joblib* to parallelize the forecasting, then this error does not happen. I think because joblib creates a separate thread for every worker, whereas Dask can have multiple threads for a worker.


nlhkh:
In more rare occasions, this solution also resulted in bad file descriptor error. I am working on a workaround.

nlhkh:
This is ready. It does not change existing behavior. It only let user turn off the suppression if desired.

hrzn:
Thank you @nlhkh, could you maybe add a (lightweight) unit test case to test the case `suppress_stdout_stderror=False` ?

nlhkh:
> Thank you @nlhkh, could you maybe add a (lightweight) unit test case to test the case `suppress_stdout_stderror=False` ?

 Thanks @hrzn . I will add it soon.

nlhkh:
@hrzn I have included the unit test as requested. I don't know why a new Prophet model is created every time `_fit` is called, but I guess the intention is to fit multiple times. If that is not the intention, I can simplify this part.

darts/models/forecasting/prophet_model.py
```diff
@@ -88,6 +91,10 @@ def __init__(
         self.country_holidays = country_holidays
         self.prophet_kwargs = prophet_kwargs
         self.model = None
+        self.suppress_stdout_stderr = suppress_stdout_stderror
+
+        self.execute_and_suppress_output = execute_and_suppress_output
+        self.model_builder = prophet.Prophet
```
hrzn, line 97: Why is this needed? It was somewhat simpler to call `prophet.Prophet` directly in the `fit()` method, no?

hrzn (commented):
Thank you, that looks good! I only have a minor comment.
Also, before we can merge, can you make sure formatting and linting checks are passing? You can follow [these instructions](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md#code-formatting-and-linting) to enable the checks. Thanks!

darts/models/forecasting/prophet_model.py
nlhkh: This is needed for unit testing, because I think we don't need to invoke the actual fit to test this feature; we only want to make sure that method fit is called. Secondly, in the previous approach, with Prophet model being constructed for every fit, I cannot really check if fit is called.

nlhkh: i.e. if I mock this `model_builder`

```
model = darts.Prophet(...)
model.model_builder = Mock(return_value=Mock(fit=Mock(return_value=True)))
```

Then I can:
1. Avoid calling the real fit
2. Check if fit is called once

nlhkh: I think another unit test is needed to make sure that the default `model_builder = prophet.Prophet`

hrzn: OK, got it, makes sense indeed, I missed this in the unit tests. Yes, maybe the extra unit test you suggest would be nice too.

hrzn: OK, got it, makes sense indeed, I missed this in the unit tests. Yes, maybe the extra unit test you suggest would be nice too.

hrzn:
> @hrzn I have included the unit test as requested. I don't know why a new Prophet model is created every time `_fit` is called, but I guess the intention is to fit multiple times. If that is not the intention, I can simplify this part.

We re-create to ensure that subsequent calls to `fit()` always start from a "clean slate" model. However to be honest I don't really remember if this is actually needed (perhaps `Prophet` already has this behaviour). If you want you can test with creating the model only once in the constructor and propose this change in this PR too.

nlhkh:
@hrzn I have ran all local lint checks, and it is okay now.

nlhkh:
> > @hrzn I have included the unit test as requested. I don't know why a new Prophet model is created every time `_fit` is called, but I guess the intention is to fit multiple times. If that is not the intention, I can simplify this part.
> 
> We re-create to ensure that subsequent calls to `fit()` always start from a "clean slate" model. However to be honest I don't really remember if this is actually needed (perhaps `Prophet` already has this behaviour). If you want you can test with creating the model only once in the constructor and propose this change in this PR too.

I checked with Facebook Prophet model, and you cannot fit a model twice, so yes, you will need a fresh model for every fit.

hrzn (approved):

hrzn:
Thank you @nlhkh - merging now :)

nlhkh:
> Thank you @nlhkh - merging now :)

Thanks @hrzn

PR #773, opened 2022-02-07
Fix/simple statistics

### LINKED ISSUE #719: Improve "Simple statistics" functions of `TimeSeries`.
At the moment our functions `mean()`, `var()` etc are simple wrappers around `pandas.DataFrame` functions, which do not work for probabilistic series, and often do not really do what we need.

--> Always take statistics over "samples" dimension.

Gian-Wiher:
Fixes #719 .

### Summary
Replaced "Simple statistics" methods with new functions that accept stochastic TimeSeries.

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -27,6 +27,7 @@ from typing import Tuple, Optional, Callable, Any, List, Union, Sequence
 from inspect import signature
 from collections import defaultdict
 from pandas.tseries.frequencies import to_offset
+from scipy.stats import skew, kurtosis
 
 from .logging import raise_log, raise_if_not, raise_if, get_logger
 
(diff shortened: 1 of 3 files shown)


darts/timeseries.py
```diff
@@ -2438,94 +2439,97 @@ def with_columns_renamed(
...
+        return self.__class__(new_xa)
 
-    def autocorr(self, lag=1) -> float:
-        """Simple wrapper around :func:`pd.DataFrame.autocorr()` for deterministic series."""
-        return self.pd_dataframe(copy=False).autocorr(lag)
-
-    def describe(self, percentiles=None, include=None, exclude=None) -> pd.DataFrame:
-        """Simple wrapper around :func:`pd.DataFrame.describe()` for deterministic series."""
-        return self.pd_dataframe(copy=False).describe(percentiles, include, exclude)
+    def quantile(self, q: float, method: str = "linear") -> "TimeSeries":
```
hrzn, line 2519: We already have a function `quantile_timeseries` which does the same thing :)
However uses a dataframe which is not great (I prefer using Numpy), and so maybe we could leverage your implementation there? And also keep the `quantile()` method too, as:
```
def quantile(...):
    return self.quantile_timeseries(...)
```

darts/timeseries.py
```diff
@@ -2438,94 +2439,97 @@ def with_columns_renamed(
...
+        new_xa = xr.DataArray(new_data, dims=self._xa.dims, coords=self._xa.coords)
+        return self.__class__(new_xa)
 
     def kurtosis(
-        self, axis=None, skipna=True, level=None, numeric_only=None, **kwargs
-    ) -> float:
-        """Simple wrapper around :func:`pd.DataFrame.kurtosis()` for deterministic series."""
-        return self.pd_dataframe(copy=False).kurtosis(
-            axis, skipna, level, numeric_only, **kwargs
+        self, fisher: bool = True, bias: bool = True, nan_policy: str = "propagate"
```
hrzn, line 2478: How about keeping all these arguments as kwargs? This way we don't have to repeat all the arguments from Numpy (and risk incompatibilities in the future). Same for the other methods where this applies.

hrzn (commented):
This looks pretty good, thanks! I added a few comments.
Could you also add some unit tests? Thanks :)

darts/timeseries.py
```diff
@@ -2438,94 +2439,97 @@ def with_columns_renamed(
...
+        return self.__class__(new_xa)
 
-    def min(
-        self, axis=None, skipna=True, level=None, numeric_only=None, **kwargs
-    ) -> float:
-        """Simple wrapper around :func:`pd.DataFrame.min()` for deterministic series."""
-        return self.pd_dataframe(copy=False).min(
-            axis, skipna, level, numeric_only, **kwargs
-        )
+    def min(self) -> "TimeSeries":
```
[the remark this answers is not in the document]
Gian-Wiher: this sounds good. Should we provide the same functionality for `mean` and `median` as well? As these are common aggregation function I think it would fit in.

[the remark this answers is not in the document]
hrzn: Yes, good idea 👍

darts/timeseries.py
```diff
@@ -2438,94 +2465,134 @@ def with_columns_renamed(
...
-    def median(
-        self, axis=None, skipna=None, level=None, numeric_only=None, **kwargs
-    ) -> float:
-        """Simple wrapper around :func:`pd.DataFrame.median()` for deterministic series."""
-        return self.pd_dataframe(copy=False).median(
-            axis, skipna, level, numeric_only, **kwargs
-        )
+    def sum(self, axis: int = 2) -> "TimeSeries":
+        """Simple wrapper around :func:`np.ndarray.sum()`."""
+        new_data = self._xa.values.sum(axis=axis, keepdims=True)
```
hrzn, line 2512: Why not relying on the DataArray functions directly when possible?
E.g.: https://xarray.pydata.org/en/stable/generated/xarray.DataArray.sum.html
Maybe it would do the coordinates management for you when possible.

darts/tests/test_timeseries.py
```diff
@@ -1452,3 +1453,113 @@ def test_time_col_convert_garbage(self):
 
         with self.assertRaises(AttributeError):
             TimeSeries.from_dataframe(df=df, time_col="Time")
+
+
+class SimpleStatisticsTestCase(DartsBaseTestClass):
+
+    times = pd.date_range("20130101", "20130110", freq="D")
+    values = np.random.rand(10, 2, 1000)
```
hrzn, line 1461: (Just in case the tests are a little slow, as there are many of them):
```suggestion
    values = np.random.rand(10, 2, 100)
```

darts/tests/test_timeseries.py
```diff
@@ -1452,3 +1453,113 @@ def test_time_col_convert_garbage(self):
 
         with self.assertRaises(AttributeError):
             TimeSeries.from_dataframe(df=df, time_col="Time")
+
+
+class SimpleStatisticsTestCase(DartsBaseTestClass):
```
hrzn, line 1458: Good unit tests, thanks 👍

hrzn (approved):
Looks good! 👍

PR #775, opened 2022-02-08
Feat/backtest sliding window

amadejkocbek:
Draft implementation of sliding backtest window, extended some simple unit tests to lightly test. I think a better test is still required.


darts/models/forecasting/forecasting_model.py
```diff
@@ -307,6 +308,8 @@ def historical_forecasts(
         num_samples
             Number of times a prediction is sampled from a probabilistic model. Should be left set to 1
             for deterministic models.
+        train_length
+            Number of time steps in our training set (size of backtesting window to train on).
```
brunnedu, line 312: We could mention that it uses all available time steps up until prediction time if set to `None`.

brunnedu (commented):
Thanks! Looks good 👍

darts/models/forecasting/forecasting_model.py
hrzn: +1. Also you can explain that the expanding window strategy is used by default with train_length=None, and otherwise the moving window strategy is used.

darts/models/forecasting/forecasting_model.py
```diff
@@ -382,7 +385,21 @@ def historical_forecasts(
 
         # iterate and forecast
         for pred_time in iterator:
-            train = series.drop_after(pred_time)  # build the training series
+            # build the training series
+            train = series.drop_after(pred_time)
+            if train_length and not isinstance(train_length, int):
```
hrzn, line 390: These checks should be done earlier in the method (as opposed to in every iteration of the loop)

darts/models/forecasting/forecasting_model.py
```diff
@@ -382,7 +385,21 @@ def historical_forecasts(
...
+                    logger,
+                )
+            elif (train_length is not None) and train_length < 1:
+                raise_log(
+                    ValueError("If not None, train_length needs to be positive."),
+                    logger,
+                )
+            if train_length and len(train) > train_length:
+                # take last train_length samples (drop_before?)
+                train = train[:train_length]
```
hrzn, line 402: Unless I'm missing something, that will take the _first_ `train_length` samples. I think you should write:
```suggestion
                train = train[-train_length:]
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -473,6 +491,8 @@ def backtest(
         num_samples
             Number of times a prediction is sampled from a probabilistic model. Should be left set to 1
             for deterministic models.
+        train_length
+            Number of time steps in our training set (size of backtesting window to train on).
```
hrzn, line 495: Same comment here

darts/tests/models/forecasting/test_backtesting.py
```diff
@@ -194,6 +223,8 @@ def test_backtest_forecasting(self):
             self.assertEqual(pred.width, 2)
             self.assertEqual(pred.end_time(), linear_series.end_time())
 
+            # TODO: adjusted window size hard to test with tcn model
```
hrzn, line 226: Is this comment still neeed?

darts/tests/models/forecasting/test_backtesting.py
```diff
@@ -100,6 +100,26 @@ def test_backtest_forecasting(self):
         )
         self.assertEqual(score, 1.0)
 
+        # very large train length should not affect the backtest
+        score = NaiveDrift().backtest(
```
hrzn, line 104: Thanks for the unit tests!
Maybe we could add one that checks that the backtest accuracy remains OK for of one of the model when a `train_length` is applied?

darts/tests/models/forecasting/test_backtesting.py
amadejkocbek: I wanted to make sure we still add a unit test of the sort you suggested, that accuracy is still OK with one of the not so simple models.

hrzn:
As discussed, can you implement `min_train_length` in the models where it's missing? Thx :)

amadejkocbek:
Now added an implementation of min train length for regression model. As for the other models, especially the network models, it looks like most correctly inherit the method from torch_forecasting_model (sum of input + output chunk lengths), and similarly random forest will inherit from regression model. So it was only crucial to override for regression model. The current implementation is more on the tentative side, but hopefully it passes the basic sanity check.

amadejkocbek:
Thanks to Dustin we have now a more correct and elegant way of calculating it. Unfortunately it’s still not covering the cases when covariates are involved, but it covers most use cases.

hrzn (approved):
Looks good thanks 👍 
You can merge once you've double checked the `min_train_length` on the `RegressionModel` with @brunnedu.

PR #777, opened 2022-02-09
Fix/int index warnings

hrzn:
Remove Int64Index support and keep only RangeIndex support.

As a consequence, this means that `TimeSeries` cannot have missing dates anymore (which was possible before with Int64Index; RangeIndex do not have holes by construction).

I had to adapt
* `TimeSeries.from_dataframe()` to handle correctly time columns made of integers (and ensure they correctly map to a RangeIndex)
* The `__getitem__()` method, because `sel()` on a `DataArray` returns an array indexed with an Int64Index (see: https://github.com/pydata/xarray/issues/6256)
* The unit tests to make sure these changes are covered.

Some warnings remain for the time being because of statsmodels.


darts/timeseries.py
```diff
@@ -15,7 +15,14 @@
       | **stochastic** (or **probabilistic**).
 
 Each series also stores a `time_index`, which contains either datetimes (:class:`pandas.DateTimeIndex`)
-or integer indices (:class:`pandas.Int64Index`).
+or integer indices (:class:`pandas.RangeIndex`).
+
+``TimeSeries`` are guaranted to:
```
brunnedu, line 20: good idea to specify that here 👍 
```suggestion
``TimeSeries`` are guaranteed to:
```

brunnedu (commented):
Thanks for this important update! 🚀 
Looks good to me, but it would be good if someone more familiar with xarray could take a look.

PR #780, opened 2022-02-10
Make TimeSeries.shift() tests pass on pandas 1.0.5

adelnick:
Fixes #.

### Summary

Currently, TimeSeries.shift(...) fails on pandas 1.0.5 if the frequency cannot be inferred. This fix solves the problem and makes it possible to use darts with pandas 1.0.5

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -1734,6 +1734,8 @@ class TimeSeries:
             new_time_index = self._time_index + n * self.freq
         else:
             new_time_index = self._time_index.map(lambda ts: ts + n * self.freq)
+            if new_time_index.freq is None:
+                new_time_index.freq = self.freq
         new_xa = self._xa.assign_coords({self._xa.dims[0]: new_time_index})
         return self.__class__(new_xa)
 


hrzn:
Thank you @adelnick. Could you update `requirements/core.txt` to correspondingly change the lower bound on the Pandas version? This way we will see here in the CI checks if all tests pass. Thanks.

adelnick:
> @hrzn Done!

hrzn:
> > @hrzn Done!

Actually I was wrong, we won't see that in the CI tests, but it's still good you made the change and I'll test it manually.

hrzn:
@adelnick I made some tests and some of the unit tests were still not passing with pandas==1.0.5. However I fixed them, and you can see the result here: https://github.com/unit8co/darts/pull/786

There is only one extra line changed in `timeseries.py` in my PR compared to yours. If you adopt this line - then we can merge your PR and I'll close mine :)

adelnick:
@hrzn Cool! Hope it's complete now.

hrzn:
> @hrzn Cool! Hope it's complete now.

Almost :) Two small requests for the `requirements/core.txt` file:
* Could you solve the conflicts with our `master` branch? (the order of the dependencies in the file has changed)
* Could you revert to `pandas>=1.0.5` (instead of `pandas>=1.0.0`)? I have made checks with 1.0.0 and it was not working (1.0.5 was).

adelnick:
@hrzn one more attempt :) I'm surprised to see that GitHub still marks for addition all the files that were pulled from 'master' (Bitbucket doesn't), but maybe that's me who did it wrong.

hrzn:
Ouch, it shouldn't be like that - seems that for some reason you are now trying to revert some recent formatting fixes that happened in many files between the moment you forked from our master branch and now.
I think at this stage the simplest might be to make a fresh fork again and re-open the PR. Would this work for you @adelnick ? Sorry for the confusion.

Issue #803, opened 2022-02-16
[BUG] NBEATS seasonality generator wrong basis


vladimir-chernykh:
**Describe the bug**
When creating the interpretable NBEATS model, the seasonality stack should contain Fourier-type basis vectors. It is described in the [original paper](https://arxiv.org/pdf/1905.10437.pdf):

![Screenshot 2022-02-16 at 15 34 31](https://user-images.githubusercontent.com/6755162/154265558-e78c5e39-07eb-4903-9560-4017515c4474.jpg)

The timesteps in **t** vector should be **normalized** by the horizon length for both trend and seasonality generators.

In DARTS implementation they are **not normalized** in `_SeasonalityGenerator`
https://github.com/unit8co/darts/blob/cd361e37ab57d74e02787d2a29cb8fa759febaf8/darts/models/forecasting/nbeats.py#L48-L69

It leads to the fact that interpretable models **do not learn seasonality at all**.

**To Reproduce**
Train any interpretable model and see that the seasonality is not learnt. Example: simplest `sin` function as data.

```python
import numpy as np
import matplotlib.pyplot as plt

from darts.models import NBEATSModel
from darts import TimeSeries

model_nbeats = NBEATSModel(
    input_chunk_length=100, 
    output_chunk_length=10,
    generic_architecture=False,
    n_epochs=1,
    batch_size=200,
)

series = TimeSeries.from_values(np.sin(np.arange(1000)))
series_train = series[:-10]
series_valid = series[-10:]

model_nbeats.fit(series_train)

forecast = model_nbeats.predict(10)

plt.figure(figsize=(20, 8))
series_train[-10:].plot(label="Real data")
series_valid.plot(label="Real data")
forecast.plot(label="Predictions")
```

![Screenshot 2022-02-16 at 15 56 59](https://user-images.githubusercontent.com/6755162/154269214-f7d96aed-c0dd-46b9-9129-e2a53d19598b.jpg)

One can also look at the first basis vector (which should be cosine function).

```python
first_basis_vector = model_nbeats.model.stacks[1].blocks[0].forecast_g.basis[1].cpu().numpy()
print(first_basis_vector)
# [1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
```

**Expected behavior**
The seasonality should be learnt as described in the paper. One needs to fix the basis vectors to achieve that.

**System (please complete the following information):**
 - Python version: 3.9
[...]

PR #804, opened 2022-02-16
Fix wrong seasonality basis generator for NBEATS

vladimir-chernykh:
Fixes https://github.com/unit8co/darts/issues/803.

### Summary

It fixes wrong seasonality basis vectors generated for NBEATS interpretable model. Now the basis vectors are normalized by the horizon length as it is done in the paper.

### Other Information

--

### DIFF
--- a/darts/models/forecasting/nbeats.py
+++ b/darts/models/forecasting/nbeats.py
@@ -50,11 +50,11 @@ class _SeasonalityGenerator(nn.Module):
         super().__init__()
         half_minus_one = int(target_length / 2 - 1)
         cos_vectors = [
-            torch.cos(torch.arange(target_length) * 2 * np.pi * i)
+            torch.cos(torch.arange(target_length) / target_length * 2 * np.pi * i)
             for i in range(1, half_minus_one + 1)
         ]
         sin_vectors = [
-            torch.sin(torch.arange(target_length) * 2 * np.pi * i)
+            torch.sin(torch.arange(target_length) / target_length * 2 * np.pi * i)
             for i in range(1, half_minus_one + 1)
         ]
 


dennisbader (commented):
Hi @vladimir-chernykh and thanks a lot for pointing this out!

I checked and agree with you. That one was easy to overlook!

I initiated the tests

darts/models/forecasting/nbeats.py
```diff
@@ -50,11 +50,11 @@ def __init__(self, target_length):
...
         half_minus_one = int(target_length / 2 - 1)
         cos_vectors = [
-            torch.cos(torch.arange(target_length) * 2 * np.pi * i)
+            torch.cos(torch.arange(target_length) / target_length * 2 * np.pi * i)
             for i in range(1, half_minus_one + 1)
         ]
         sin_vectors = [
-            torch.sin(torch.arange(target_length) * 2 * np.pi * i)
+            torch.sin(torch.arange(target_length) / target_length * 2 * np.pi * i)
             for i in range(1, half_minus_one + 1)
```
hrzn, line 58: Good catch!
Now, looking again at Eq (3) in the paper, I'm wondering whether we shouldn't also change the range of `i`, as it should start from 0 for the cosines and go from ceil(H/2) to ceil(H-1) for the sines.

So something like changing the range to this for both sine and cosine:
```
for i in range(0, half_minus_one)
```
for i in range(half_minus_one + 1, 
```

And adding `target_length / 2` to `i` in the sines loop.

Unless I'm missing something?

vladimir-chernykh:
@hrzn,

I believe the range of `i` should start from 0 for both sines and cosines.

Screenshot 2022-02-17 at 12 32 22

What you're talking about I think is the theta coefficients. For cosines we take first floor(H/2-1) values from theta vector and for sines we take next floor(H/2-1) values starting from floor(H/2). But this is already done correctly in the DARTS. So in my opinion no more changes needed here.

hrzn:
> @hrzn,
> 
> I believe the range of `i` should start from 0 for both sines and cosines.
> 
> Screenshot 2022-02-17 at 12 32 22
> 
> What you're talking about I think is the theta coefficients. For cosines we take first floor(H/2-1) values from theta vector and for sines we take next floor(H/2-1) values starting from floor(H/2). But this is already done correctly in the DARTS. So in my opinion no more changes needed here.

You are right - I convinced myself too between my earlier comment and now. Thanks for checking - and good catch! We will release it (along with a planned patch release) asap.

Issue #802, opened 2022-02-16
[BUG] ValueError: 'cuda' is not a valid DistributedType


gsamaras:
**Describe the bug**
Suddenly, upon relaunching my notebook I wasn't able to train N-Beats on GPU and got `ValueError: 'cuda' is not a valid DistributedType`, without me changing anything in the code.

**To Reproduce**
Install like this in a Jupyter Notebook:

`!pip install 'u8darts[torch]`

and then try to train any model in GPU, e.g. an N-Beats model like this:

```
from darts.models import NBEATSModel

model_nbeats = NBEATSModel(
    input_chunk_length=2,
    output_chunk_length=1,
    generic_architecture=True,
    num_stacks=2,
    num_blocks=1,
    num_layers=1,
    layer_widths=2,
    n_epochs=20,
    nr_epochs_val_period=1,
    batch_size=2,
    random_state=0,
    optimizer_cls = optim.Adam,
    optimizer_kwargs={"lr": 1e-3},
    lr_scheduler_cls = optim.lr_scheduler.ReduceLROnPlateau,
    lr_scheduler_kwargs= {"optimizer": optim.Adam, "threshold": 0.0001, "verbose": True},
    torch_device_str="cuda:0",
)
```

which gives the error:

```
ValueError                                Traceback (most recent call last)
Input In [6], in <module>
    167 train, val = series.split_after(trainset_size)
    169 model_nbeats = define_NBEATS_model(train_set=train, val_set=val, gridsearch=False)
--> 170 model_nbeats.fit(series=train, val_series=val, verbose=True)
    172 pred_series = model_nbeats.historical_forecasts(
    173     series,
    174     start=trainset_size,
   (...)
    178     verbose=True,
    179 )
    180 display_forecast(pred_series, series, "1 horizon", start_date=trainset_size)

File /opt/conda/lib/python3.9/site-packages/darts/utils/torch.py:70, in random_method.<locals>.decorator(self, *args, **kwargs)
     68 with fork_rng():
     69     manual_seed(self._random_instance.randint(0, high=MAX_TORCH_SEED_VALUE))
---> 70     return decorated(self, *args, **kwargs)

File /opt/conda/lib/python3.9/site-packages/darts/models/forecasting/torch_forecasting_model.py:726, in TorchForecastingModel.fit(self, series, past_covariates, future_covariates, val_series, val_past_covariates, val_future_covariates, trainer, verbose, epochs, max_samples_per_ts, num_loader_workers)
[...]

Issue #801, opened 2022-02-16
[BUG] Not using GPU by default in RNNModel


rocket-yg:
**Describe the bug**
I am trying to train a RNN model for a time series problem statement. Using Google Colab GPU backed instance, when I train, it doesn't use the GPU by itself as shown in training logs:

```
WARNING:darts.models.forecasting.torch_forecasting_model:DeprecationWarning: kwarg `verbose` is deprecated and will be removed in a future Darts version. Instead, control verbosity with PyTorch Lightning Trainer parameters `enable_progress_bar`, `progress_bar_refresh_rate` and `enable_model_summary` in the `pl_trainer_kwargs` dict at model creation.
GPU available: True, used: False
TPU available: False, using: 0 TPU cores
```

When i explicitly pass the 'cuda:0'  string to the `torch_device_str` argument, I get a value error message:

`ValueError: 'cuda' is not a valid DistributedType`

**Expected behavior**
It should be able to train on the gpu instance.

gsamaras:
darts version 0.17.0 was released yesterday.
The TorchForecastingModel are now built on top of PyTorch Lightning.

So try changing:

    torch_device_str="cuda:0"

to: 

```
pl_trainer_kwargs={
        "accelerator": "gpu",
        "gpus": [0]
}
```

PR #806, opened 2022-02-16
Fix/ptl fixes

dennisbader:
Addresses #802, #801

### Summary

fixes:
- issue with ReduceLROnPlateau
- issue with torch_device_str

adds:
- check if user passed invalid parameters at model creation
- deprecation info to model docs

### DIFF
--- a/darts/models/forecasting/block_rnn_model.py
+++ b/darts/models/forecasting/block_rnn_model.py
@@ -213,13 +213,21 @@ class BlockRNNModel(PastCovariatesTorchModel):
             Number of epochs to wait before evaluating the validation loss (if a validation
             ``TimeSeries`` is passed to the :func:`fit()` method).
         torch_device_str
-            Optionally, a string indicating the torch device to use. (default: ``None``. Set "cuda:0" if a GPU
-            is available, otherwise "cpu")
+            Optionally, a string indicating the torch device to use. By default, ``torch_device_str`` is ``None``
+            which will run on CPU. Set it to ``"cuda"`` to use all available GPUs or ``"cuda:i"`` to only use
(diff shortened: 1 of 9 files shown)


darts/models/forecasting/block_rnn_model.py
```diff
@@ -213,8 +213,15 @@ def __init__(
...
             ``TimeSeries`` is passed to the :func:`fit()` method).
         torch_device_str
-            Optionally, a string indicating the torch device to use. (default: "cuda:0" if a GPU
+            Optionally, a string indicating the torch device to use. (default: ``None``. Set "cuda:0" if a GPU
             is available, otherwise "cpu")
+
+            .. deprecated:: v0.17.0
+                ``torch_device_str`` has been deprecated in v0.17.0 and will be removed in a future version.
+                Instead, specify this with keys ``"accelerator", "gpus", "auto_select_gpus"`` in your
+                ``pl_trainer_kwargs`` dict. For more info, see here:
```
hrzn, line 222: Could you maybe write down which device will be used by default? 
Also, you could maybe add a small example here. e.g. "use: `gpus=-1` to use all available GPUs" or something along those lines. Or maybe using "auto"?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -389,9 +398,29 @@ def _extract_torch_devices(torch_device_str) -> Tuple[str, Optional[list], bool]
             )
         return accelerator, gpus, auto_select_gpus
 
-    @staticmethod
-    def _extract_torch_model_params(**kwargs):
+    @classmethod
+    def _validate_model_kwargs(cls, **kwargs):
```
hrzn, line 402: 👍

hrzn (approved):
Nice, thanks!

Issue #814, opened 2022-02-21
Model can't find encoding_available variable [BUG]


Joe-TheBro:
**Describe the bug**
After fitting TCN model, when attempting to backtest program spits out error regarding ```self.encoders.encoding_available```

**Error**
```
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_1955/95313221.py in <module>
      7     retrain=False,
      8     num_samples=1,
----> 9     verbose=True,
     10 )

/lib/python3.7/site-packages/darts/utils/utils.py in sanitized_method(self, *args, **kwargs)
    170 
    171                 getattr(self, sanity_check_method)(*only_args.values(), **only_kwargs)
--> 172             return method_to_sanitize(self, *only_args.values(), **only_kwargs)
    173 
    174         return sanitized_method

/lib/python3.7/site-packages/darts/models/forecasting/forecasting_model.py in historical_forecasts(self, series, past_covariates, future_covariates, num_samples, start, forecast_horizon, stride, retrain, overlap_end, last_points_only, verbose)
    421                 past_covariates=past_covariates,
    422                 future_covariates=future_covariates,
--> 423                 num_samples=num_samples,
    424             )
    425 

/lib/python3.7/site-packages/darts/models/forecasting/forecasting_model.py in _predict_wrapper(self, n, series, past_covariates, future_covariates, num_samples)
   1008             past_covariates=past_covariates,
   1009             future_covariates=future_covariates,
-> 1010             num_samples=num_samples,
   1011         )
   1012 

/lib/python3.7/site-packages/darts/utils/torch.py in decorator(self, *args, **kwargs)
     68         with fork_rng():
     69             manual_seed(self._random_instance.randint(0, high=MAX_TORCH_SEED_VALUE))
---> 70             return decorated(self, *args, **kwargs)
     71 
     72     return decorator

/lib/python3.7/site-packages/darts/models/forecasting/torch_forecasting_model.py in predict(self, n, series, past_covariates, future_covariates, trainer, batch_size, verbose, n_jobs, roll_size, num_samples, num_loader_workers)
   1076         )
   1077 
-> 1078         if self.encoders.encoding_available:
   1079             past_covariates, future_covariates = self.encoders.encode_inference(
   1080                 n=n,

AttributeError: 'NoneType' object has no attribute 'encoding_available'
```

**To Reproduce**
```
backtest_series = model.historical_forecasts(
    series_transformed,
[...]

PR #821, opened 2022-02-26
removed checkpoint from prediction

### LINKED ISSUE #820: [BUG] Torch model with checkpoints is not saved independent of checkpoint
**Describe the bug**
When fitting a torch based model with checkpoints and then using save_model and load_model to persist it will result in an error during loading because the model will contain a reference to the checkpoints and tries to load the last checkpoint during prediction.

**To Reproduce**
```python
import pandas as pd

# Read data:

# Create training and validation sets:
train, val = series.split_after(pd.Timestamp("19590101"))

# Normalize the time series (note: we avoid fitting the transformer on the validation set)
transformer = Scaler()
train_transformed = transformer.fit_transform(train)
val_transformed = transformer.transform(val)
series_transformed = transformer.transform(series)

model_name = "Air_RNN"
my_model = RNNModel(
    n_epochs=20,
    model_name=model_name,
    random_state=42,
    input_chunk_length=14,
    training_length=20,
    save_checkpoints=True,
)

my_model.fit(
    train_transformed,
    val_series=val_transformed,
)

best_model = RNNModel.load_from_checkpoint(model_name, best=True)
best_model.save_model("best_model.pth.tar")

# delete logs
!rm -rf darts_logs

# reload model
best_model = RNNModel.load_model("best_model.pth.tar")

# this will fail because best_model.load_ckpt_path != None but the checkpoints are deleted
best_model.predict(1, val_transformed)
# Raises FileNotFoundError: Checkpoint at ....ckpt not found. Aborting training.
```

**Expected behavior**
loaded model predicts without needing the checkpoint.

 - Python version: 3.7
 - darts version 0.17.1

**Additional context**
For us this is somewhat critical that the model can not be loaded outside its training context. However the easy workaround seems to be to just set `best_model.load_ckpt_path = None`

From my limited perspective there is also no reason why the checkpoint should be needed for prediction in the first place. I can prepare a PR that just skips loading it if that helps.

dennisbader:
Fixes #820.

### Summary

- Removed checkpoint from prediction. We already load the model state when calling `load_from_checkpoint()`, which is why we can remove it from `predict()`
- Updated docs for loading methods as suggested in #814.

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -1223,13 +1223,8 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         trainer = trainer if trainer is not None else self.trainer
         self._setup_trainer(trainer=trainer, verbose=verbose, epochs=self.n_epochs)
 
-        # if model checkpoint was loaded without calling fit afterwards (when `load_ckpt_path is not None`),
-        # trainer needs to be instantiated here
-        ckpt_path = self.load_ckpt_path
-        self.load_ckpt_path = None
-
         # prediction output comes as nested list: list of predicted `TimeSeries` for each batch.
-        predictions = self.trainer.predict(self.model, pred_loader, ckpt_path=ckpt_path)
+        predictions = self.trainer.predict(self.model, pred_loader)
         # flatten and return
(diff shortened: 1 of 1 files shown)


hrzn (approved):
Thanks!

PR #822, opened 2022-02-26
return metric score along with untrained best models & params

### LINKED ISSUE #776: Gridsearch should return metric score along with the best untrained model and parameters.
When performing gridsearch, we also want to know how good the best parameters can perform. I can have this insight if I can access the score as the result of gridsearch. It is redundant to have to run backtest again to get the score.

In the `gridsearch` method, return the metric score in addition to the model and parameters.

Create another method `gridsearch_with_score`, so that it won't affect existing code.

**Additional context**
The score is needed to evaluate how much I should trust the best parameter. If is it not good, then perhaps it can hint that we should use a different forecasting approach.

nlhkh:
Fixes #776.

### Summary

The gridsearch method also returned the best metric score, aka the minimum error, along with the best model and parameters.


hrzn (commented):
This is a breaking change, but I think we can do it. The other option would be to introduce a parameter in the function to optionally return the best metric, but that would inflate the signature. So I think we could merge this change. Could you maybe just add a short unit test that verifies the actual value being returned in `test_backtesting.py` ?

nlhkh:
> This is a breaking change, but I think we can do it. The other option would be to introduce a parameter in the function to optionally return the best metric, but that would inflate the signature. So I think we could merge this change. Could you maybe just add a short unit test that verifies the actual value being returned in `test_backtesting.py` ?

There is a function `compare_best_against_random` in file "test_backtesting.py", which seems to perform some score comparison between best params and random params. I was thinking that instead of recalculating `best_score_1` and `best_score_2`, I can use the returned score from the new `gridsearch` function. Would that be a good approach in this case? One concern that I have is that `best_model_1` gridsearch uses `stride`, but the `best_score_1` backtesting only uses a default stride, so the scores might be different. Do you think we should add the usage of stride in the subsequent backtesting?

Another approach is to leave this as-is, and create another test where I also manually backtest the best model, and compare the score with the one returned by gridsearch function.

nlhkh:
Out of curiosity, there are some tests that you mark as *skipUnless TORCH_AVAILABLE*. I don't recall models such as RandomForest from sklearn requires PyTorch.

hrzn:
> > This is a breaking change, but I think we can do it. The other option would be to introduce a parameter in the function to optionally return the best metric, but that would inflate the signature. So I think we could merge this change. Could you maybe just add a short unit test that verifies the actual value being returned in `test_backtesting.py` ?
> 
> There is a function `compare_best_against_random` in file "test_backtesting.py", which seems to perform some score comparison between best params and random params. I was thinking that instead of recalculating `best_score_1` and `best_score_2`, I can use the returned score from the new `gridsearch` function. Would that be a good approach in this case? One concern that I have is that `best_model_1` gridsearch uses `stride`, but the `best_score_1` backtesting only uses a default stride, so the scores might be different. Do you think we should add the usage of stride in the subsequent backtesting?
> 
> Another approach is to leave this as-is, and create another test where I also manually backtest the best model, and compare the score with the one returned by gridsearch function.

I think in case of doubt it is better to create a new unit test dedicated to test this new functionality. My only concern is that ideally it shouldn't be too long-running; so maybe take a small series (e.g. air passengers), a model very quick to fit (e.g. Theta), and not too large a parameter space.

hrzn:
> Out of curiosity, there are some tests that you mark as _skipUnless TORCH_AVAILABLE_. I don't recall models such as RandomForest from sklearn requires PyTorch.

RandomForest does not depend on PyTorch, but it used to (we used to rely on PyTorch `Datasets` to train them, but that is no longer the case). It might be that we forgot to remove a couple of `skipUnless` clauses from the unit tests. If you find some and you are confident there's no PyTorch-based models involved, you can remove them in this PR, and I'll try testing (it will be tested by our CI pipelines upon merging anyway).

nlhkh:
Hello @hrzn . Anything else I should add? 😄

PR #825, opened 2022-02-28
added val loss to progess bar and updated docs

dennisbader:
### Summary

- added train and val loss to progress bar
- changed default `nr_epochs_val_period` from 10 to 1
- updated TorchForecastingModel docs with default values. 

### Other Information
- concerning automatically updating the TFM docs: it is possible with a wrapper around `__init__`. This would then only be visible on the documentation website and not in the source code. I personally like having the docs also in the source code so I left it as is for now.


hrzn (approved):
👍

PR #829, opened 2022-03-01
Fix/encoders with fit from ds

hrzn:
`predict()` is not working if called after `fit_from_dataset()`. This fixes it.


dennisbader (approved):
Looks good, thanks!

PR #833, opened 2022-03-03
Feat/stochastic inputs

hrzn:
Handle stochastic series for `fit()` and `predict()` in Torch-based models.

When fitting: a sample is drawn from the series uniformly at random (instead of taking always the 1st sample, as before). This way the training procedure can naturally be carried over the stochastic samples.

When predicting: *one* sample is drawn from the series uniformly at random. Unfortunately at the moment it would be quite complicated to batch the forward passes over all of a series' samples, so each call to `predict()` will only return the forecast related to this one sample. However, calling `sample()` many times and concatenating the results allows to get stochastic forecasts even with deterministic models (not using a likelihood) when predicting stochastic series (or using stochastic covariates). Here's an example:

image

### DIFF
--- a/darts/tests/models/forecasting/test_probabilistic_models.py
+++ b/darts/tests/models/forecasting/test_probabilistic_models.py
@@ -242,3 +242,24 @@ class ProbabilisticTorchModelsTestCase(DartsBaseTestClass):
                     "The difference between the mean forecast and the mean series is larger "
                     "than expected on component 1 for distribution {}".format(lkl),
                 )
+
+        def test_stochastic_inputs(self):
+            model = RNNModel(input_chunk_length=5)
+            model.fit(self.constant_ts, epochs=2)
+
+            # build a stochastic series
+            target_vals = self.constant_ts.values()
+            stochastic_vals = np.random.normal(
+                loc=target_vals, scale=1.0, size=(len(self.constant_ts), 100)
+            )
+            stochastic_vals = np.expand_dims(stochastic_vals, axis=1)
+            stochastic_series = TimeSeries.from_times_and_values(
+                self.constant_ts.time_index, stochastic_vals
+            )
+
+            # A deterministic model forecasting a stochastic series
+            # should return stochastic samples
+            preds = [model.predict(series=stochastic_series, n=10) for _ in range(2)]
+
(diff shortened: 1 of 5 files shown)


darts/tests/models/forecasting/test_probabilistic_models.py
```diff
@@ -242,3 +242,24 @@ def _get_avgs(series):
...
+            stochastic_series = TimeSeries.from_times_and_values(
+                self.constant_ts.time_index, stochastic_vals
+            )
+
+            # A deterministic model forecasting a stochastic series
+            # should return stochastic samples
+            preds = [model.predict(series=stochastic_series, n=10) for _ in range(2)]
+
+            # random samples should differ
+            self.assertFalse(np.alltrue(preds[0].values() == preds[1].values()))
```
dennisbader, line 265: This is nitpicking but isn't there a slim chance that the predictions will be identical? :D we could add a seed to make sure they are not.

darts/timeseries.py
```diff
@@ -1176,6 +1176,28 @@ def values(self, copy=True, sample=0) -> np.ndarray:
         else:
             return self._xa.values[:, :, sample]
 
+    def random_component_values(self, copy=True) -> np.array:
```
dennisbader, line 1179: What do you think about an additional optional to randomly sample a quantile per sample? 

I would assume this is similar to quantile regression based on stochastic input (without needing QuantileRegression as a likelihood)

dennisbader (approved):
Looks good, thanks :) I had a question about a potential additional option for randomly sampling the quantiles

darts/tests/models/forecasting/test_probabilistic_models.py
hrzn: Yes, normally there is a seed earlier in the file which I believe should apply here :)

darts/timeseries.py
hrzn: Oh, you mean returning the exact quantile instead of sampling a sample at random? That's an interesting idea :) But would require a bit of adaptation elsewhere as we would still need to ensure this is applied on the output targets only, and not e.g. on the inputs or covariates. Let's keep the idea though.

PR #845, opened 2022-03-14
Feat/more losses

hrzn:
Add two new PyTorch loss functions (`SmapeLoss` and `MapeLoss`), which can provide different criteria and could for instance be used to replicate some of the M3/M4 competition results.

### DIFF
--- a/darts/tests/utils/test_losses.py
+++ b/darts/tests/utils/test_losses.py
@@ -1,49 +1,41 @@
-import numpy as np
+import torch
 
-from darts.dataprocessing.transformers import Scaler
-from darts.datasets import AirPassengersDataset
-from darts.metrics import smape
-from darts.models import NBEATSModel
 from darts.tests.base_test_class import DartsBaseTestClass
-from darts.utils.losses import MapeLoss, SmapeLoss
+from darts.utils.losses import MAELoss, MapeLoss, SmapeLoss
 
 
 class LossesTestCase(DartsBaseTestClass):
-    air = AirPassengersDataset().load().astype(np.float32)
-    scaler = Scaler()
-    air_s = scaler.fit_transform(air)
-    air_train, air_val = air_s[:-36], air_s[-36:]
+    x = torch.tensor([1.1, 2.2, 0.6345, -1.436])
+    y = torch.tensor([1.5, 0.5])
+
+    def helper_test_loss(self, exp_loss_val, exp_w_grad, loss_fn):
+        W = torch.tensor([[0.1, -0.2, 0.3, -0.4], [-0.8, 0.7, -0.6, 0.5]])
+        W.requires_grad = True
+        y_hat = W @ self.x
+        lval = loss_fn(y_hat, self.y)
+        lval.backward()
+
+        print(lval)
+        self.assertTrue(torch.allclose(lval, exp_loss_val, atol=1e-3))
+        self.assertTrue(torch.allclose(W.grad, exp_w_grad, atol=1e-3))
 
     def test_smape_loss(self):
-        model = NBEATSModel(
-            input_chunk_length=24,
-            output_chunk_length=12,
-            num_stacks=4,
-            num_blocks=1,
-            layer_widths=64,
-            loss_fn=SmapeLoss(),
-            random_state=42,
+        exp_val = torch.tensor(0.7753)
+        exp_grad = torch.tensor(
+            [[-0.2843, -0.5685, -0.1640, 0.3711], [-0.5859, -1.1718, -0.3380, 0.7649]]
         )
-
-        model.fit(self.air_train, epochs=150)
-
-        pred = model.predict(n=36)
-
-        self.assertLess(smape(pred, self.air_s), 16.5)
+        self.helper_test_loss(exp_val, exp_grad, SmapeLoss())
 
     def test_mape_loss(self):
-        model = NBEATSModel(
-            input_chunk_length=24,
-            output_chunk_length=12,
-            num_stacks=4,
-            num_blocks=1,
-            layer_widths=64,
-            loss_fn=MapeLoss(),
-            random_state=42,
+        exp_val = torch.tensor(1.2937)
+        exp_grad = torch.tensor(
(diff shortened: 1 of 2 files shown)


darts/utils/losses.py
```diff
@@ -0,0 +1,64 @@
...
+import torch.nn as nn
+
+
+def _divide_no_nan(a, b):
+    """
+    a/b where the resulted NaN or Inf are replaced by 0.
+    """
+    result = a / b
+    result[result != result] = 0.0
+    result[result == np.inf] = 0.0
```
amadejkocbek, line 18: Is the case -Inf covered or needed?

darts/tests/utils/test_losses.py
```diff
@@ -0,0 +1,49 @@
...
+from darts.utils.losses import MapeLoss, SmapeLoss
+
+
+class LossesTestCase(DartsBaseTestClass):
+    air = AirPassengersDataset().load().astype(np.float32)
+    scaler = Scaler()
+    air_s = scaler.fit_transform(air)
+    air_train, air_val = air_s[:-36], air_s[-36:]
+
+    def test_smape_loss(self):
```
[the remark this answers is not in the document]
brunnedu: Yes, I thought of that as well. Although actually using the loss functions for fitting might reveal some problems that we wouldn't notice otherwise.

darts/utils/losses.py
```diff
@@ -0,0 +1,65 @@
...
+        self.block_denom_grad = block_denom_grad
+
+    def forward(self, inpt, tgt):
+        num = torch.abs(tgt - inpt)
+        if self.block_denom_grad:
+            with torch.no_grad():
+                denom = torch.abs(tgt) + torch.abs(inpt)
+        else:
+            denom = torch.abs(tgt) + torch.abs(inpt)
+        return torch.mean(_divide_no_nan(num, denom))
```
brunnedu, line 53: Am I correct that we drop the factor of 2 because it doesn't matter for the minimization?

darts/utils/losses.py
```diff
@@ -0,0 +1,65 @@
...
+class MapeLoss(nn.Module):
+    """
+    MAPE loss as defined in: https://en.wikipedia.org/wiki/Mean_absolute_percentage_error
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, inpt, tgt):
+        return torch.mean(torch.abs(inpt - tgt))
```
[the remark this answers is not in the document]
brunnedu: +1

brunnedu (commented):
Nice addition! 👍

darts/tests/utils/test_losses.py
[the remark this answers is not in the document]
hrzn: Yes it's a tiny bit better to test the fitting to make sure the gradients are kept where they should, so we can leave it like that for now.

darts/utils/losses.py
hrzn: Good catch! I've added it.

darts/utils/losses.py
hrzn: Yes, it's even a factor of 200, but it impacts only the magnitude of the gradients. I've added a note in the docstring.

darts/utils/losses.py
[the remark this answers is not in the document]
hrzn: I have changed it to make it optional to ignore the denominator (but it's taken into account by default). This is because the denominator is a constant (wrt. the weights) that changes only the magnitude of the gradients, so I think in some cases it may make sense to keep the numerator only (and I was having slightly better results when doing so).
When denominator is ignored this is indeed equivalent to MAE. I've also added MAE to the list now (unit test on its way) :)

darts/tests/utils/test_losses.py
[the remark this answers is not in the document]
hrzn: On second thoughts, I think your idea is better, as long as we're also checking the loss gradients. I've changed the tests to do that now, thanks for the suggestion 👍

darts/tests/utils/test_losses.py
```diff
@@ -0,0 +1,41 @@
...
+from darts.utils.losses import MAELoss, MapeLoss, SmapeLoss
+
+
+class LossesTestCase(DartsBaseTestClass):
+    x = torch.tensor([1.1, 2.2, 0.6345, -1.436])
+    y = torch.tensor([1.5, 0.5])
+
+    def helper_test_loss(self, exp_loss_val, exp_w_grad, loss_fn):
+        W = torch.tensor([[0.1, -0.2, 0.3, -0.4], [-0.8, 0.7, -0.6, 0.5]])
+        W.requires_grad = True
```
dennisbader, line 13: Nice tests +1

darts/tests/utils/test_losses.py
```diff
@@ -0,0 +1,41 @@
...
+        y_hat = W @ self.x
+        lval = loss_fn(y_hat, self.y)
+        lval.backward()
+
+        print(lval)
```
dennisbader, line 18: should be removed

darts/utils/losses.py
```diff
@@ -0,0 +1,95 @@
...
+            \\sum_{t=1}^{T}{\\frac{\\left| y_t - \\hat{y}_t \\right|}{y_t}}.
+
+        The results of divisions yielding NaN or Inf are replaced by 0. Note that we drop the coefficient of
+        100 usually used for computing MAPE values, as it impacts only the magnitude of the gradients
+        and not their direction.
+        """
+        super().__init__()
+
+    def forward(self, inpt, tgt):
+        return torch.mean(_divide_no_nan(torch.abs(tgt - inpt), tgt))
```
dennisbader, line 75: according to the wikipedia link shouldn't it be like this?

```suggestion
        return torch.mean(torch.abs(_divide_no_nan(tgt - inpt, tgt)))
```

dennisbader (approved):
Very nice, thanks a lot!
After addressing the last suggestions, it can be merged

darts/utils/losses.py
hrzn: I think it is correct - see first definition on Wikipedia (also matching Hyndman's recommendations):
image

PR #860, opened 2022-03-19
Feat/add fp16 support

hrzn:
This is a tentative PR: to be discussed whether it is a good idea to merge.

It adds support for np.float16 format for TimeSeries

Advantages:
* This is supported with AMP on PyTorch (on GPUs only at the moment AFAIK), which means that some deep learning models are susceptible to become faster

Disadvantages:
* It breaks some functionalities of Darts. For instance Scalers won't work (out of the box) on TimeSeries that are in float16
* I haven't yet seen speedups on training deep learning models, but I haven't experimented much.

To be discussed if we want it. On the one hand it breaks some functionalities, on the other hand, we can probably assume that users who call `astype(np.float16)` know sufficiently what they are doing; and in some cases it could speed up deep learning models.


hrzn:
Closing for now, we may revisit later

PR #864, opened 2022-03-21
fix/check torch import

gnwhr:
### Summary
Now checks if `torch` is available before running `utils/test_losses.py`.


hrzn (approved):
Thank you!

PR #874, opened 2022-03-25
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
### Summary

Import ABC from `collections.abc` for Python 3.10 compatibility. Using `collections` was deprecated and removed in Python 3.10


hrzn (approved):
Thank you 👍

tirkarthi:
Thanks @hrzn

PR #888, opened 2022-04-05
Fix/ptl1.6.0

hrzn:
Fixes two issues that arised with Pytorch Lightning >= 1.6.0

* The `current_epoch` is now "fixed" in PTL, so we have to make our own fix dynamic and check at runtime the version we are running
* For some reason saving models using `torch.save()` only did not save the state of the PTL trainer correctly. So we revert to saving the PTL module directly using PTL checkpointing mechanism. Pros: it fixes the issue, and also will be more robust because not dependent e.g. on the current path. Con: It saves 2 files instead of one, and the parameters are saved twice. (I'll open a separate issue to address this if we can).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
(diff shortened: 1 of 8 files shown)


dennisbader (approved):
Looks good, thanks a lot!

PR #893, opened 2022-04-05
Feat/statsforecasts

hrzn:
I've implemented a couple of models coming from [Statsforecasts](https://github.com/Nixtla/statsforecast). The most interesting is probably their implementation of AutoARIMA, which uses Numba (and jit compilation) and can in some cases be much faster than pmdarima (after the first call to `fit()`). I would propose that we add it first as a separate model, and if everything goes well and it proves superior to the current AutoARIMA version, we could even replace it entirely at some point (it is not quite the same model, though).


darts/models/forecasting/sf_auto_arima.py
```diff
@@ -0,0 +1,80 @@
...
+        super()._predict(n, future_covariates, num_samples)
+        forecast_df = self.model.predict(
+            h=n,
+            X=future_covariates.values(copy=False) if future_covariates else None,
+            level=68,  # ask one std for the confidence interval. Note, we're limited to int...
+        )
+
+        mu = forecast_df["mean"].values
+        if num_samples > 1:
+            std = 2 * (forecast_df["hi_68%"].values - mu)
```
gnwhr, line 64: Why is there a factor 2? Isn't hi_68 - mu equal to 1 std?

darts/models/forecasting/croston.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+import numpy as np
+from statsforecast.models import croston_classic, croston_optimized
+
+from darts.models.forecasting.forecasting_model import ForecastingModel
+from darts.timeseries import TimeSeries
+
+
+class Croston(ForecastingModel):
```
dumjax, line 13: Apparently TSB and croston_sba are improved version of Croston (https://www.lancaster.ac.uk/pg/waller/pdfs/Intermittent_Demand_Forecasting.pdf), might be worth proposing each of the 3 options in the same function as they already exist in statsforecast?

darts/models/forecasting/sf_auto_arima.py
dumjax: Yes, same question

hrzn: Unless I missed something, With a 68% forecast interval, it means that "hi - mu" represents 34% of the mass. (and "mu - low" represents the other 34%). So I multiply by 2 to get back one std (2x 34%). Makes sense?

darts/models/forecasting/croston.py
hrzn: Good idea, done.

darts/models/forecasting/sf_auto_arima.py
hrzn: OK my bad that was right ;) Now fixed.

PR #909, opened 2022-04-13
xai init

dumjax:
Explanability: Shap with Regression models #609

### Summary

First draw of what should be explainability in darts. We use shap library to explain time series, in the case of RegressionModel.


brunnedu (commented):
Really cool addition to darts! 🚀  Added a few comments, may look at it in more detail later.

darts/explainability/explainability.py
```diff
@@ -0,0 +1,171 @@
...
+                "A background future covariates is not provided, but the model needs future covariates.")
+        
+        self.model = model
+        self.past_steps_explained = past_steps_explained
+        self.n = n
+
+        # For now we won't consider further time step that output_chunk_length, even though we could in 
+        # theory explain anything in the future by auto regressive process.
+        if hasattr(self.model, 'output_chunk_length'):
+            self.n = self.model.output_chunk_length
```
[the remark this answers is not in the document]
dumjax: Good point

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        self, 
+        model: GlobalForecastingModel,
+        background_series: Union[TimeSeries, Sequence[TimeSeries]],
+        background_past_covariates: Union[TimeSeries, Sequence[TimeSeries]],
+        background_future_covariates: Union[TimeSeries, Sequence[TimeSeries]],
+        background_nb_samples: Optional[int] = None
+    ):
+
+        self.model = model
+        self.multioutput = self.model.model._get_tags()['multioutput']
```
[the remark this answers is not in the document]
dumjax: I tested on lgbm and all kind of scikit-learn, seems to exist to all scikit-learn regression type model even with lgbm, but I am going to double check.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+            (len(foreground_future_covariates) == len(foreground_series)),
+            'target, past and future covariates must have the same length.'
+        )
+
+        raise_if_not(
+            all(foreground_series.time_index == foreground_past_covariates.time_index) and \
+            all(foreground_past_covariates.time_index == foreground_future_covariates.time_index) and \
+            all(foreground_future_covariates.time_index == foreground_series.time_index),
+            'The target, past and future covariates must have the same time index.'
+        )
```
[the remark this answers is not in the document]
dumjax: Indeed it should be enough!

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+                time.
+        background_past_covariates
+            A past covariates TimeSeries or list of TimeSeries that the model needs once fitted.
+        background_future_covariates
+            A future covariates TimeSeries or list of TimeSeries that the model needs once fitted.
+        n
+            Optionally, Number of predictions ahead we want to explain. It will provide complete explanation for each 
+            future n steps.
+            If we have existence of output_chunk_length, then n := output_chunk_length automatically.
+            (TODO Not sure it is really needed. Indeed, if there is no output_chunk_length, would be just n=1
```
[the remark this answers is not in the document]
dumjax: Yes, I just limited the scope for now, as probably users will want to explain the direct output first. But we might want to extend it to any n steps in the future.

darts/explainability/explainability.py
```diff
@@ -0,0 +1,171 @@
...
+            self.past_steps_explained = self.model.input_chunk_length
+
+        if not self.test_stationarity():
+            logger.warning("One time series component of the background time series is not stationary."
+            "Beware of wrong interpretation with chosen explainability")
+
+    @abstractmethod
+    def explain_from_input(        
+        foreground_series: TimeSeries,
+        foreground_past_covariates: Optional[TimeSeries],
```
hrzn, line 108: Those should have default values set to `None` I think.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        self.cache_foreground_series = None
+        self.cache_foreground_past_covariates = None
+        self.cache_foreground_future_covariates = None
+        self.cache_foreground_X = None
+        self.foreground_changed = True
+
+    def shap_values(
+        self,
+        foreground_series: Union[TimeSeries, Sequence[TimeSeries]],
+        foreground_past_covariates: Union[TimeSeries, Sequence[TimeSeries]],
```
hrzn, line 318: shouldn't the covariates be optional?

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+                    shap_values_dict[h][t].time_index,
+                    shap_values_dict[h][t].values,
+                    columns= shap_values_dict[h][t].feature_names
+                )
+
+        return shap_values_dict
+    
+    def shap_values(
+        self,
+        foreground_series: TimeSeries,
```
hrzn, line 146: maybe we could drop the `foreground_` here?

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        if horizons is None:
+            horizons = range(self.model.output_chunk_length)
+
+        for t in target_names:
+            for h in horizons:
+                shap_values.append(self.explainers.shap_values_from_X(foreground_X_sampled ,h ,t))
+                plt.title('Target: `{}` - Horizon: {}'.format(t, 't+' + str(h)))
+                shap.summary_plot(shap_values[-1], foreground_X_sampled,  plot_type=plot_type)
+
+class RegressionShapExplainers:
```
hrzn, line 258: I would make this class private

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+            background_past_covariates,
+            background_future_covariates,
+            background_nb_samples
+        )
+
+        self.target_dict = {c:i for i, c in enumerate(background_series.columns)}
+
+        if (self.target_dim > 1 or self.model.output_chunk_length > 1) and (not self.multioutput):
+            self.explainers = {}
+            for i in range(self.model.output_chunk_length):
```
hrzn, line 291: How does this related to `n` defined in `ShapeExplainer`?

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+                self.explainers[i] = {}
+                for j in range(self.target_dim):
+                    self.explainers[i][j] = shap.Explainer(
+                        self.model.model.estimators_[i+j], 
+                        self.background_X
+                    )
+                    # Special case of trees, where we don't want to depend on a background (faster in general)
+                    # Warning: Randomforest is quite slow.
+                    if isinstance(self.explainers[i][j], shap.explainers._tree.Tree):
```
hrzn, line 300: Isn't there another way to check this than accessing a private member?

hrzn: It seems to me that here we are first building an explainer, then checking its type, and then changing it based on the result. Shouldn't we build the right type of explainer right away? E.g. by checking whether the initial estimator object is a tree?

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        self.foreground_changed = True
+
+    def shap_values(
+        self,
+        foreground_series: Union[TimeSeries, Sequence[TimeSeries]],
+        foreground_past_covariates: Union[TimeSeries, Sequence[TimeSeries]],
+        foreground_future_covariates: Union[TimeSeries, Sequence[TimeSeries]],
+        horizon: Optional[int]=None, 
+        target_name: Optional[str]=None
+    ) -> shap._explanation.Explanation:
```
hrzn, line 322: Here too, better to remove the annotation than relying on a private member.

darts/explainability/shap_explainer.py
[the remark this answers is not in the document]
hrzn: You can maybe re-use `darts.utils.utils.retain_period_common_to_all()`

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+
+        raise_if_not(
+            all(foreground_series.time_index == foreground_past_covariates.time_index) and \
+            all(foreground_past_covariates.time_index == foreground_future_covariates.time_index) and \
+            all(foreground_future_covariates.time_index == foreground_series.time_index),
+            'The target, past and future covariates must have the same time index.'
+        )
+
+        # We don't recompute if the foreground is the same as the last computation
+        if (self.cache_foreground_series != foreground_series) or \
```
hrzn, line 339: What does this cache buy us? It introduces some extra complexity and state in the class, and I'm not sure it's worth the price. I would put this burden on the user (not calling twice the same method with the same inputs is user's responsibility), but maybe I'm missing something.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        X,
+        horizon: Optional[int]=None, 
+        target_name: Optional[str]=None
+    ) -> shap._explanation.Explanation:
+        
+        # Multioutput case
+        if (self.target_dim > 1 or self.model.output_chunk_length > 1):
+            
+            # MultioutputRegressor case
+            if isinstance(self.explainers, dict):
```
hrzn, line 376: There are many such `isinstance` checks. In general we should try to really keep those at a minimum (they are considered bad practice whenever they can be avoided). Here, I have the feeling that it would be cleaner to make a couple of such checks in the contructor method of the class, once and for all, and then store everything in a unified way. For instance even if the model is made of a single estimator, you can store it as `self.models = {0: estimator}` or something. This way you always have the same format everywhere after that (for instance dict-based), and you don't rely on these checks in other parts of the code.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+            
+            # X: covariate lags
+            for idx, (df_cov, lags) in enumerate(covariates):
+                if lags:
+                    for lag in lags:
+                        df_tmp = df_cov.shift(-lag)
+                        if idx == 0:
+                            cov_type = 'past'
+                        else:
+                            cov_type = 'fut'
```
hrzn, line 462: Maybe keep all these strings as constants at the top of the file.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        )
+
+    def explain_from_input(
+        self,
+        foreground_series: TimeSeries, 
+        foreground_past_covariates: Optional[TimeSeries], 
+        foreground_future_covariates: Optional[TimeSeries], 
+        horizons: Optional[Sequence[int]] = None, 
+        target_names: Optional[Sequence[str]] = None
+        ) -> Sequence[Sequence[TimeSeries]]:
```
hrzn, line 119: I think here it would be nice to explain how the model relates to this foreground series - i.e., that some "features" (i.e., "input sub-slices") are extracted from this series in a way that is similar to how the training samples are obtained, and the shap values are computed on those samples as they are shifted over time. And explain the result type also (btw it doesn't seem to be a `Sequence[Sequence[TimeSeries]]`).

hrzn:
It looks like a great start and I really like the result :) I also like the structure with separate explainer objects of different kinds living in the darts.explainability submodule.

A couple of points:
* I know it's because it's still a draft, you will have to pay attention to the arguments for each methods. I saw many cases that fail currently (when some argument is None, or not provided for instance).
* You should rethink a bit what goes in the super-class vs the implementation (I think the super-class should not contain much logic except what is very general and common to all present and future explainers).
* Writing unit tests will help you catch some of the issues.
* Also beware of the listing, you should follow the setup [here](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md#code-formatting-and-linting)
* Are you sure you need the class `RegressionShapExplainers`? I'm wondering if it brings something, as to me it looks like an extra layer and some extra code in a place where we don't really need the encapsulation, as its state is more or less the same as `ShapExplainer`. But I might be missing something.

darts/explainability/explainability.py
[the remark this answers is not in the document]
dumjax: n and past_steps_explained were here to be able to also make a choice for autoregressive models. 
past_steps_explained can probably be inferred with any kind of models (arima for example will be the max order etc), but n can be arbitrarily long for a given model. We might then just decide n = 1 for autoregressive ones and n = output_chunk_length for others, n = depending on the model ?

darts/explainability/explainability.py
```diff
@@ -0,0 +1,171 @@
...
+        foreground_future_covariates: Optional[TimeSeries],
+        horizons: Optional[Sequence[int]]=None, 
+        target_names: Optional[Sequence[str]]=None
+    ) -> Union[TimeSeries, Sequence[TimeSeries]]:
```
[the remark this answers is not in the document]
dumjax: Yes, I wrongly paste some of docstrings explaining it is shap Explanation dictionary but here this is the generic method of ForecastingModelExplainer, and should return time series of importance or explainability values from any explainer model.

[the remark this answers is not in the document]
dumjax: Actually a dictionary of dictionary of TimeSeries, not the present one

darts/explainability/shap_explainer.py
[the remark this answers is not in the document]
dumjax: Actually `all(foreground_series.time_index == foreground_past_covariates.time_index == foreground_future_covariates.time_index)` doesn't work if the length are not the same. I change slightly this test by using `                `len(background_series.time_index.difference(background_past_covariates.time_index)`

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,474 @@
...
+        target_series, 
+        past_covariates, 
+        future_covariates, 
+        n_samples = None
+    ):
+        """
+        Helper function that creates training/validation matrices (X and y as required in sklearn), given series and
+        max_samples_per_ts.
+
+        Partially adapted from _create_lagged_data funtion in regression_model
```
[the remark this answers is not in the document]
hrzn: Either this, or maybe (cleaner) factor it out somewhere else; e.g. in `darts.utils.data`, where similar functionalities for neural nets already live.

darts/explainability/explainability.py
```diff
@@ -0,0 +1,222 @@
...
+                        background_future_covariates[idx],
+                    ]
+                )
+
+            self.background_series = background_series
+            self.background_past_covariates = background_past_covariates
+            self.background_future_covariates = background_past_covariates
+
+        self.target_names = self.background_series.columns
+        self.past_covariates_names = self.background_past_covariates.columns
```
hrzn, line 126: This crashes when no covariates are specified.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,511 @@
...
+        # We add one property to the shap._explanation.Explanation which is the index of time steps we explain
+        shap_values.time_index = X.index
+
+        # When MultiOutputRegressor, or when pure univariate and output_chun_legth = 1, we need to ravel base_values
+        # to make work force plot.
+        shap_values.base_values = shap_values.base_values.ravel()
+
+        return shap_values
+
+    def _create_RegressionModel_shap_X(
```
hrzn, line 425: case to correct

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,511 @@
...
+            foreground_future_covariates,
+        ) = retain_period_common_to_all(
+            [
+                foreground_series,
+                foreground_past_covariates,
+                foreground_future_covariates,
+            ]
+        )
+
+        # We don't recompute if the foreground is the same as the last computation
```
hrzn, line 365: Is it necessary to handle caching here? I would assume that it is users responsibility to avoid calling twice the function with the same inputs (but might be missing something)

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,635 @@
...
+    def force_plot_from_ts(
+        self,
+        foreground_series: TimeSeries,
+        foreground_past_covariates: Optional[TimeSeries] = None,
+        foreground_future_covariates: Optional[TimeSeries] = None,
+        horizons: Optional[Sequence[int]] = None,
+        target_names: Optional[Sequence[str]] = None,
+    ):
+        """
+        Display a shap force_plot per target and per horizon.
```
hrzn, line 271: It would be nice to put a couple of sentences explaining what the force plot shows (in this context) for people not familiar with Shap.

hrzn (commented):
It looks great overall @dumjax , nice one!

I have some comments, but mostly unimportant things. The only somewhat important things IMO are
* Adding unit tests
* Fixing the case where no past or future covariates are given
* Improving the docstrings in a few places (I've made small suggestions, but I think we can improve the explanations and formatting in a few places)
* Factoring out the X/y computation as discussed

I'm really looking forward for us to release it 🚀

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,635 @@
...
+    ADDITIVE = 8
+
+
+ShapMethod = NewType("ShapMethod", _ShapMethod)
+
+
+class ShapExplainer(ForecastingModelExplainer):
+    def __init__(
+        self,
+        model: ForecastingModel,
```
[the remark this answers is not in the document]
dumjax: I want to keep it as general as possible for now, as in the long run we want to explain every kind of model. You are talking about RegressionModel right?

darts/explainability/explainability.py
```diff
@@ -0,0 +1,305 @@
...
+        ] = None,
+        horizons: Optional[Sequence[int]] = None,
+        target_names: Optional[Sequence[str]] = None,
+    ) -> Union[
+        Dict[integer, Dict[str, TimeSeries]],
+        Sequence[Dict[integer, Dict[str, TimeSeries]]],
+    ]:
+        """
+        Main method of the ForecastingExplainer class.
+        Return a dictionary of dictionaries of (mutivariates) TimeSeries instances
```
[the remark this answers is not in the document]
dumjax: Yes, very good idea, I'm going to implement that.

darts/explainability/shap_explainer.py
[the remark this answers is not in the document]
hrzn: I think we could leave the type hints as precise as possible for now, and change them later to `ForecastingModel` when the method becomes more general.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,635 @@
...
+    PARTITION = 5
+    LINEAR = 6
+    PERMUTATION = 7
+    ADDITIVE = 8
+
+
+ShapMethod = NewType("ShapMethod", _ShapMethod)
+
+
+class ShapExplainer(ForecastingModelExplainer):
```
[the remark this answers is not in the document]
hrzn: OK, let's keep this name then.

darts/explainability/explainability.py
```diff
@@ -0,0 +1,305 @@
...
+                        == self.future_covariates_names
+                        if self.background_future_covariates is not None
+                        else True,
+                    ]
+                ),
+                "Columns names must be identical between TimeSeries list components (multi-TimeSeries).",
+            )
+
+        # the number of samples we will build for explanation is:
+        # sum(len(intersection(target, fut_cov, past_cov))- min_length+1). We compare this to a fixed constant min.
```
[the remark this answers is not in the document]
dumjax: the first goal was to check that the time index of each of tuple (series , pas , future covariates) were corresponding more or less. Indeed the tabularization function doesn't provide such checks on time index. On top of that I introduced at this level a MIN_BACKGROUND_SAMPLE, independently of which kind of models we use (regression model, torch, type of explainability). If not I would have to check after each different required transformation depending on the model.

darts/explainability/shap_explainer.py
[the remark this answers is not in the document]
dumjax: 👍🏻

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+            A string naming the target name or component we want to plot.
+        foreground_series
+            Target timeseries we want to explain. Can be multivariate.
+        foreground_past_covariates
+            Optionally, past covariate timeseries if needed by model.
+        foreground_future_covariates
+            Optionally, future covariate timeseries if needed by model.
+            Optionally, A list of string naming the target names we want to plot.
+
+
```
dennisbader, line 332:
```suggestion
        horizon
            An integer for the point/step in the future we want to explain, starting from the first 
            prediction step at 0. Currently, only forecasting models are supported which provide an 
            `output_chunk_length` parameter. `horizons` must not be larger than `output_chunk_length`.
        target_name
            The target component name to plot.
        foreground_series
            The target series to explain. Can be multivariate.
        foreground_past_covariates
            Optionally, a past covariate series if required by the forecasting model.
        foreground_future_covariates
            Optionally, a future covariate series if required by the forecasting model.
        **kwargs
            Optionally, additional keyword arguments passed to `shap.force_plot()`.
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+                shap.summary_plot(
+                    shaps_[h][t],
+                    foreground_X_sampled,
+                    plot_type=plot_type,
+                    **kwargs,
+                )
+
+    def force_plot_from_ts(
+        self,
+        horizon: int = None,
```
dennisbader, line 301: Should this be optional (or initialized to 0 instead of None)? I believe it will fail for horizon=None

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+        self.future_covariates_names = future_covariates_names
+
+        self.n = n
+        self.shap_method = shap_method
+        self.background_series = background_series
+        self.background_past_covariates = background_past_covariates
+        self.background_future_covariates = background_future_covariates
+
+        self.single_output = False
+        if (self.n == 1) and (self.target_dim) == 1:
```
dennisbader, line 469:
```suggestion
        if self.n == 1 and self.target_dim == 1:
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+        Parameters
+        ----------
+        foreground_X
+            the Dataframe of lags features specific of darts RegressionModel.
+        horizons
+            Optionally, a list of integer values representing which elements in the future
+            we want to explain, starting from the first timestamp prediction at 0.
+            For now we consider only models with output_chunk_length and it can't be bigger than output_chunk_length.
+        target_names
+            Optionally, A list of string naming the target names we want to explain.
```
dennisbader, line 516:
```suggestion
        horizons
            Optionally, a list of integers representing which points/steps in the future we want to explain,
            starting from the first prediction step at 0. Currently, only forecasting models are supported which
            provide an `output_chunk_length` parameter. `horizons` must not be larger than `output_chunk_length`.
        target_names
            Optionally, a list of strings with the target components we want to explain.
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+
+        """
+
+        # Creation of an unified dictionary between multiOutputRegressor estimators and
```
dennisbader, line 520:
```suggestion
        # create a unified dictionary between multiOutputRegressor estimators and
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+            )
+        elif shap_method == _ShapMethod.KERNEL:
+            explainer = shap.KernelExplainer(
+                model_sklearn.predict, background_X, keep_index=True, **kwargs
+            )
+        elif shap_method == _ShapMethod.LINEAR:
+            explainer = shap.LinearExplainer(model_sklearn, background_X, **kwargs)
+        elif shap_method == _ShapMethod.DEEP:
+            explainer = shap.LinearExplainer(model_sklearn, background_X, **kwargs)
+        elif shap_method == _ShapMethod.ADDITIVE:
```
dennisbader, line 596: Shouldn't we add an `else` condition to catch other types and raise an error?

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,669 @@
...
+        return explainer
+
+    def _create_regression_model_shap_X(
+        self,
+        target_series,
+        past_covariates,
+        future_covariates,
+        n_samples=None,
+        train=False,
+    ):
```
dennisbader, line 610:
```suggestion
    ) -> pd.DataFrame:
```

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # We need at least 8 points for force_plot
+        self.assertTrue(
+            ShapExplain.force_plot_from_ts(
+                2,
+                "power",
+                self.target_ts[100:108],
+                self.past_cov_ts[100:108],
+                self.fut_cov_ts[100:108],
+            )
```
dennisbader, line 343: does this method return True?

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -0,0 +1,348 @@
...
+
+        # Missing a past covariate if you choose to use a new background
+        with self.assertRaises(ValueError):
+            ShapExplainer(
+                m, self.target_ts, background_future_covariates=self.fut_cov_ts
+            )
+
+        # good type of explainers
+        ShapExplain = ShapExplainer(m)
```
dennisbader, line 186: variable names should be lowercase and seperated by "_".
Could you also adapt the same cases further down?
```suggestion
        shap_explain = ShapExplainer(m)
```

requirements/core.txt
```diff
@@ -17,3 +17,4 @@ statsmodels>=0.13.0
 tbats>=1.1.0
 tqdm>=4.60.0
 xarray>=0.17.0
+shap>=0.40.0
```
dennisbader, line 20: Did you check that this is the minimum version which we can support?

dennisbader (commented):
Really cool, thanks a lot @dumjax!
Will be great to use the explainer in the future :)

In general, it looks very good!
The main part of my suggestions is actually only about docstring formatting and some rephrasing/trying to simplify a couple of the "complex" sections.

darts/explainability/explainability.py
```diff
@@ -0,0 +1,336 @@
...
+                - In general we want to keep the training_series of the model and this is the default one,
+                but in case of multiple time series training (global or meta learning) the ForecastingModel doesn't
+                save them. In this case we need to feed a background time series.
+                - We might want to consider a reduced well chosen background in order to reduce computation
+                time.
+        background_past_covariates
+            A past covariates TimeSeries or list of TimeSeries that the model needs once fitted.
+        background_future_covariates
+            A future covariates TimeSeries or list of TimeSeries that the model needs once fitted.
+        """
```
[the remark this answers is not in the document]
hrzn: +1, replacing "(list of)" by (sequence of)" :D

darts/explainability/explainability.py
```diff
@@ -0,0 +1,336 @@
...
+        foreground_past_covariates: Optional[
+            Union[TimeSeries, Sequence[TimeSeries]]
+        ] = None,
+        foreground_future_covariates: Optional[
+            Union[TimeSeries, Sequence[TimeSeries]]
+        ] = None,
+        horizons: Optional[Sequence[int]] = None,
+        target_names: Optional[Sequence[str]] = None,
+    ) -> ExplainabilityResult:
+        """
```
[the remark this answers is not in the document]
hrzn: This probably should be adapted to reflect that it's now an ExplainabilityResult (and not a dict) being returned.

darts/utils/timeseries_generation.py
```diff
@@ -668,8 +668,8 @@ def datetime_attribute_timeseries(
     else:
         if cyclic:
             if attribute == "day":
-                periods = [time_index[i].days_in_month for i in time_index.month]
-                freq = 2 * np.pi * np.reciprocal(periods)
+                periods = time_index.days_in_month.values
+                freq = 2 * np.pi * np.reciprocal(periods.astype(float))
```
dennisbader, line 672: We must use `dtype` here so that TorchForecastingModels won't crash
```suggestion
                freq = 2 * np.pi * np.reciprocal(periods.astype(dtype))
```

dennisbader (changes requested):

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,681 @@
...
+            horizons, target_names
+        )
+
+        shap_values_list = []
+
+        for idx, foreground_ts in enumerate(foreground_series):
+
+            foreground_X = self.explainers._create_regression_model_shap_X(
+                foreground_ts,
+                foreground_past_covariates[idx],
```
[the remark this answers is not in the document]
dumjax: solved

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+            foreground_past_cov_ts = None
+            foreground_future_cov_ts = None
+
+            if foreground_past_covariates:
+                foreground_past_cov_ts = foreground_past_covariates[idx]
+
+            if foreground_future_covariates:
+                foreground_future_cov_ts = foreground_future_covariates[idx]
+
+            foreground_X = self.explainers._create_regression_model_shap_X(
```
hrzn, line 217: maybe avoid relying on the default values and specify train=False here

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+            shap_ = self.explainers.shap_explanations(
+                foreground_X, horizons, target_names
+            )
+
+            shap_values_dict = {}
+            for h in horizons:
+                tmp = {}
+                for t in target_names:
+                    tmp[t] = TimeSeries.from_times_and_values(
+                        shap_[h][t].time_index,
```
hrzn, line 232: I guess having the horizons starting at 1 would mean indexing `h-1` here.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+    def summary_plot(
+        self,
+        horizons: Optional[Sequence[int]] = None,
+        target_names: Optional[Sequence[str]] = None,
+        nb_samples: Optional[int] = None,
+        plot_type: Optional[str] = "dot",
+        **kwargs,
+    ):
+        """
+        Display a shap plot summary per (target, horizon)).
```
hrzn, line 254:
```suggestion
        Display a shap plot summary for each horizon and each component dimension of the target.
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+        We here reuse the initial background data as foreground (potentially sampled) to give a general importance
```
hrzn, line 255:
```suggestion
        This method reuses the initial background data as foreground (potentially sampled) to give a general importance
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+        plot for each feature.
+        If no target names and/or no horizons are provided, we plot all summary plots.
```
hrzn, line 257:
```suggestion
        If no target names and/or no horizons are provided, all summary plots are produced.
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+
+        Parameters
+        ----------
+        horizons
+            Optionally, a list of integers representing which points/steps in the future we want to explain,
+            starting from the first prediction step at 0. Currently, only forecasting models are supported which
```
hrzn, line 263: Is this sentence necessary? As anyway the whole class is limited to RegressionModels.

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+            provide an `output_chunk_length` parameter. `horizons` must not be larger than `output_chunk_length`.
+        target_names
+            Optionally, a list of strings with the target components we want to explain.
```
hrzn, line 266:
```suggestion
            Optionally, a list of strings with the target components to be explained.
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+        horizon: int = None,
+        target_name: str = None,
+        foreground_series: TimeSeries = None,
+        foreground_past_covariates: Optional[TimeSeries] = None,
+        foreground_future_covariates: Optional[TimeSeries] = None,
+        **kwargs,
+    ):
+        """
+        Display a shap force_plot for one target and one horizon.
+        It displays SHAP values of each lag/covariate with an additive force layout.
```
hrzn, line 312: (for consistency with the rest of the docs)
```suggestion
        It displays shape values of each lag/covariate with an additive force layout.
```

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+        Here the inputs are the foreground series (and not the shap explanations objects)
```
hrzn, line 313: "(and not the shap explanations objects)" --> I don't understand this. What are the other methods receiving shap explanation objects as inputs?

darts/explainability/shap_explainer.py
```diff
@@ -0,0 +1,690 @@
...
+        'original sample ordering' has to be selected to observe the time series chronologically.
+
+        Parameters
+        ----------
+        horizon
+            An integer for the point/step in the future we want to explain, starting from the first
+            prediction step at 0. Currently, only forecasting models are supported which provide an
```
hrzn, line 320: I think this sentence could be removed.

darts/utils/data/tabularization.py
```diff
@@ -0,0 +1,136 @@
...
+from darts.logging import raise_if
+from darts.timeseries import TimeSeries
+
+
+def create_lagged_data(
+    target_series: Union[TimeSeries, Sequence[TimeSeries]],
+    output_chunk_length: int,
+    past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+    future_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
+    lags_list: Optional[Sequence[int]] = None,
```
hrzn, line 15: could we remove the `_list` from the variable names here and below? Also adapt the type hints, if they are indeed lists.

darts/utils/data/tabularization.py
```diff
@@ -0,0 +1,136 @@
+from typing import Optional, Sequence, Union
+
+import numpy as np
+import pandas as pd
+
+from darts.logging import raise_if
+from darts.timeseries import TimeSeries
+
+
+def create_lagged_data(
```
hrzn, line 10: I think we should probably keep this function private (unless we have a reason not to?)

darts/utils/data/tabularization.py
```diff
@@ -0,0 +1,136 @@
...
+    max_samples_per_ts.
+
+    X has the following structure:
+    lags_target | lags_past_covariates | lags_future_covariates
+
+    Where each lags_X has the following structure (lags_X=[-2,-1] and X has 2 components):
+    lag_-2_comp_1_X | lag_-2_comp_2_X | lag_-1_comp_1_X | lag_-1_comp_2_X
+
+    y has the following structure (output_chunk_length=4 and target has 2 components):
+    lag_+0_comp_1_target | lag_+0_comp_2_target | ... | lag_+3_comp_1_target | lag_+3_comp_2_target
```
hrzn, line 32: It would be nice to document each of the parameters (even if, assuming we make the method private, it won't appear in the documentation pages)

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -0,0 +1,434 @@
...
+            "1_fut_cov_lag0",
+            "hour_fut_cov_lag0",
+            "dayofweek_fut_cov_lag0",
+            "relative_idx_fut_cov_lag0",
+        ]
+
+        results = shap_explain.explain()
+
+        # all the features explained are here, in the right order
+        self.assertTrue(
```
hrzn, line 349: nice one

hrzn (changes requested):
I feel we're 95% there, but there are still a few important things to adapt and to polish before we can merge.
I've made lots of small comments, mainly about the documentation (pls double-check how it renders in HTML before merging), variable naming (I think `target_names` is confusing, we should use `target_components`) and also having the horizons start at 1 instead of 0 to be consistent with our forecasting models.

hrzn (approved):

PR #917, opened 2022-04-15
Feat/separate doc workflow

hrzn:
There is a small fix to do in the user guide (wrong image being used), and I would like to release the documentation again without releasing a new version of the library, so I wrote a new workflow to build and release only the doc.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -6,7 +6,6 @@ on:
 jobs:
   deploy-docs:
     runs-on: ubuntu-latest
-    needs: [release]
     steps:
       - name: "1. Clone repository"
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dennisbader (commented):
Thanks, that will certainly help for the future docs fixes :) Looks good to me. Maybe @tomasvanpottelbergh can give the approval as I'm not super familiar with workflow files?

tomasvanpottelbergh (approved):

.github/workflows/doc.yml
```diff
@@ -0,0 +1,59 @@
+name: darts documentation build and release workflow
+
+on:
+  workflow_dispatch:
+
+jobs:
+  deploy-docs:
+    runs-on: ubuntu-latest
+    needs: [release]
```
tomasvanpottelbergh, line 9: I think this might need to be removed?

hrzn: Good catch, thanks

PR #924, opened 2022-04-20
Recognize DARTS_CONFIGURE_MATPLOTLIB=0

### LINKED ISSUE #892: [BUG] Darts matplotlib style settings should be optional
**Describe the bug**
Importing anything from darts results in significant changes to the global matplotlib configuration.  The style is pretty good, but it should be optional.

**To Reproduce**
```python
import darts
plt.figure()
```

**Expected behavior**
There is a way to make the figure reflect the user's personal matplotlibrc without manually undoing everything listed in `darts/__init__.py`.

 - Python version: 3.7.9
 - darts==0.17.1

**Additional context**
A possible fix would be to suppress [this call](https://github.com/unit8co/darts/blob/d5dd29611221ccceb35236f018e94cd7dd918043/darts/__init__.py#L40) if something like `export DARTS_CONFIGURE_MATPLOTLIB=0` is set.

zgana:
Fixes #892.

### Summary

This PR makes it possible to import darts without changing the global matplotlib configuration.

* By default, darts still applies matplotlib configurations
* If the environment variable DARTS_CONFIGURE_MATPLOTLIB=0 is set, then darts.u8plots_mplstyle is still defined but not applied.

### Other Information

Normally it would be best to add tests of the new behavior, but it's difficult to test behavior in the root `__init__.py`.  Both the pre-existing behavior and the adjustment here would be easier to test if the plotting configuration were factored out into some helper module.  That would be a larger (though still fairly small) change.

A larger question would be: _should_ darts change the global matplotlib configuration by default?  Making this opt-in would obviously cause an aesthetic mismatch for anyone following documentation produced prior to that change.  But it would arguably be cleaner (and more testable) to require e.g. `darts.apply_u8plots_mplstyle()` or otherwise respect users' matplotlibrc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Darts is still in an early development phase and we cannot always guarantee back
 
 ## [Unreleased](https://github.com/unit8co/darts/tree/master)
 [Full Changelog](https://github.com/unit8co/darts/compare/0.19.0...master)
+### For users of the library:
+
+**Improved**
+- Option to avoid global matplotlib configuration changes.
+  [#924](https://github.com/unit8co/darts/pull/924) by [Mike Richman](https://github.com/zgana).
 
 ## [0.19.0](https://github.com/unit8co/darts/tree/0.19.0) (2022-04-13)
 ### For users of the library:
--- a/darts/__init__.py
+++ b/darts/__init__.py
@@ -3,6 +3,8 @@ darts
 -----
 """
 
+import os
+
 import matplotlib as mpl
 from matplotlib import cycler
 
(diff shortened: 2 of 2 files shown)


hrzn (approved):
Thanks for this, LGTM.
I agree with your comment that opting-in using something along the lines of `darts.apply_u8plots_mplstyle()` would be better eventually. Modifying the global config is not great. We could modify `TimeSeries.plot()` in order to still apply some style in those plots only, maybe.

zgana:
Thanks @hrzn! I just realized I did not add an entry to the change log. Should I push another commit with an entry under "Unreleased"?

hrzn:
Yes that'd be great - thanks!

hrzn (approved):

hrzn:
Thank you @zgana, merging now, it will be part of the next release :)

PR #931, opened 2022-04-28
Correct the docstring of the TransformerModel

shaido987:
No related issue since it's a very minor fix that is easily done.

Corrects the docstring of the `TransformerModel` to make the parameter description match the actual parameter values.
1) the `model` parameter is no longer supported and raises an error when used.
2) `input_chunk_length` and `output_chunk_length` do not have default values.
3) multiple parameters had incorrect default values.

Also fixed some capitalization and punctuation issues and changed "the multiheadattention model" to "the multi-head attention mechanism".


hrzn (approved):
Good catch, thank you!

PR #958, opened 2022-05-13
Enforce argument names when calling `raise_if` and `raise_if_not`

### LINKED ISSUE #956: Bug: wrong parameter passing to function `raise_if_not`
https://github.com/unit8co/darts/blob/5213854b71e3d56d41dcdffac59140a8da9c10a7/darts/utils/utils.py#L227

In this `utils.py` file, in line 227, the signature when calling `raise_if_not` is incorrect.
```python
# darts/utils/utils.py
raise_if_not(n.start >= 0, logger)

# darts/logging.py
def raise_if_not(
    condition: bool,
    message: str = "",
    logger: logging.Logger = get_logger("main_logger"),
)
```

When calling `raise_if_not`, the logger object is passed in the place of the message object. As such, when the condition is `False`, there is an attempt to concatenate a string and a logger object `logger.error("ValueError: " + message)`, where message is now a logger object, resulting in an error
```
TypeError: can only concatenate str (not "Logger") to str
```

The fix should be to update the calling signature
```python
# darts/utils/utils.py
raise_if_not(n.start >= 0, logger=logger)
```
It is also useful to include a message too, although I would need advice what should an appropriate message be.

nlhkh:
Fixes #956 

### Summary

The 2 functions `raise_if` and `raise_if_not` require specifying argument names for message and logger, so that developers do not mistakenly specify a logger in place of a message.


darts/logging.py
```diff
@@ -57,6 +57,7 @@ def raise_deprecation_warning(
 
 def raise_if_not(
     condition: bool,
+    *,
```
hrzn, line 60: Why is this needed now?

hrzn (commented):
Thanks for this. In fact, I'm wondering whether we should drop the logger argument altogether? WDYT?
Also: could you check what's causing one of the tests to fail?
Thanks

nlhkh:
> Thanks for this. In fact, I'm wondering whether we should drop the logger argument altogether? WDYT?
> 
> Also: could you check what's causing one of the tests to fail?
> 
> Thanks

I will by early next week.

hrzn:
> > Thanks for this. In fact, I'm wondering whether we should drop the logger argument altogether? WDYT?
> > Also: could you check what's causing one of the tests to fail?
> > Thanks
> 
> I will by early next week.

fyi: the tests are most likely not failing due to this PR, but rather to some external dependency that got updated.

darts/logging.py
nlhkh: It is to force using parameter name, so that the caller has to use `raise_if_not(..., msg="msg")` instead of `raise_if_not(..., "msg")`. If you think this style is bad, I can remove it.

hrzn: OK I see. I think it'd be slightly better to remove it so we keep things a bit simpler. There's no need to enforce the keyword arguments too strongly IMO.

nlhkh:
> @nlhkh what is the status of this PR?

@hrzn I think it is ready. I keep the change to minimum to only avoid the error when passing the log argument.

hrzn:
> > @nlhkh what is the status of this PR?
> 
> @hrzn I think it is ready. I keep the change to minimum to only avoid the error when passing the log argument.

OK, then I'll merge. Thanks!

nlhkh:
> > > @nlhkh what is the status of this PR?
> > 
> > 
> > @hrzn I think it is ready. I keep the change to minimum to only avoid the error when passing the log argument.
> 
> OK, then I'll merge. Thanks!

Thank you!

Issue #617, opened 2021-12-01
Add new datasets to `darts.datasets`


hrzn:
It would be nice to have more datasets - in particular "parts" and "traffic" used in some of the Amazon research papers.

A good source: https://archive.ics.uci.edu/ml/datasets.php

* Traffic: UCI Machine Learning Repository: PEMS-SF Data Set 
https://archive.ics.uci.edu/ml/datasets/PEMS-SF

Issue #597, opened 2021-12-01
Add support for static covariates


hrzn:
Needs to be broken down.

* Add static covariates to `TimeSeries`.
* Adapt some of the models that can exploit it.

DeastinY:
Most relevant would probably be TFT, which is also already implemented into pytorch-forecasting. It encodes static information at every step anew.

manavchoudhary:
Are there any plans to include support for static covariates in Temporal Fusion Transformer (TFT) model? Thanks!

felipegssantos:
I believe the RegressionModel would also be a good test candidate. But indeed, adapting the TimeSeries to handle static covariates seems like a hard task. Are there any discussions around this issue going on? I would very much like to participate if possible.

PR #960, opened 2022-05-15
Feat/New Datasets (ETT and Electricity)

gdevos010:
Fixes part of #617.

The PEMS-SF dataset is in an odd format and Im not sure how to include it.

### Summary
Added the ETT and Electricity Dataset. 
Added a method for processing other .zip datasets.

### Other Information
None


gdevos010:
@hrzn I also need help understanding why this one failed? It's falling on the Electricity dataset but not the ETT ones.
`FAILED darts/tests/datasets/test_dataset_loaders.py::DatasetLoaderTestCase::test_ok_dataset`

hrzn:
> 

The tests fail because a newer version of some dependency is introducing an issue. It's not due to this PR - it's a separate issue that we'll fix asap.

darts/datasets/__init__.py
```diff
@@ -308,3 +309,130 @@ def __init__(self):
...
+                hash="7687e47825335860bf58bccb31be0c56",
+                header_time="date",
+                format_time="%Y-%m-%d %H:%M:%S",
+            )
+        )
+
+
+class ElectricityDataset(DatasetLoaderCSV):
+    """
+    Measurements of electric power consumption in one household with a one-minute sampling rate over a period of almost
```
hrzn, line 412: A few things seem off here. The time granularity is 15-mins and not minute, and also, looking at the dataset description on UCI ML repository, there is "one column per client". I understand this as meaning there are 370 distinct housholds, and this is their electricity consumption (vs 370 different electrical quantities of one household). Let me know if I misunderstood this? Also, I think that in this case we should probably return a list of 370 `TimeSeries` rather than a single `TimeSeries` - I somehow expect this to be a more natural representation (each household being one "data point" -vs- one data point of 370 dimensions). Let me know what you think.

darts/datasets/dataset_loaders.py
```diff
@@ -113,6 +121,29 @@ def _download_dataset(self):
                 "Could not download the dataset. Reason:" + e.__repr__()
             ) from None
 
+    def _download_zip_dataset(self):
+        os.makedirs(self._root_path, exist_ok=True)
+        extracted_dir = Path(DatasetLoader._DEFAULT_DIRECTORY, "temp_dir")
```
hrzn, line 126: It'd be safer to use [this way of creating temporary directories](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). Also using a context manager can ensure the directory is deleted.

darts/datasets/dataset_loaders.py
```diff
@@ -19,12 +22,14 @@ class DatasetLoaderMetadata:
...
+    header_time: Optional[str]
     # used to convert the string date to pd.Datetime
     # https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior
-    format_time: str = None
+    format_time: Optional[str] = None
     # used to indicate the freq when we already know it
-    freq: str = None
+    freq: Optional[str] = None
+    # a custom function to handling non-csv based datasets
+    pre_process_fn: Optional[Callable] = None
```
hrzn, line 32: Is this one also used in case we're not using a zip file?

darts/datasets/dataset_loaders.py
```diff
@@ -113,6 +121,29 @@ def _download_dataset(self):
...
+        extracted_dir = Path(DatasetLoader._DEFAULT_DIRECTORY, "temp_dir")
+        zip_path = Path(DatasetLoader._DEFAULT_DIRECTORY, "temp.zip")
+
+        try:
+            request = requests.get(self._metadata.uri)
+            with open(zip_path, "wb") as f:
+                f.write(request.content)
+            with zipfile.ZipFile(zip_path, "r") as zip_ref:
+                zip_ref.extractall(extracted_dir)
+            self._metadata.pre_process_fn(extracted_dir, self._get_path_dataset())
```
hrzn, line 135: what happens if `pre_process_fn` is `None` ?

hrzn (commented):
Thanks, it's a very nice initiative to add more datasets :)
I have a few comments before we can merge this one.

darts/datasets/__init__.py
```diff
@@ -308,3 +309,130 @@ def __init__(self):
                 format_time="%Y-%m-%d",
             )
         )
+
+
+class ETTh1Dataset(DatasetLoaderCSV):
```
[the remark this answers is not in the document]
gdevos010: I was a bit too careless copying the info from the ETT repo. You are right; each ETT file is for 1 transformer. I added the column descriptions.

darts/datasets/dataset_loaders.py
gdevos010: I did not know about the `tempfile` library. I was using `finally` to clean up the directory, but this is much nicer.

darts/datasets/dataset_loaders.py
gdevos010: `pre_process_fn` is required for any non-csv file, even if it is to move the file from the directory to the datasets folder. I added a test to check for this.

darts/datasets/__init__.py
gdevos010: Im not sure where I found that description, but it is definitely wrong. I have updated it with the correct information. 

I think the list vs. single `TimeSeries` is related to the multi-series vs. multivariate discussion. As a small anecdote, I get better results with the ice cream & heater dataset if I `fit` using the dataset as a multivariate, even if Im making univariate predictions.

In general, this feels like a decision that should be left to the user. The tutorials show how to turn quickly turn a multivariate dataset into a list of `TimeSeries`. I could add some example code in the docstring.

darts/datasets/__init__.py
```diff
@@ -308,3 +309,130 @@ def __init__(self):
...
+    ----------
+    .. [1] https://github.com/zhouhaoyi/ETDataset
+    .. [2] https://arxiv.org/abs/2012.07436
+    """
+
+    def __init__(self):
+        super().__init__(
+            metadata=DatasetLoaderMetadata(
+                "ETTh1.csv",
+                uri="https://raw.githubusercontent.com/zhouhaoyi/ETDataset/main/ETT-small/ETTh1.csv",
```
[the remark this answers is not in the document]
gdevos010: I have added the files to the datasets folder. The `test_ok_dataset` test case now fails because they are not in master yet.

dennisbader:
Those will be nice additions, thanks for that!

From a quick look at the data, these are multivariate datasets without any static covariates (such as household id or similar), right?

Could we imagine also adding a dataset for static covariates? I'm currently using the one from below:

This one [here](https://www.kaggle.com/datasets/utathya/future-volume-prediction) includes some of Stallion & Co.'s beer sales data from different wholesalers (Agencies) and SKUs (Stock Keeping Units).

*Context: Country Beeristan, a high potential market, accounts for nearly 10% of Stallion & Co.’s global beer sales. Stallion & Co. has a large portfolio of products distributed to retailers through wholesalers (agencies). There are thousands of unique wholesaler-SKU/products combinations. In order to plan its production and distribution as well as help wholesalers with their planning, it is important for Stallion & Co. to have an accurate estimate of demand at SKU level for each wholesaler.*

I'm interested in the salespromotion.csv: *($/hectoliter) Holds the monthly price, sales & promotion data in dollar value per hectoliter at Agency-SKU-month level from January 2013 - December 2017.* 
This results in 350 TimeSeries at Agency-SKU level. They share the same time index so we could used it for both multivariate and multiple TS.

### Additional Info
Agency and SKU are categorical, so we would have to encode them. There are 58 unique Agencies and 28 unique SKUs. For testing if everything works with static covariates I'm just converting them to numerical right now as below (will try later on with one hot encoding).

```
df = pd.read_csv(fin)
for col in ["Agency", "SKU"]:
    df[col] = pd.Categorical(df[col]).codes
```

darts/datasets/__init__.py
hrzn: OK fair enough we can leave it like that. It'd be awesome if you add an example code in the docstring - making it clear what is returned is a single multivariate `TimeSeries` containing 370 components, which can be transformed to 370 univariate `TimeSeries` (from experience it's a constant source of confusion among users).

darts/datasets/dataset_loaders.py
hrzn: It could be nice to use `pre_process_fn` in any case (zip file or not) if it is provided, and also support the case where it is not provided (assuming the zipped CSV file is already of the right format). WDYT?

darts/datasets/__init__.py
gdevos010: I added some sample code. let me know what you think

gdevos010:
@dennisbader this seems like a separate PR. Can we move this discussion to #966 or #597 ? I agree that this is an important next step for datasets.

darts/datasets/dataset_loaders.py
gdevos010: I think it would be nice to have a use case before adding code. Do you have any in mind?

hrzn: In fact I didn't have in mind to add any code, just re-arrange things slightly so that `pre_process_fn` is being called in all execution branches (not the zip file one only). I'm proposing this because I find it potentially slightly confusing to have such a parameter `pre_process_fn` and have it silently ignored in some cases (plain CSV file). But maybe I'm missing something?

hrzn:
> @dennisbader this seems like a separate PR. Can we move this discussion to #966 or #597 ? I agree that this is an important next step for datasets.

I also agree with that. @dennisbader could we address it in a separate PR once your static covariates one is merged?

darts/datasets/dataset_loaders.py
gdevos010: Maybe Im missing something, or I should rename it? the goal of the function is to take in a temp direction (the unzipped folder) and move it to the darts database folder. here's what I was trying to accomplish with the electricity dataset. First, we have to extract the zip, then move the file to the `.darts/datasets` folder. But unfortunately, the dataset is not in the standard comma-separated standard, so each comma needs to be changed to a period and each semi-colon to a comma. This way, the dataset conforms to the format of all the other datasets and can reuse the load function. 

https://github.com/unit8co/darts/blob/61eba9b9ba504239764a98257e81570cda2cb600/darts/datasets/__init__.py#L423-L427

If we move it outside the zip section, that temp folder is going to be empty so I'm not sure what processing would happen
```python 
        if not self._is_already_downloaded():
            with tempfile.TemporaryFile() as tf:
                with tempfile.TemporaryDirectory() as td:
                    if self._metadata.uri.endswith(".zip"):
                        self._download_zip_dataset(tf, td)
                    else:
                        self._download_dataset()
                    self._metadata.pre_process_fn(td, self._get_path_dataset())
        self._check_dataset_integrity_or_raise()
        return self._load_from_disk(self._get_path_dataset(), self._metadata)
```

Potential solutions would be to rename it to be less confusing and to throw an exception if the arg is provided with a csv file. I also might just not understand.

hrzn: OK I see. What I had in mind was to separate *moving* the file to the right path from *pre-processing* its content (e.g. to change periods in semi-columns). This way the `pre_process_fn` could do only the pre-processing on any CSV file, and we could move the file first (using e.g. `shutil.move`). We could also easily apply `pre_process_fn` in the function `_download_dataset()` too so that it would also be applied to any CSV file even if this CSV file has not previously been zipped. However, decoupling this way also makes things slightly more complex which is perhaps unnecessary at this point. So let's leave it as you propose. Perhaps one improvement would be to rename `pre_process_fn` to `pre_process_zipped_csv_fn` to make it clear it applies only to CSVs that are zipped. Makes sense to you?

gdevos010: renamed the parameter and added a warning if the param is used with a csv file.

gdevos010:
@hrzn The dataset tests won't pass because the ETT files are not yet in main. Should we merge or handle it a different way

darts/datasets/__init__.py
```diff
@@ -308,3 +309,205 @@ def __init__(self):
...
+                        fout.write(line.replace(",", ".").replace(";", ","))
+
+        super().__init__(
+            metadata=DatasetLoaderMetadata(
+                "Electricity.csv",
+                uri="https://archive.ics.uci.edu/ml/machine-learning-databases/00321/LD2011_2014.txt.zip",
+                hash="acfe6783eea43905e510f537add940fd",
+                header_time="Unnamed: 0",
+                format_time="%Y-%m-%d %H:%M:%S",
+                pre_process_fn=pre_proces_fn,
```
hrzn, line 511:
```suggestion
                pre_process_zipped_csv_fn=pre_proces_fn,
```

hrzn (approved):

hrzn:
> @hrzn The dataset tests won't pass because the ETT files are not yet in main. Should we merge or handle it a different way

I think we have no other choice but to merge and hope for the best :) We'll open a fix PR later on if it fails for some reason.

PR #966, opened 2022-05-20
Feat/static covs

### LINKED ISSUE #597: Add support for static covariates

dennisbader:
Fixes #597.

Edit:
### Summary
- adds static covariate support for TimeSeries: static covariates are stored in the TimeSeries object itself as a pd.DataFrame
- adds static covariates to torch datasets (additionally, unified the variable names between all datasets: training and inference datasets)
- adds static covariate support for TFTModel
- adds method `TimeSeries.with_static_covariates()` to add static covariates to any TimeSeries
- adds method `TimeSeries.from_group_dataframe()`: group dataframes -> Basically multiple unit specific time series DataFrames that are vertically stacked. These can now be read and converted to multiple TimeSeries. 
- additional methods/properties: `TimeSeries.static_covariates_values()`, `TimeSeries.static_covariates`


darts/timeseries.py
```diff
@@ -496,6 +497,119 @@ def from_dataframe(
...
+                if not isinstance(static_cov_vals, tuple)
+                else static_cov_vals
+            )
+            # check that for each group there is only one unique value per column in `static_cols`
+            if static_cols:
+                static_cols_valid = [
+                    len(group[col].unique()) == 1 for col in static_cols
+                ]
+                if not all(static_cols_valid):
+                    # encountered performance issues when evaluating the error message from below in every
```
hrzn, line 577: 🤔  Any idea why?

darts/timeseries.py
```diff
@@ -496,6 +497,119 @@ def from_dataframe(
...
+            splits.append(
+                (
+                    pd.DataFrame(static_cov_vals, index=static_cov_cols),
+                    group.drop(columns=static_cov_cols),
+                )
+            )
+
+        # create a list with multiple TimeSeries and add static covariates
+        return [
+            TimeSeries.from_dataframe(
```
hrzn, line 602: A possibility could be to directly accept `static_covariates` in all (or most of) our `from_XXX` methods. This way we would get a chance to build the DataArray directly right, containing the covariates, and it'd avoid copying it later on in in the method to set the static covariates (see my other comments there to keep TimeSeries immutable).

But that's just a thought - it might be overkill and avoiding this copy not really needed for now.

darts/timeseries.py
```diff
@@ -2019,6 +2142,32 @@ def with_values(self, values: np.ndarray) -> "TimeSeries":
...
+                n_components > 1 and n_components != self.n_components,
+                "When passing a multi-column pandas DataFrame, the number of columns must match the number of "
+                "components of the TimeSeries object (multivariate static covariates must map to each TimeSeries "
+                "component).",
+                logger,
+            )
+        elif isinstance(covariates, pd.Series):
+            covariates = covariates.to_frame()
+        self._xa.attrs["static_covariates"] = (
+            covariates.astype(self.dtype) if covariates is not None else covariates
```
hrzn, line 2167: I think we could drop this type conversion and later datasets and/or models deal with it downstream - assuming some models such as LGBM can support for instance string types, while others (such as torch models) cannot (we could add a type check with an informative message in the datasets).

hrzn (commented):
I've only reviewed `TimeSeries` so far - will cover the rest later.

darts/tests/models/forecasting/test_TFT.py
```diff
@@ -163,6 +165,34 @@ def test_mixed_covariates_and_accuracy(self):
...
+        def test_static_covariates_support(self):
+            target = tg.sine_timeseries(length=2, freq="h")
+            target.set_static_covariates(pd.Series([0.0, 1.0], index=["st1", "st2"]))
+
+            # should work with cyclic encoding for time index
+            model = TFTModel(
+                input_chunk_length=1,
+                output_chunk_length=1,
+                add_encoders={"cyclic": {"future": "hour"}},
+                pl_trainer_kwargs={"fast_dev_run": True},
```
hrzn, line 177: Nice - we could maybe use this in more places :)

hrzn (commented):
It looks really great 💯💯. A few comments:
* Let's keep `TimeSeries` immutable and check static covariates format (e.g. DataFrame of right shape and maybe right dtype upon construction of `TimeSeries`).
* On dimensionality of static covariates: I think what you have now is good and we can leave it like that. Maybe add some good documentation on any method to access those, so that it's clear what the dimensions are.
* You need add support for concatenation/stacking (hopefully quite easy) and maybe grep all parts of the code that calls `values()` or `all_values()` to make sure static covariates are propagated where needed.
* I don't think we need to scale the static covariates - at least not for now. But let's mention this in the scaler docstring.
* I think you can already pass on the static covariates tensors to all models and leave it up to the models (currently all of them but TFT) to ignore them.
* Categorical support: as discussed, let's be flexible on the types allowed in static covariates and not limit ourselves to numeric types. We can then throw a runtime error if a string is used in e.g. a torch model, and leave it up to the models (or datasets) to do these checks and use the covariates as intended (as discussed, one of the first thing to add later on will be support for categorical features in LGBM).

darts/utils/data/inference_dataset.py
```diff
@@ -168,39 +178,57 @@ def __getitem__(
...
+            )
+            future_covariate = (
+                future_covariate
+                if future_covariate is not None and len(future_covariate) > 0
+                else None
             )
 
-        return past_target, cov_past, cov_future, target_series
+        # TODO: we need think about the dimensionality of static covariates
+        static_covariate = (
```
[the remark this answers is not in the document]
dennisbader: Sounds good. I checked and `to_numpy(copy=False)` returns a view only when all values/variable have the same dtype in the DataFrame. Otherwise, it returns a copy.

darts/models/forecasting/tft_model.py
```diff
@@ -411,27 +410,21 @@ def forward(self, x: Tuple[torch.Tensor, Optional[torch.Tensor]]) -> torch.Tenso
...
         # Embedding and variable selection
-        if static_covariates is not None:
-            # TODO: impelement static covariates
-            # # static embeddings will be constant over entire batch
-            # static_embedding = {name: input_vectors[name][:, 0] for name in self.static_variables}
-            # static_embedding, static_covariate_var = self.static_covariates_vsn(static_embedding)
-            raise NotImplementedError("Static covariates have yet to be defined")
+        if self.static_variables:
+            static_embedding = {
+                name: x_static[:, 0, i].unsqueeze(-1)
```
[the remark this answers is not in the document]
dennisbader: Good catch! 👍

darts/timeseries.py
```diff
@@ -496,13 +563,131 @@ def from_dataframe(
...
+        fillna_value: Optional[float] = None,
+    ) -> Union["TimeSeries", List["TimeSeries"]]:
+        """
+        Build a list of TimeSeries instances grouped by a selection of columns from a DataFrame.
+        One column (or the DataFrame index) has to represent the time,
+        a list of columns `group_cols` must be used for extracting the individual TimeSeries by groups,
+        and a list of columns `value_cols` has to represent the values for the individual time series.
+        Values from columns ``group_cols`` and ``static_cols`` are added as static covariates to the resulting
+        TimeSeries objects. These can be viewed with `my_series.static_covariates`. Different to `group_cols`,
+        `static_cols` only adds the static values without using the to extract the TimeSeries groups.
```
hrzn, line 585:
```suggestion
        `static_cols` only adds the static values but are not used to extract the TimeSeries groups.
```

darts/timeseries.py
```diff
@@ -496,13 +563,131 @@ def from_dataframe(
...
+        freq
+            Optionally, a string representing the frequency of the Pandas DateTimeIndex. This is useful in order to
+            fill in missing values if some dates are missing and `fill_missing_dates` is set to `True`.
+        fillna_value
+            Optionally, a numeric value to fill missing values (NaNs) with.
+
+        Returns
+        -------
+        TimeSeries
+            A univariate or multivariate deterministic TimeSeries constructed from the inputs.
```
hrzn, line 621:
```suggestion
            A list containing a univariate or multivariate deterministic TimeSeries per group in the DataFrame.
```

darts/timeseries.py
```diff
@@ -705,6 +919,28 @@ def from_pickle(cls, path: str) -> "TimeSeries":
     ==========
     """
 
+    @property
+    def static_covariates(self) -> Optional[pd.DataFrame]:
+        """
+        Returns the static covariates contained in the series as a pandas DataFrame.
+        The columns represent the static variables and the rows represent the components of the uni/multivariate
+        series. If a single-row DataFrame, the covariates are globally 'applied' to all components of the
```
hrzn, line 927: Do we need to keep this sentence and the rest of the docstring below? seems copy-pasted from the constructor.

darts/timeseries.py
```diff
@@ -346,6 +394,14 @@ def from_csv(
             fill in missing values if some dates are missing and `fill_missing_dates` is set to `True`.
         fillna_value
             Optionally, a numeric value to fill missing values (NaNs) with.
+        static_covariates
+            Optionally, a set of static covariates to be added to the TimeSeries. Either a pandas Series or a pandas
+            DataFrame. If a Series, the index represents the static variables. The covariates are globally 'applied'
+            to all components of the TimeSeries. If a DataFrame, the columns represent the static variables and the
+            rows represent the components of the uni/multivariate TimeSeries. If a single-row DataFrame, the covariates
+            are globally 'applied' to all components of the TimeSeries. If a multi-row DataFrame, the number of
```
hrzn, line 402: When you write "If a single-row DataFrame, the covariates are globally applied" --> That's a good idea. But is it implemented? In the constructor, I only see a check that the length of the DataFrame matches the number of components. Maybe we should "tile" the DataFrame if it has length 1?

darts/timeseries.py
```diff
@@ -3366,10 +3730,69 @@ def _set_freq_in_xa(xa_: xr.DataArray):
...
+
+    raise_if_not(
+        all([len(ts.static_covariates) == ts.n_components for ts in series])
+        and all(
+            [
+                ts.static_covariates.columns.equals(series[0].static_covariates.columns)
+                for ts in series
+            ]
+        ),
+        "Concatenation of multiple TimeSeries with static covariates requires all `static_covariates` "
```
hrzn, line 3779: 👍

darts/timeseries.py
```diff
@@ -1241,6 +1485,29 @@ def univariate_values(self, copy=True, sample=0) -> np.ndarray:
         else:
             return self._xa[:, 0, sample].values
 
+    def static_covariate_values(self, copy: bool = True) -> Optional[np.ndarray]:
```
hrzn, line 1488: Shall this one maybe be called `static_covariates_values` (i.e., with an `s` at `covariates`)? For consistency.

hrzn (approved):
This is really great work @dennisbader - beautiful :)
And it's an excellent basis to start using static covariates in more places :)
I've had a few more very minor comments, but then from my perspective we can merge as soon as you feel confident you've covered all the cases (e.g. with the `_xa` accesses).

darts/timeseries.py
dennisbader: In the constructor it checks the following:
https://github.com/unit8co/darts/blob/6dc7ff886099e775699e5e2dd60fbf8ccd38e75e/darts/timeseries.py#L220

so either the df has 1 or `n_components` components.

The component names are then set in the following lines:
https://github.com/unit8co/darts/blob/6dc7ff886099e775699e5e2dd60fbf8ccd38e75e/darts/timeseries.py#L235

When the number of static cov components is equal to `n_components`, it simply assign the component names to the static covs.
If the number is not equal `n_components`, it must be a single global component and it gets the following component name:
https://github.com/unit8co/darts/blob/6dc7ff886099e775699e5e2dd60fbf8ccd38e75e/darts/timeseries.py#L55

I don't think we need to tile the DataFrame as it doesn't add any new information. WDYT?

If we have a multivariate series with the global staic covariates and we select on component from the series, the static covariate component name will change from `DEFAULT_GLOBAL_STATIC_COV_NAME ` to the name of the univariate component:

```
>>> sc = pd.Series([0., 1.], index=["sc1", "sc2"])
>>> ts = TimeSeries.from_values(values=np.random.rand(5, 2), columns=["a", "b"]).with_static_covariates(sc)
>>> print(ts.static_covariates)
>>> print("")
>>> print(ts["a"].static_covariates)

static_covariates  sc1  sc2
global_components  0.0  1.0

static_covariates  sc1  sc2
component                  
a                  0.0  1.0
```

darts/tests/test_timeseries_static_covariates.py
```diff
@@ -507,35 +539,125 @@ def test_operations(self):
...
-        ts_new = ts**3
-        assert ts_new.static_covariates.equals(ts.static_covariates)
-        ts_new = ts + 3
-        assert ts_new.static_covariates.equals(ts.static_covariates)
-        ts_new = ts - 3
-        assert ts_new.static_covariates.equals(ts.static_covariates)
-        ts_new = abs(ts)
-        assert ts_new.static_covariates.equals(ts.static_covariates)
+        # arithmetics with series (left) and non-series (right)
+        self.helper_test_cov_transfer(ts, ts / 3)
```
hrzn, line 543: 💯

hrzn (approved):
🚀

darts/timeseries.py
hrzn: OK I see. I had in mind to tile it so that whenever we access the covariates values (e.g. from training datasets), we always get a tensor with dimension matching the number of components.

PR #996, opened 2022-06-07
Feat/torchmetrics

### LINKED ISSUE #995: Torch Metrics
Pytorch Lightning has created an excellent metrics library. It can provide metrics feedback while training. 
It can also provide better pruning when using external tuning libraries like Ray tune and Optuna.

https://torchmetrics.readthedocs.io/en/stable/

I will open a PR soon

gdevos010:
Fixes #995.

### Summary

Added Torch Metrics to training and validation


darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -230,6 +257,45 @@ def _compute_loss(self, output, target):
...
+                    )
+                    raise e
+
+    def calculate_metrics(self, y, y_hat, tag):
+        metrics = []
+        for metric, metric_str, metric_params in zip(
+            self.metrics, self.metrics_str, self.metrics_params
+        ):
+            if self.likelihood:
+                _metric = metric(y_hat, self.likelihood.sample(y), **metric_params)
```
gdevos010, line 281: Is `sample` the correct method to use here?

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -230,6 +257,45 @@ def _compute_loss(self, output, target):
...
+            for metric in self.metrics_str:
+                try:
+                    self.metrics.append(getattr(task_module, metric))
+                except AttributeError as e:
+                    raise_log(
+                        ValueError(
+                            f"{metric} is not a valid functional metric defined in the torchmetrics.functional module"
+                        )
+                    )
+                    raise e
```
hrzn, line 273: Is this needed?

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -100,6 +108,23 @@ def __init__(
...
+                len(self.metrics_params) != len(self.metrics_str),
+                "Number of metrics parameters must be equal to number of metrics.",
+                logger,
+            )
+        # create empty dict for each metric
+        if self.metrics_str and metrics_params is None:
+            for _ in self.metrics_str:
+                self.metrics_params.append(dict())
+
+        self._setup_metrics()
```
hrzn, line 126: Here `_self_metrics()` has side effects - it relies on the right `self` variables being set (e.g. `self.metrics_str`) and sets `self.metrics`. I would find it cleaner to keep it without side effects (more functional), doing something like
```python
self.metrics = PLForecastingModule._create_metrics(self.metrics_str)
```
with the method static.

darts/models/forecasting/pl_forecasting_module.py
hrzn: That could be right, but will mean that the metric will be evaluated on random samples of the output distributions, which typically mean the metrics will be noisier. I'm guessing we could leave it like that. Could you add a unit test to test the metrics when the model has a likelihood?

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -230,6 +257,45 @@ def _compute_loss(self, output, target):
...
+
+            self.log(
+                f"{tag}_{metric_str}",
+                _metric,
+                on_epoch=True,
+                on_step=False,
+                logger=True,
+                prog_bar=True,
+            )
+        return metrics
```
hrzn, line 297: Why do we need to maintain and return `metrics` in this function? It seems it's never used.

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -230,6 +257,45 @@ def _compute_loss(self, output, target):
             # last dimension of model output, for properly computing the loss.
             return self.criterion(output.squeeze(dim=-1), target)
 
+    def _setup_metrics(self):
+        if self.metrics_str:
+            self.metrics = []
+            task_module = torchmetrics.functional
```
hrzn, line 263: couldn't we directly do something like `import torchmetrics.functional as func_torch_metrics` at the top of the file? No need to redeclare it here, unless I'm missing something.

hrzn (commented):
Thanks, that looks like a good idea. Maybe it'd be nice to augment the new early stopping subsection of the User Guide to also include an example showing how to use a custom metric for early stopping? WDYT?

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -29,10 +30,12 @@ def __init__(
         input_chunk_length: int,
         output_chunk_length: int,
         loss_fn: nn.modules.loss._Loss = nn.MSELoss(),
+        torch_metrics: Optional[List[str]] = None,
```
dennisbader, line 33: What do you think of accepting instances of `torchmetrics.Metric` here (the parent class from TorchMetrics for all built-in and custom metrics, similar to `loss_fn`) instead of strings? Something similar to `loss_fn`.
```suggestion
# import Metric at beginning
from torchmetrics import Metric
...
        torch_metrics: Optional[List[Metric]] = None,
```
Like this we won't have to setup the metrics, and can check the metrics at model creation:

We might have to group the metrics in a [MetricCollection](https://torchmetrics.readthedocs.io/en/latest/pages/overview.html#metriccollection) for PyTorch Lightning to properly identify them as part of the module.

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -29,10 +30,12 @@ def __init__(
...
+        metrics_params: Optional[List[Dict]] = None,
```
dennisbader, line 34: I believe that the class based metrics (see my comment from above) don't require any additional parameters at metric computation. They are already given at object creation.

If so, we could also remove this parameter

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -230,6 +257,45 @@ def _compute_loss(self, output, target):
...
+                except AttributeError as e:
+                    raise_log(
+                        ValueError(
+                            f"{metric} is not a valid functional metric defined in the torchmetrics.functional module"
+                        )
+                    )
+                    raise e
+
+    def calculate_metrics(self, y, y_hat, tag):
+        metrics = []
```
dennisbader, line 276: If we switch to class based metrics, you can follow [this section on MetricCollection](https://torchmetrics.readthedocs.io/en/latest/pages/overview.html#metriccollection) for logging

dennisbader (changes requested):
Thanks already for another nice idea and the contribution :)

I think this could be a valuable addition for more feedback on model performance during training.

My main suggestion is that we could/should(?) try to base the approach metrics on instances of torchmetrics.Metric. This can potentially reduce complexity on our side such as setting up metrics, need for additional function call parameters, ...

gdevos010:
@hrzn @dennisbader The model now accepts a `TorchMetric` or a `MetricCollection`. I also added the likelihood test and updated the early stop example

hrzn (commented):
This is definitely a very nice addition, thanks a lot!
I just have one last doubt before we can merge. It seems you tested with univariate series only. I guess it might not work out of the box for multivariate series? That is, if `y` and `y_hat` have some dimension `(batch, time, dim, n_params)` with `dim > 1`. In this case, my guess would be that we need to call `metrics()` for each of the `y[:, i, :]` and somehow aggregate/reduce the results. We could take the mean as a default reduction - so what is tracked in those cases is the average "metric" over all components.
It'd be nice to also add a test case for the univariate cases.

gdevos010:
The metrics work with multivariate series. You can calculate regression metrics on multidimensional data as long as they have the same shape. The aggregate/reductions happens in the metric calculation, the same as calculating multiple time steps. Hopefully that explanation makes sense. 

I added test cases for multivariate time series

hrzn (approved):

hrzn:
> The metrics work with multivariate series. You can calculate regression metrics on multidimensional data as long as they have the same shape. The aggregate/reductions happens in the metric calculation, the same as calculating multiple time steps. Hopefully that explanation makes sense.
> 
> I added test cases for multivariate time series

OK perfect, thanks!

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -100,6 +108,23 @@ def __init__(
...
+            except AttributeError:
+                raise_log(
+                    AttributeError(
+                        "torch_metrics only accepts type torchmetrics.Metric or torchmetrics.MetricCollection"
+                    ),
+                    logger,
+                )
+        else:
+            self.train_metrics = torchmetrics.MetricCollection([])
+            self.val_metrics = torchmetrics.MetricCollection([])
```
dennisbader, line 126: Let's avoid try/except with something like below.

```suggestion
        if torch_metrics is None:
            torch_metrics = torchmetrics.MetricCollection([])
        elif isinstance(torch_metrics, torchmetrics.Metric):
            torch_metrics = torchmetrics.MetricCollection([torch_metrics])
        elif isinstance(torch_metrics, torchmetrics.MetricCollection):
            pass
        else:
            raise_log(
                AttributeError(
                    "`torch_metrics` only accepts type torchmetrics.Metric or torchmetrics.MetricCollection"
                ),
                logger,
            )
        self.train_metrics = torch_metrics.clone(prefix="train_")
        self.val_metrics = torch_metrics.clone(prefix="val_")
```

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -226,12 +253,28 @@ def _compute_loss(self, output, target):
         if self.likelihood:
             return self.likelihood.compute_loss(output, target)
         else:
-            # If there's no likelihood, nr_params=1 and we need to squeeze out the
+            # If there's no likelihood, nr_params=1, and we need to squeeze out the
             # last dimension of model output, for properly computing the loss.
             return self.criterion(output.squeeze(dim=-1), target)
 
+    def _calculate_metrics(self, output, target, metrics):
+        if self.likelihood:
```
dennisbader, line 261: let's check if there is a metric before computing anything that is not required (such as sampling)

```suggestion
        if not len(metrics):
            return
        
        if self.likelihood:
            ...
```

dennisbader (changes requested):
Very nice thank you :) Looks great!

I left two suggestions two avoid try/except and avoid computing anything that is not required when there is no metric available (such as sampling).

After these it's ready to merge :) 👍

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -58,6 +62,9 @@ def __init__(
             PyTorch loss function used for training.
             This parameter will be ignored for probabilistic models if the ``likelihood`` parameter is specified.
             Default: ``torch.nn.MSELoss()``.
+        torch_metrics
```
dennisbader, line 65: Could you copy paste the docstring into all our TorchForecastingModels (TFTModel, NBEATSModel, ...)?  Otherwise this will not be shown in the model documentation.

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -76,7 +83,8 @@ def __init__(
         super().__init__()
 
         # save hyper parameters for saving/loading
-        self.save_hyperparameters()
+        # do not save type nn.Module params
+        self.save_hyperparameters(ignore=["loss_fn", "torch_metrics"])
```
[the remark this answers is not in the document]
gdevos010: Pytorch Lightning gives a warning for saving nn.Modules because they are already saved as a part of the model. Adding the ignore prevents them from being saved twice.

darts/models/forecasting/pl_forecasting_module.py
gdevos010: Added

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -226,12 +252,32 @@ def _compute_loss(self, output, target):
...
+            # If there's no likelihood, nr_params=1, and we need to squeeze out the
             # last dimension of model output, for properly computing the loss.
             return self.criterion(output.squeeze(dim=-1), target)
 
+    def _calculate_metrics(self, output, target, metrics):
+        if not len(metrics):
+            return
+        
+        if self.likelihood:
+            ...
```
dennisbader, line 264: Removing the dots will probably make the tests pass
```suggestion
```

dennisbader (approved):

gdevos010:
Are we able to merge this branch?

hrzn:
> Are we able to merge this branch?

Yes. Thanks! Will be part of the next release (in a few days).

PR #999, opened 2022-06-08
fixed verbose overwriting trainer parameters

### LINKED ISSUE #991: `'enable_progress_bar': False` ignored?
**Describe the bug**
`'enable_progress_bar': False` in `pl_trainer_kwargs` does not seem to produce the expected behavior.

**To Reproduce**
```
import numpy as np
import pandas as pd

from darts.models import TCNModel

import warnings
warnings.filterwarnings('ignore')

# Read data:
series = series.astype(np.float32)

# Create training and validation sets:
train, val = series.split_after(pd.Timestamp('19590101'))

# Normalize the time series (note: we avoid fitting the transformer on the validation set)
transformer = Scaler()
train_transformed = transformer.fit_transform(train)
val_transformed = transformer.transform(val)
series_transformed = transformer.transform(series)

model = TCNModel(
    pl_trainer_kwargs={
        'enable_progress_bar': False,
    },
    n_epochs=30,
    input_chunk_length=14,
    output_chunk_length=1,
)

model.fit(
    train_transformed,
    val_series=val_transformed,
)
```

**Expected behavior**
I expected to not see the progress bar, which appears anyway.

 - Python version: 3.10.4
 - darts version: 0.19.0

dennisbader:
Fixes #991.

## Summary
- prevents setting `verbose` from overwriting trainer parameters in TorchForecastingModels

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -514,7 +514,10 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
             )
 
     def _setup_trainer(
-        self, trainer: Optional[pl.Trainer], verbose: Optional[bool] = None, epochs: int = 0
+        self,
+        trainer: Optional[pl.Trainer],
+        verbose: Optional[bool] = None,
+        epochs: int = 0,
     ) -> None:
         """Sets up the PyTorch-Lightning trainer for training or prediction."""
         trainer_params = {key: val for key, val in self.trainer_params.items()}
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -522,13 +522,6 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         """Sets up the PyTorch-Lightning trainer for training or prediction."""
         trainer_params = {key: val for key, val in self.trainer_params.items()}
         if verbose is not None:
-            raise_deprecation_warning(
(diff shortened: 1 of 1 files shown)


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -514,17 +514,28 @@ def _init_model(self, trainer: Optional[pl.Trainer] = None) -> None:
...
     ) -> None:
         """Sets up the PyTorch-Lightning trainer for training or prediction."""
-
-        self.trainer_params["enable_model_summary"] = (
-            verbose if self.model.epochs_trained == 0 else False
-        )
-        self.trainer_params["enable_progress_bar"] = verbose
+        trainer_params = {key: val for key, val in self.trainer_params.items()}
+        if verbose is not None:
+            raise_deprecation_warning(
```
hrzn, line 525: Could this be time to phase out `verbose` in next release maybe?

hrzn (approved):

darts/models/forecasting/torch_forecasting_model.py
dennisbader: I thought about this and personally I actually like the flexibility and simplicity of setting verbose at trainnig/inference time. 

What again was the motivation behind removing it?

hrzn: Actually you're right, I also kind of like it. Perhaps we could un-deprecate it then? :D

dennisbader: Sounds good :)

PR #1000, opened 2022-06-09
HNiTS rename to match model naming convection 

### LINKED ISSUE #947: [BUG] Consistent model naming
**Describe the bug**
NHiTS class should be renamed to NHiTSModel to match naming convention of all other models.

(https://github.com/unit8co/darts/blob/611e4119d4bdd7c42cd20847039fed5ee7b0ff67/darts/models/forecasting/nhits.py#L421-L434)

gdevos010:
Fixes #947.


dennisbader (approved):
Very nice, thanks a lot! 💯

hrzn (commented):
It was actually intentional to not put the `Model` part (we also have other models without this - and when I added NHiTS I thought it is redundant, and somewhat came to regret that we have it on the other models). But if you think there are good reasons to have it, we could leave it I guess.
Was this motivated by any practical concern?

gdevos010:
It's for consistency. When I first started using the model, the name threw me off. They should all have it or none of them

hrzn:
> It's for consistency. When I first started using the model, the name threw me off. They should all have it or none of them

OK, I guess I agree.

hrzn (approved):

Issue #739, opened 2022-01-14
Extending pre-commit functionality and linting 


gdevos010:
**Is your feature request related to a current problem? Please describe.**
no

**Describe proposed solution**
including more git pre-commit hooks can increase code clarity and prevent bugs.

**Describe potential alternatives**
N/A

**Additional context**
- flake8 can enforce code formatting not automatically fixed by black
- the pre-commit-hook repo can fix many basic annoyance
- pyupgrade can update deprecated methods
- mypy is a static typing 
- isort  or reorder_python_imports improves imports

**example**
```
repos:
-   repo: https://github.com/psf/black
    rev: 21.12b0
    hooks:
    -   id: black
        language_version: python3.7
        args: [--safe, "--line-length", "100"]

-   repo: https://github.com/PyCQA/flake8
    rev: 4.0.1
    hooks:
    -   id: flake8
        language_version: python3

-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.0.1
    hooks:
    -   id: check-yaml
    -   id: end-of-file-fixer
    -   id: trailing-whitespace
    -   id: check-added-large-files
    -   id: check-builtin-literals
    -   id: check-json
    -   id: check-shebang-scripts-are-executable
    -   id: check-xml
    -   id: end-of-file-fixer
    -   id: requirements-txt-fixer
    -   id: mixed-line-ending
    -   id: check-ast
    -   id: pretty-format-json
        args: ['--autofix', '--no-sort-keys']

-   repo: https://github.com/asottile/pyupgrade
    rev: v2.29.1
    hooks:
    -   id: pyupgrade
        args: ['--py37-plus']

-    repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.910-1
     hooks:
     -   id: mypy
         args: [--ignore-missing-imports]
         additional_dependencies: [types-PyYAML, types-requests, types-PyMySQL]


-   repo: https://github.com/asottile/reorder_python_imports
    rev: v2.6.0
    hooks:
    -   id: reorder-python-imports
        args: [--py37-plus]

#-   repo: https://github.com/pre-commit/mirrors-isort
#    rev: v5.9.3
#    hooks:
#    -   id: isort
#        args: [-e, "--profile", "black"]
```

I would be willing to work on it but want to know if it would be merged

hrzn:
[...]

PR #1001, opened 2022-06-10
Fix/ updated black to use 120 char limit to match flake8

### LINKED ISSUE #967: [BUG] black is using 80 line length
**Describe the bug**
flake8 has 120 line length but black's default is 
black needs this line added to it `args: ["--line-length", "120"]`
https://github.com/unit8co/darts/blob/155c653a5243af3dd56faad9ffa539bd58c8e050/.pre-commit-config.yaml#L1-L6

It also should be added here 
https://github.com/unit8co/darts/blob/155c653a5243af3dd56faad9ffa539bd58c8e050/build.gradle#L100-L103

@tomasvanpottelbergh Would you be able to add this? I feel like I didn't grasp the two commit setup

gdevos010:
Fixes #967.


hrzn:
80 characters is what's [recommended in PEP-8](https://peps.python.org/pep-0008/#maximum-line-length), shall we change flake8 instead? I believe it'd be one line to change in `setup.cfg` only. WDYT?

gdevos010:
We settled on 120 to match PyTorch when we first added the pre-commit code. It can be a developer preference. Let me know what you think. 

https://github.com/unit8co/darts/issues/739#issuecomment-1016370204

PR #1002, opened 2022-06-10
fix/ updated .git_blame-ignore-revs

### LINKED ISSUE #967: [BUG] black is using 80 line length

gdevos010:
Fixes #967.

- [ ] Add commit hash once #1001 is merged

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -6,3 +6,4 @@
 38cc6712a6f701703074a7a7c82ce0252fe869ee
 # Fix last isort issues and update Black to 22.1.0
 8158d3eaef9d9f6e04f219b029e306d1f1be46d5
+# Update Black to use 120 char line length


tomasvanpottelbergh:
Won't do this for now (see https://github.com/unit8co/darts/issues/967#issuecomment-1156142445)

PR #1003, opened 2022-06-10
Feat/Uber TLC Dataset

gdevos010:
Fixes #.

### Summary

New dataset based on the [FiveThirtyEight](https://github.com/fivethirtyeight/uber-tlc-foil-response) dataset.
Supports both daily and hourly sampling.


darts/datasets/__init__.py
```diff
@@ -511,3 +517,92 @@ def pre_proces_fn(extracted_dir, dataset_path):
...
+    ..
+
+
+    References
+    ----------
+    .. [1] https://github.com/fivethirtyeight/uber-tlc-foil-response
+
+    """
+
+    def __init__(self, sample_freq="hourly"):
```
hrzn, line 566: How about a parameter specifying whether we want this dataset to return a single multivariate series, or multiple series (could be a multivariate one by default). That would be nice and also would remove the need to copy/paste some of the code from the docstring example in case someone needs the multiple univariate option.
We could even have such a parameter for all the multivariate series we currently have - I guess (hope? :D) that this would be simple and adding this parameter to the metadata parameters and adapting the DatasetLoaderCSV class with a code similar the one that you put in the docstring could be enough.

hrzn (commented):
Once more, a very nice addition, thanks a lot!
I just have a comment concerning leaving it as an option to return a multivariate series, or multiple series. Let me know if you think that's a good idea.

darts/datasets/__init__.py
gdevos010: Let me know what you think of the implementation. If it looks good, I will add it to the other series

darts/datasets/__init__.py
```diff
@@ -511,3 +519,96 @@ def pre_proces_fn(extracted_dir, dataset_path):
...
+    """
+
+    def __init__(self, sample_freq: str = "hourly", multivariate: bool = True):
+        """
+        Parameters
+        ----------
+        sample_freq: str
+            The sampling frequency of the data. Can be "hourly" or "daily". Default is "hourly".
+        multivariate: bool
+            Whether to return a multivariate timeseries or a list of univariate timeseries. Default is True.
```
hrzn, line 548:
```suggestion
            Whether to return a single multivariate timeseries - if False returns a list of univariate TimeSeries. Default is True.
```

hrzn (approved):
Cool, only one minor suggestion left then we can merge :)

gdevos010:
I added the functionality to the electricity dataset too

darts/datasets/__init__.py
hrzn: Yes, looks very good.

gdevos010:
Is something else needed, or can we merge this branch?

hrzn:
> Is something else needed, or can we merge this branch?

I think we can merge soon. The test coverage seems to have decreased. Are you able to have a look why?

hrzn:
> @hrzn It's probably the `_to_multi_series`. I can add a test for them, but the Electricity's `_to_multi_series` is **very very** slow.

You could maybe add a test that calls `_to_multi_series` with a small DataFrame, and just have a sanity check that what the function returns has the right format.

darts/tests/datasets/test_dataset_loaders.py
```diff
@@ -97,6 +97,15 @@
         pre_process_zipped_csv_fn=None,
     )
 )
+ele_multi_series_dataset = DatasetLoaderCSV(
+    metadata=DatasetLoaderMetadata(
+        "Electricity_test.csv",
+        uri=_DEFAULT_PATH + "/Electricity_test.csv",
```
hrzn, line 103: Could we maybe put this one in `datasets/test/Electricity_test.csv` ? i.e., we create a test subfolder.

hrzn:
Thanks @gdevos010. I assume the tests should pass once it is merged, right? The only last tiny thing I'd ask is to put the test dataset in a subfolder `datasets/test/`.

gdevos010:
@hrzn Yes, the tests passed locally. I moved the file to `tests/datasets`.

gdevos010:
This will be my last PR for about a month. I will be traveling without a laptop.

hrzn:
> @hrzn Yes, the tests passed locally. I moved the file to `tests/datasets`.

Thanks! I was referring to `datasets/Electricity_test.csv` though, which I think is still living under `datasets/`. But it's fine, I will merge the PR and we can change it later on.

hrzn:
> This will be my last PR for about a month. I will be traveling without a laptop.

Enjoy :)

Issue #927, opened 2022-04-21
Remove the warnings upfront when importing the light-weight versions


aaronsl-hku:
**Is your feature request related to a current problem? Please describe.**
Yes. Currently using the "u8darts[torch]" version (for all the chaos for installing prophet but that's another story), every time I run my program with darts imported, I would received the warnings from [darts.models.__init__.py](https://github.com/unit8co/darts/blob/master/darts/models/__init__.py).  

It would not be a problem if the logger is instantiated directly from `logging` module, but with levels, format and `addHandler` hardcoded in `darts.logging.get_logger()` there's basically nothing I can do.

**Describe proposed solution**
The best solution would be completely replace the `darts.logging.get_logger()` function with standard `logging.getLogger()` (unless there're any very special reasons).

**Describe potential alternatives**
- Do not add handlers other than `NullHandler` as mentioned [here](https://docs.python.org/3/howto/logging.html#library-config)
- Raise the warnings in a lazy manner, if that is plausible.

brunnedu:
Hi @aaronsl-hku, we would really like to help you resolve this issue. At the same time, we think that it is important to inform our users about missing modules. 

Therefore we would suggest giving you the option to opt-out of those warnings. Would it be possible for you to set an environment variable such as `DISABLE_DARTS_WARNINGS` in your terminal? We could then use this environment variable to adjust the logging behavior of darts.

aaronsl-hku:
Hello @brunnedu . Nice to receive your reply.

We might have some misunderstanding here: I am **not** looking into disable all warnings as a whole. What I am proposing is not to hardcode the handler and formatter into the `get_logger` function. 

Long story short, if my project has a logging config and the formatting is not the same darts, my log would be easily screwed by the hardcoded formatter.

May I know if there is any particular reason not to use the standard logging.get_logger as-is?

> Hi @aaronsl-hku, we would really like to help you resolve this issue. At the same time, we think that it is important to inform our users about missing modules. 
> 
> 
> 
> Therefore we would suggest giving you the option to opt-out of those warnings. Would it be possible for you to set an environment variable such as `DISABLE_DARTS_WARNINGS` in your terminal? We could then use this environment variable to adjust the logging behavior of darts.

martinb-bb:
> @martinb-bb WDYT of this solution?

I think this could work.

brunnedu:
Thanks for your suggestions, everyone! 

[...]

PR #1010, opened 2022-06-16
Opt-out of darts logging based on environment variable

brunnedu:
### OUTDATED, SEE BELOW FOR UPDATE

Addresses #927 

### Summary

- Implemented option to opt-out of darts logging based on the environment variable `DISABLE_DARTS_LOGGING`
- If `DISABLE_DARTS_LOGGING` is set to the string `"1"`, the `StreamHandler` is replaced with the `NullHandler` in the `darts.logging.get_logger()` function.

### Other Information

I'm open to any suggestions on how we could handle this better.


brunnedu:
Addresses #927 

### Summary

After some helpful suggestions from the community, we have now decided that the best solution is to stop adding custom handlers or formatters to our loggers.

- `darts.logging.get_logger(name)` is now a simple wrapper of `logging.getLogger(name)`
- `darts.logging.get_logger()` sets the level of newly created loggers to `INFO` which is the default logging level of Darts

hrzn (approved):
Thanks

brunnedu:
@martinb-bb Unfortunately using `logging.basicConfig()` wouldn't work, because after creating a new logger in `get_logger()` we set its level to `INFO` overwriting whatever level was set in `logging.basicConfig()`. Because of that we now decided to not use `logger.setLevel(logging.INFO)` anymore (#1034) which would then allow you to disable the warnings using:
```python
import logging
# set level to ERROR, not logging levels WARNING and below anymore
logging.basicConfig(level=logging.ERROR)
import darts
```
Another option that works in either case would be to use:
```python
import contextlib
import os
with open(os.devnull,'w') as handle:
    with contextlib.redirect_stderr(handle):
        import darts
```

PR #1011, opened 2022-06-16
Add torchmetric dependency

hrzn:
@gdevos010 could you have a quick look?


gdevos010:
Yes, that looks good to me. If you look at the pip output for the passing tests, torch metrics is a dependency of PyTorch lightning so I'm not why its not on main.

gdevos010 (approved):

PR #1023, opened 2022-06-21
Add Python 3.10 & Remove prophet dependency

hrzn:


tomasvanpottelbergh:
I think you missed some references in `requirements/dev-all.txt` and `build.gradle`

PR #1040, opened 2022-06-26
Adds drop_columns method to TimeSeries

### LINKED ISSUE #880: Adding of a TimeSeries .drop() method
I haven't found in the TimeSeries documentation anything similar to a pd.DataFrame .drop() method and I think it would be compfy to have such.
Surely It ait't something critical, but in order to remove some componentes from multivariate TimeSeries, would be more comfy to my opinion.

At the moment I worked around it using `set()` differences between the set of components of the TimeSeries and the compinent I want to remove:
```ts[list(set(ts.columns) - set(['stop_time']))] ```

I hope not to be wasting your time if such operation is already implemented through slicing operations `[ ]`

shaido987:
Fixes #880.

### Summary

Adds a function to drop one or multiple columns to the `TimeSeries`. I tried follow the format and style of the other `TimeSeries` methods and also added a test to the `tests/test_timeseries_multivariate` file (since it does not make sense to drop columns in a single column time series).

### DIFF
--- a/darts/tests/test_timeseries_multivariate.py
+++ b/darts/tests/test_timeseries_multivariate.py
@@ -243,3 +243,15 @@ class TimeSeriesMultivariateTestCase(DartsBaseTestClass):
         self.assertEqual(
             self.series3.univariate_component(1).last_values().tolist(), [20]
         )
+
+    def test_drop_column(self):
+        # testing dropping a single column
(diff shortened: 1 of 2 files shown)


hrzn (approved):
LGTM, thanks!

PR #1053, opened 2022-07-01
speedup static covariates when no casting is needed

hrzn:
Static covariates of very high dimensionality can cause performance issues due to casting
This PR avoids the casting where/when possible.


dennisbader (approved):
Looks good, thanks 👍

PR #1054, opened 2022-07-01
reintegrate prophet in core deps

hrzn:
An attempt to (maybe) reintegrate Prophet (with version 1.1)


tomasvanpottelbergh (approved):
Looks good! We shouldn't forget to update the docs at the next release.

hrzn:
I ran some tests with conda envs, Py39:

* Base environment size: 114 MB
* Size of u8darts core install, without Prophet: 984 MB
* Size of u8darts core install, with prophet==1.1: 1019 MB
* Size of u8darts pmdarima install, with prophet==1.1: 1022 MB
* Size of full install without Prophet: 1.5 GB
* Size of full install with prophet==1.1: 1.5 GB

Other observation: pmdarima depends on Cython, which is also a dependency of Prophet. [It doesn't seem to have any other potentially problematic dependency](https://github.com/alkaline-ml/pmdarima/blob/master/requirements.txt), and installed without issue with pip in my tests.

Conclusions:
* It's safe to re-integrate Prophet as its size doesn't inflate that of Darts too much
* It's probably safe to drop our pmdarima flavour and just put pmdarima in the core dependencies, and maintain only one flavour for PyTorch. @tomasvanpottelbergh thoughts?

PR #1056, opened 2022-07-04
Update coloring for graphs

colin99d:
### Summary

Currently the charting in Darts messes up if you are using custom settings in matplotlib. I am creating a default colors argument so that users can choose whether they want Dart's default colors. If this is a desired feature I can add it to the rest of the charting functions in this file, and any other files you point out.


colin99d:
Thanks for the feedback. All requested changes have been implemented.

hrzn (approved):
Thanks!

Issue #936, opened 2022-05-02
Add generic save/load functions to all darts models


sotte:
**Is your feature request related to a current problem? Please describe.**
`darts` offers a nice interface to train/evaluate models independent of the implementation (`statsmodels`, `torch`, `prophet`, etc.), but it does not offer the same interface for serializing/deserializing models. It would be great if darts models offered generic save/load functions.

Right now only (?) the `pyroch` models have the save/load feature, other models mostly rely on pickling the models.

**Describe proposed solution**
`darts` implements save/load for each model framework that it supports.

* use the current implementation for pytorch models
* joblib pickle for the sklearn like models ([details](https://scikit-learn.org/stable/modules/model_persistence.html))
* pickle for statsmodels (?)
* etc.

**Additional context**
This feature would make it easier to add `darts` to framework independent model serving tools like `bentoml`, see https://github.com/bentoml/BentoML/issues/2368 for details.

hrzn:
+1
Thanks for raising! Adding to our backlog :)
Would you want to contribute something around this?

sotte:
As usual there are time constraints regarding contributions :(

Also right now this issue is not well defined yet. Do we just want to use existing serialization formats? Do we want to define wrap existing formats into a darts format (just a fancy name, but maybe all serialized models look like `model.darts` instead of framework specific names)? Do we want to save additional meta data? Do we want to create just one file or a folder?
Do you @hrzn and your folks have any opinions on this?

hrzn:
@sotte https://github.com/unit8co/darts/pull/1070 introduced generic `save()` and `load()` methods for all forecasting models. PyTorch models rely on `torch.save()` and other models rely on pickle by default. Hopefully we're in a better state than before already.

PR #1070, opened 2022-07-12
Feat/generic save load

brunnedu:
Addresses #936

First version of generic save and load methods for `ForecastingModels`

For now, we have decided to implement generic save and load methods for all `ForecastingModels` using joblib, as long as there is no better alternative for a particular model subclass, as is currently the case only for `TorchForecastingModels`. We decided to take this approach because most libraries either suggest doing it this way ([sklearn](https://scikit-learn.org/stable/model_persistence.html)) or wrap a very similar solution themselves ([statsmodels](https://www.statsmodels.org/dev/_modules/statsmodels/iolib/smpickle.html)). 

There are obvious downsides to this solution regarding maintainability and security such as the following ones:
- `ForecastingModels` saved in one version of darts might not be loadable in another version of darts or they could exhibit unexpected behavior.
- One should never call `ForecastingModel.load_model()` on untrusted data as it could lead to malicious code being executed upon loading.

I'm open to hearing your opinion on the current implementation and the additional functionality you believe those methods should contain.


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1278,20 +1278,18 @@ def save_model(self, path: str) -> None:
...
 
         # We save the whole object to keep track of everything
         with open(path, "wb") as f_out:
             torch.save(self, f_out)
 
         # In addition, we need to use PTL save_checkpoint() to properly save the trainer and model
         if self.trainer is not None:
-            base_path = path[:-8]
-            path_ptl_ckpt = base_path + "_ptl-ckpt.pth.tar"
+            base_path = path.split(".")[0]
```
hrzn, line 1291: Shouldn't you set aside the last token instead? I.e. make it work if the provided path is e.g. `my.model.v.0.pt`

darts/models/forecasting/forecasting_model.py
```diff
@@ -12,14 +12,17 @@
...
+import os
 import time
 from abc import ABC, ABCMeta, abstractmethod
 from collections import OrderedDict
 from itertools import product
 from random import sample
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, BinaryIO, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
+import joblib
```
hrzn, line 25: Do joblib or BinaryIO require new dependencies?

darts/models/forecasting/forecasting_model.py
```diff
@@ -896,6 +899,63 @@ def model_params(self) -> dict:
             self._model_params if hasattr(self, "_model_params") else self._model_call
         )
 
+    def save_model(
```
hrzn, line 902: I'd have preferred to call them `save()` and `load()`, but maybe it is too late...
We could perhaps deprecate these names with a warning message, and mention in the future only `save()` and `load()` will be available. WDYT?

darts/models/forecasting/forecasting_model.py
```diff
@@ -896,6 +899,63 @@ def model_params(self) -> dict:
...
+        ----------
+        path
+            Path under which to save the model at its current state.
+        joblib_kwargs
+            Keyword arguments passed to `joblib.dump()`
+        """
+
+        if path is None:
+            # default path
+            path = f"{type(self).__name__}_{datetime.now().strftime('%Y-%m-%d-%H-%M-%S')}.pkl"
```
hrzn, line 917: You could factor out a function giving you default names.

darts/models/forecasting/forecasting_model.py
```diff
@@ -896,6 +899,63 @@ def model_params(self) -> dict:
...
+
+        if isinstance(path, str):
+            raise_if_not(
+                os.path.isdir(os.path.dirname(path)),
+                f"The dictionary {os.path.dirname(path)} doesn't exist",
```
hrzn, line 922:
```suggestion
                f"The directory {os.path.dirname(path)} doesn't exist",
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -896,6 +899,63 @@ def model_params(self) -> dict:
...
+        Parameters
+        ----------
+        path
+            Path from which to load the model.
+        """
+
+        if isinstance(path, str):
+            raise_if_not(
+                os.path.exists(path),
+                f"The path {path} doesn't exist",
```
hrzn, line 946:
```suggestion
                f"The file {path} doesn't exist",
```

darts/models/forecasting/forecasting_model.py
brunnedu: no, joblib is a dependency of sklearn and BinaryIO is part of the typing module.

darts/models/forecasting/forecasting_model.py
brunnedu: Yeah, I'd be in favor of doing this. I think as long as it is only models that we can call these methods on we don't need the "_model" suffix.

brunnedu:
Yeah this sounds like a good idea, I'll make a few trials 👍

hrzn:
> Yeah this sounds like a good idea, I'll make a few trials 👍

I don't have a strong opinion so feel also free to leave joblib.

brunnedu:
Did a few comparisons between `pickle` and `joblib` but wasn't able to tell a difference, so I think we should use `pickle` for now.

hrzn:
@brunnedu what's the status of this PR? Anything more remaining to be done?

darts/models/forecasting/forecasting_model.py
```diff
@@ -908,6 +911,55 @@ def model_params(self) -> dict:
...
+                os.path.exists(path),
+                f"The file {path} doesn't exist",
+                logger,
+            )
+
+            # Load the object using joblib
+            with open(path, "rb") as handle:
+                model = pickle.load(file=handle)
+        else:
+
```
hrzn, line 958:
```suggestion
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1300,6 +1300,36 @@ def save_model(self, path: str) -> None:
...
+            torch.save(self, f_out)
+
+        # In addition, we need to use PTL save_checkpoint() to properly save the trainer and model
+        if self.trainer is not None:
+            path_ptl_ckpt = path + ".ckpt"
+            self.trainer.save_checkpoint(path_ptl_ckpt)
+
+    def save_model(self, path: str) -> None:
+        """
+        `TorchForecastingModel.save_model()` is deprecated and will be removed in a future darts version.
```
hrzn, line 1318: You can make this look nicer using a numpydoc deprecationwarning, something like this:
```
.. deprecated:: 0.21.0
          `TorchForecastingModel.save_model()` is deprecated and will be removed in a future darts version.
          Please use `TorchForecastingModel.save()` instead.
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1300,6 +1300,36 @@ def save_model(self, path: str) -> None:
         # TODO: the parameters are saved twice currently, once with complete
         # object, and once with PTL checkpointing.
 
+        if path is None:
+            # default path
+            path = f"{type(self).__name__}_{datetime.datetime.now().strftime('%Y-%m-%d-%H-%M-%S')}.pt"
```
hrzn, line 1305: Could we maybe extract this out in a `ForecastingModel.get_default_save_path()` method?
This would avoid having this default convention written twice (in ForecastingModel and TorchForecastingModel).

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1300,6 +1300,36 @@ def save_model(self, path: str) -> None:
...
+        ----------
+        path
+            Path under which to save the model at its current state.
+        """
+
+        save_warning = """`TorchForecastingModel.save_model()` is deprecated and will be removed in a future darts version.
+        Please use `TorchForecastingModel.save()` instead."""
+        raise_deprecation_warning(save_warning, logger)
+
         raise_if_not(
```
hrzn, line 1333: Don't you want to call `self.save()` instead of running the code below?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1327,7 +1357,42 @@ def load_model(path: str) -> "TorchForecastingModel":
...
                 from darts.models import RNNModel
 
-                model_loaded = RNNModel.load_model("my_model.pth.tar")
+                model_loaded = RNNModel.load("my_model.pth.tar")
+            ..
+
+        Parameters
+        ----------
+        path
+            Path under which to save the model at its current state. The path should end with '.pth.tar'
```
hrzn, line 1366: Do we want to keep enforcing that the path ends with `.pth.tar`? The default save path does not seem to satisfy this.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1327,7 +1357,42 @@ def load_model(path: str) -> "TorchForecastingModel":
...
+        if os.path.exists(path_ptl_ckpt):
+            model.model = model.model.__class__.load_from_checkpoint(path_ptl_ckpt)
+            model.trainer = None
+
+        return model
+
+    @staticmethod
+    def load_model(path: str) -> "TorchForecastingModel":
+        """
+        `TorchForecastingModel.load_model()` is deprecated and will be removed in a future darts version.
```
hrzn, line 1383: Same here, a `deprecationwarning` would look nicer in the compiled docs.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1336,6 +1401,12 @@ def load_model(path: str) -> "TorchForecastingModel":
             Path under which to save the model at its current state. The path should end with '.pth.tar'
         """
 
+        load_warning = (
+            "`TorchForecastingModel.load_model()` is deprecated and will be removed in a future darts "
+            "version. Please use `TorchForecastingModel.load()` instead."
+        )
+        raise_deprecation_warning(load_warning, logger)
+
         raise_if_not(
```
hrzn, line 1410: Same comment here, don't you want to replace the code below by a call to `self.load()` ?

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -164,6 +173,45 @@ def test_save_model_parameters(self):
...
+                # test save
+                model.save()
+                model.save(model_path)
+
+                self.assertTrue(os.path.exists(full_model_path))
+
+                # test load
+                loaded_model = type(model).load(model_path)
+
+                self.assertEqual(
```
hrzn, line 209: Nice 👍

hrzn (commented):
Looks good! Only a few minor comments :)

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -164,6 +173,45 @@ def test_save_model_parameters(self):
...
+                forecasting_horizon = 4
+                model_path = type(model).__name__
+
+                full_model_path = os.path.join(self.temp_work_dir, model_path)
+
+                model.fit(self.ts_pass_train)
+                model_prediction = model.predict(forecasting_horizon)
+
+                # test save
+                model.save()
```
dennisbader, line 201: we might want to check here as well that the file is saved at the desired location when not giving a path.
You can see everything that was saved with `os.walk(self.temp_work_dir)`.

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -97,13 +101,56 @@ class LocalForecastingModelsTestCase(DartsBaseTestClass):
...
+            model_prediction = model.predict(self.forecasting_horizon)
+
+            # test save
+            model.save()
+            model.save(model_path)
+            with open(model_path_f, "wb") as f:
+                model.save(f)
+
+            for full_model_p in full_model_paths:
+                self.assertTrue(os.path.exists(full_model_p))
```
dennisbader, line 140: Same comment as for TorchForecastingModels

darts/models/forecasting/torch_forecasting_model.py
dennisbader: I think Dustin had a look at this and found that `.pth.tar` is more of a relic. It looks like `.pt` is now recommended. With the current code I don't think we have to enforce any file ending and just append `.ckpt` to the checkpoint file.

The parameter description needs to be updated though :)

dennisbader (commented):
Looks good to me as well :) Thanks for this, finally we have built in saving and loading for all models!
Just had a few minor comments/additions to @hrzn's comments

darts/models/forecasting/torch_forecasting_model.py
brunnedu: Sounds like a good idea. I didn't do it because of the different filename extensions (.pt and .pkl) but now I'll just add the filename extensions separately and use `ForecastingModel.get_default_save_path()` for the main part.

darts/models/forecasting/torch_forecasting_model.py
hrzn: OK, what you suggest sounds good 👍

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1289,71 +1289,117 @@ def _batch_collate_fn(batch: List[Tuple]) -> Tuple:
...
 
-    def save_model(self, path: str) -> None:
-        """Saves the model under a given path. The path should end with '.pth.tar'
+    def save(self, path: Optional[str] = None) -> None:
+        """
+        Saves the model under a given path.
+
+        Creates two files under ``path`` (model object) and ``path``.ckpt (checkpoint).
+
+        Example for saving and loading a :class:`RNNModel`:
```
hrzn, line 1298: 💯

hrzn (approved):
Awesome 🚀

Issue #1075, opened 2022-07-14
Static Covariates after from fill_missing_values


nydaniel:
When using the function fill_missing_values the returned TimeSeries does not have the static covariates of the original TimeSeries

```
values = np.array([0, 1, 2, np.nan])
static_series = pd.Series([10, 20, 30], index=['a', 'b', 'c'])
ts = darts.TimeSeries.from_values(values, static_covariates=static_series)
print(ts.has_static_covariates)
new_ts = fill_missing_values(ts, fill=0.0)
print(new_ts.has_static_covariates)
```

This code outputs `True` for the first print statement and `False` for the second print statement.

nydaniel:
Happy to help and thanks a lot for the development effort

hrzn:
This one was fixed in https://github.com/unit8co/darts/pull/1076 and released.

PR #1076, opened 2022-07-14
Propagate static covs and hierarchy in missing value filler

hrzn:
Solves #1075


dennisbader (approved):
Looks great, thanks for quickly fixing these issues with static covs :)

PR #1106, opened 2022-07-31
Fix/fix various issues

hrzn:
* Fixes a recent issue with xarray 2022-06, which changed `reset_index()` behaviour (see https://github.com/pydata/xarray/issues/6607). We now re-build the DataArrays from scratch with proper coordinates instead of relying on `reset_index()`.

* Fixes an issue due to the latest release of StatsForecast, which contained a breaking change, by fixing the version number.

* Small improvements to TimeSeries user guide.


dennisbader (approved):
Thanks a lot for fixing these issues! Looks good 👍

Issue #1101, opened 2022-07-28
Implement min_train_series_length() method for the FourTheta model


rijkvandermeulen:
**Is your feature request related to a current problem? Please describe.**
The FourTheta model extends the ForecastingModel class which has a class property defining the minimum required length for the training series (min_train_series_length). This property defautlts to 3 but can be overriden for each child class inherenting from ForecastingModel . Now, I'm missing the method that overrides this min_train_series_length in the case of the FourTheta model. I'm pretty sure the Theta model requires at least 2 * seasonality_period in case seasonality is used (i.e. season_mode != SeasonalityMode.NONE).

When you currently try to fit a seasonal FourTheta model when you have < 2*seasonality_period observations, it errors out when it tries to do a seasonal_decomposition in extract_trend_and_seasonality(). Would it be an idea to already capture this in an earlier stage by implementing the min_train_series_length() method 

**Describe proposed solution**
Implement min_train_series_length() method for the FourTheta model overwriting the default of 3. Something like: 

`
        if self.season_mode.value:
            return self.seasonality_period * 2
        else:
            return 3
`
Might be a bit more nuance to the abovementioned solution since seasonality_period is an optional param and can also be retrieved from the time series. Hence, if seasonality_period=None we need to first retrieve the seasonality_period.

**Describe potential alternatives**

**Additional context**

hrzn:
Thanks, that makes sense. Would you be willing to open a PR with this change?

rijkvandermeulen:
Sure, let me give that a try!

PR #1111, opened 2022-08-01
#1101 Implemented min_train_series_length for Theta and FourTheta

rijkvandermeulen:
This PR implements the min_train_series_length() method for the FourTheta and Theta models that overwrites the minimum default of 3 training samples by 2*seasonal_period when appropriate.

### Summary
When seasonality is used in the Theta and FourTheta models the minimum number of training samples should be 2*seasonal_period. Since the min_train_series_length() method wasn't implemented for the theta models, however, it always defaulted to 3 (for seasonal as well as non-seasonal models).

### Other Information
Note that there is one case that the newly implemented min_train_series_length() method does not address: namely when season_mode != SeasonalityMode.NONE and seasonality_period==None. In this case, for us to be able to determine the minimum number of training samples, we need to first retrieve the seasonality_period from the time series. We can actually do this (for example by using the check_seasonality() function in utils.statistics). However, I wasn't entirely sure whether this would be desirable. The issue that I saw was that you lose your strict decoupling of models and time series. I.e., you cannot retrieve the min_train_series_length property from the model alone anymore - you'd have to pass the time series itself as well. Another possibility would be to make the seasonal_period a mandatory argument (probably not desired though I guess).

Looking forward to hearing your thoughts and feedback

Best,
Rijk


darts/tests/models/forecasting/test_4theta.py
```diff
@@ -76,3 +76,51 @@ def test_best_model(self):
         self.assertTrue(
             mape(val_series, forecast_best) <= mape(val_series, forecast_random)
         )
+
+    def test_min_train_series_length_with_seasonality(self):
```
hrzn, line 80: Thank you for adding unit tests :)

hrzn (approved):
LGTM, thanks!
Concerning the adaptation when the seasonality is not known beforehand, I agree with your observation that we should leave "data" (time series) and models decoupled. So I think we can live with that for now.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Darts is still in an early development phase and we cannot always guarantee back
 - An issue with arguments being reverted for the `metric` function of gridsearch and backtest [#989](https://github.com/unit8co/darts/pull/989) by [Clara Grotehans](https://github.com/ClaraGrthns).
 - An error checking whether `fit()` has been called in global models [#944](https://github.com/unit8co/darts/pull/944) by [Julien Herzen](https://github.com/hrzn).
 - An error in Gaussian Process filter happening with newer versions of sklearn [#963](https://github.com/unit8co/darts/pull/963) by [Julien Herzen](https://github.com/hrzn).
+- Implemented the min_train_series_length method for the FourTheta and Theta models that overwrites the minimum default of 3 training samples by 2*seasonal_period when appropriate [#1101](https://github.com/unit8co/darts/pull/1101) by [Rijk van der Meulen](https://github.com/rijkvandermeulen)
```
hrzn, line 37: @rijkvandermeulen here too, could you move this to the `Unreleased` section? Thanks.

rijkvandermeulen: Jup - done!

PR #1124, opened 2022-08-06
reset ptl trainer when loading torch models

### LINKED ISSUE #1116: Unable to load a saved pytorch model due to trainer being required [BUG]
**Describe the bug**

When trying to run `NBEATSModel.load_model` on a file I saved using `save_model`, I get the following error:

```
RuntimeError: _NBEATSModule is not attached to a `Trainer`.
```

**To Reproduce**

```python
model = NBEATSModel(**my_params)
model.fit(my_data)
model.save('my_path.pth.tar')
NBEATSModel.load_model('my_path.pth.tar')
```

**Expected behavior**

The model should get successfully loaded and not raise any errors.

 - Python version: 3.9.13
 - darts version 0.20.0
 - MacOS 12.4

**Additional information**

Relevant part of the traceback:

```
  File "<PYTHON>/site-packages/darts/models/forecasting/torch_forecasting_model.py", line 1331, in load_model
    model.trainer = model.model.trainer
  File "<PYTHON>/site-packages/pytorch_lightning/core/module.py", line 179, in trainer
    raise RuntimeError(f"{self.__class__.__qualname__} is not attached to a `Trainer`.")
```

dennisbader:
Fixes #1116.

### Summary

- Fixes error when loading TorchForecastingModels where trainer is not associated with a model.
- fixed issue where custom feed forward moduels were ignored in previous PyTorch version

### DIFF
--- a/darts/models/forecasting/transformer_model.py
+++ b/darts/models/forecasting/transformer_model.py
@@ -9,7 +9,7 @@ from typing import Optional, Tuple
 import torch
 import torch.nn as nn
 
-from darts.logging import get_logger, raise_if_not
+from darts.logging import get_logger, raise_if, raise_if_not
 from darts.models.components import glu_variants
 from darts.models.components.glu_variants import GLU_FFN
 from darts.models.forecasting.pl_forecasting_module import PLPastCovariatesModule
@@ -20,6 +20,20 @@ logger = get_logger(__name__)
 
 BUILT_IN = ["relu", "gelu"]
 FFN = GLU_FFN + BUILT_IN
+
+
+class _CustomFeedForwardEncoderLayer(nn.TransformerEncoderLayer):
+    # overwrite the feed forward block
+    def _ff_block(self, x):
+        x = self.activation(x)
+        return self.dropout2(x)
+
+
+class _CustomFeedForwardDecoderLayer(nn.TransformerDecoderLayer):
+    # overwrite the feed forward block
+    def _ff_block(self, x):
+        x = self.activation(x)
+        return self.dropout2(x)
 
 
 # This implementation of positional encoding is taken from the PyTorch documentation:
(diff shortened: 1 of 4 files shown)


darts/models/forecasting/transformer_model.py
```diff
@@ -22,6 +22,20 @@
 FFN = GLU_FFN + BUILT_IN
 
 
+class _CustomFeedForwardEncoderLayer(nn.TransformerEncoderLayer):
+    # overwrite the feed forward block
+    def _ff_block(self, x):
+        x = self.activation(x)
+        return self.dropout2(x)
```
gdevos010, line 29: The glu variants have dropout built-in

darts/models/forecasting/transformer_model.py
```diff
@@ -142,10 +156,43 @@ def __init__(
...
+                custom_encoder is not None or custom_decoder is not None,
+                "Cannot use `custom_encoder` or `custom_decoder` along with an `activation` from "
+                f"{GLU_FFN}",
+                logger=logger,
+            )
             # use glu variant feedforward layers
             self.activation = getattr(glu_variants, activation)(
                 d_model=d_model, d_ff=dim_feedforward, dropout=dropout
             )
+            encoder_layer = _CustomFeedForwardEncoderLayer(
```
gdevos010, line 169: the glu variants are full feedforward layers. It wasn't fully correct before to set the activation = glu_variants. 
For this to be fully correct, I think there should be a separate arg for ff_bock and leave activation alone. 

This customFeedForwardEncoderLayer makes more sense.

darts/models/forecasting/transformer_model.py
```diff
@@ -142,10 +156,43 @@ def __init__(
...
+                d_model=d_model,
+                nhead=nhead,
+                dim_feedforward=dim_feedforward,
+                dropout=dropout,
+                activation=self.activation,
+            )
+            encoder_norm = nn.LayerNorm(d_model)
```
gdevos010, line 176: I can add the new norms when this gets merged

darts/models/forecasting/transformer_model.py
dennisbader: Yes, but I think we should add the second dropout following the original implementation (see [here](https://github.com/pytorch/pytorch/blob/a61c96492b042947467bb603244301400f0e8d1a/torch/nn/modules/transformer.py#L536)):

```
    # feed forward block
    def _ff_block(self, x: Tensor) -> Tensor:
        x = self.linear2(self.dropout(self.activation(self.linear1(x))))
        return self.dropout2(x)
```

then per TransformerEncoderLayer:
```
    def forward(...):
        x = self.norm1(x + self._sa_block(x, src_mask, src_key_padding_mask))
        x = self.norm2(x + self._ff_block(x))
    return x
```

This is equivalent to what they do in the [Annotated Transfomer](https://nlp.seas.harvard.edu/2018/04/03/attention.html#encoder-and-decoder-stacks) with SublayerConnection, EncoderLayer and Position-wise Feed-Forward Networks.

Our current FeedForward class does (changed a bit so it's easier to compare):
```
    def forward(self, x)
        x = self.linear2(self.dropout(self.activation(self.linear1(x))))
        return x
```
So when we overwrite the PyTorch TransformerEncoderLayer, the last dropout would be missing.

hrzn (approved):
LGTM (though I didn't check the correctness in details)
Thanks!

PR #1126, opened 2022-08-07
fix typo in timeseries

### LINKED ISSUE #1115: Just a small typo
https://github.com/unit8co/darts/blob/caccce1dc71f58180142d96e70bf06024d68798f/darts/timeseries.py#L652

It would be 

> ValueError: If time_col is not specified, the DataFrame must be indexed either witha DatetimeIndex, or with a RangeIndex.

, which `witha` should be `with a`.

blacksnail789521:
Fixes #1115.

### Summary

https://github.com/unit8co/darts/blob/caccce1dc71f58180142d96e70bf06024d68798f/darts/timeseries.py#L652

It would be 

> ValueError: If time_col is not specified, the DataFrame must be indexed either witha DatetimeIndex, or with a RangeIndex.

, which `witha` should be `with a`.

### Other Information

None

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -649,7 +649,7 @@ class TimeSeries:
         else:
             raise_if_not(
                 isinstance(df.index, VALID_INDEX_TYPES),
-                "If time_col is not specified, the DataFrame must be indexed either with"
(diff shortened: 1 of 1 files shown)


hrzn (approved):
Thanks

PR #1128, opened 2022-08-07
Fix/filter static cov component

hrzn:
`TimeSeries.univariate_component()` was broken when the series contained static covariates, this fixes it (and simplifies the method).

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -2702,6 +2702,9 @@ class TimeSeries:
         Retrieve one of the components of the series
         and return it as new univariate ``TimeSeries`` instance.
 
+        This drops the hierarchy (if any), and retains only the relevant static
+        covariates column.
+
         Parameters
         ----------
         index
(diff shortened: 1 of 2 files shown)


darts/timeseries.py
```diff
@@ -2713,11 +2716,10 @@ def univariate_component(self, index: Union[str, int]) -> "TimeSeries":
...
         """
-        if isinstance(index, int):
-            new_xa = self._xa.isel(component=index).expand_dims(DIMS[1], axis=1)
-        else:
-            new_xa = self._xa.sel(component=index).expand_dims(DIMS[1], axis=1)
-        return self.__class__(new_xa)
+
+        return self.__getitem__(
+            index if isinstance(index, str) else self.components[index]
+        )
```
dennisbader, line 2722: in other places we use `[]` directly
```suggestion
        return self[index if isinstance(index, str) else self.components[index]]
```

dennisbader (approved):
Thanks, looks good! Added a minor comment

darts/timeseries.py
hrzn: good point, thanks :)

PR #1131, opened 2022-08-07
Query Selector Model

### LINKED ISSUE #1120: Query Selector
Add the [Query Selector](https://github.com/moraieu/query-selector) model.

I have a working non-probabilistic version working. Working on making it probabilistic.

gdevos010:
Fixes #1120.

### Summary
This model is a direct adapted from the author's [repo](https://github.com/moraieu/query-selector
).
There are also a few more minor documentation fixes mostly recommended by PyCharm.

### Other Information

- [x] convert to probabilistic 
- [ ] more model documentation
- [ ] tests
- [ ] document license (the apache2.0 license feels too long for the top of the file, so Im not sure the best way).
- [ ] add in layer norm variants. 
- [ ] add error handling


darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+        MonteCarloDropout(p=dropout),
+        nn.Linear(d_ff, d_model),
+        MonteCarloDropout(p=dropout),
+    )
+
+
+class FeedForwardBlock(nn.Module):
+    def __init__(self, d_model: int, d_ff: int, **kwargs):
+        super().__init__()
+        self.norm = nn.LayerNorm(d_model)
```
hrzn, line 189: Do you want to make this one parameterizable?

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+        n_heads=1,
+        attn_type="full",
+        norm=nn.LayerNorm,
+    ):
+        super().__init__()
+        self.attn = MultiHeadAttentionBlock(
+            dim_val, dim_attn, n_heads, attn_type=attn_type
+        )
+
+        self.norm1 = norm(dim_val)
```
hrzn, line 221: These normalization don't have parameters, so I don't think you need to define two of them, one should be enough.

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+    def forward(self, x):
+        a = self.attn(x)
+        x = self.norm1(x + a)
+
+        a = self.ffn(x)
+        x = self.norm2(x + a)
+
+        return x
+
+    def record(self):
```
hrzn, line 235: Why is this needed?

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+    ):
+        super().__init__()
+        self.attn1 = MultiHeadAttentionBlock(
+            dim_val, dim_attn, n_heads, attn_type=attn_type
+        )
+        self.attn2 = MultiHeadAttentionBlock(
+            dim_val, dim_attn, n_heads, attn_type=attn_type
+        )
+
+        self.norm1 = norm(dim_val)
```
hrzn, line 257: Same comment here, I think only one `self.norm` should be enough.

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+    def __init__(
+        self,
+        input_chunk_length: int,
+        output_chunk_length: int,
+        dim_val: int = 32,
+        dim_attn: int = 128,
+        dec_seq_len: int = 12,
+        n_encoder_layers: int = 3,
+        n_decoder_layers: int = 3,
+        enc_attn_type="query_selector_0.8",
```
hrzn, line 420: Could we keep the fraction f of the query selector as a separate float argument?

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+        self.debug = debug
+
+    def forward(self, x_in: tuple):
+        x, _ = x_in
+        # x: [Batch, Input length, Channel]
+
+        # encoder
+        e = self.encs[0](self.pos(self.enc_dropout(self.enc_input_fc(x))))
+
+        for enc in self.encs[1:]:
```
hrzn, line 382: I guess this will crash if there's only one encoder layer.

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+            e = enc(e)
+        if self.debug:
+            print(f"Encoder output size: {e.shape}")
+        # decoder
+        decoded = self.dec_input_fc(x[:, -self.dec_seq_len :])
+
+        d = self.decs[0](decoded, e)
+        for dec in self.decs[1:]:
```
hrzn, line 390: here too, this would fail if there's only one decoder layer.

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+            d = dec(d, e)
+
+        # output
+        x = self.out_fc(d.flatten(start_dim=1))
+
+        x = torch.reshape(x, (x.shape[0], -1, self.output_dim))
+        x = x.view(
```
hrzn, line 397: I see they also do it like that in the reference implementation, and we also do it like that in our current `Transformer` implementation, but I'm not sure that's the best way.
Currently, in a single forward pass we output one large vector of size `self.output_chunk_length * self.output_dim * self.nr_params`, which we reshape in order to get a time series of length `self.output_chunk_length`. However, if you look at the transformer paper, I believe a more natural way to do this would be to auto-regressively output the next time steps; i.e., instead of outputing `output_chunk_length` all at once, output first `t+1` in a forward pass, then feed this as an input to the decoder and output `t+2`, etc. So perform `output_chunk_length` forward passes. I believe this might be better as it would avoid depending solely on the last FC layer to capture everything needed about an arbitrary number of future time steps. That might come at the price of increase run time for inference though. Perhaps we could leave this as an option?

darts/models/forecasting/query_selector.py
```diff
@@ -0,0 +1,650 @@
...
+        dec_seq_len : int
+            the embedding size for the decoder (default=12).
+        n_encoder_layers : int
+            The number of encoder layers in the encoder (default=3).
+        n_decoder_layers : int
+            The number of decoder layers in the encoder (default=3).
+        enc_attn_type : str
+            "full" attention or query selector with a reduction f factor. Example "query_selector_0.4".
+            Default to "query_selector_0.8"
+        dec_attn_type : str
```
hrzn, line 451: Could we refactor our current `Transformer` model to somehow leverage this and only be a particular case of query selector with full matrix?

hrzn: Could we also by any chance incorporate `Informer` as another particular implementation here?

hrzn (commented):
Thanks, that looks quite good! My main comment revolve around the fact that I believe it might make sense to leave an option allowing the outputs to be generated in an auto-regressive way with the Transformer decoder (as in the original "Attention is all you need" paper).

I also have one higher level comment/question. What is your impression of the model? I'm slightly hesitant on whether this one would represent a good addition to Darts. I quickly went over the paper and it looks very drafty, without many experiments, no analysis, ablation study, etc. But if you tell us that you had great results with this model over other competitive ones, then it's an argument to have it.

I would also like if we could factor some of our transformers implementations together; e.g., refactor `Transformer` to be the particular `"full"` case of this one. In general I would also prefer if we avoid too much code duplication. E.g., having the same modules redefined in several places (such as positional encoding); that will make maintenance more difficult.

gdevos010:
Im currently look for an apartment, I will address these comments when I have a bit more free time

hrzn:
> Im currently look for an apartment, I will address these comments when I have a bit more free time

Sure, no rush

gdevos010:
@hrzn I originally wanted to add it because it looked to do reasonably well on [benchmarks](https://paperswithcode.com/paper/long-term-series-forecasting-with-query). I have run a bunch of tests and haven't been impressed. I mostly get forecasts that look like these plots below. I have done some tuning and cant seem to get it to "learn" the data. I will move the typos I made to a new branch.


Forecast_Hydro Energy QuerySelector (default) (LayerNorm)
Forecast_Sunspots QuerySelector (default) (LayerNorm)

hrzn:
> @hrzn I originally wanted to add it because it looked to do reasonably well on [benchmarks](https://paperswithcode.com/paper/long-term-series-forecasting-with-query). I have run a bunch of tests and haven't been impressed. I mostly get forecasts that look like these plots below. I have done some tuning and cant seem to get it to "learn" the data. I will move the typos I made to a new branch.
> 
> Forecast_Hydro Energy QuerySelector (default) (LayerNorm) Forecast_Sunspots QuerySelector (default) (LayerNorm)

OK, thanks for investigating! In general, I think we should indeed favour models that either perform very well, or are accompanied by a solid publication demonstrating its value. Your results are not great, however looking at them I would tend to think they look more like a bug in the implementation than an issue with the model itself. We can still park this PR/branch until someone has more time to investigate.

hrzn:
@gdevos010 I will close this one for now, but let's be open about re-opening if the time comes later on. Hope that's OK for you.

Issue #623, opened 2021-12-01
Extend backtesting methods to retrain after configurable number of time steps


hrzn:
See: https://github.com/unit8co/darts/issues/135#issuecomment-709979882

chefPony:
Hi, I would like to contribute to this issue. The solution I have in mind is very straightforward with minimal impact I think. The main method impacted will be `historical_forecasts` in the `ForecastingModel` base class, basically I would change method signature to: `retrain: int = 1` (actual behaviour) instead of `bool`. Then, before the prediction loop at line 430, I would define a `counter=0` variable which will get increment at each iteration. Finally I would change the train if at line 438 to something like `if (retrain or not self._fit_called) and (counter % retrain == 0)`, and this should do it.
Let me know what you think, if you think this can work I can make a pr soon.
Thanks!

FBruzzesi:
Hi @hrzn, I would like to propose a more general approach, which I find particularly useful for retraining.
As a naive example to justify this, suppose that you want to retrain the model every 1st of the month, then the above approach wouldn't work.

 The way to do that is to pass a callable to determine on which conditions you should retrain the model, leaving the user all the flexibility, i.e. something like:

```python
if (not self._fit_called 
    or (retrain 
         and retrain_on_condition(pred_time, train_series, past_covariates, future_covariates))
    ):
    # retrain the model here
```
Where the function `retrain_on_condition` should be provided by the user, here a list of few examples:

```python
import pandas as pd
from darts import TimeSeries

def retrain_on_condition(
    pred_time: pd.Timestamp = None,
    train_series: TimeSeries = None,
    past_covariates: TimeSeries = None,
    future_covariates: TimeSeries = None
):
    """
    Examples:
        return pred_time.hour == 0  # retrain every midnight (for freq="H") 
        return pred_time.day == 1  # retrain every month start (for freq="D") 
        return pred_time.hour == 0 & pred_time.day == 1  # retrain every month start (for freq="H")
        return pred_time.hour % N == 0  # retrain every N hours (for freq="H")
        return pred_time.hour == 0 & pred_time.dayofweek == 0  # retrain every week (for freq="H")
    """
    return pred_time.hour == 0 & pred_time.day == 1  # retrain every month start (for freq="H")
```

This adds a layer of complexity for the user, yet it is very flexible to similate and backtest what would happen in a production environment since it leaves the possibility to implement logics on the data series.

As an example, we retrain models every month first OR when we get some kind of alert on the data quality/drift

[...]

Issue #135, opened 2020-07-10
`backtest()`: Retrain every n steps


aschl:
Recently found your library and it is really great and handy!
The evaluation of the forecasting performance is best done using a cross-validation approach. The `backtest_forecasting()`-function does that - although it currently iterates and re-trains the model on every single time step. In my application, I am training ten-thousands of different time series and it becomes computationally unfeasible to retrain on every time step. Another approach already implemented in scikit-learn is `TimeSeriesSplit()`, which generates a k-fold cross-validation split designed for time series.
https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html

One solution to the issue would be to add support for different cross-validation techniques implemented in scikit-learn (such as TimeSeriesSplit()). Another solution I see could be to add a `stride` parameter in `backtest_forecasting()` that specifies how many time steps it should skip while backtesting.

hrzn:
Thanks for pointing this out @aschl. We are currently re-working quite substantially the backtesting functionality, and indeed our plan is to introduce a `stride` parameter like you suggest. We will also introduce moving windows for the training set (instead of expending only, as now). These should come in a release sometime soon.

aschl:
Awesome! Looking forward to seeing this feature. The idea of incorporating the moving window is also great and very useful.
When do you approximately plan to release this?

guillaumeraille:
Hello @aschl, while we don't have a date yet for the release of this feature. If this something you would like to try as soon as possible the stride parameter has already been added to the `backtest_forcasting` method on the develop branch. Keep in mind that we are currently working on a quite heavy refactor of the whole `backtesting.py` file and the API for backtesting might change in the upcoming release.

aschl:
I checked `backtesting.py` in the `develop` branch. If I understand it correctly, the `stride` parameters determines the distance between two consecutive predictions. This might be helpful if e.g. you only want to predict January values (and skip Feb-Dec forecast). I think there are quite a few use cases when you need all prediction points but would like to train the model only on every `x` time steps. This would increase efficiency quite a lot without necessarily loosing predictive power (adding a few more samples to a series of few hundert observations doesn't substantially improve the model if the distribution remains similar).

The idea behind scikitlearn's `TimeSeriesSplit()` is exactly this. You split the time series in `k` cross-validation pieces and retrain the model only `k` times. This way you obtain a relatively unbiased estimate of the forecasting error with reasonable computational efforts.

[...]

PR #1139, opened 2022-08-10
Feature/historical retrain on condition

FBruzzesi:
Addresses [#135](https://github.com/unit8co/darts/issues/135) and [#623](https://github.com/unit8co/darts/issues/623)

### Summary

Implemented support for new types in `retrain` argument of `historical_forecasts` method. Now it accepts `bool`, (positive) `int` and `Callable` (returning a `bool`) data types.

The new behaviour is as follows (copying from the updated docstring):

- In the case of ``bool``: retrain the model at each step (`True`), or never retrains the model (`False`).
   Not all models support setting `retrain` to `False`.
   Notably, this is supported by neural networks based models.
- In the case of ``int``: the model is retrained every `retrain` iterations.
- In the case of ``Callable``: the model is retrained whenever callable returns `True`.
   Notice that the arguments passed to the callable are as follows:
   - `pred_time (pd.Timestamp)`: next timestamp to predict (retraining happens before)
   - `series (TimeSeries)`: train series up to `pred_time`
   - `past_covariates (TimeSeries)`: past_covariates series up to `pred_time`
   - `future_covariates (TimeSeries)`: future_covariates series up to `pred_time + series.freq * forecast_horizon`
 

### Other Information

1. In order to achive this behaviour, I also added a decorator function, called `_retrain_wrapper`, in `utils/utils.py`, which passes to the wrapped function only the original signature arguments, and raises a `ValueError` in the case that the provided function doesn't return a boolean value.
2. To test the behaviour, I wrote a test, called `test_backtest_retrain`, in `test_local_forecasting_models.py` as suggested in [#623](https://github.com/unit8co/darts/issues/623)


darts/models/forecasting/forecasting_model.py
```diff
@@ -353,8 +354,20 @@ def historical_forecasts(
...
+            Whether and/or on which condition to retrain the model before predicting.
+            This parameter supports 3 different datatypes: ``bool``, (positive) ``int``, and
+            ``Callable`` (returning a ``bool``).
+            In the case of ``bool``: retrain the model at each step (`True`), or never retrains the model (`False`).
+            Not all models support setting `retrain` to `False`.
+            Notably, this is supported by neural networks based models.
+            In the case of ``int``: the model is retrained every `retrain` iterations.
+            In the case of ``Callable``: the model is retrained whenever callable returns `True`.
+            Notice that the arguments passed to the callable are as follows:
+                - `pred_time (pd.Timestamp)`: next timestamp to predict (retraining happens before)
```
hrzn, line 366:
```suggestion
                - `pred_time (pd.Timestamp or int)`: timestamp of forecast time (end of the training series)
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -353,8 +354,20 @@ def historical_forecasts(
...
+                - `series (TimeSeries)`: train series up to `pred_time`
```
hrzn, line 367: How about calling it `train_series` ?

darts/models/forecasting/forecasting_model.py
```diff
@@ -405,6 +419,22 @@ def historical_forecasts(
                 logger,
             )
 
+        if isinstance(retrain, bool) or (isinstance(retrain, int) and retrain >= 0):
+            retrain_func = _retrain_wrapper(
```
hrzn, line 423: In this case, could we bypass the call to `retrain_func` altogether when deciding to retrain?
I think it would be slightly better in some cases for performance reasons, because each call to `retrain_func()` requires some (potentially) relatively expensive calls to `drop_after()` on some series, whereas when `retrain` is bool or int we wouldn't need to produce these time series.

darts/models/forecasting/forecasting_model.py
```diff
@@ -431,15 +461,27 @@ def historical_forecasts(
...
 
-            if retrain or not self._fit_called:
+            if (not self._fit_called) or retrain_func(
+                counter=_counter,
+                pred_time=pred_time,
+                series=train,
+                past_covariates=past_covariates.drop_after(pred_time)
+                if past_covariates
+                else None,
+                future_covariates=future_covariates.drop_after(
```
hrzn, line 479: This will fail in some cases (and causes the quickstart notebook to fail), because the actual forecasts do not go further than the end of `series` in backtesting. So you should probably do something like
```python
future_covariates=future_covariates.drop_after(
    min(pred_time + series.freq * forecast_horizon, series.end_time()))
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -353,8 +354,20 @@ def historical_forecasts(
         stride
             The number of time steps between two consecutive predictions.
         retrain
-            Whether to retrain the model for every prediction or not. Not all models support setting
-            `retrain` to `False`. Notably, this is supported by neural networks based models.
+            Whether and/or on which condition to retrain the model before predicting.
+            This parameter supports 3 different datatypes: ``bool``, (positive) ``int``, and
+            ``Callable`` (returning a ``bool``).
+            In the case of ``bool``: retrain the model at each step (`True`), or never retrains the model (`False`).
+            Not all models support setting `retrain` to `False`.
```
hrzn, line 361: Could you move these two sentences down in the `retrain` doc? And slightly modify it, e.g.,
```
Note: some models do require being retrained every time
and do not support anything else than `retrain=True`.
```

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -336,3 +338,43 @@ def test_statsmodels_dual_models(self):
             # check backtesting with retrain=False
             model: TransferableDualCovariatesForecastingModel = model_cls(**kwargs)
             model.backtest(series1, future_covariates=exog1, retrain=False)
+
+    def test_backtest_retrain(self):
```
hrzn, line 342: You could also add a test to make sure `retrain` is being called as expected (with expected arguments, e.g., past/future covariates etc) when it is a `Callable`. I think this should be pretty easy to do with `unittest.mock`, you can check out examples here: https://docs.python.org/3/library/unittest.mock.html

hrzn (commented):
Looks quite good, thanks! I have a couple of comments.

darts/tests/models/forecasting/test_local_forecasting_models.py
FBruzzesi: I wasn't able to use `Mock`, yet implemented a test with "fake" `train_series`, `past_covariates` and `future_covariates` all equal to the series itself, but testing on some "proper" condition.

hrzn: This can also work, however your current test does not test correctness, or even that the function has been called. If you want to do it this way (using conditions on the series within the function), I would maybe cook a small toy example where the result of the backtest is not the same depending (for instance) on the quantile quantities computed on the covariates series. This way you could then test that 
* The output of the backtesting procedure is not the same if different `future_covariates` are used
* The output is as you expect in some pre-computed case (e.g., the historical forecasts is a series containing values [x, y, z]).

That being said, I still think it might be easier to simply test whether the function is called by using mocking. You can find an example of how we are doing it in other tests, for instance [here](https://github.com/unit8co/darts/blob/master/darts/tests/models/forecasting/test_torch_forecasting_model.py#L51), where we then check [here](https://github.com/unit8co/darts/blob/master/darts/tests/models/forecasting/test_torch_forecasting_model.py#L82) whether a given function has been called or not.

FBruzzesi: Just to explain myself: it's easy to mock a retrain function (i.e. a callable returning a boolean), but the function itself is called at each iteration (and I am able to check for that). 
I would be more interested to verify that `_fit_wrapper` method is called multiple times. 
However when trying to patch/mock `_fit_wrapper` and `_predict_wrapper`, I get all sort of errors.

FBruzzesi: @hrzn do you have any new suggestion on this?

hrzn: Then if mocking does not work my suggestion would be to at least test that the output of `model_cls.backtest()` is what you expect it to be for a couple of simple cases (e.g., when you retrain each 1st of the month).

FBruzzesi: Found few tricks to make it work using `mock.patch`. Looking forward to the review.

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -336,3 +339,124 @@ def test_statsmodels_dual_models(self):
...
+                with self.assertRaises(ValueError):
+                    _ = model_cls.historical_forecasts(series, retrain=retrain)
+
+            else:
+
+                if isinstance(retrain, Mock):
+                    # resets patch_retrain_func call_count to 0
+                    retrain.call_count = 0
+                    retrain.side_effect = [True, False] * (len(series) // 2)
+                    # retrain.return_value = True
```
hrzn, line 433: this line can be removed

hrzn (commented):
Thanks, the tests look better with the mocking! 
There were a couple more comments to address and then I think we will be able to merge.

darts/models/forecasting/forecasting_model.py
hrzn: @FBruzzesi did you get a chance yet to work on these last remaining small changes?

FBruzzesi: I probably didn't mentioned it anywhere. In line 441 I added a variable containing the original retrain function signature. Then in the control flow I am passing the past/future series (hence calling `drop_after` method) only if it is contained in such variable. If it's not, then I just pass `None`

darts/models/forecasting/forecasting_model.py
```diff
@@ -357,8 +358,20 @@ def historical_forecasts(
...
         retrain
-            Whether to retrain the model for every prediction or not. Not all models support setting
-            `retrain` to `False`. Notably, this is supported by neural networks based models.
+            Whether and/or on which condition to retrain the model before predicting.
+            This parameter supports 3 different datatypes: ``bool``, (positive) ``int``, and
+            ``Callable`` (returning a ``bool``).
+            In the case of ``bool``: retrain the model at each step (`True`), or never retrains the model (`False`).
+            In the case of ``int``: the model is retrained every `retrain` iterations.
+            In the case of ``Callable``: the model is retrained whenever callable returns `True`.
+            Arguments passed to the callable are as follows:
```
hrzn, line 367: We need to add blank lines before list for the rst text to render well

```suggestion
            Arguments passed to the callable are as follows:
            
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -357,8 +358,20 @@ def historical_forecasts(
...
+                - `pred_time (pd.Timestamp or int)`: timestamp of forecast time (end of the training series)
+                - `train_series (TimeSeries)`: train series up to `pred_time`
+                - `past_covariates (TimeSeries)`: past_covariates series up to `pred_time`
+                - `future_covariates (TimeSeries)`: future_covariates series up
+                to `min(pred_time + series.freq * forecast_horizon, series.end_time())`
```
hrzn, line 372:
```suggestion
                  to `min(pred_time + series.freq * forecast_horizon, series.end_time())`
                  
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -556,8 +601,20 @@ def backtest(
...
         retrain
-            Whether to retrain the model for every prediction or not. Not all models support setting
-            `retrain` to `False`. Notably, this is supported by neural networks based models.
+            Whether and/or on which condition to retrain the model before predicting.
+            This parameter supports 3 different datatypes: ``bool``, (positive) ``int``, and
+            ``Callable`` (returning a ``bool``).
+            In the case of ``bool``: retrain the model at each step (`True`), or never retrains the model (`False`).
+            In the case of ``int``: the model is retrained every `retrain` iterations.
+            In the case of ``Callable``: the model is retrained whenever callable returns `True`.
+            Arguments passed to the callable are as follows:
```
hrzn, line 610:
```suggestion
            Arguments passed to the callable are as follows:
            
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -556,8 +601,20 @@ def backtest(
...
+                - `pred_time (pd.Timestamp or int)`: timestamp of forecast time (end of the training series)
+                - `train_series (TimeSeries)`: train series up to `pred_time`
+                - `past_covariates (TimeSeries)`: past_covariates series up to `pred_time`
+                - `future_covariates (TimeSeries)`: future_covariates series up
+                to `min(pred_time + series.freq * forecast_horizon, series.end_time())`
```
hrzn, line 615:
```suggestion
                  to `min(pred_time + series.freq * forecast_horizon, series.end_time())`
                  
```

hrzn (commented):
Looks good! I just have 4 tiny commit suggestions and then I think we'll be good.

hrzn (approved):
Thanks @FBruzzesi ! I'll merge it :)

PR #1141, opened 2022-08-11
Feat/add australian tourism dataset

hrzn:
Adds a new dataset, which will be used to show hierarchical reconcilation.
Tests are expected to fail as the CSV is not on `master` yet.


brunnedu (approved):
LGTM!

gdevos010:
@hrzn I think you are missing the [1] reference

hrzn:
> @hrzn I think you are missing the [1] reference

Good catch, thanks

Issue #1144, opened 2022-08-12
[BUG] lightgbm model with validation set 


bawiek:
**Describe the bug**
Hi, I get an error for the LGBM model with a validation set with output_chunk_length >1  
With a output_chunk_length=1 it works fine, is this normal ?

![image](https://user-images.githubusercontent.com/50851904/184353380-7dbaa365-e2e2-4aa4-bba4-63e3098d708a.png)

**System :**
 - Python version: 3.8.10
 - darts version: 20.0.0

**To Reproduce**
```
from darts.datasets import AirPassengersDataset
series_air = AirPassengersDataset().load()
scaler_air, scaler_milk = Scaler(), Scaler()
series_air_scaled = scaler_air.fit_transform(series_air)
train_air, val_air = series_air_scaled[:-36], series_air_scaled[-36:]

model_lightGBM = LightGBMModel( lags=10, 
                                lags_past_covariates=None, 
                                lags_future_covariates=None, 
                                output_chunk_length=2, 
                                likelihood=None, 
                                quantiles=None, 
                                random_state=None)
model_lightGBM.fit(
    series = train_air,
    val_series = val_air,
    #past_covariates = None,
    #val_past_covariates = None,
    #future_covariates = None,
    #val_future_covariates = None,
    verbose=True
)
```
_TypeError: Wrong type(ndarray) for label.
It should be list, numpy 1-D array or pandas Series_

**Expected behavior**
[1]	valid_0's l2: 0.222919
[2]	valid_0's l2: 0.210642

hrzn:
Hi, could you please post the full stack trace of the error? Thank you.

bawiek:
Hi @hrzn,
Thank you for your time!

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
C:\Users\BWIECZ~1\AppData\Local\Temp\2/ipykernel_2512/315192144.py in <module>
     12                                 quantiles=None,
     13                                 random_state=None)
---> 14 model_lightGBM.fit(
     15     series = train_air,
     16     val_series = val_air,

c:\Users\bwieczorek\Desktop\lib\site-packages\darts\models\forecasting\gradient_boosted_model.py in fit(self, series, past_covariates, future_covariates, val_series, val_past_covariates, val_future_covariates, max_samples_per_ts, **kwargs)
    160             return self
    161 
--> 162         super().fit(
    163             series=series,
    164             past_covariates=past_covariates,

[...]

PR #1154, opened 2022-08-16
MultiOutputRegressor that handles validation data in fit()

brunnedu:
Addresses #1144 

### Summary

This PR solves an issue where validation data that was passed to the `LightGBMModel` or the `CatBoostModel` in `fit()` wasn't handled correctly by the original sklearn `MultiOutputRegressor`.

The solution is a slightly modified `MultiOutputRegressor` class that slices the validation data together with the training data.

- Added `MultiOutputRegressor` in `darts.utils.multioutput` that inherits from sklearn `MultiOutputRegressor` and overwrites its `fit()` method
- `MultiOutpuRegressor.fit()` now slices validation data together with training data as long as it is passed in the parameter `eval_set` as part of `**fit_params`
- Added unit tests that check for this case

### DIFF
--- a/darts/utils/multioutput.py
+++ b/darts/utils/multioutput.py
@@ -1,0 +1,93 @@
+from joblib import Parallel
+from sklearn.base import is_classifier
+from sklearn.multioutput import MultiOutputRegressor as sk_MultiOutputRegressor
+from sklearn.multioutput import _fit_estimator
+from sklearn.utils.fixes import delayed
+from sklearn.utils.multiclass import check_classification_targets
+from sklearn.utils.validation import _check_fit_params, has_fit_parameter
+
+
+class MultiOutputRegressor(sk_MultiOutputRegressor):
+    """
+    :class:`sklearn.utils.multioutput.MultiOutputRegressor` with a modified ``fit()`` method that also slices
+    validation data correctly. The validation data has to be passed as parameter ``eval_set`` in ``**fit_params``.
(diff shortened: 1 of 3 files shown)


brunnedu:
I also wanted to add some tests for the new `MultiOutputRegressor` class but not sure when I'll get to it. For now, I just added a new test in `test_regression_models.py`.

darts/utils/multioutput.py
```diff
@@ -0,0 +1,93 @@
+from joblib import Parallel
+from sklearn.base import is_classifier
+from sklearn.multioutput import MultiOutputRegressor as sk_MultiOutputRegressor
+from sklearn.multioutput import _fit_estimator
+from sklearn.utils.fixes import delayed
+from sklearn.utils.multiclass import check_classification_targets
+from sklearn.utils.validation import _check_fit_params, has_fit_parameter
+
+
+class MultiOutputRegressor(sk_MultiOutputRegressor):
```
hrzn, line 10: Looks good! I wonder if that could be of any use to propose this as a PR to modify the `_MultiOutputEstimator` class in sklearn directly. WDYT?

hrzn (approved):
Thanks, looks great!

darts/utils/multioutput.py
brunnedu: That sounds like a good idea, I'll see what I can do 👍

PR #1164, opened 2022-08-22
Fix for pandas.Int64Index is deprecated and will be removed from pandas in a future version

tranquilitysmile:
```
FutureWarning: pandas.Int64Index is deprecated and will be removed from pandas in a future version. Use pandas.Index with the appropriate dtype instead.
  from pandas import MultiIndex, Int64Index
```

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -4076,9 +4076,7 @@ class TimeSeries:
 
             # restore a RangeIndex if needed:
             time_idx = xa_.get_index(self._time_dim)
-            if time_idx.is_integer() and not isinstance(
-                time_idx, pd.RangeIndex
-            ):
+            if time_idx.is_integer() and not isinstance(time_idx, pd.RangeIndex):
                 xa_ = xa_.assign_coords(
                     {self._time_dim: pd.RangeIndex(start=key, stop=key + 1)}
                 )
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -4076,7 +4076,7 @@ class TimeSeries:
 
             # restore a RangeIndex if needed:
             time_idx = xa_.get_index(self._time_dim)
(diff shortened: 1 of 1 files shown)


hrzn (approved):
Awesome, thanks!

darts/timeseries.py
```diff
@@ -4076,7 +4076,7 @@ def _set_freq_in_xa(xa_: xr.DataArray):
 
             # restore a RangeIndex if needed:
             time_idx = xa_.get_index(self._time_dim)
-            if isinstance(time_idx, pd.Int64Index) and not isinstance(
+            if isinstance(time_idx, pd.Index) and not isinstance(
```
dennisbader, line 4079: pd.DatetimeIndex() is also an instance of pd.Index. For series with a DatetimeIndex, we would return a new time series with a Range Index instead of a DatetimeIndex. So I believe we need to check the following:
```suggestion
            if time_idx.is_integer() and not isinstance(
```

dennisbader (changes requested):
Thanks for this. I think we need to slightly change the condition so that series with a DatetimeIndex will return a new series with a DatetimeIndex istead of a RangeIndex.

hrzn:
> Thanks for this. I think we need to slightly change the condition so that series with a DatetimeIndex will return a new series with a DatetimeIndex istead of a RangeIndex.

Oh good catch! I trusted the unit tests too blindly before accepting. We need to add a unit test for that!

hrzn:
@tranquilitysmile are you willing to try and find a workaround?

tranquilitysmile:
> @tranquilitysmile are you willing to try and find a workaround?

With the latest commit, the warnings are gone, I did not find any problems in the work of applications

hrzn (approved):
LGTM
@dennisbader wdyt?

dennisbader:
Looks good, yes we can merge!

dennisbader (approved):

PR #1171, opened 2022-08-25
Feat/statsforecasts ets

hrzn:
* Update Croston and StatsForecastAutoARIMA to make it work with latest versions of statsforecast
* Add new model `StatsForecastETS` containing ETS
* Updated docs, added examples
* Bumped `statsforecast` version


dumjax (approved):
nice add, I don't see any problem

Issue #1181, opened 2022-09-02
[BUG] Can't convert np.ndarray of type numpy.object_ 


vattyamp:
**Describe the bug**
When fitting to any of the global models I get an error about being unable to convert a datatype. Training works for LightGBM model but for none of the other models. My data set is a list of multivariate time series `[s1,s2 ...]` from the function `TimeSeries.from_group_dataframe`. 
 
**To Reproduce**
```python
from darts import TimeSeries

from darts.utils.model_selection import train_test_split
from darts.models import RNNModel,TCNModel,NBEATSModel,BlockRNNModel,LightGBMModel

df_timeseries = TimeSeries.from_group_dataframe(df.toPandas(),group_cols='ID',time_col='Date',value_cols=['Total','A','B','C'], freq='D',fill_missing_dates=True)

y_train,y_val = train_test_split(df_timeseries_filled,test_size=7, axis=1)

model_cov = RNNModel(input_chunk_length=14, output_chunk_length=3, n_epochs=100, random_state=0)
model_cov.fit(series=y_train)

```

error message:
_TypeError: can't convert np.ndarray of type numpy.object_. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool._

```
TypeError: can't convert np.ndarray of type numpy.object_. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uint8, and bool.

TypeError                                 Traceback (most recent call last)
<command-4487064413187948> in <cell line: 10>()
      8 #regressor.fit(series=y_train,future_covariates=x_train)
      9 
---> 10 model_cov.fit(series=y_train)

/local_disk0/.ephemeral_nfs/envs/pythonEnv-7de0310b-4a5c-4a5b-97b4-744bb60f3c8e/lib/python3.9/site-packages/darts/utils/torch.py in decorator(self, *args, **kwargs)
    110         with fork_rng():
    111             manual_seed(self._random_instance.randint(0, high=MAX_TORCH_SEED_VALUE))
--> 112             return decorated(self, *args, **kwargs)
    113 
    114     return decorator

[...]

PR #1183, opened 2022-09-03
Fix/ds no static covs

dennisbader:
Addresses #1181

### Summary
Currently our torch datasets use/include static covariates even if the model does not support it.
- adds option to ignore static covariates for darts training and inference torch datasets
- TorchForecastingModels' Datasets now only use static covariates if the model supports it
- avoids errors from non-numeric static covariates even if the model doesn't use them.


hrzn (approved):
Thanks!

louisoutin:
Would be nice to also have use_static_covariates for non torch models (linear regression, random forest etc).
Btw thanks for this package, really useful!

PR #1186, opened 2022-09-05
Fix/#1185 corrected type hinting ExponentialSmoothing model

rijkvandermeulen:
Fixes mistake in type hinting ExponentialSmoothing model

### Summary

`seasonal: Optional[SeasonalityMode]` iso `seasonal: Optional[ModelMode]`

### Other Information

Potentially useful to add mypy to pipeline and pre-commit hooks?

### DIFF
--- a/darts/models/forecasting/exponential_smoothing.py
+++ b/darts/models/forecasting/exponential_smoothing.py
@@ -34,7 +34,7 @@ class ExponentialSmoothing(ForecastingModel):
         <https://www.statsmodels.org/stable/generated/statsmodels.tsa.holtwinters.ExponentialSmoothing.html>`_;
         we refer to this link for the original and more complete documentation of the parameters.
 
-        `model_mode` must be a ``ModelMode`` Enum member. You can access the Enum with
+        `trend` must be a ``ModelMode`` Enum member. You can access the Enum with
          ``from darts.utils.utils import ModelMode``.
         `seasonal` must be a ``SeasonalityMode`` Enum member. You can access the Enum with
         ``from darts.utils.utils import SeasonalityMode``.
--- a/darts/models/forecasting/exponential_smoothing.py
+++ b/darts/models/forecasting/exponential_smoothing.py
@@ -21,7 +21,7 @@ class ExponentialSmoothing(ForecastingModel):
         self,
         trend: Optional[ModelMode] = ModelMode.ADDITIVE,
         damped: Optional[bool] = False,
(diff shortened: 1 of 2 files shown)


rijkvandermeulen:
Pipeline seems to break when executing CodeCov on macos (but not on ubuntu). Not sure exactly what's going on. Have you seen this before?

darts/models/forecasting/exponential_smoothing.py
```diff
@@ -34,7 +34,10 @@ def __init__(
         <https://www.statsmodels.org/stable/generated/statsmodels.tsa.holtwinters.ExponentialSmoothing.html>`_;
         we refer to this link for the original and more complete documentation of the parameters.
 
-        `model_mode` must be a ``ModelMode`` Enum member. You can access the Enum with ``from darts import ModelMode``.
+        `model_mode` must be a ``ModelMode`` Enum member. You can access the Enum with
```
hrzn, line 37: Do we still need this line?

darts/models/forecasting/exponential_smoothing.py
```diff
@@ -34,7 +34,10 @@ def __init__(
...
+         ``from darts.utils.utils import ModelMode``.
+        `season_mode` must be a ``SeasonalityMode`` Enum member. You can access the Enum with
```
hrzn, line 39: Shouldn't it be
```suggestion
        `seasonal` must be a ``SeasonalityMode`` Enum member. You can access the Enum with
```

hrzn:
> Pipeline seems to break when executing CodeCov on macos (but not on ubuntu). Not sure exactly what's going on. Have you seen this before?

Sometimes the codecov upload fails, I've just relaunched it.

darts/models/forecasting/exponential_smoothing.py
rijkvandermeulen: Indeed! Changed it.

darts/models/forecasting/exponential_smoothing.py
rijkvandermeulen: Yes would say so as the ModelMode enum is still used for the trend argument in the ES model

hrzn: Then I think we should write `trend` must be a ``ModelMode`` ... no?

rijkvandermeulen: Jup of course - apologies for being sloppy here. Fixed it now.

hrzn (approved):
thanks!

PR #1193, opened 2022-09-07
return single encoded covariates if called on single target

dennisbader:
### Summary
- fixes bug where `add_encoders` in RegressionModels was used in combination with a single `target` series in fit/predict
- SequentialEncoder (called by model's `add_encoders`) now returns covariates as single TimeSeries when called on single series


dumjax (commented):
good for me

hrzn (approved):
Thanks @dennisbader.
Do you know what was the impact? Were the encoders simply not working for single series (though I guess we would have cought it earlier)?

dennisbader:
Reason was that SequentialEncoder always returned a list of TimeSeries even for single series input (@dumjax found this when testing for Shap integration).

For TorchForecastingModels this wasn't an issue as the series were internally converted to a list of series before using the encoders.
For RegressionModels the conversion isn't applied before the encodings are generated.

I only added RegressionModel encoder tests for the multi-ts case, which is why they didn't catch it.
Should I add another test for single-ts case?

hrzn:
> I only added RegressionModel encoder tests for the multi-ts case, which is why they didn't catch it.
> Should I add another test for single-ts case?

If you have a moment, this is always welcome, so we make sure the issue does not reproduce again.

hrzn:
As discussed, still to do:
* Transform everything as a list in RegressionModels
* Unify API for accessing model.encode_inference() in torch and regression models.

dennisbader:
> As discussed, still to do:
> 
> * Transform everything as a list in RegressionModels
> * Unify API for accessing model.encode_inference() in torch and regression models.

I will open a new PR for these points as this is a bug fix.

> If you have a moment, this is always welcome, so we make sure the issue does not reproduce again.

I checked and I had already added a test for single series in this PR: https://github.com/unit8co/darts/blob/6ff4e22b169cd177defeeee90d4b011dea5fe7cf/darts/tests/models/forecasting/test_regression_models.py#L1058

PR #1204, opened 2022-09-11
First mypy refactorings

rijkvandermeulen:
### Current situation

I did a first investigation into the existing mypy errors and the amount of refactoring that would be required. Running mypy on the whole directory results in over **650** "errors". Some of these errors are super easy to fix, others will probably require slightly more effort.

### Proposal

Curious to learn how you guys look at this, but in my opinion fixing >650 errors is quite a substantial effort for one PR. I'm therefore wondering whether it would make more sense to split this up into more "manageable" chunks? The way of working would then look something like:

- I've created a mypy config file (mypy.ini) where we a.o. ignore the mypy errors of the part of the repo that we did not refactor yet.
- This allows us to already add mypy to the pre-commit hooks as I did as part of this PR (since it will only check the part of the repo that we actually expect to comply with mypy).
- Having this set-up in place, we can create seperate tickets to address individual parts of the repo (e.g., we refactor darts/timeseries/py and after that we remove: `[mypy-darts.timeseries.*]
ignore_errors = True`. This would allow us to step by step improve.

To make all of this more practical I've already took a shot on resolving the mypy issues for darts/dataprocessing/dtw and darts/dataprocessing/pipeline.py. As you'll see in the mypy.ini these directories/files are thus excluded from the "exception list" and hence mypy will check these files after each commit.

Looking forward to hearing your thoughts on this approach!


hrzn:
Thanks @rijkvandermeulen. It looks like a good way to do it. We'll try to review it sometime soon (we are a bit behind these days, apologies). We also will have to think whether the benefits (better type checks) over-weight the small cost (tracking all folders and dependencies in the ignore list, maintain the pre-commit hook). In any case I like your step-by-step approach!

Borda (approved):
just would be curious what is the runtime on whole project `run --all-files`

dennisbader:
Closing this, since it's quite outdated. At some point we will work on this using `ty`.

PR #1208, opened 2022-09-12
Fix bug when using latest version of Prophet

hrzn:
Solves a bug with a breaking change in latest version of Prophet 1.1.1


dennisbader (approved):
LGTM, thanks!

PR #1209, opened 2022-09-12
fix plotting with length=1

hrzn:
Before, plotting a series of length 1 would crash. Now we have a behaviour like this (example for a stochastic series):

image


dumjax:
It seems correct to me, but shouldn't we also add a specific test in unit test for 1 point plot? I am not even sure if there are some tests on plot in general.

hrzn:
> It seems correct to me, but shouldn't we also add a specific test in unit test for 1 point plot? I am not even sure if there are some tests on plot in general.

We don't have tests for plotting unfortunately (and I think adding some would be too complex for the benefits, at least for the time being).

PR #1220, opened 2022-09-17
add progress bar when loading datasets made of multiple series

hrzn:


eliane-maalouf:
I don't see an issue here, I ran the unittest for data loading and they pass.

eliane-maalouf (commented):
I don't see an issue here, I ran the unittest for data loading and they pass.

PR #1221, opened 2022-09-18
Fix/regr model multi series

dennisbader:
### Summary
- coverts single timeseries to list of timeseries at beginning of fit/predict in regression models
- general reordering of things in RegressionModel & TorchForecastingModels methods `fit()` and `predicT()`
  - i.e. compare predict now between the two models. they are now almost identical. we could think of further factoring out the code duplication
- added component names to encoders to keep track of the encoded components inside the covariates
- abstracted away encode_train and encode_inference behavior inside models into GlobalForecastingModel methods `generate_fit_encodings()` and `generate_predict_encodings()`
  - opens up encoding generation to public
  - allows to generate the encodings in a simple/unified way through the models
- refactored parts of encoders to overwrite encoded components if they are in the input (overwriting the output not the input)
  - this simplifies how we can use encoders in forecasting models (i.e. when predicting without `series`, and covariates, the model loads the covariates which were saved at fitting time in the single `series` case. The encodings were saved with these covariates and we don't want to generate duplicate encodings for inference, ...)
  - added fit before predict logic for encoders that require it
  - some additional comments/docs where needed

### Additional info:
- fixes CatboosModel always writing `catboost_info` files. By default models now do not write this output


darts/utils/data/encoder_base.py
```diff
@@ -238,9 +239,46 @@ def _merge_covariate(
         """
         return covariate.stack(encoded) if covariate is not None else encoded
 
+    @staticmethod
+    def _drop_encoded_components(
+        covariate: Optional[TimeSeries], components: pd.Index
+    ) -> Optional[TimeSeries]:
+        """Avoid pitfalls: `encode_train()` or `encode_inference()` can be called multiple times or chanined.
```
hrzn, line 246:
```suggestion
        """Avoid pitfalls: `encode_train()` or `encode_inference()` can be called multiple times or chained.
```

darts/utils/data/encoder_base.py
```diff
@@ -238,9 +239,46 @@ def _merge_covariate(
...
+        """
+        if covariate is None:
+            return covariate
+
+        duplicate_components = components[components.isin(covariate.components)]
+        # case 1: covariate only consists of encoded components
+        if len(duplicate_components) == len(covariate.components):
+            covariate = None
+        # case 2: covariate has also non-encoded components
+        elif not duplicate_components.empty:
```
hrzn, line 257: Shouldn't this be something like
```suggestion
        elif len(duplicate_components) < len(covariate.components):
```

darts/utils/data/encoder_base.py
```diff
@@ -238,9 +239,46 @@ def _merge_covariate(
...
+
+    @property
+    def fit_called(self) -> bool:
+        """Returns whether the `Encoder` object has been fitted."""
+        return self._fit_called
+
+    @property
+    @abstractmethod
+    def requires_fit(self) -> bool:
+        """Whether the `Encoder` sub class must be fit with `Enocder.encode_train()` before inference
```
hrzn, line 275:
```suggestion
        """Whether the `Encoder` sub class must be fit with `Encoder.encode_train()` before inference
```

darts/tests/models/forecasting/test_encoders.py
```diff
@@ -292,6 +312,284 @@ def helper_encoder_from_model(
...
+            "transformer": Scaler(),
+        }
+        # given `add_encoders` dict, we expect encoders to generate the following components
+        comps_expected_past = pd.Index(
+            [
+                "month_sin",
+                "month_cos",
+                "day_sin",
+                "day_cos",
+                "month",
```
hrzn, line 464: Question: with this new components-checking logic, will the encoders avoid generating a component if there's already a component named `month` in the original series? If yes, could be worth maybe renaming the encoded components with something not likely to collide - e.g. `darts_enc_month` or something.

hrzn (approved):
Thanks a lot @dennisbader , looks very good!
I just got a few relatively minor comments, and one question related to conflicting names between encoded components and the original components.

darts/tests/models/forecasting/test_encoders.py
dennisbader: Yep that makes sense, thanks

dennisbader: With this implementation it's actually a bit different: if `month` is already in the original series, we remove it, generate `month` encodings from scratch, and add it back to the original series.

In a first approach I implemented it like you described. But in the single TS `fit()` case, we save the covariates along with the encodings to the model. Then when we call `predict(n=10)`, the model uses the saved covariates to generate the encodings -> the encodings would not be regenerated as already part of input covariates -> we would not have all required timestamps for prediction (as `encode_inference()` is basically bypassed).
We could handle this by removing the encodings from the saved covariates in case they were loaded, but i think it's not the cleanest way to do. WDYT @hrzn?

hrzn: Indeed it doesn't sound very clean. Let's leave it like that then. It might be good to document this in a docstring somewhere though, as some users will probably wonder how such cases are handled.

PR #1244, opened 2022-09-26
fix documentation to match pytorch-lightning

gdevos010:
the `gpus` arg in PyTorch Lightning has been deprecated in favor of `devices`

fixed broken PyTorch Lightning link

https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#gpus


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -170,13 +170,13 @@ def __init__(
...
-                Instead, specify this with keys ``"accelerator", "gpus", "auto_select_gpus"`` in your
+                Instead, specify this with keys ``"accelerator", "devices", "auto_select_gpus"`` in your
                 ``pl_trainer_kwargs`` dict. Some examples for setting the devices inside the ``pl_trainer_kwargs``
                 dict:
 
                 - ``{"accelerator": "cpu"}`` for CPU,
-                - ``{"accelerator": "gpu", "gpus": [i]}`` to use only GPU ``i`` (``i`` must be an integer),
-                - ``{"accelerator": "gpu", "gpus": -1, "auto_select_gpus": True}`` to use all available GPUS.
+                - ``{"accelerator": "gpu", "devices": [i]}`` to use only GPU ``i`` (``i`` must be an integer),
+                - ``{"accelerator": "gpu", "devices": -1, "auto_select_gpus": True}`` to use all available GPUS.
```
dennisbader, line 179: could you adapt the code parts as well were we add the "gpus"?
- https://github.com/unit8co/darts/blob/06824b29c094250b55bd121cd44d42dac87f08af/darts/models/forecasting/torch_forecasting_model.py#L326

We probably should check the pytorch lightning version to either add "gpus" or "device". You can do to this similarly to https://github.com/unit8co/darts/blob/06824b29c094250b55bd121cd44d42dac87f08af/darts/models/forecasting/pl_forecasting_module.py#L397

dennisbader (changes requested):
Thanks again for another PR :) 
We would have to adapt the parts of the code where we generate params dicts for PyTorch Lightning Trainer with "gpus". Otherwise the deprecation warning will not disappear.

I added a suggestion.

darts/models/forecasting/torch_forecasting_model.py
gdevos010: updated

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -297,7 +301,7 @@ def __init__(
             pass
 
         # TODO: remove below in the next version ======>
-        accelerator, gpus, auto_select_gpus = self._extract_torch_devices(
+        accelerator, devices, auto_select_gpus = self._extract_torch_devices(
```
hrzn, line 304: Could we actually remove support for `torch_device_str`, and also remove these lines (as suggested in the TODO comment)?

docs/userguide/timeseries.md
```diff
@@ -97,7 +97,7 @@ Optionally, `TimeSeries` objects can contain a hierarchy, which specifies how it
 
 For instance, the following hierarchy means that the two components `"a"` and `"b"` add up to `"total"`:
 ```python
-hierarchy = {"a": ["total"], "b", ["total"]}
+hierarchy = {"a": ["total"], "b": ["total"]}
```
hrzn, line 100: Thanks, good catch!

hrzn (commented):
LGTM, however we could consider moving these comments to actual docstring, and remove the support for `torch_device_str`, WDYT?

dennisbader:
> LGTM, however we could consider moving these comments to actual docstring, and remove the support for `torch_device_str`, WDYT?

Agreed 👍

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -81,6 +81,10 @@
 
 logger = get_logger(__name__)
 
+# Check whether we are running pytorch-lightning >= 1.7.0 or not:
+tokens = pl.__version__.split(".")
+pl_170_or_above = int(tokens[0]) >= 1 and int(tokens[1]) >= 7
```
hrzn, line 86: Where is this used?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -351,62 +338,6 @@ def __init__(
         # pl_module_params must be set in __init__ method of TorchForecastingModel subclass
         self.pl_module_params: Optional[dict] = None
 
-    @staticmethod
-    def _extract_torch_devices(
```
hrzn, line 355: Less code 🥳

hrzn (approved):
LGTM, thanks @gdevos010 !
Just have a small question about the pytorch lightning check in `torch_forecasting_model.py`: is this needed?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -323,9 +313,6 @@ def __init__(
 
         # setup trainer parameters from model creation parameters
         self.trainer_params = {
-            "accelerator": accelerator,
-            "gpus": gpus,
-            "auto_select_gpus": auto_select_gpus,
```
dennisbader, line 328: We still need these 3 params (including your "device" logic for >= 1.7) as we decided to always use CPU by default (see my comments below on the default params)

dennisbader (changes requested):
Nice :) Only thing is that we should set the 3 parameters by default "accelerator", "gpus" (or "devices"), "auto_select_gpus", as we decided to always use CPU by default.

After that we can merge :)

darts/models/forecasting/block_rnn_model.py
```diff
@@ -274,6 +256,17 @@ def __init__(
...
+            "devices", and "auto_select_gpus"``. Some examples for setting the devices inside the ``pl_trainer_kwargs``
+            dict:
+
+                - ``{"accelerator": "cpu"}`` for CPU,
+                - ``{"accelerator": "gpu", "devices": [i]}`` to use only GPU ``i`` (``i`` must be an integer),
+                - ``{"accelerator": "gpu", "devices": -1, "auto_select_gpus": True}`` to use all available GPUS.
+
+                For more info, see here:
+                https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#trainer-flags , and
+                https://pytorch-lightning.readthedocs.io/en/stable/accelerators/gpu_basic.html#train-on-multiple-gpus
```
dennisbader, line 269: Could you apply one lower indent level here and in all other docstrings? Also add a new line at beginning and end for better paragraph separation

```suggestion

            - ``{"accelerator": "cpu"}`` for CPU,
            - ``{"accelerator": "gpu", "devices": [i]}`` to use only GPU ``i`` (``i`` must be an integer),
            - ``{"accelerator": "gpu", "devices": -1, "auto_select_gpus": True}`` to use all available GPUS.

            For more info, see here:
            https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#trainer-flags , and
            https://pytorch-lightning.readthedocs.io/en/stable/accelerators/gpu_basic.html#train-on-multiple-gpus
            
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -351,62 +338,6 @@ def __init__(
...
-        """This method handles the deprecated `torch_device_str` and should be removed in a future Darts version.
-
-        Returns
-        -------
-        Tuple
-            (accelerator, gpus, auto_select_gpus)
-        """
-
-        if torch_device_str is None:
-            return "cpu", None, False
```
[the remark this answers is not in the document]
gdevos010: These are PyTorch Lightning default values, so I don't think we should specify them. WDYT?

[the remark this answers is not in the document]
dennisbader: True, I wasn't sure anymore if they automatically used GPU. I checked it now as well and you're right :)
You can leave it as it is (without setting the params)
Thanks! 👍

darts/models/forecasting/torch_forecasting_model.py
gdevos010: it was used before I removed the torch_device_str. Its now removed

darts/models/forecasting/torch_forecasting_model.py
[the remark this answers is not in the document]
gdevos010: done

gdevos010:
@dennisbader we should be good to merge now

dennisbader (approved):
Yes, looks perferct now 👍 Thanks a lot for all the effort @gdevos010!

PR #1247, opened 2022-09-27
DLinear and NLinear models

### LINKED ISSUE #1119: DLinear
Add the [Dlinear ](https://github.com/cure-lab/DLinear) model from cure-lab.

I have a working non-probabilistic version working. Working on making it probabilistic.

gdevos010:
Fixes #1119.

### Summary

Added DLinear and NLinear models. These are extremely small models, yet are impressively accurate. 

https://github.com/cure-lab/LTSF-Linear
### Other Information

I need help making them probabilistic i.e., Im not sure how to incorporate `nr_params`.


hrzn:
Thanks @gdevos010, this looks very promising and I like the pitch for these models (although I still have to read the paper). I'll try to provide a thorough review sometime soon.

In the meantime, regarding `nr_params`, you basically should treat it as a multiplier for the number of outputs of the network. A typical way to do it is, if the last layer is fully connected (as is often the case), multiply its output dimensionality by `nr_params`. You can see an example of how it's done for BlockRNNModel [here](https://github.com/unit8co/darts/blob/master/darts/models/forecasting/block_rnn_model.py#L100). Other models do the same kind of things.

Could you also add tests for these new models?
Thanks!

darts/models/forecasting/nlinear.py
```diff
@@ -0,0 +1,246 @@
...
+    def forward(self, x_in: tuple):
+        x, _ = x_in
+        # x: [Batch, Input length, Channel]
+
+        seq_last = x[:, -1:, :].detach()
+        x = x - seq_last
+        x = self.Linear(x.permute(0, 2, 1)).permute(0, 2, 1)
+        x = x + seq_last
+
+        x = x.view(
```
gdevos010, line 72: @hrzn, so I've been working on this for a while. `x` is of shape `batch_size, output_chunk_length * nr_params, input_dim`. `view` cannot split up the dims into `batch_size, output_chunk_length, input_dim, nr_params`. I can do a permute right before this, but the accuracy goes way down. I've also tried  `reshape` and a few other things but the predictions get way worse. 

Do you have any thoughts?

hrzn: Thanks for the update @gdevos010. I'll have a look and make a suggestion.

gdevos010:
closed for @1292

PR #1256, opened 2022-10-04
Anomaly Detection (anomaly model, scorer, detector, aggregator)

julien12234:
Fixes #.


darts/anomaly_detection/anomaly_model.py
```diff
@@ -0,0 +1,214 @@
+from cProfile import label
```
hrzn, line 1: Don't import anything that is not used

hrzn: Pay attention to the order of the imports - see [isort](https://pycqa.github.io/isort/) and more in general [formatting and linting for Darts](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md#code-formatting-and-linting)

darts/anomaly_detection/anomaly_model.py
```diff
@@ -0,0 +1,214 @@
...
+from darts.logging import raise_if, raise_if_not
+
+from darts.anomaly_detection.score import _TrainableScorer
+from darts.anomaly_detection import score
+
+from darts.models.forecasting.forecasting_model import ForecastingModel
+from darts.models.filtering.filtering_model import FilteringModel
+
+from darts.models import NaiveSeasonal, MovingAverage, NBEATSModel, KalmanFilter, GaussianProcessFilter, ARIMA, RegressionModel
+
```
hrzn, line 19: Pay attention to PEP-8 (2 blank lines expected before a class)

darts/anomaly_detection/anomaly_model.py
```diff
@@ -0,0 +1,214 @@
...
+        if isinstance(model, ForecastingModel):
+            self.model = model 
+        else :
+            raise ValueError(
+            "Model must be a darts.models.forecasting not a {}".format(type(model))
+            )
+
+        self.scorer = scorer
+        
+   def fit(self, series, model_fit_kwargs={}, scorer_fit_kwargs={}):
```
hrzn, line 44: Don't use mutable default arguments, see: https://docs.python-guide.org/writing/gotchas/#:~:text=Python's%20default%20arguments%20are%20evaluated,to%20the%20function%20as%20well.

test_jupyter.ipynb
```diff
@@ -0,0 +1,295 @@
+{
```
hrzn, line 1: Don't commit your test code/notebooks

hrzn (commented):
Just a first round of comments, mostly about formatting for now :)

darts/anomaly_detection/anomaly_model.py
```diff
@@ -0,0 +1,357 @@
...
+            hist_forecasts_params = {}
+
+        raise_if_not(
+            self.model._fit_called,
+            f"Model {self.model} has not been trained. Please call .fit()",
+        )
+
+        pred = self.model.historical_forecasts(
+            series, retrain=False, **hist_forecasts_params
+        )
```
dumjax, line 155: Here we can't input any parameters for hist_forecasts_params, as if we choose `last_points_only=False` we will have a list of TimeSeries and not just a TimeSeries. The slice_intersect will not work in this case.

darts/anomaly_detection/anomaly_model.py
```diff
@@ -0,0 +1,357 @@
...
+                    type(scorer)
+                ),
+            )
+
+    def fit(
+        self,
+        series: TimeSeries,
+        model_fit_params: Optional[Dict[str, Any]] = None,
+        hist_forecasts_params: Optional[Dict[str, Any]] = None,
+    ):
```
dumjax, line 89: Subject for future bainstorming:

- For now we want to keep everything general, but isn't this hist_forecasts_params somehow a property of the scorer? And so, when we instantiate a scorer, we would have to input  a subset  of the `historical_forecast` parameters. Typically, the `forecast_horizon`, which. would give a prediction window on which we compute the metrics. 
- Also, where are the covariates ? (or maybe we have chosen to just do univariate for now, but in this case you have to check the type of model and that doesn't accept past and future covariates)
- Can we also have by default the fitting and the scoring training on the fitted model saved series, and then have series in input as optional?

darts/anomaly_detection/anomaly_model.py
```diff
@@ -0,0 +1,357 @@
...
+        else:
+            return anomaly_scores
+
+    def score_metric(
+        self,
+        series: TimeSeries,
+        true_anomalies: TimeSeries,
+        hist_forecasts_params: Optional[Dict[str, Any]] = None,
+        return_metric="AUC_ROC",
+    ):
```
dumjax, line 173: Same here, do we want to add covariates as params?

darts/anomaly_detection/score.py
```diff
@@ -0,0 +1,412 @@
...
+            raise_if_not(
+                self._fit_called,
+                "The Scorer has not been fitted yet. Call `fit()` first",
+            )
+
+            if series_2 is None:
+                series = series_1
+            else:
+                series = self._diff(series_1, series_2)
+
```
dumjax, line 108: Where is _diff define in the base class?

darts/anomaly_detection/anomaly_model.py
hrzn: +1 with these 3 points. We should not expose `hist_forecasts_params`, but rather only expose whichever parameters are relevant to this particular `fit()` function (if any). Also we should accept covariates and multiple series here.

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,362 @@
...
+
+        Parameters
+        ----------
+        series : Darts TimeSeries
+        filter_params: dict, optional
+            parameters of the Darts `.filter()` filtering model
+
+        Returns
+        -------
+        Darts TimeSeries
```
hrzn, line 302: Adapt to e.g. `Union[TimeSeries, List[TimeSeries]]`

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,362 @@
...
+
+        if filter_params is None:
+            filter_params = {}
+
+        pred = self.filter.filter(series, **filter_params)
+
+        anomaly_scores = []
+
+        for i, scorer in enumerate(self.scorers):
+            anomaly_scores.append(scorer.compute(pred, series))
```
hrzn, line 314: What's the best here for downstream use (e.g. with detectors): return a list of univariate series, or a single multivariate series?

hrzn (commented):
I've made a round of comments.

Overall, I think it's a great start and taking shape. Here are the most important points:
* Please, think of what happens both for (i) multivariate series and (ii) multiple series. It seems to me that the code is not general enough in few places about that.
* Unless I'm missing something, I think most scorers (e.g. L2, but also KMeans and others) should act in the original D-dimensional space for D-dimensional multivariate series, and always return univariate series (I don't think it's really done like that currently).
* Somehow, I would expect KMeansScorer and other similar scorers to work over certain rolling windows (kind of what you have for Wasserstein). This is probably where the characteristic time length should play (rather than in some functions keeping e.g. the max values over certain windows as a postprocessing step for all scorers - I would suggest removing this).
* I would let the scorers compute the distance themselves if they need it, rather than feeding them with (some variants) of residuals.
* I haven't looked at Detectors yet
* I have pushed a couple of commits with small improvements (mainly to docstrings), but more corrections need to be made there

darts/ad/anomaly_model.py
julien12234: The problem is that scores do not return the same length because of windowing (for example, WassersteinScorer will return a shorter anomaly score than L2Scorer). So we can not return a multivariate time series containing the anomaly scores of the scorers (unless we add Nan values to make them the same length?)

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,360 @@
...
+        """
+
+        if model_fit_params is None:
+            model_fit_params = {}
+
+        if hist_forecasts_params is None:
+            hist_forecasts_params = {}
+
+        # fit forecasting model
+        if hasattr(self.model, "fit"):
```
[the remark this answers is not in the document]
julien12234: ok

darts/ad/score.py
```diff
@@ -0,0 +1,567 @@
...
+                )
+            ]
+            return TimeSeries.from_times_and_values(
+                series._time_index[self.characteristic_length :], values
+            )
+
+    def score(
+        self,
+        series: Union[TimeSeries, Sequence[TimeSeries]],
+        actual_anomalies: Union[TimeSeries, Sequence[TimeSeries]],
```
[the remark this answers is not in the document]
julien12234: definitely, ok

darts/ad/score.py
```diff
@@ -0,0 +1,567 @@
...
+
+    def __init__(self, characteristic_length) -> None:
+        super().__init__(characteristic_length=characteristic_length)
+        self._fit_called = False
+        self.trainable = True
+
+    def compute(
+        self,
+        series_1: Union[TimeSeries, Sequence[TimeSeries]],
+        series_2: Union[TimeSeries, Sequence[TimeSeries]] = None,
```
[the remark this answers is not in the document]
julien12234: If you would like to compute Kmeans on the original TS itself. Let's suppose the TS is d-dimensional and has a finite number of states. Then applying Kmeans with a window of 1 can be interesting in detecting anomalies. 

For now, only trainable Scorer have this option, every non-trainable Scorer only expects 2 inputs (to compute a norm you need 2 inputs)

darts/ad/score.py
```diff
@@ -0,0 +1,567 @@
...
+        pass
+
+    def fit(
+        self,
+        series_1: Union[TimeSeries, Sequence[TimeSeries]],
+        series_2: Union[TimeSeries, Sequence[TimeSeries]] = None,
+    ) -> Union[float, Sequence[float]]:
+        """Fits the scorer on the given time series input.
+
+        If 2 time series are given, a distance function, given as input, will be applied
```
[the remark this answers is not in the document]
julien12234: What about: "If 2 time series are given, a distance function, given as a parameter in the __init__ method 
        (distance_function), will be applied to transform the 2 time series into 1. Default: "l1""

darts/ad/score.py
```diff
@@ -0,0 +1,567 @@
...
+            Anomaly score time series
+        """
+        self.sanity_check(series_1, series_2)
+
+        series_1, series_2 = self.return_intersect(series_1, series_2)
+
+        return self.window_adjustment_series(self._compute_core(series_1, series_2))
+
+
+class TrainableScorer(Scorer):
```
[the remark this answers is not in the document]
julien12234: So also NonTrainableScorer -> NonFittableScorer ?

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,360 @@
...
+            )
+
+        if i == 0:
+            return anomaly_scores[0]
+        else:
+            return anomaly_scores
+
+
+class FilteringAnomalyModel(AnomalyModel):
+    def __init__(self, filter: FilteringModel, scorer: Union[Scorer, Sequence[Scorer]]):
```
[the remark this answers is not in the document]
julien12234: Okk np, but I was thinking: we have the luxury of having two separate classes, one for filter (FilteringAnomalyModel) and the other one for forecasting (ForecastingAnomalyModel). So why not differentiate them in their input (one accepts a "filter" and the other a "model", maybe we can change "model" into "forecasting" or "forecasting_model")?

darts/ad/score.py
```diff
@@ -0,0 +1,567 @@
...
+            Fitted Scorer.
+        """
+
+        self.sanity_check(series_1, series_2)
+
+        if series_2 is None:
+            series = series_1
+        else:
+            series_1, series_2 = self.return_intersect(series_1, series_2)
+            series = self._diff(series_1, series_2)
```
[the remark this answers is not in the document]
julien12234: done

darts/ad/score.py
```diff
@@ -0,0 +1,567 @@
...
+    def _compute_core(
+        self, series: Union[TimeSeries, Sequence[TimeSeries]]
+    ) -> Union[TimeSeries, Sequence[TimeSeries]]:
+        # return distance to the clostest centroid
+        distance = []
+
+        for i in range(len(series) - self.characteristic_length + 1):
+            distance.append(
+                wasserstein_distance(
+                    self.training_data,
```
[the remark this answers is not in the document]
julien12234: Does it make sense to have a window for the training data? 

- The idea here is to compare the distribution of the residuals of the training data vs a window w in the testing data. (here we would be talking about residuals). For now, we take the entire training data. If we only take a window of the training data, we reduce the number of samples (however, the more, the better because we would like to have the true distribution of our training data). 
- Secondly, if we go with this idea of using a window for the training data, what would be the rule to compare the train and test distribution? Should we compare the window in the training and testing that have identical timestamps? But if the series is stationary (does not depend on time) then it doesn't make sense to compare at the same time range. The signal in the training data at time 5 to 15 has maybe nothing to do with the signal in the testing data at time 5 to 15. 
- Thirdly, what about if we train with a list of time series? 

Were you thinking of partitioning the training data into n window of size w and averaging the distributions? So that in the Wasserstein distance, we compare the same number of samples?

darts/ad/scorers.py
```diff
@@ -0,0 +1,567 @@
...
+        self, series: Union[TimeSeries, Sequence[TimeSeries]]
+    ) -> Union[TimeSeries, Sequence[TimeSeries]]:
+        # return distance to the clostest centroid
+        np_anomaly_score = np.abs(
+            self.model.score_samples(series.all_values().flatten().reshape(-1, 1))
+        )
+        return TimeSeries.from_times_and_values(series._time_index, np_anomaly_score)
+
+
+class L2(NonTrainableScorer):
```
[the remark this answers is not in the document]
julien12234: in process :)

hrzn (commented):
This is only a round of comments (I have mainly reviewed `anomaly_model.py` so far) @julien12234 . It looks very good! I have mostly minor comments. I have also pushed a commit with some improvements to the docstring.

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+        )
+        self.model = model
+
+        super().__init__(model=model, scorer=scorer)
+
+    def fit(
+        self,
+        series: Union[TimeSeries, Sequence[TimeSeries]],
+        allow_model_training: bool = False,
+        model_fit_params: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
julien12234: i changed the name to model_fit_kwargs, and put the parameter at the end. 

Your suggested order has allow_model_training (optional parameter) before series (required parameter). It raises an error, hence I opted for this new order: ```self, series, past_covariates, future_covariates, allow_model_training, forecast_horizon, start, num_samples, model_fit_kwargs```

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+            Internal parameter for error message, a string indicating if it is a "future" or "past" covariates.
+
+        Returns
+        -------
+        Sequence[TimeSeries]
+            Covariate time series
+        """
+
+        if covariates is not None:
+            covariates = (
```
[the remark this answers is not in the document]
julien12234: I checked initially, but there is a slight difference: 

series2seq -> return [ts] if isinstance(ts, TimeSeries) else ts
I need ->         return [ts] if not isinstance(ts, Sequence) else ts

It's important because I want the user to be able to input time-series and sequence of time-series.  
And the first thing I do is convert the input into a sequence, and then I iterate through its element to check if they are of type time-series. With series2seq, the user could input something other than a time-series or sequence, and then my iteration will not work, because I may not have a sequence.

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+            )
+
+        # generate the historical_forecast() prediction of the model on the train set
+        if self.scorers_are_trainable:
+
+            # check if the window size of the scorers are lower than the max size allowed
+            self._check_window_size(list_series, start)
+
+            list_pred = []
+            for idx, series in enumerate(list_series):
```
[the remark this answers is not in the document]
julien12234: yes, waiting on the new version of``` historical_forecasts``` !

darts/ad/utils.py
```diff
@@ -0,0 +1,679 @@
...
+
+    return series_1, series_2
+
+
+def show_anomalies_from_scores(
+    series: TimeSeries,
+    model_output: TimeSeries = None,
+    anomaly_scores: Union[TimeSeries, Sequence[TimeSeries]] = None,
+    window: Union[int, Sequence[int]] = 1,
+    name_of_scorers: Union[str, Sequence[str]] = None,
```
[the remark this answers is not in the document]
julien12234: Why not allow str? if there is only one scorer?

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+        Parameters
+        ----------
+        series
+            The series given to the .historical_forecast()
+        start
+            Parameter of the .historical_forecast(): first point of time at which a prediction is computed
+            for a future time.
+        """
+        if start is None:
+            start = 0.5
```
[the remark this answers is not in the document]
julien12234: Ok! 

In the new version of the historical forecast, the minimal possible start value will be available for each model. I would like to set the default value of ```start``` to its minimal value. For now, it is set to 0.5 if no value is given for ```start```.

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+
+        for scorer in self.scorers:
+            for s in series:
+                max_possible_window = len(s) - len(
+                    s.drop_after(s.get_timestamp_at_point(start))
```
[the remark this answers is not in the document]
julien12234: It gives the same result; I've replaced it with your formula

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+                    start=start,
+                    num_samples=num_samples,
+                )
+            )
+
+        list_anomaly_scores = []
+        for scorer in self.scorers:
+            list_anomaly_scores.append(
+                scorer.score(
+                    list_pred if len(list_pred) > 1 else list_pred[0],
```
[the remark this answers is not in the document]
julien12234: I recently added this condition, to respect a rule: the return type of a function must match the given input type. 

The scorer does accept series OR a list of series. And the rule is:
-  if a series is given -> will return an anomaly series  
- if a list of series is given -> will return a list of anomaly series (even if the series contains only 1 element)
I wanted that the function returns the same type of input. 

The situation here is that our anomaly_model accepts series OR a list of series. And if a series is given, it will be converted into a list, which means that anomaly_model will always manipulate list. As a consequence, when the anomaly_model calls the scorers, it will give as input a list for list_pred and list_series.

However, if the users input was a series and not a list for the anomaly_model, the scorer will nevertheless receive a list, and will return a list of anomaly scores rather than a series of anomaly scores. 

To avoid that, i added these conditions here...
I hope it's clear! And hopefully, there is a cleaner way to solve this issue.

darts/ad/anomaly_model.py
```diff
@@ -0,0 +1,1003 @@
...
+            Number of times a prediction is sampled from a probabilistic model. Should be left set to 1 for
+            deterministic models.
+        metric
+            Optionally, Scoring function to use. Must be one of "AUC_ROC" and "AUC_PR".
+            Default: "AUC_ROC"
+
+        Returns
+        -------
+        Union[float, Sequence[float], Sequence[Sequence[float]]]
+            Score for the time series
```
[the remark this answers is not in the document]
julien12234: Yes of course, just added:
```
float -> if one series is given and the anomaly model contains one scorer 
Sequence[float] -> if one series is given and the anomaly model contains more than one scorer 
Sequence[Sequence[float]]] -> if multiple series are given 
```

darts/ad/anomaly_model.py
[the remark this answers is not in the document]
hrzn: Couldn't you do something like
```python
for scorer in self.scorers:
    list_anomaly_scores += scorer.score(list_pred, list_series)
```
This way all scorers always return a list (could be of length 1), and these lists get concatenated into `list_anomaly_scores`.

darts/ad/utils.py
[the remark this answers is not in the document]
hrzn: Sorry my bad; I thought you were checking for Sequence only later in the method but you also allow strings. Please disregard my comment :)

darts/ad/anomaly_model.py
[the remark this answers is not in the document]
hrzn: Looks good 👍

darts/ad/anomaly_model.py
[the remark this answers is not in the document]
hrzn: I see what you mean. In this case I'm wondering whether we should maybe change `series2seq` itself, because its intended use is (AFAIK) exactly this (we use it throughout the code base in places where we want to seamlessly deal with either single TimeSeries or Sequences of TimeSeries).

darts/ad/scorers.py
```diff
@@ -0,0 +1,1582 @@
...
+            1st sequence of time series
+        series_2:
+            2nd sequence of time series
+
+        Returns
+        -------
+        TimeSeries
+            Sequence of series of length n
+        """
+        list_series = []
```
hrzn, line 342: You seem to be working a lot with lists everywhere.
You could perhaps consider replacing those by generators in some places for efficiency. For instance here:
```python
return map(self._diff, zip(list_series_1, list_series_2))
```

This is more elegant than always relying on for loops, and can be more efficient in some cases as results are not proactively computed. You can instantiate a list at the very end (when returning the result to the user), which triggers the actual computation on the whole generator.

darts/ad/scorers.py
```diff
@@ -0,0 +1,1582 @@
...
+
+
+# FittableAnomalyScorer
+
+
+class GaussianMixtureScorer(FittableAnomalyScorer):
+    """GaussianMixtureScorer anomaly score
+
+    Wrapped around the GaussianMixture scikit-learn function.
+    Source code: <https://github.com/scikit-learn/scikit-learn/blob/f3f51f9b6/sklearn/mixture/_gaussian_mixture.py#L456>
```
hrzn, line 386: I think it would be better to link to the sklearn doc rather than the raw source code. Also check out how to put references (or links) in rst format.

darts/ad/scorers.py
```diff
@@ -0,0 +1,1582 @@
...
+    def __init__(
+        self,
+        window: Optional[int] = None,
+        n_components: int = 1,
+        reduced_function=None,
+        component_wise: bool = False,
+    ) -> None:
+
+        """
+        A Gaussian mixture model is trained on the training data when the ``fit()`` method is called.
```
hrzn, line 398: I find these kinds of comments not very clear. What is the "training data" and what are the "samples" that the log-likelihood is computed on? I think you could rephrase, to explain how exactly the GMM is trained when `fit(actual_series, pred_series)` and `score(actual_series, pred_series)` are called. I think the comment you have below (detailing training) is much more clear. 

If you want to give a gist you could e.g. write something along the lines:
```
Upon calling fit(actual_series, pred_series), the difference between actuals and predictions is obtained, then a moving window is applied, which results in a set of vectors of size W, where W is the window size. The GMM model is trained on these vectors.
```

darts/ad/scorers.py
```diff
@@ -0,0 +1,1582 @@
...
+        The ``score()`` method will compute the log-likelihood of each sample.
+
+        TODO: stride in training is equal to w, and in score stride is equal to 1. Give the option to change
+        these parameters.
+
+        If 2 time series are given in the ``fit()`` or ``score()`` methods, a reduced function, given as a parameter
+        in the __init__ method (reduced_function), will be applied to transform the 2 time series into 1.
+        Default: "abs_diff"
+
+        component_wise is a boolean parameter in the __init__ method indicating how the model should behave with input
```
hrzn, line 408: No need to refer to the `__init__` method. You can simply say it's a parameter of the model (or of the constructor method).

darts/ad/scorers.py
```diff
@@ -0,0 +1,1582 @@
...
+        that is a multivariate series. If set to True, the model will treat each width/dimension of the series
+        independently. If the series has a width of d, the model will train and store the d Gaussian mixture models
+        and fit them on each dimension. If set to False, the model will concatenate the widths in the considered window
+        and compute the score using only one trained Gaussian mixture model.
+
+        Training:
+
+        The input can be a series (univariate or multivariate) or a list of series. The series will be partitioned
+        into equal size subsequences. If the series is multivariate (width>1), then the subsequence will be of size
+        window * width, with window being a given parameter.
```
hrzn, line 418: Here you use "window" and later you define "w" as a parameter. I would recommend you introduce all your notation upfront (e.g. saying say the input is a multivariate series of length `n`, dimension `d`, and we apply a window of size `w`, then ....).
Also use single bactickts for the variable names, it'll be clearer when the doc is rendered.

hrzn (commented):
A round of small comments on the scorers (I haven't yet reviewed them all :) )

darts/ad/scorers/gamma_nll_scorer.py
```diff
@@ -0,0 +1,36 @@
...
+    def __str__(self):
+        return "GammaNLLScorer"
+
+    def _score_core_NLlikelihood(
+        self,
+        deterministic_values: np.ndarray,
+        probabilistic_estimations: np.ndarray,
+    ) -> np.ndarray:
+
+        # TODO: takes a very long time to compute, understand why
```
hrzn, line 31: Here (as all the other likelihoods) you have to vectorize your code. Drop the iteration over the time steps in the for-loop below, estimate the parameters of the beta in a vectorized way (there must be a way :) and then call `logpdf()` only once.

darts/ad/scorers/kmeans_scorer.py
```diff
@@ -0,0 +1,216 @@
...
+
+from darts.ad.scorers.scorers import FittableAnomalyScorer
+from darts.logging import raise_if_not
+from darts.timeseries import TimeSeries
+
+
+class KMeansScorer(FittableAnomalyScorer):
+    def __init__(
+        self,
+        window: Optional[int] = None,
```
hrzn, line 27: Is a window of None equivalent to 1? If yes, set the default value of 1 here for clarity (it's OK to use default values for non-mutable types :) )

darts/ad/scorers/kmeans_scorer.py
```diff
@@ -0,0 +1,216 @@
...
+        k: int = 2,
```
hrzn, line 28: This parameter is so important that I wonder if we shouldn't force the user to provide it.

darts/ad/scorers/kmeans_scorer.py
```diff
@@ -0,0 +1,216 @@
...
+        component_wise: bool = False,
+        diff_fn=None,
```
hrzn: Set the default here (and similar for the other scorers where this applies)
```suggestion
        diff_fn= "abs_diff",
```

darts/ad/scorers/kmeans_scorer.py
```diff
@@ -0,0 +1,216 @@
...
+        `component_wise`.
+
+        A window of size W is rolled on the series with a stride equal to 1. It is the same size window W used during
+        the training phase. At each timestamp, the previous W values will form a vector of size W * D
+        of the series (with D being the series dimensions). The KMeans model will then retrieve the closest centroid
+        to this vector and compute the euclidean distance between the centroid and the vector. The output will be a
+        series of dimension one and length N-W+1, with N being the length of the input series. Each value will represent
+        how anomalous the sample of the W previous values is.
+
+        If a list is given, a for loop will iterate through the list, and the function ``_score_core()`` will be
```
hrzn, line 87: That's too many details for the users (they don't need to know there's a `_score_core()` function etc).

darts/ad/scorers/kmeans_scorer.py
```diff
@@ -0,0 +1,216 @@
...
+        else:
+            models = []
+            for width in range(self.width_trained_on):
+                model = KMeans(n_clusters=self.k)
+                model.fit(
+                    np.concatenate(
+                        [
+                            np.array(
+                                [
+                                    np.array(np_series[i : i + self.window, width])
```
hrzn, line 161: You should definitely try to find way to vectorize this. We can leave it like that for now in this PR, but this should be high prio for later :) Numpy offers ways to perform windowing. Our own `TimeSeries` also does. You should try to rely on those.

In general having for loops over either timestamps or dimensions (width) should be a red flag!

hrzn (commented):
Here's a round of comments. I have looked at anomaly models and scorers for now. I still have to look at detectors, aggregators and utils.py.

Main comments (I have also pushed some commits):
* As far as I could tell the unit tests are mainly concerned about checking argument and return types. Please also include unit tests that do test the correctness of the returned values. For instance create some small end-to-end examples with synthetic data that check what you're doing on your notebooks. It's important that if I modify a scorer later on and inadvertently make the computation incorrect a unit test is able to flag it to me.

* Please move all the computations of NLL to using scipy wherever possible and vectorize the calls (see my comment in the Gaussian NLL class)

* In general there are lots of for loops over time and components of the series in some scorers which will make this code very slow for big series. We can leave some of it like that for now, but it is very important that you work on vectorizing the code wherever possible afterwards.

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,844 @@
...
+)
+from darts.logging import get_logger, raise_if_not
+
+logger = get_logger(__name__)
+
+
+class AnomalyScorer(ABC):
+    "Base class for all anomaly scorers"
+
+    def __init__(self, returns_UTS: bool, window: Optional[int] = None) -> None:
```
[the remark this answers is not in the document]
julien12234: Ok, I changed it to `univariate_scorer` (it makes more sense, I think: it indicates if the scorer will always output a univariate scorer). 

Also changed the function name ``check_return_UTS`` to ``check_univariate_scorer``.

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,352 @@
...
+        raise_if_not(
+            type(return_model_prediction) is bool,
+            f"'return_model_prediction' must be Boolean, found type: {type(return_model_prediction)}",
+        )
+
+        list_series = _to_list(series)
+
+        preds = [self.filter.filter(s, **filter_kwargs) for s in list_series]
+        scores = [
+            [sc.score_from_prediction(s, p) for sc in self.scorers]
```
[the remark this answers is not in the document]
julien12234: What you are suggesting is how it was implemented in the previous version. The reason I changed is the following:

When we have more than 1 scorer and more than 1 series, the output is a Sequence[Sequence[TimeSeries]]. 
Current version:  The outer sequence is over the series, and the inner sequence is over the scorers.
Previous version:  The outer sequence is over the scorers, and the inner sequence is over the series.

I thought it was always better to have the outer sequence over the series (like in the other parts of Darts). 

So we could call score_from_prediction() directly on Sequence[TimeSeries], but then we need to reorder the results. And I haven't found an efficient way (without using flatten(), and reshape()).

For example, if we have two scorers, and 3 series we would need to transform: [[1,2,3],[4,5,6]] -> [[1,4],[2,5],[3,6]]

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,848 @@
...
+
+        Returns
+        -------
+            TimeSeries
+        """
+
+        if self.window == 1:
+            return series
+        else:
+            values = [
```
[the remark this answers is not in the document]
julien12234: Should I leave it to the user with a parameter window_function? (only when window is bigger than 1)

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,848 @@
...
+        list_actual_series, list_pred_series = _to_list(actual_series), _to_list(
+            pred_series
+        )
+        _same_length(list_actual_series, list_pred_series)
+
+        list_fit_series = []
+        for idx, (s1, s2) in enumerate(zip(list_actual_series, list_pred_series)):
+            _sanity_check_2series(s1, s2)
+            s1 = self._check_deterministic(s1, "actual_series")
+            s2 = self._check_deterministic(s2, "pred_series")
```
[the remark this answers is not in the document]
julien12234: For now, we don't have a fittable anomaly scorer that works on stochastic predictions. 

To be cautious, should I put these checks in the _score_core function directly?

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,848 @@
...
+    def _score_core_from_prediction(
+        self,
+        actual_series: TimeSeries,
+        pred_series: TimeSeries,
+    ) -> TimeSeries:
+        """For each timestamp of the inputs:
+            - the parameters of the considered distribution are fitted on the samples of the probabilistic time series
+            - the negative log-likelihood of the determinisitc time series values are computed
+
+        If the series are multivariate, the score will be computed on each width independently.
```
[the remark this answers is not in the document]
julien12234: What I meant:

If the series is univariate or multivariate, the score will be computed on each component independently.
If it is multivariate, it will also be computed on each width independently.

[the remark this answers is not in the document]
julien12234: I have changed the comment to "The score will be computed on each component independently. Additionally, if the series is multivariate, the score will be computed on each width independently."

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,848 @@
...
+        TimeSeries
+        """
+        actual_series = self._check_deterministic(actual_series, "actual_series")
+        self._check_stochastic(pred_series, "pred_series")
+
+        np_actual_series = actual_series.all_values(copy=False)
+        np_pred_series = pred_series.all_values(copy=False)
+
+        np_anomaly_scores = []
+        for width in range(pred_series.width):
```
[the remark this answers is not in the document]
julien12234: I don't understand why it should be called component. We are talking about the width/dimension of the series. 

The idea is that:

_score_core_from_prediction() --> for loops over the widths and calls _score_core_nllikelihood()

_score_core_nllikelihood() --> for loops over the timestamps and at each timestamp, estimate the parameters of the distribution and compute the likelihood 

There is room for improvement! I'll try to remove both of the for loops tomorrow.

darts/ad/scorers/scorers.py
```diff
@@ -765,7 +765,8 @@ def _score_core_from_prediction(
             - the parameters of the considered distribution are fitted on the samples of the probabilistic time series
             - the negative log-likelihood of the determinisitc time series values are computed
 
-        If the series are multivariate, the score will be computed on each width independently.
+        The score will be computed on each component independently. Additionally, if the series is
+        multivariate, the score will be computed on each width independently.
```
hrzn, line 769: Replace "each width" by "each component" (the width is the number of components)

darts/ad/scorers/pyod_scorer.py
```diff
@@ -0,0 +1,136 @@
...
+        if not self.component_wise:
+            self.model.fit(
+                np.concatenate(
+                    [
+                        np.array(
+                            [
+                                np.array(np_series[i : i + self.window])
+                                for i in range(len(np_series) - self.window + 1)
+                            ]
+                        ).reshape(-1, self.window * len(np_series[0]))
```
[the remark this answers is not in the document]
julien12234: A stochastic series will be transformed into a deterministic one in the score() function (by taking the median), + a warning will be thrown (located in the class FittableAnomalyScorer)

darts/ad/scorers/kmeans_scorer.py
julien12234: Agree. It is actually the same code used in:
- Pyod Scorer
- Kmeans Scorer 
- Wasserstein Scorer 

for fitting() and scoring(). So one solution could be used 6 times!

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,848 @@
...
+        else:
+            return anomaly_scores
+
+    def fit(
+        self,
+        series: Union[TimeSeries, Sequence[TimeSeries]],
+    ):
+        """Fits the scorer on the given time series input.
+
+        If sequence of series is given, the scorer will be fitted on the concatenation of the sequence.
```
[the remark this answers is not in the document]
julien12234: For now, our 3 fittable scorers (PyODScorer, KMeansScorer, WassersteinScorer) use this concatenation process when a sequence is given for fitting. 

I thought that "Concatenation of the sequence" == "Union of the windows in all series" ? 

I don't want to talk about the parameter window and the machinery that comes with it at this stage. It is explained in the docstring of each scorer. I wanted only to give a general idea here.

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,696 @@
...
+        start
+            Parameter of the .historical_forecast(): first point of time at which a prediction is computed
+            for a future time.
+        """
+        # biggest window of the anomaly_model scorers
+        max_window = max(scorer.window for scorer in self.scorers)
+
+        for s in series:
+            max_possible_window = (
+                len(s.drop_before(s.get_timestamp_at_point(start))) + 1
```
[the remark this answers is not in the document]
julien12234: I tried to think about a way to compute the number of steps that the historical forecast will be outputing without creating a new series, but it is not trivial: it all depends on the frequency of the series. And there exist a lot of different possible frequencies (so make a condition for each one of them?).

I need to spend more time on it!

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+    def __init__(
+        self,
+        model: FilteringModel,
+        scorer: Union[AnomalyScorer, Sequence[AnomalyScorer]],
+    ):
+        """Filtering-based Anomaly Detection Model
+
+        The filtering model may or may not be already fitted. The underlying assumption is that this model
+        should be able to adequately filter the series in the absence of anomalies. For this reason,
+        it is recommend to either provide a model that has already been fitted and evaluated to work
```
eliane-maalouf, line 32:
```suggestion
        it is recommended to either provide a model that has already been fitted and evaluated to work
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+        In addition, calling :func:`fit()` will also fit the fittable scorers, if any.
+
+        Parameters
+        ----------
+        filter
+            A filtering model from Darts that will be used to filter the actual time series
+        scorer
+            One or multiple scorer(s) that will be used to compare the actual and predicted time series in order
+            to obtain an anomaly score ``TimeSeries``.
+            If a list of `N` scorer is given, the anomaly model will call each
```
eliane-maalouf, line 47:
```suggestion
            If a list of `N` scorers is given, the anomaly model will call each
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def fit(
+        self,
+        series: Union[TimeSeries, Sequence[TimeSeries]],
+        allow_model_training: bool = False,
+        **filter_fit_kwargs,
+    ):
+        """Fit the underlying filtering model (if applicable) and the fittable scorers, if any.
+
+        Train the filter (if not already fitted and `allow_filter_training` is set to True)
```
eliane-maalouf, line 67:
```suggestion
        Train the filter (if not already fitted and `allow_model_training` is set to True)
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+        and the scorer(s) on the given time series.
+
+        The filter model will be applied to the given series, and the results will be used
+        to train the scorer(s).
+
+        Parameters
+        ----------
+        series
+            The (sequence of) series to be trained on.
+        allow_filter_training
```
eliane-maalouf, line 77:
```suggestion
        allow_model_training
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+        series: TimeSeries,
+        actual_anomalies: TimeSeries = None,
+        names_of_scorers: Union[str, Sequence[str]] = None,
+        title: str = None,
+        metric: str = None,
+        **score_kwargs,
+    ):
+        """Plot the results of the anomaly model.
+
+        Computes the score on the given series input. And shows the different anomaly scores with respect to time.
```
eliane-maalouf, line 154:
```suggestion
        Computes the score on the given series input and shows the different anomaly scores with respect to time.
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+
+        The plot will be composed of the following:
+
+        - the series itself with the output of the filtering model
+        - the anomaly score of each scorer. The scorer with different windows will be separated.
```
eliane-maalouf, line 159: > The scorer with different windows will be separated.

Not clear to me what this means. What is meant by separation?

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+        - give personalized names for the scorers with `names_of_scorers`
+        - show the results of a metric for each anomaly score (AUC_ROC or AUC_PR), if the actual anomalies is given
+
+        Parameters
+        ----------
+        series
+            The series to visualize anomalies from.
+        actual_anomalies
+            The ground truth of the anomalies (1 if it is an anomaly and 0 if not)
+        names_of_scorers
```
eliane-maalouf, line 174: > names_of_scorers

maybe this parameter could be renamed to "scorers_names"

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+
+        - the series itself with the output of the filtering model
+        - the anomaly score of each scorer. The scorer with different windows will be separated.
+        - the actual anomalies, if given.
+
+        It is possible to:
+
+        - add a title to the figure with the parameter `title`
+        - give personalized names for the scorers with `names_of_scorers`
+        - show the results of a metric for each anomaly score (AUC_ROC or AUC_PR), if the actual anomalies is given
```
eliane-maalouf, line 166:
```suggestion
        - show the results of a metric for each anomaly score (AUC_ROC or AUC_PR), if the actual anomalies are given
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+
+        Parameters
+        ----------
+        series
+            The series to visualize anomalies from.
+        actual_anomalies
+            The ground truth of the anomalies (1 if it is an anomaly and 0 if not)
+        names_of_scorers
+            Name of the scores. Must be a list of length equal to the number of scorers in the anomaly_model.
```
eliane-maalouf, line 175:
```suggestion
            Name of the scorers. Must be a list of length equal to the number of scorers in the anomaly_model.
```

darts/ad/anomaly_model/filtering_am.py
```diff
@@ -0,0 +1,348 @@
...
+
+        return self
+
+    def show_anomalies(
+        self,
+        series: TimeSeries,
+        actual_anomalies: TimeSeries = None,
+        names_of_scorers: Union[str, Sequence[str]] = None,
+        title: str = None,
+        metric: str = None,
```
eliane-maalouf, line 149: add Optional type ?

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+    def __init__(
+        self,
+        model: ForecastingModel,
+        scorer: Union[AnomalyScorer, Sequence[AnomalyScorer]],
+    ):
+        """Forecasting-based Anomaly Detection Model
+
+        The forecasting model may or may not be already fitted. The underlying assumption is that `model`
+        should be able to accurately forecast the series in the absence of anomalies. For this reason,
+        it is recommend to either provide a model that has already been fitted and evaluated to work
```
eliane-maalouf, line 37:
```suggestion
        it is recommended to either provide a model that has already been fitted and evaluated to work
```

darts/ad/anomaly_model/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+--------------
+
+Anomaly models make it possible to use any of Darts' forecasting
+or filtering models to detect anomalies in time series.
+
+The basic idea is to compare the predictions produced by a fitted model (the forecasts
+or the filtered series) with the actual observations, and to emit an anomaly score
+describing how "different" the observations are from the predictions.
+
+An anomaly model takes as parameter a model and one or multiple scorer objects.
```
eliane-maalouf, line 12:
```suggestion
An anomaly model takes as parameters a model and one or multiple scorer objects.
```

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+        In addition, calling :func:`fit()` will also fit the fittable scorers, if any.
+
+        Parameters
+        ----------
+        model
+            An instance of a Darts forecasting model.
+        scorer
+            One or multiple scorer(s) that will be used to compare the actual and predicted time series in order
+            to obtain an anomaly score ``TimeSeries``.
+            If a list of `N` scorer is given, the anomaly model will call each
```
eliane-maalouf, line 52:
```suggestion
            If a list of `N` scorers is given, the anomaly model will call each
```

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+        Parameters
+        ----------
+        series
+            One or multiple (if the model supports it) target series to be
+            trained on (generally assumed to be anomaly-free).
+        allow_model_training
+            Boolean value that indicates if the forecasting model needs to be fitted on the given series.
+            If set to False, the model needs to be already fitted.
+            Default: False
+        model_fit_params
```
eliane-maalouf, line 94:
```suggestion
        
```

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+            Parameters to be passed on to the forecast model ``fit()`` method.
```
eliane-maalouf, line 95:
```suggestion
           
```

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+
+        Train the model (if not already fitted and ``allow_model_training`` is set to True) and the
+        scorer(s) (if fittable) on the given time series.
+
+        Once the model is fitted, the series historical forecasts are computed,
+        representing what would have been forecasted by this model on the series.
+
+        The prediction and the series are then used to train the scorer(s).
+
+        Parameters
```
eliane-maalouf, line 85: the order of the parameters below does not follow the same order as in the function signature, could be a bit confusing when reviewing

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+
+        return self
+
+    def _prepare_covariates(
+        self,
+        covariates: Union[TimeSeries, Sequence[TimeSeries]],
+        series: Sequence[TimeSeries],
+        name_covariates: str,
+    ) -> Sequence[TimeSeries]:
+        """Convert `covariates` into Sequence, if not already, and checks if its length is equal to the one of `series`.
```
eliane-maalouf, line 223:
```suggestion
        """Convert `covariates` into Sequence, if not already, and checks if their length is equal to the one of `series`.
```

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -0,0 +1,673 @@
...
+        start: Union[pd.Timestamp, float, int] = 0.5,
+        num_samples: int = 1,
+        actual_anomalies: TimeSeries = None,
+        names_of_scorers: Union[str, Sequence[str]] = None,
+        title: str = None,
+        metric: str = None,
+    ):
+        """Plot the results of the anomaly model.
+
+        Computes the score on the given series input. And shows the different anomaly scores with respect to time.
```
eliane-maalouf, line 271:
```suggestion
        Computes the score on the given series input and shows the different anomaly scores with respect to time.
```

darts/ad/scorers/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+- ``fit()`` and ``score()``: directly on a series to uncover the relationships between the different
+  dimensions (over timesteps within windows and/or over dimensions of multivariate series).
+- ``fit_from_prediction`` and ``score_from_prediction``: which will compute a difference (residuals)
+  between the prediction (coming e.g., from a forecasting model) and the series itself.
+  When scoring, the scorer will attribute a higher score to residuals that are distant
+  from the clusters found during the training phase.
+    
+Note that `Anomaly Models <https://unit8co.github.io/darts/generated_api/darts.ad.anomaly_model.html>`_
+can be used to conveniently combine any of Darts forecasting and filtering model with one or multiple scorers.
```
eliane-maalouf, line 35:
```suggestion
can be used to conveniently combine any of Darts forecasting and filtering models with one or multiple scorers.
```

darts/ad/anomaly_model/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+works with Darts filtering models.
+The anomaly models also can be fitted by calling :func:`fit()`, which trains the scorer(s)
+(in case some are trainable), and potentially the model as well.
+
+The function :func:`eval_accuracy()` is the same as :func:`score()`, but outputs the score of an agnostic
+threshold metric ("AUC-ROC" or "AUC-PR"), between the predicted anomaly score time series, and some known binary
+ground-truth time series indicating the presence of actual anomalies.
+Finally, the function :func:`show_anomalies()` can also be used to visualize the predictions
+(in-sample predictions and anomaly scores) of the anomaly model.
+"""
```
eliane-maalouf, line 26: shouldn't there be some imports here ?

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,754 @@
...
+                + f" timeseries (number of samples must be equal to 1, found: {series.n_samples}). The "
+                + "series will be converted to a deterministic series by taking the median of the samples.",
+            )
+            series = series.quantile_timeseries(quantile=0.5)
+
+        return series
+
+    @abstractmethod
+    def __str__(self):
+        "returns the name of the scorer"
```
eliane-maalouf, line 114:
```suggestion
        """returns the name of the scorer"""
```

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,754 @@
...
+
+    @abstractmethod
+    def score_from_prediction(self, actual_series: Any, pred_series: Any) -> Any:
+        pass
+
+    def show_anomalies_from_prediction(
+        self,
+        actual_series: TimeSeries,
+        pred_series: TimeSeries,
+        name_of_scorer: str = None,
```
eliane-maalouf, line 171: > name_of_scorer

it might be better to call this parameter scorer_name

darts/ad/aggregators/aggregators.py
```diff
@@ -0,0 +1,300 @@
...
+# - create show_all_combined (info about correlation, and from what path did
+#   the anomaly alarm came from)
+
+from abc import ABC, abstractmethod
+from typing import Any, Sequence, Union
+
+import numpy as np
+
+from darts import TimeSeries
+from darts.ad.utils import _to_list, eval_accuracy_from_binary_prediction
```
eliane-maalouf, line 19: about _to_list() , there already exists in Darts a function called series2seq() in darts.utils.utils.py, might be worth reusing ?

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,754 @@
...
+        self,
+        actual_series: Union[TimeSeries, Sequence[TimeSeries]],
+        pred_series: Union[TimeSeries, Sequence[TimeSeries]],
+    ) -> Union[TimeSeries, Sequence[TimeSeries]]:
+        """Computes the anomaly score on the two (sequence of) series.
+
+        The function ``diff_fn`` passed as a parameter to the scorer, will transform `pred_series` and `actual_series`
+        into one "difference" series. By default, ``diff_fn`` will compute the absolute difference
+        (Default: "abs_diff").
+        If actual_series and pred_seriesare sequences, ``diff_fn`` will be applied to all pairwise elements
```
eliane-maalouf, line 503:
```suggestion
        If actual_series and pred_series are sequences, ``diff_fn`` will be applied to all pairwise elements
```

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,754 @@
...
+
+    def fit(
+        self,
+        series: Union[TimeSeries, Sequence[TimeSeries]],
+    ):
+        """Fits the scorer on the given time series input.
+
+        If sequence of series is given, the scorer will be fitted on the concatenation of the sequence.
+
+        The assumptions is that the series `series` used for training are generally anomaly-free.
```
eliane-maalouf, line 555:
```suggestion
        The assumption is that the series `series` used for training are generally anomaly-free.
```

darts/ad/scorers/scorers.py
```diff
@@ -0,0 +1,754 @@
...
+            else:
+                return series.window_transform(
+                    transforms={
+                        "window": self.window,
+                        "function": "mean",
+                        "mode": "rolling",
+                        "min_periods": self.window,
+                    },
+                    treat_na="dropna",
+                )
```
eliane-maalouf, line 740: 🥇 for the window_transform() ;)

darts/ad/scorers/kmeans_scorer.py
```diff
@@ -0,0 +1,201 @@
...
+        Each value in the score series thus represents how anomalous the sample of the `W` previous values is.
+
+        Parameters
+        ----------
+        window
+            Size of the window used to create the subsequences of the series.
+        k
+            The number of clusters to form as well as the number of centroids to generate by the KMeans model.
+        diff_fn
+            Optionally, reduced function to use if two series are given. It will transform the two series into one.
```
eliane-maalouf, line 92:
```suggestion
            Optionally, reduction function to use if two series are given. It will transform the two series into one.
```

darts/ad/scorers/nll_exponential_scorer.py
```diff
@@ -0,0 +1,37 @@
...
+        self,
+        deterministic_values: np.ndarray,
+        probabilistic_estimations: np.ndarray,
+    ) -> np.ndarray:
+
+        # This is the ML estimate for 1/lambda, which is what scipy expects as scale.
+        mu = np.mean(probabilistic_estimations, axis=1)
+
+        # This is ML estimate for the loc - see:
+        # https://github.com/scipy/scipy/blob/de80faf9d3480b9dbb9b888568b64499e0e70c19/scipy/stats/_continuous_distns.py#L1705
```
eliane-maalouf, line 34: Useful to have the above comment in the scorers docstring

hrzn: I don't know... don't want to overload users with details of how things are computed :D

eliane-maalouf (commented):
AD is a great addition to the Darts @julien12234 and @hrzn . Thanks for this colossal work :)

darts/ad/scorers/nll_cauchy_scorer.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, window: int = 1) -> None:
+        super().__init__(window=window)
+
+    def __str__(self):
+        return "CauchyNLLScorer"
+
+    def _score_core_nllikelihood(
+        self,
+        deterministic_values: np.ndarray,
+        probabilistic_estimations: np.ndarray,
```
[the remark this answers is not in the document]
hrzn: You're right but this is a private function, so I'd leave that for later :)

hrzn (approved):
🥳

PR #1257, opened 2022-10-04
Fix encoder transformer and remove absolute index encoder

dennisbader:
### Summary
Fixes the following issues:
- encoder transformer failed when only using a cyclic encoder
- encoder transformer was fit on all series encodings from fit(). This lead to bad results when the series used at predict() do not represent the same series. Now the transformer is globally fit on all observed values per encoded component
- Removed absolute index position encoder and only keep relative index position encoder. Relative basically carries the same information as absolute just that it uses the end of the target series as the reference point. This also makes more sense for transformations, as the maximum value is always 0 when fitting. Users can always implement their own absolute idx logic through the „custom“ encoder (CallableIndexEncoder)


hrzn (approved):
LGTM, nice job @dennisbader !

BlackFireAlex:
Relative doesn't necessarily carry the same info as absolute gives you the lifespan. Am I wrong ?

PR #1265, opened 2022-10-06
Fix/ts gaps

### LINKED ISSUE #629: Add option to choose 'or' mode in `TimeSeries.gaps()`
Right now for multivariate TS, gaps are only considered if all components are nan, but it should be possible to treat periods where any value is nan as gap too.

to be discussed with @pennfranc

madtoinou:
Fixes #629, #647 and #1225 

### Summary

Took code suggested by @phubermi in #629, corrected it and added the corresponding tests.


darts/tests/test_timeseries.py
```diff
@@ -1111,13 +1116,20 @@ def test_gaps(self):
                 )
             ).all()
         )
+        gaps7 = series7.gaps()
+        self.assertTrue(gaps7.empty)
```
hrzn, line 1120: cool, thanks for these tests!

hrzn (commented):
Only have some comments related to rendering of the docstrings. Otherwise it looks great, thanks @madtoinou !

hrzn (approved):
Nice, thanks!

PR #1203, opened 2022-09-09
Issue 1079: Add windowed features

adamkells:
Implements the windowing of features. Relates to [issue 1079](https://github.com/unit8co/darts/issues/1079).

### Summary

Adds a function within RegressionModel to create new windowed features from a user dictionary which can contain either a user function or a string to implement a pre-defined function.

PR #1269, opened 2022-10-09
Feat/window transformer

### LINKED ISSUE #1079: window transformer
In addition to "lag features" which we already have, it'd be nice to add "window features", specifying window characteristics and corresponding function(s) to apply to create features dynamically in regression models. For instance, it is often helpful to use the trailing mean and variance of the last N points as features. We could also imagine having a way to have fairly generic windows (e.g., "last month", "last week", "the N points starting N-k time steps ago", etc...

eliane-maalouf:
Fixes #1079. ~Draft for code review.~

### Summary

Implement window features generation as transformer that can be called as a standalone transformation, from pipeline ~or from forecasting model~.

### Other Information

Implementation started by @adamkells in https://github.com/unit8co/darts/pull/1203#:~:text=from-,adamkells,feat/window%2Dfeatures,-%2B675%20%E2%88%920


darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                    all(
+                        c_idx <= series[s_idx].width - 1
+                        for s_idx in range(len(series))
+                        for c_idx in transformation["comp_id"]
+                    ),
+                    "Some components are not available in the provided series.",
+                )
+
+                series_subset += [
+                    (s.univariate_component(c), transformation, builtins)
```
hrzn, line 382: supporting component names instead of indexes, you could do something like `s[c].values()`

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                            builtins,
+                        )
+                        for (s_id, c_id) in itertools.product(
+                            transformation["series_id"], transformation["comp_id"]
+                        )
+                    ]
+
+        return iter(series_subset)  # the iterator object for ts_transform function
+
+    def ts_transform(
```
hrzn, line 442: could be nice to add the `@staticmethod` decorator for clarity here too

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+            "bfill"
+            if (
+                ("navalue" not in kwargs)
+                or ("navalue" in kwargs and kwargs["navalue"] is None)
+            )
+            else None,
+        )
+        navalue = kwargs.get("navalue", None)
+
+        BUILTIN_TRANSFORMS = builtins
```
hrzn, line 558: Couldn't you maybe define the builtins outside of the class as a constant, in order to access them easily from everywhere?

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+
+        fn = transformation["function"]
+
+        df_ts = series.pd_dataframe(
+            copy=True
+        )  # get the series values in a dataframe. TODO: check if copy is necessary
```
hrzn, line 564: If you don't mutate `df_ts` later on you can (should) use `copy=False`

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                        ),
+                        "Some components are not available in the provided series.",
+                    )
+                    series_subset += [
+                        (
+                            series[s_idx].univariate_component(c_idx),
+                            transformation,
+                            builtins,
+                        )
+                        for (s_idx, c_idx) in itertools.chain(
```
hrzn, line 408: This is a little hard to read 😅 Some intermediate variables could be welcome.

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                    copy_transformation["window"] = window
+                    # get function_group and function kwargs
+                    function_group_kwargs, function_kwargs = _get_function_kwargs(
+                        copy_transformation, BUILTIN_TRANSFORMS
+                    )
+
+                    if "closed" not in function_group_kwargs:
+                        function_group_kwargs[
+                            "closed"
+                        ] = "left"  # to garantee that the latest value is not included in the window: forecasting safe
```
hrzn, line 588: How about when "closed" is in the kwargs? Would this still be "forecasting safe"?

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+            function_name = BUILTIN_TRANSFORMS[fn][1]
+
+            # if rolling, we need to run through window list
+            if function_group == "rolling":
+                transf_ts = []  # list to get all windows resulting transformations
+                for window in transformation["window"]:
+                    copy_transformation = copy.deepcopy(
+                        transformation
+                    )  # to avoid writing the original dict
+                    copy_transformation["window"] = window
```
hrzn, line 579: This live copying and editing of the dict with slightly different types seems a bit funny.. couldn't we find a cleaner way?

darts/dataprocessing/transformers/window_transformer.py
hrzn: How about implementing most of this logic in a `TimeSeries` function - e.g. `TimeSeries.window_transform()` ? That could be nice because it would allow users to obtain transformed series very simply without even instantiating a transformer (similar to how they can call `TimeSeries.map()` to apply a function on the series). WDYT?

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+
+from darts.dataprocessing.transformers import BaseDataTransformer
+from darts.logging import get_logger, raise_if_not, raise_log, raise_user_warning
+from darts.timeseries import TimeSeries
+from darts.utils.utils import series2seq
+
+logger = get_logger(__name__)
+
+
+class ForecastingWindowTransformer(BaseDataTransformer):
```
hrzn, line 15: Two high level questions.
1) It seems somehow this doesn't work on multivariate series? I'd have expected to receive a multivariate series back with two windowed components when doing this:
image

2) What's the expected behavior when using several window functions? in the example below I see only one series being returned:
image

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                        "Only one or the other can be provided. You can either set one of them "
+                        "to None or not provide it as keyword in the transform() function call."
+                    ),
+                    logger,
+                )
+
+        return TimeSeries.from_dataframe(transf_ts)
+
+    # TODO: move to forecasting model class
+    def construct_transformation_dictionary(cls, series_configurations):
```
hrzn, line 670: Can be removed from here

hrzn (commented):
This looks like a good start @eliane-maalouf. My main comments are mainly
* I'm not sure I understand the behavior with multivariate series. I'd have expected window transformed applied on a multivariate series to return a new multivariate series with windowing applied component-wise.
* I see you iterate over series (that's OK) and components when extracting the values. Would be nicer for performance reasons to extract values for all components at once.
* How about moving the core of the windowing logic to a function `TimeSeries.window_transform()` ? It would allow users to simply get windowed series without instantiating a transformer, similar to `TimeSeries.map()`. Then this transformer could simply call `TimeSeries.window_transform()`. (see `Mapper` transformer). WDYT?

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+            (i.e., pandas.DataFrame.rolling or pandas.DataFrame.ewm) or for the specific function in that group
+            (i.e., pandas.DataFrame.rolling.mean/std/max/min... or pandas.DataFrame.ewm.mean/std/sum).
+            Example of use:
+
+                    .. highlight:: python
+                    .. code-block:: python
+                        from darts.dataprocessing.transformers import ForecastingWindowTransformer
+                        all_series = [series_1, series_2, series_3] # each series could have multiple components
+                        window_transformations_1 = [{'function':'mean','window':[3],'series_id': 0,'comp_id':[0,1,2]},
+                                                    {'function':'quantile', 'window':[3, 5, 30], 'quantile':0.5}]
```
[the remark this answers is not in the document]
eliane-maalouf: true, I agree

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: yes, can do

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: yep, can do

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: if the user provides a value for closed that is not "left" then no, we cannot guarantee forecasting safe. Should we warn the user about this if they provide the "closed" item? We can give the user a warning, but the computation can still go through, or we can raise an Error and block the computation. I guess we have to always provide forecasting safe behaviour, correct?

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: yes, true. The issue is that I have filtered components on which I apply the transformations and this loses the structure of the original series, because now they become only lists of univariate_series. I will revisit the filtering part and make sure that series structures are preserved. 
and for the multiple transformation case, there should be two resulting transformed series. I am not sure why they are not plotted both. Thank you for bringing that up.

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: yes

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+        n_jobs: int = 1,
+        verbose: bool = False,
+    ):
+        """
+        A transformer that applies window transformation to a TimeSeries or a Sequence of TimeSeries. It expects a
+        dictionary or a list of dictionaries.
+
+        Parameters
+        ----------
+        window_transformations
```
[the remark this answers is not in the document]
eliane-maalouf: indeed, if the user's function is provided without a rolling key, I assume that the user's function is taking care of creating a window and moving it or applying some other form of mapping. 
We can also only provide a default rolling behaviour to apply a user's function.

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+    # - key: name of the function to be provided by the user
+    # - value: (function_group, function, ([mandatory arguments for the function_group],
+    #                                       [mandatory arguments for the function]))
+    BUILTIN_TRANSFORMS = {
+        "count": (
+            pd.DataFrame.rolling,
+            "count",
+            (["window"], []),
+        ),
+        "sum": (pd.DataFrame.rolling, "sum", (["window"], [])),
```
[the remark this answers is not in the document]
eliane-maalouf: done and moved also to timeseries.py file

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+             1) the function is provided along a 'rolling':True item and a 'window' value: in this case the function is
+             applied to a rolling window of the input series.
+             2) no 'rolling':True item is provided, the function is applied as is to the input. The provided function
+              should take a pandas.DataFrame as input and return a pandas.DataFrame as output with the necessary column
+              names and index formatting to allow for correct following casting into a TimeSeries object. The provided
+              function does not need to cast the pandas.DataFrame into a TimeSeries.
+
+            When using the builtin functions, the 'window' key should be provided for the
+            pandas.DataFrame.rolling functions group.
+            The 'window' value should be a positive integer representing the size of the window to be used for the
```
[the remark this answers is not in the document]
eliane-maalouf: window lists will not be supported as per previous previous review comments

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+    #  standalone or from model (useful to set column names for example)
+    def __init__(
+        self,
+        window_transformations: Union[dict, List[dict]],
+        name: str = "ForecastingWindowTransformer",
+        n_jobs: int = 1,
+        verbose: bool = False,
+    ):
+        """
+        A transformer that applies window transformation to a TimeSeries or a Sequence of TimeSeries. It expects a
```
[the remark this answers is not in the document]
eliane-maalouf: updated in timeseries.py window_transform() function

darts/dataprocessing/transformers/window_transformer.py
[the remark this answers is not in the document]
eliane-maalouf: update: 'rolling' is not a required key anymore for user provided functions, as we will only support rolling windows over the underlying dataframe. please check updated version in timeseries.py window_transform()

darts/dataprocessing/transformers/window_transformer.py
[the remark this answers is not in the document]
eliane-maalouf: 'comp_id' becomes 'components'

no window list support anymore 

series_id list will only be available when using the transformer class

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: logic moved to TimeSeries.window_transform()

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: I added user warning and forecasting_safe enforcement in TimeSeries.window_transform(). The default configuration of the function is to be forecasting safe

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: update : those issued should be solved now in TimeSeries.window_transform()

- if multivariate series, user can choose on which component to apply transformation and can choose if to return a TimeSeries with only the transformed components or the transformed components along the non transformed ones. The original components of the transformed ones are not returned.
- if series is stochastic, all samples are transformed identically, as per the specified corresponding component transformation 
-  if multiple transformations are desired, the result is a multivariate series with the different transformed data as new components. Similar behavior as previous for selecting and filtering components to apply transformation or not.

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: not relevant anymore

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                    else:
+                        raise_log(
+                            ValueError(
+                                f"`window_transformation` at index {idx} must contain a positive integer or a list "
+                                f"of positive integers for the 'window'. "
+                            ),
+                            logger,
+                        )
+
+                if "step" in transformation:
```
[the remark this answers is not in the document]
eliane-maalouf: update : I kept 'step' as a key for this argument, because it is being automatically filtered along other potentially user provided arguments that I cannot tell in advance (more flexibility, but user should know a bit more if they want to configure more things). This key is an advanced usage, ideally the user should know what they are doing with it, if not, we set it to the safe value of 1 and they should not worry about it.

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+
+        # check if series_id and comp_id exist in the series list
+        # run through the transformations
+        series_subset = []
+        for idx, transformation in enumerate(self.window_transformations):
+            # check whether different transformations are to be performed for different series and different components
+            if "series_id" not in transformation and "comp_id" not in transformation:
+                # apply the transformation to all series and all components
+                series_subset += [
+                    (s.univariate_component(c), transformation, builtins)
```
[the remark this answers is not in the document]
eliane-maalouf: I always need to return TimeSeries here since window_transform() is applied on TimeSeries

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,691 @@
...
+                    for s_idx in transformation["series_id"]
+                    for c in range(series[s_idx].width)
+                ]
+
+            elif "series_id" not in transformation and "comp_id" in transformation:
+                # apply the transformation to all series on a specific component only in each series
+                # test that component exists in each relevant series
+                raise_if_not(
+                    all(
+                        c_idx <= series[s_idx].width - 1
```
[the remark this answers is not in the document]
eliane-maalouf: It now supports component names only, do we need to add ids ? 
with component names, some scenarios might not make much sense

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: with the new version, supporting component names, we don't need to extract values, I pass the whole series, all the transformation are then managed in TimeSeries.window_transform()

darts/dataprocessing/transformers/window_transformer.py
eliane-maalouf: simplified with the new version, should be clearer

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+                    ],
+                    axis=1,
+                )
+
+            elif isinstance(
+                transformation["function"], Callable
+            ):  # user provided function
+                new_columns.extend(
+                    [
+                        f"userFn{transformation['window'] if 'window' in transformation else idx}_{comp_name}"
```
hrzn, line 3528:
```suggestion
                        f"udf_{transformation['window'] if 'window' in transformation else idx}_{comp_name}"
```
("udf" is a common shorthand for "user-defined function")

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+                    [
+                        resulting_transformations,
+                        ts_df[filter_df_columns]
+                        .rolling(**function_group_kwargs)
+                        .apply(fn, **function_kwargs),
+                    ],
+                    axis=1,
+                )
+
+            else:
```
hrzn, line 3542: wasn't this checked before?

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+                raise_log(
+                    ValueError("The transformation function is not valid."), logger
+                )
+
+            # filter out non_transformed components
+            if keep_non_transformed:
```
hrzn, line 3548: The logic here seems to have some issues. For instance I'd have expected all components to be kept with `keep_non_transformed` is set. Yet in the example below it keeps only one of the two:
image

hrzn: Also in this example it seems the user provided function is the same as mean (note: I've set both transforms to apply on first component only) ?
image

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+                    f"NAs were replaced by provided value {treat_na}.",
+                    logger,
+                )
+        elif forecasting_safe:
+            resulting_transformations.dropna(inplace=True)
+            if not suppress_warnings:
+                raise_user_warning(
+                    forecasting_safe,
+                    "Enforcing forecasting safe. Resulting series is truncated by dropping NAs.",
+                    # TODO: ASSUMES NAs ARE AT THE BEGINNING OF THE SERIES, IS THIS A SAFE ASSUMPTION?
```
hrzn, line 3584: Good point. We don't want to drop all NAs I think, as some might come later in the series. Could we only drop the first few time stamps affected (by looking at the window length, "center" and "closed"?

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+        verbose
+            Whether to print operations progress
+        """
+        super().__init__(name, n_jobs, verbose)
+
+        # dictionary checks are mostly implemented in TimeSeries.window_transform()
+        # here we only need to verify that the input is not None
+        # and that 'series_id', if provided, is a list of positive integers
+
+        if window_transformations is None:
```
hrzn, line 122: Similar as for TimeSeries, no need to check this explicitly I think

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+                ValueError(
+                    "window_transformations argument should be provided and "
+                    "must be a non-empty dictionary or a non-empty list of dictionaries."
+                ),
+                logger,
+            )
+
+        if window_transformations is not None:
+
+            raise_if_not(
```
hrzn, line 133: to remove (rely on TimeSeries checks)

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+                    window_transformations
+                ]  # if only one dictionary, make it a list
+
+            for idx, transformation in enumerate(window_transformations):
+                raise_if_not(
+                    isinstance(transformation, dict),
+                    f"`window_transformations` must contain dictionaries. Element at index {idx} is not a dictionary.",
+                )
+
+                if (
```
hrzn, line 156: to remove

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+                    "series_id" in transformation
+                    and transformation["series_id"] is None
+                ):
+                    window_transformations[idx].pop("series_id")
+
+                if (
+                    "series_id" in transformation
+                    and transformation["series_id"] is not None
+                ):
+                    raise_if_not(
```
hrzn, line 166: I think the 20 lines below could be replaced by something like
```python
series_id = transformation["series_id"]
series_id = series_id if isinstance(series_id, list) else [series_id]
raise_if_not(all([isinstance(i, int) and i >= 0 for i in series_id), 
             "series_id must be positive integers")
```

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+                        ),
+                        f"`window_transformation` at index {idx} must contain a positive integer >= 0 for the "
+                        f"'series_id', or a non-empty list containing positive integers >= 0. ",
+                    )
+                    if isinstance(transformation["series_id"], int):
+                        window_transformations[idx]["series_id"] = [
+                            transformation["series_id"]
+                        ]  # make list
+
+                if (
```
hrzn, line 186: to remove

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+        series_subset = []
+        for idx, transformation in enumerate(self.window_transformations):
+
+            if "series_id" not in transformation and "components" not in transformation:
+                # apply the transformation to all series and all components
+                series_subset += [(s, transformation) for s in series]
+
+            elif "series_id" in transformation and "components" not in transformation:
+                # apply the transformation to a specific series and all its components
+                raise_if_not(
```
hrzn, line 210: Let the sequence type catch this for you :)

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+        self, series: Union[TimeSeries, Sequence[TimeSeries]]
+    ) -> Iterator[Tuple[TimeSeries, dict]]:
+
+        series = series2seq(series)
+
+        # run through the transformations
+        series_subset = []
+        for idx, transformation in enumerate(self.window_transformations):
+
+            if "series_id" not in transformation and "components" not in transformation:
```
hrzn, line 204: I think you could probably simplify all these cases by going along those lines (approximatively):
```python

all_series_ids = list(range(len(series)))

for idx, transformation in enumerate(self.window_transformations):
  series_ids = transformation["series_id"] if "series_id" in transformation else all_series_ids
  for s in series:
    if "components" in transformation:
      yield [(s[transformation["components"]], transformation)]
    else:
      yield [(s, transformation)]
```
(note also using `yield` for the iterator, if that's possible).

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+
+            else:
+                # apply the transformation to a specific component in a specific series
+                # if a different component is provided for each selected series
+                if all(
+                    isinstance(x, list) for x in transformation["components"]
+                ) and len(transformation["components"]) == len(
+                    transformation["series_id"]
+                ):
+                    # testing that components exist in the corresponding series is implemented
```
hrzn, line 234: Yes, also really you should let the TimeSeries component-indexation mechanism crash and give you the relevant error - no need to check component correctness yourself ever in the windowing mechanism.

hrzn (commented):
That's some good foundations for the windowing @eliane-maalouf, thanks for that! I think the core of it, relying on pandas and separating the stochastic series over several columns is good.

There are still quite a few things to change, and behaviour that might need correction (I think I spotted a couple of bugs).

As an overall comment, I would really recommend that you try to *simplify* your code to the maximum possible. For instance you do many checks, many of which can be either simply skipped (delegated to another part of the code responsible for raising the exception) or drastically simplified. I've put quite a few examples of how that could look like in the comments. Each line of code is a small debt for our future selves who'll have to maintain it, so it's good to minimise it when we can. Small is often beautiful :)

Also in terms of structure, I think you could separate a bit more two steps in the TimeSeries function:
* First do all the checks required (should hold in few lines of code only hopefully)
* Then assume everything is checked and correct, and apply the logic.
This way you can avoid inter-mingling some computation code with correctness- or type-checking code.

darts/timeseries.py
```diff
@@ -1401,25 +1432,48 @@ def pd_dataframe(self, copy=True) -> pd.DataFrame:
...
+            comp_name = list(self._xa.get_index(DIMS[1]))
+            samples = range(self.n_samples)
+            df_col_names = [
+                "_s".join((comp_name, str(sample_id)))
+                for comp_name, sample_id in itertools.product(comp_name, samples)
+            ]
+
+            if copy:
+                return pd.DataFrame(
+                    self._xa.stack(data=(DIMS[1], DIMS[2])).values.copy(),
```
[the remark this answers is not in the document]
eliane-maalouf: `stack()` returns a view on the underlying numpy ndarray 

Here is the verification I ran : 

```
import pandas as pd
import numpy as np

series = TimeSeries.from_times_and_values(pd.date_range("20200101", "20200106"), np.array([1,2,3,4,5,6]))
series = series.stack(series + 5)
series = series.concatenate(series, axis = 2) # multi-variate and stochastic

orig_values = series._xa.values # original underlying numpy ndarray

DIMS = ("time", "component", "sample")
stacked_values = series._xa.stack(data=(DIMS[1], DIMS[2])).values # stacked numpy ndarray as produced in pd_dataframe()

print(orig_values.base) # = None 
print(stacked_values.base) # = orig_values
print(stacked_values.base is orig_values) # = True
print(np.shares_memory(stacked_values, orig_values)) # = True
```

darts/timeseries.py
```diff
@@ -2721,7 +2775,7 @@ def with_hierarchy(self, hierarchy: Dict):
             )
         )
 
-    def stack(self, other: "TimeSeries") -> "TimeSeries":
+    def stack(self, other: "TimeSeries", axis=1) -> "TimeSeries":
```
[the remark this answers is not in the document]
eliane-maalouf: reverted, no actual need for it

darts/timeseries.py
```diff
@@ -1401,25 +1432,48 @@ def pd_dataframe(self, copy=True) -> pd.DataFrame:
...
+                raise_user_warning(
+                    True,
+                    "You are transforming a stochastic TimeSeries (i.e., contains several samples). "
+                    "The resulting DataFrame is a 2D object with all samples on the columns. "
+                    "If this is not the expected behavior consider calling a function "
+                    "adapted to stochastic TimeSeries like quantile_df().",
+                    logger,
+                )
+
+            comp_name = list(self._xa.get_index(DIMS[1]))
```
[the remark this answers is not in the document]
eliane-maalouf: :sweat_smile:

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
 
         return self.__class__(new_xa)
 
+    def window_transform(
+        self,
+        window_transformations: Union[Dict, Sequence[Dict]],
```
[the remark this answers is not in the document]
eliane-maalouf: :ok_hand:

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+        treat_na: Optional[Union[str, Union[int, float]]] = None,
+        forecasting_safe: Optional[bool] = True,
+        target: Optional["TimeSeries"] = None,
+        keep_non_transformed: Optional[bool] = False,
+        store_window_transformation: Optional[bool] = False,
```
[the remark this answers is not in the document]
eliane-maalouf: yes indeed, I added it for troubleshooting, will remove it when I am done with the modifications

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+            All other dictionary items provided will be treated as keyword arguments for the function group
+            (i.e., ``pandas.DataFrame.rolling`` or ``pandas.DataFrame.ewm``) or for the specific function
+            in that group (i.e., ``pandas.DataFrame.rolling.mean/std/max/min...`` or
+            ``pandas.DataFrame.ewm.mean/std/sum``).
+            This allows for more flexibility in configuring how the window slides over the data, by providing for
+            example:
+
+            * ``"center"``: ``True``/``False`` to set the observation at the current timestep at the
+              center of the windows (default is False),
+            * ``"closed"``: ``"right"``/``"left"``/``"both"``/``"neither"`` to specify whether the right,
```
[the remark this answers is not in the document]
eliane-maalouf: true my doc mixed the two. 'closed' means where the interval is closed (hence whether the values at the borders of the interval are included or not). Since we allow the current time step to be included in the window, then this parameter will not be important anymore, we can actually leave it to default ('right') or specify that both ends are included ('both').

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+            For user-provided functions, extra arguments in the transformation dictionary are passed to the function.
+            Darts expects user-provided functions to receive numpy arrays as input. This can be modified by adding
+            item ``"raw": False`` in the transformation dictionary. It is expected that the function returns a single
+            value for each window. Other possible configurations can be found in the
+            `pandas.DataFrame.Rolling().apply() documentation: <https://pandas.pydata.org/docs/reference/window.html>`_.
+
+        treat_na
+            String to specify how to treat missing values in the resulting TimeSeries. Can be 'dropna' to truncate the
+            TimeSeries and drop observations with missing values, 'bfill' to specify that NAs should be filled with the
+            last valid observation. Can also be a value, in which case NAs will be filled with this value.
```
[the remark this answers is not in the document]
eliane-maalouf: I am not sure what is a good default for`treat_na`. If nothing is specified, I would assume the user wants to leave the NAs in the series?! is this a good assumption?

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+
+        forecasting_safe
+            If True, Darts enforces that the resulting TimeSeries is safe to be used in forecasting models as target
+            or as features. This parameter guarantees that the window transformation will not include any future values
+            in the current timestep and will not fill NAs with future values. Default is True.
+            Only pandas.DataFrame.Rolling functions can be currently guaranteed to be forecasting-safe.
```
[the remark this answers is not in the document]
eliane-maalouf: I verified the ewm behavior, it is safe by default, it computes the window from observation at time step 0 until the current timestep. It does not look into the future.

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+                            itertools.chain(
+                                *_builtin_transforms[transformation["function"]][2]
+                            )
+                        ).intersection(set(transformation.keys()))
+                        == set(
+                            itertools.chain(
+                                *_builtin_transforms[transformation["function"]][2]
+                            )
+                        ),
+                        f"`window_transformation` dictionary at index {idx} must contain at least the following keys "
```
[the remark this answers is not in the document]
eliane-maalouf: the error message that would be raised here for missing arguments , without the above code, would come from pandas. I have no problem with that. I thought it might be clearer/easier to control the set of mandatory arguments on Darts side and notify the user, but I see that this is creating more redundancy with pandas than true usefulness. I can remove those checks and this indeed removes the need to generate the _builtin_transforms dictionary, which main use was to check the mandatory arguments!

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+
+                             * Based on ``pandas.DataFrame.rolling`` windows, ``"function"`` should map to one of
+                               ``{"count", "sum", "mean", "median", "min", "max", "std", "var", "skew", "kurt",
+                               "corr", "cov", "quantile", "sem", "rank"}``.
+                             * Based on ``pandas.DataFrame.ewm`` (Exponentially-weighted window),
+                               ``"function"`` should be one of ``{"mean", "var", "std", "corr", "cov", "sum",
+                               "skew"}``, prefixed by ``"ewm_"``. For example, ``"function": "ewm_mean"``
+                               or ``"function": "ewm_sum"``.
+
+            :``"window"``: the size of the window (in number of time steps) to be used for the transformation.
```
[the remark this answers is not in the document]
eliane-maalouf: this would be supported by default, since it is managed by pandas

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+
+            :``"function"``: the name of one of the builtin transformation functions, or to a callable function
+                             that can be applied to the input series by using ``pandas.DataFrame.rolling object``.
+
+                             The two following options are available for built-in transformation functions:
+
+                             * Based on ``pandas.DataFrame.rolling`` windows, ``"function"`` should map to one of
+                               ``{"count", "sum", "mean", "median", "min", "max", "std", "var", "skew", "kurt",
+                               "corr", "cov", "quantile", "sem", "rank"}``.
+                             * Based on ``pandas.DataFrame.ewm`` (Exponentially-weighted window),
```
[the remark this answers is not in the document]
eliane-maalouf: I added the "group" key to allow the selection of "rolling", "ewm" or "expanding".

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+                               applied to all components.
+
+            All other dictionary items provided will be treated as keyword arguments for the function group
+            (i.e., ``pandas.DataFrame.rolling`` or ``pandas.DataFrame.ewm``) or for the specific function
+            in that group (i.e., ``pandas.DataFrame.rolling.mean/std/max/min...`` or
+            ``pandas.DataFrame.ewm.mean/std/sum``).
+            This allows for more flexibility in configuring how the window slides over the data, by providing for
+            example:
+
+            * ``"center"``: ``True``/``False`` to set the observation at the current timestep at the
```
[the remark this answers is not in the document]
eliane-maalouf: True, probably the best thing to do here is to make sure center=False with forecasting_safe = True. For closed, if user doesn't specify anything, then just enforce closed=both to have both ends of the window are included.

darts/timeseries.py
eliane-maalouf: I set `keep_non_transformed` such that we return only the non_transformed components along the transformed ones and not all components. Wouldn't there be possible issues if we return the original components as well if say we might also have lagged data in a model? or am I just thinking too much :sweat_smile:

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+
+        # run through all transformations in window_transformations
+        for idx, transformation in enumerate(window_transformations):
+            fn = transformation["function"]
+
+            if (
+                "closed" not in transformation or forecasting_safe
+            ):  # forecasting_safe takes over "closed"
+                if not suppress_warnings:
+                    raise_user_warning(
```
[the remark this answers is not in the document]
eliane-maalouf: ok , raised with conflicting center instead of closed

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,258 @@
...
+            1) Based on pandas.DataFrame.Rolling windows, the 'function' key should have one of
+                {BUILTIN_TRANSFORMS_WINDOW}.
+            2) Based on pandas.DataFrame.ewm (Exponentially-weighted window), the 'function' key should have one of
+                {BUILTIN_TRANSFORMS_EWM} prefixed by 'ewm_'. For example, 'function': 'ewm_mean', 'ewm_sum'.
+
+            The 'window' key should be provided for built-in and for user provided callable functions.
+            The 'window' value should be a positive integer representing the size of the window to be used for the
+            transformation.
+
+            Two optional keys can be provided for more flexibility: 'components' and 'series_id':
```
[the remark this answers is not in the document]
eliane-maalouf: the case here is to allow a sequence of timeseries as input to the transformer. If not, then user needs to build a separate transformer for each series indeed. I think the decision here will depend on what we allow other transforms in a pipeline to do, if other transformers accept sequence of timeseries, then we have to allow this transformer to accept a sequence of timeseries, if not, it won't fit in a pipeline, or is there another way to do it and I don't see it ?

darts/timeseries.py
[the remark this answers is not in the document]
eliane-maalouf: I defaulted it to "both"

darts/timeseries.py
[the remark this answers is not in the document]
eliane-maalouf: default here would be None , meaning user wants to leave NaNs in the series

darts/timeseries.py
```diff
@@ -3005,6 +3037,598 @@ def map(
...
+            if fn == "quantile":
+                return (pd.DataFrame.rolling, fn, (["window"], ["quantile"]))
+            else:
+                return (pd.DataFrame.rolling, fn, (["window"], []))
+
+        def _mk_entry_ewm(fn):
+            return (pd.DataFrame.ewm, fn, ([], []))
+
+        _builtin_transforms = {
+            fn: _mk_entry_rolling(fn) for fn in BUILTIN_TRANSFORMS_WINDOW
```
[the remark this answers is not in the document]
eliane-maalouf: nope :)

eliane-maalouf:
@hrzn what do you think of this new version?

darts/timeseries.py
```diff
@@ -2766,7 +2792,7 @@ def stack(self, other: "TimeSeries") -> "TimeSeries":
         Return
         ------
         TimeSeries
-            A new multivariate TimeSeries instance.
+            A new multivariate and/or multi_sample TimeSeries instance.
```
hrzn, line 2795: I think this change could be reverted.
```suggestion
            A new multivariate TimeSeries instance.
```

darts/timeseries.py
```diff
@@ -3066,6 +3092,379 @@ def map(
...
+            in that mode (i.e., ``pandas.DataFrame.rolling.mean/std/max/min...`` or
+            ``pandas.DataFrame.ewm.mean/std/sum``).
+            This allows for more flexibility in configuring the transformation, by providing for
+            example:
+
+            * :``"window"``: Size of the moving window for the "rolling" mode.
+                            If an integer, the fixed number of observations used for each window.
+                            If an offset, the time period of each window.
+            * :``"min_periods"``: The minimum number of observations in the window required to have a value (otherwise
+                NaN). Darts reuses pandas defautls of 1 for "rolling" and "expanding" modes and of 0 for "ewm" mode.
```
hrzn, line 3140:
```suggestion
                NaN). Darts reuses pandas defaults of 1 for "rolling" and "expanding" modes and of 0 for "ewm" mode.
```

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,150 @@
...
+
+from darts.dataprocessing.transformers import BaseDataTransformer
+from darts.logging import get_logger
+from darts.timeseries import TimeSeries
+from darts.utils.utils import series2seq
+
+logger = get_logger(__name__)
+
+
+class ForecastingWindowTransformer(BaseDataTransformer):
```
hrzn, line 16: Since it's not necessarily for forecasting only, could we rename it to `WindowTransformer` ?

darts/timeseries.py
```diff
@@ -3066,6 +3092,379 @@ def map(
...
+        original_components = self.components
+        n_samples = self.n_samples
+        original_index = self._time_index
+
+        resulting_transformations = pd.DataFrame()
+        new_columns = []
+        added_na = []
+
+        # run through all transformations in transforms
+        for idx, transformation in enumerate(transforms):
```
hrzn, line 3356:
```suggestion
        for transformation in transforms:
```

darts/tests/dataprocessing/transformers/test_window_transformations.py
```diff
@@ -0,0 +1,416 @@
...
+
+        transformed_ts = self.series_multi_det.window_transform(
+            transforms=window_transformations
+        )
+        self.assertEqual(len(transformed_ts.components), 2)
+
+        transformed_ts = self.series_multi_det.window_transform(
+            transforms=window_transformations, keep_non_transformed=True
+        )
+        self.assertEqual(len(transformed_ts.components), 6)
```
hrzn, line 179: (nitpicking) it would be a tiny bit better to test the actual component names (incl the original ones)

hrzn (commented):
This looks really great @eliane-maalouf ! I think it's nearly ready to be merged :) I've added a bunch of tiny comments, nothing important. I have also pushed 1 commit to make the transformer be listed in the documentation page.

Could you remove the draft notebook before merging? We can later transform it into a nice user-facing example notebook :) [I've created a new task for that](https://github.com/unit8co/darts/issues/1389)

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -0,0 +1,150 @@
...
+        self,
+        transforms: Union[dict, List[dict]],
+        treat_na: Optional[Union[str, Union[int, float]]] = None,
+        forecasting_safe: Optional[bool] = True,
+        keep_non_transformed: Optional[bool] = False,
+        name: str = "ForecastingWindowTransformer",
+        n_jobs: int = 1,
+        verbose: bool = False,
+    ):
+        """
```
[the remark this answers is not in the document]
eliane-maalouf: 👍

darts/tests/dataprocessing/transformers/test_window_transformations.py
eliane-maalouf: not sure I understand, I proposed something based on what I understood. Using sets to work with the column names was messing up the order in fact, good thing you asked me to check the component names.

eliane-maalouf:
> This looks really great @eliane-maalouf ! I think it's nearly ready to be merged :) I've added a bunch of tiny comments, nothing important. I have also pushed 1 commit to make the transformer be listed in the documentation page.
> 
> Could you remove the draft notebook before merging? We can later transform it into a nice user-facing example notebook :) [I've created a new task for that](https://github.com/unit8co/darts/issues/1389)

Thanks @hrzn for the feedback. I removed the draft notebook and included your other comments.

hrzn (approved):

jope35:
nice to see that this feature is in!
very good work all

PR #1291, opened 2022-10-14
Refactor/1080 onehsot regressionmodels

### LINKED ISSUE #1080: "one-shot" RegressionModels
Currently, our RegressionModels will create `k` internal models when `output_chunk_length=k`. This can scale poorly for larger values of `k`. 

Another option, which (AFAIK) is often used in Kaggle competitions, is to train only one model, taking as input features that can all be computed at least `k` time steps in advance such as:
* Sufficiently old lags
* Sufficiently old window features
* Future covariates

Pros:
* Seems to work very well in many cases (often relying on good future covariates capturing time axis too)
* Scales very well

Cons:
* Cannot access recent lags, but this does not always matter

It'd be nice to provide such an option in Darts. It could be either an option in `RegressionModel`, or potentially a new kind of `RegressionModel` altogether.

It might be nice to couple it with adding `encoders` for RegressionModels, which could dynamically create the future covariates (e.g., time axis features) on the fly.

eliane-maalouf:
Fixes #1080.

### Summary
Refactors the RegressionModel class to provide a k_step ahead prediction with one model. 
The option is still available to allow a model per future lag.


darts/models/forecasting/regression_model.py
```diff
@@ -508,7 +517,9 @@ def predict(
 
         # check that the input sizes of the target series and covariates match
         pred_input_dim = {
-            "target": series[0].width,
+            "target": series[
+                0
+            ].width,  # DO ALL SERIES IN A GIVEN SEQUENCE HAVE THE SAME WIDTH?
```
hrzn, line 522: Yes, that's a pre-requisite.

darts/models/forecasting/regression_model.py
```diff
@@ -578,7 +599,10 @@ def predict(
         series_matrix = None
         if "target" in self.lags:
             series_matrix = np.stack(
-                [ts[self.lags["target"][0] :].values(copy=False) for ts in series]
+                [
+                    ts[self.lags["target"][0] - shift :].values(copy=False)
```
hrzn, line 603: (nitpicking) I think it would be better to first extract the Numpy array and then index it, rather than doing the indexing on the TimeSeries first. That's because indexing a TimeSeries creates a new TimeSeries, which can be relatively expensive.

hrzn (commented):
This looks very solid to me as far as I can tell. Great job @eliane-maalouf 👏 and a super nice improvement!!

Before we can merge, could you add unit tests?
Also, you need to propagate the support to LinearRegressionModel, LGBM and Catboost (it's just about adding the parameters there).

darts/models/forecasting/regression_model.py
eliane-maalouf: I see, I did not want to change too much the existing logic, but I can correct that

darts/utils/data/tabularization.py
```diff
@@ -86,10 +87,17 @@ def _create_lagged_data(
...
+                    df_target.shift(-future_target_lag).rename(
+                        columns=lambda x: f"{x}_horizon_lag{future_target_lag}"
+                    )
+                )
+        else:
             df_y.append(
-                df_target.shift(-future_target_lag).rename(
-                    columns=lambda x: f"{x}_horizon_lag{future_target_lag}"
+                df_target.shift(-output_chunk_length + 1).rename(
+                    columns=lambda x: f"{x}_horizon_lag{output_chunk_length}"
```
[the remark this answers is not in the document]
eliane-maalouf: indeed

hrzn (approved):
LGTM, great job @eliane-maalouf

PR #1292, opened 2022-10-15
Feat/dlinear nlinear

hrzn:
PR ready for review.

Builds on top of https://github.com/unit8co/darts/pull/1247 effort started by @gdevos010 

* Add probabilistic support for both DLinear and NLinear
* I've extracted a few hyperparameters (whether to init layers with a constant, share the weights across components, and the kernel size for DLinear).
* The case with weights sharing cannot support covariates (since it implies a 1-1 mapping between input and output dimensions). I've adapted the docstrings and weight sharing defaults to False
* I have (I believe) improved the efficiency of the implementation w.r.t. [the original repo](https://github.com/cure-lab/LTSF-Linear) by removing the for-loop iteration over components in the "individual" case and replaced this by a single linear layer of appropriate dimensions. It changes a bit the `forward()` logic too.
* In my understanding, the normalization part of NLinear does not apply to the probabilistic case in general (or at least not without a bit of extra logic), as it does not make sense to add original domains values to arbitrary distribution parameters (e.g. variance) => for the time being, I made the normalization optional, and disabled support for probabilistic likelihoods in case it is enabled.

@gdevos010 Could you test this and let me know what you think?

We can close your original PR (your commits are on this PR anyway) if you agree with these changes.

Done:
- [x] Unit tests
- [x] An attempt to make these models support future covariates and maybe static covariates. I think it shouldn't be too hard.


gdevos010:
@hrzn Thank you for the help, I will test it out soon

hrzn:
Did you get any change to test it @gdevos010 ?

gdevos010:
@hrzn Ive been super busy these past two weeks but should have time this weekend

gdevos010:
@hrzn These changes look great! Thanks for the help, I know I've been MIA. Here are the results of the test I ran.

[plots.zip](https://github.com/unit8co/darts/files/9934791/plots.zip)

eliane-maalouf:
Hello, during the writing of some unittests for this model (nbeats like tests) I noticed that Dlinear and Nlinear models have trouble forecasting a linear series, which is a bit unexpected given their architecture. 
Attached are the tests, the thing is that they require 5x more epochs than nbeats for a simple problem as this one and still their results are quite off (+ increase in input_chunk_length). 
I also changed that random_state from 42 to 30, 100, 50 , nbeats results worsen with all random_state values except 42 ! Screenshot 2022-11-04 at 17 23 21
Screenshot 2022-11-04 at 17 18 20

gdevos010:
i tried to update some of the documentation but it looks like I don't have write permission to this branch

hrzn:
> i tried to update some of the documentation but it looks like I don't have write permission to this branch

@gdevos010 I think the simplest would be for you to fork off this branch, and then open a PR into this branch (i.e., select `feat/dlinear-nlinear` instead of `master` as the target when opening the PR).

hrzn:
> Hello, during the writing of some unittests for this model (nbeats like tests) I noticed that Dlinear and Nlinear models have trouble forecasting a linear series, which is a bit unexpected given their architecture. Attached are the tests, the thing is that they require 5x more epochs than nbeats for a simple problem as this one and still their results are quite off (+ increase in input_chunk_length). I also changed that random_state from 42 to 30, 100, 50 , nbeats results worsen with all random_state values except 42 ! Screenshot 2022-11-04 at 17 23 21 Screenshot 2022-11-04 at 17 18 20

With enough epochs I found it to perform good enough on those kinds of linearly growing series (both NLinear and DLinear)
image

PR #1307, opened 2022-10-22
Feat/encoder cleanup

dennisbader:
### Summary

This is the first of two PRs for encoder improvements (two PRs to make it more simple to review). It only covers variable renaming and moving of encoder files.

- move encoders from `darts/utils/data/` into `darts/dataprocessing/encoders`
- rename variables to be consistent with forecasting models (i.e. covariate -> covariates)

Should I rename `target` to `series` as well?


hrzn (approved):
LGTM, thanks!
IMO you can leave the name `target`.

PR #1311, opened 2022-10-24
#1310 Fixed ShapExplainer for regression models with lags_future_covariates

### LINKED ISSUE #1310: [BUG] Output ShapExplainer for regression models too short when lags_future_covariates equals (x,y) where y > 1
First of all, big thanks for addding the ShapExplainer for regression models in the latest release - this is super useful! It also find it to work in a straigtforward and user friendly way - great work :).

**Description of problem**
The behaviour that I found odd was that the TimeSeries representing the explanations was shorter than I expected when we use the ShapExplainer with a regression model (e.g., LightGBM) with lags_future_covariates (x,y) where y > 1. More specifcially, the max_date in the TimeSeries was smaller than what I'd expect. When I changed the lags_future_covariates to (x,1) or omitted future covariates entirely I did not notice this behaviour (and got back a longer TimeSeries). 

**Believed root cause**
The root cause (I think) is that at a certain moment we are truncating the covariates dataframe (df_cov) by reindexing it to df_target. This is fine for past_covariates but troublesome for future_covariates. Why? Well, in a later step we shift covariates in df_cov. However, since df_cov has been truncated, we cannot shift back the values, resulting in (unnessary) NaNs that are filtered out later on. Here's an example as illustration:

df_cov with future covariates before calling reindex():
image

df_cov with future covariates after calling reindex():
image

When we then do the shifting we get NaNs for the last row. These NaNs are then filtered out causing , down the line, the  TimeSeries representing the explanations to be shorter than it needs to be.

image

Here's the part of the `_create_lagged_data` helper where the reindexing occurs:
   
```
        for idx, (df_cov, lags_cov) in enumerate(covariates):

            if idx == 0:
                covariate_name = "past"
            else:
                covariate_name = "future"
            if lags_cov:

                if not is_training:
                    # We extend the covariates dataframes
                    # to have the same timestamps as the target at the end
                    #  so that when we create the lags with shifts
                    # we don't have nan on the last rows. Only useful for inference.
                    df_cov = df_cov.reindex(df_target.index)

                for lag in lags_cov:
                    df_X.append(
                        df_cov.shift(-lag).rename(
                            columns=lambda x: f"{x}_{covariate_name}_cov_lag{lag}"
                        )
                    )
```

**Proposed solution**
I think the solution would be to just do the reindexing only for past_covariates (and not for future convarites). Or potentially do a dynamic check at runtime to see if the end_time() of one of the covariates frames is smaller than the target series and only then do the reindexing.

 - Python version: [e.g. 3.9]
 - darts version [e.g. 0.22.0]

rijkvandermeulen:
Fixes #1310 

### Summary
ShapExplainer with a regression model (e.g., LightGBM) with lags_future_covariates (x,y) where y > 1 resulted in a TimeSeries representing the explanations that is shorter than it needs to be. For more info please check the linked issue.

The problem was related to reindexing that always happened whereas in fact we only want this to happen if `df_cov.index[-1] < df_target.index[-1]`

### Other Information
Also added two units tests


dumjax:
Hi @rijkvandermeulen , nice catch!

More generally speaking, I also think we have the same problem if covariates have older timestamps than target - we miss some lags probably at the beginning too. Would a more general approach like : 

- We get the union of  the cov DateTimeIndex (or RangeIndex) and target DateTimeIndex , so we have the extended index which works in all cases
- we do a reindex on this union range or datetime index.

WDYT?

rijkvandermeulen:
> Hi @rijkvandermeulen , nice catch!
> 
> More generally speaking, I also think we have the same problem if covariates have older timestamps than target - we miss some lags probably at the beginning too. Would a more general approach like :
> 
> * We get the union of  the cov DateTimeIndex (or RangeIndex) and target DateTimeIndex , so we have the extended index which works in all cases
> * we do a reindex on this union range or datetime index.
> 
> WDYT?

@dumjax good point - I guess we indeed have a similar issue if covariates have older timestamps than target and the `lags` parameter in the regression model is None. I think your suggestion of reindexing on the union of the datetime indices covers all (edge) cases and is the more elegant way of doing it. See changes in latest commits :)

dumjax (approved):
Nice add @rijkvandermeulen , many thanks for the effort!

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -372,6 +373,105 @@ def test_explain(self):
 
         self.assertTrue(isinstance(shap_explain.explain(), ExplainabilityResult))
 
+    def test_explain_with_lags_future_covariates_series_of_same_length_as_target(self):
```
hrzn, line 376: I like it a lot when the tests are much longer than the added code :)

hrzn (approved):
Thanks a lot @rijkvandermeulen and good catch 👏

PR #1318, opened 2022-10-26
init

dumjax:
### Summary

Add the possibility to historical forecast, backtest and compute residuals on multiple time series.

Also automatically detects the starting point if `start` is not specified

### DIFF
--- a/darts/models/forecasting/forecasting_model.py
+++ b/darts/models/forecasting/forecasting_model.py
@@ -41,6 +41,7 @@ from darts.utils.timeseries_generation import (
     _build_forecast_series,
     _generate_new_dates,
 )
+from darts.utils.utils import series2seq
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


maximilianreimer:
Thanks for the great work! Is there a apporximate time line for when this is going to be part of a realease?

darts/models/forecasting/forecasting_model.py
```diff
@@ -241,6 +260,230 @@ def min_train_series_length(self) -> int:
...
+            if max_future_cov_lag
+            else 0
+            - min(
+                min_target_lag if min_target_lag else 0,
+                min_past_cov_lag if min_past_cov_lag else 0,
+                min_future_cov_lag if min_future_cov_lag else 0,
+            )
+        )
+
+    def _get_historical_forecastable_time_index(
```
hrzn, line 367: If this is called only from `historical_forecasts()`, how about moving it inside the `historical_forecasts()` function?

darts/models/forecasting/forecasting_model.py
```diff
@@ -241,6 +260,230 @@ def min_train_series_length(self) -> int:
...
+        """
+        Private function that returns the largest time_index representing the subset of each timestamps
+        for which historical forecasts can be made, given the model's properties, the training series
+        and the covariates.
+            - If ``None`` is returned, there is no point where a forecast can be made.
+
+            - If ``is_training=False``, it returns the time_index subset of predictable timestamps.
+
+            - If ``is_training=True``, it returns the time_index subset of trainable timestamps. A trainable
+            timestamp is a timestamp that have at least a training sample of length ``self.training_sample_length``
```
hrzn, line 383:
```suggestion
            timestamp is a timestamp that has a training sample of length at least ``self.training_sample_length``
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -241,6 +260,230 @@ def min_train_series_length(self) -> int:
...
+            Optionally, a future covariates.
+        is_training
+            Whether the returned time_index should be taking into account the training.
+
+        Returns
+        -------
+        Union[pd.DatetimeIndex, pd.RangeIndex, None]
+            The longest time_index that can be used for historical forecasting.
+
+        Examples
```
hrzn, line 403: +1 for the example (unfortunately the whole docstring won't be rendered for private methods, but still useful)

darts/models/forecasting/forecasting_model.py
```diff
@@ -331,11 +578,14 @@ def historical_forecasts(
         Parameters
         ----------
         series
-            The target time series to use to successively train and evaluate the historical forecasts.
+            The (or a sequence of) target time series to use to successively train and evaluate
```
hrzn, line 581:
```suggestion
            One or multiple target time series to use to successively train and evaluate
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -353,6 +603,12 @@ def historical_forecasts(
             `series` that will be used as first prediction time.
             In case of ``pandas.Timestamp``, this time stamp will be used to determine the first prediction time
             directly.
+            If `start` is not specified, the first prediction time will automatically be set to :
+                 - the first predictable point if `retrain` is False
+
+                 - the first trainable point if `retrain` is True and `train_length` is None
+
+                 - the first trainable point + `train_length` otherwise
```
hrzn, line 611: Maybe a small sentence afterwards explaining precisely what is a "predictable point" and a "trainable point" might be welcome here (users may wonder).

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+        past_covariates = series2seq(past_covariates)
+        future_covariates = series2seq(future_covariates)
 
-        # build the prediction times in advance (to be able to use tqdm)
-        last_valid_pred_time = self._get_last_prediction_time(
-            series, forecast_horizon, overlap_end
-        )
+        # If the model has never been fitted before using historical_forecasts,
+        # we need to know if it uses past or future covariates. The only possible assumption is that
+        # the user is using the same covariates as he would in the fit method.
```
hrzn, line 705: :) 
```suggestion
        # the user is using the same covariates as they would in the fit method.
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -384,18 +640,12 @@ def historical_forecasts(
             Whether to print progress
         Returns
         -------
-        TimeSeries or List[TimeSeries]
-            By default, a single ``TimeSeries`` instance created from the last point of each individual forecast.
-            If `last_points_only` is set to False, a list of the historical forecasts.
+        TimeSeries or List[TimeSeries] (or Sequence[TimeSeries] or Sequence[List[TimeSeries]])
```
hrzn, line 643: Actually the returned Sequence will always be a List, right? You could indicate it in the type hints.

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+        # If the model has never been fitted before using historical_forecasts,
+        # we need to know if it uses past or future covariates. The only possible assumption is that
+        # the user is using the same covariates as he would in the fit method.
+        if self._fit_called is False:
+            if past_covariates is not None:
+                self._uses_past_covariates = True
+            if future_covariates is not None:
+                self._uses_future_covariates = True
+
+        forecasts_list = []
```
hrzn, line 712: Couldn't we maybe use a generator (yielding the results one-by-one) rather than keeping everything in memory and blocking the call until everything is computed?

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+                    self.training_sample_time_index_length
+                    if retrain
+                    else self.predict_sample_time_index_length
+                )
+                * series_.freq
+            )
+
+            if historical_forecasts_time_index is None:
+                raise_log(
+                    ValueError(
```
hrzn, line 741: Might be good to indicate which of the series (idx) causes an issue.

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+            # prepare the start parameter -> pd.Timestamp
+            if start is not None:
+
+                historical_forecasts_time_index = drop_before_index(
+                    historical_forecasts_time_index,
+                    series_.get_timestamp_at_point(start),
+                )
+
+            else:
+                if (retrain is not False) or (not self._fit_called):
```
hrzn, line 758:
```suggestion
                if retrain or (not self._fit_called):
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+                            historical_forecasts_time_index,
+                            historical_forecasts_time_index[0] + 1 * series_.freq,
+                        )
+
+                    # if retrain is False, fit hasn't been called yet and train_length None,
+                    # it means that the entire backtesting will be based on a set of two training samples
+                    # at the first step, so we warn the user.
+                    if (
+                        (not self._fit_called)
+                        and (retrain is False)
```
hrzn, line 785:
```suggestion
                        and not retrain
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+                        and (not train_length)
+                    ):
+                        logger.warning(
```
hrzn, line 788: I wonder whether we should not rather raise an exception in this case. It sounds like almost always a bad (and unexpected) behavior to have a single training done on 2 samples.

darts/models/forecasting/forecasting_model.py
```diff
@@ -444,105 +696,202 @@ def historical_forecasts(
...
+                            "Model has not been fitted yet, `start` and train_length are not specified. "
+                            " The model is not retraining during the historical forecasts. Hence the "
+                            "the first and only training will be done on 2 samples which could lead to "
+                            "hazardous results."
+                        )
+
+            # build the prediction times in advance (to be able to use tqdm)
```
hrzn, line 795: Shall we also (maybe optionally) use tqdm at the outer-loop level? If you keep a list and don't move to an iterator, I would say it'll be very important for users to be able to track progress at the outer loop level.

We could have maybe two levels of iterations tracking when verbose=True for the two outer and inner loops (I think it should be possible).

darts/models/forecasting/forecasting_model.py
```diff
@@ -554,13 +903,14 @@ def backtest(
         ] = metrics.mape,
         reduction: Union[Callable[[np.ndarray], float], None] = np.mean,
         verbose: bool = False,
-    ) -> Union[float, List[float]]:
+    ) -> Union[float, List[float], Sequence[float], List[Sequence[float]]]:
 
         """Compute error values that the model would have produced when
```
hrzn, line 908: Same comments as for historical_forecasts() docstring

darts/models/forecasting/forecasting_model.py
```diff
@@ -934,52 +1303,49 @@ def residuals(
...
+            logger,
         )
 
+        residuals_list = []
         # compute residuals
-        series_trimmed = series.slice_intersect(p)
-        residuals = series_trimmed - (
-            p.quantile_timeseries(quantile=0.5) if p.is_stochastic else p
-        )
+        for idx, target_ts in enumerate(series):
```
hrzn, line 1322: Here too a generator might be cleaner and more scalable. If we go this way, we could change the check above (`all([serie.is_univariate for serie in series])` and check each series one-by-one inside the loop, instead of having the eager version.

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -0,0 +1,735 @@
...
+from darts.utils import timeseries_generation as tg
+from darts.utils.likelihood_models import GaussianLikelihood
+
+try:
+    import torch
+
+    TORCH_AVAILABLE = True
+except ImportError:
+    logger = get_logger(__name__)
+    logger.warning("Torch not installed - will be skipping Torch models tests")
```
hrzn, line 30:
```suggestion
    logger.warning("Torch not installed - will be skipping historical forecasts tests")
```
Could we maybe simply skip torch models from the tests but keep the other tests in case torch is not available?

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -0,0 +1,735 @@
...
+                    + 1  # because we include the first element
+                )
+
+                self.assertTrue(
+                    len(forecasts[0]) == len(forecasts[1]) == theorical_forecast_length,
+                    f"Model {model_cls.__name__} does not return the right number of historical forecasts in the case "
+                    "of retrain=True and overlap_end=False, and a time index of type DateTimeIndex.",
+                )
+
+        @unittest.skipUnless(
```
hrzn, line 453: I think anyway the interpreter won't reach this line if torch is not available.

docs/source/conf.py
```diff
@@ -52,7 +52,7 @@
     + "TransferableFutureCovariatesLocalForecastingModel,GlobalForecastingModel,TorchForecastingModel,"
     + "PastCovariatesTorchModel,FutureCovariatesTorchModel,DualCovariatesTorchModel,MixedCovariatesTorchModel,"
     + "SplitCovariatesTorchModel,TorchParametricProbabilisticForecastingModel,"
-    + "min_train_series_length,uses_future_covariates,uses_past_covariates,"
```
hrzn, line 55: Actually, could this become a documented property, which users could query to determine whether a given instance accepts or expects certain covariates? WDYT? Might be useful.

hrzn (commented):
Looks great @dumjax, I only put minor comments.
Maybe the main one is a proposal to use a generator instead of eagerly filling up a list. I think this could scale better and avoid blocking calls until all results are computed. However, I think it could be done in a later PR, as this could maybe be done in other parts of Darts as well.
The second fairly important comment would be managing the progress tracking of the outer loop.

darts/models/forecasting/forecasting_model.py
dumjax: Actually the test is not False is needed, as retrain can be something else than False or True, typically 0. (I had such problems in unit tests)

darts/models/forecasting/forecasting_model.py
dumjax: it actually what I meant, as  if `the user would have made the same choice if he had used fit before historical_forecasts()`

darts/models/forecasting/forecasting_model.py
dumjax: yes, we might want to do that indeed. More generally, probably not allowing a training under a certain threshold of sample numbers? I fixed 2, but it might be more like a percentage of the total length of the predictable forecast timeindex?

darts/models/forecasting/forecasting_model.py
hrzn: Oh, OK!

darts/models/forecasting/forecasting_model.py
hrzn: I think I understand your sentence, but that was an attempt to make it gender neutral as not all our users are male :D

darts/models/forecasting/forecasting_model.py
hrzn: Any other threshold except the trivial one (as here) is risky IMO... how would we choose the percentage?

darts/models/forecasting/forecasting_model.py
hrzn: In the meantime I "fixed" this by having a tqdm iterator on the outer loop only if there are >1 series (and otherwise using tqdm in the inner loop).

hrzn (approved):
Nice one @dumjax 👏 
I have made a few changes directly in your branch. I think we can merge...

PR #1320, opened 2022-10-27
Solve an issue reusing an existing name

hrzn:
We were reusing the name `idx` of the outer loop in an inner loop. I've fixed it and slightly changed the iterator.


dennisbader (approved):
Looks good, thanks!

PR #1324, opened 2022-10-28
xai corrections + _create_lagged_data corrections

dumjax:
Some corrections on xai and tabularization

### DIFF
--- a/darts/utils/data/tabularization.py
+++ b/darts/utils/data/tabularization.py
@@ -92,6 +92,9 @@ def _create_lagged_data(
                     columns=lambda x: f"{x}_horizon_lag{future_target_lag}"
                 )
             )
+        # we add an empty row if we are in predict mode, as we also want the potential predictable next point
+        if not is_training:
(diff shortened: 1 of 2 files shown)


hrzn:
Is this draft or ready for review?

darts/utils/data/tabularization.py
```diff
@@ -92,6 +92,9 @@ def _create_lagged_data(
                     columns=lambda x: f"{x}_horizon_lag{future_target_lag}"
                 )
             )
+        # we add an empty row if we are in predict mode, as we also want the potential predictable next point
+        if not is_training:
+            df_target.loc[df_target.index[-1] + 1 * df_target.index.freq, :] = np.nan
```
hrzn, line 97: Is this unit-tested? Is there always a `freq` attribute to the `df_target.index` ?

dumjax: Indeed, well spotted. I actually realised the unit test are not treating the RangeIndex case. I will add that in this branch as well

Issue #1261, opened 2022-10-05
[BUG] Deepcopy fails on list of TimeSeries


DavidKleindienst:
**Describe the bug**
Deepcopy fails on a list of TimeSeries with the following Error:

`TypeError: __deepcopy__() takes 1 positional argument but 2 were given`

Presumably this is caused by TimeSeries.\_\_deepcopy\_\_ accepting only `self` as argument, while \_\_deepcopy\_\_ should accept `memo` as second argument (see [Documentation](https://docs.python.org/3/library/copy.html))

**To Reproduce**

```python
from darts.timeseries import TimeSeries
import pandas as pd
import numpy as np
import datetime as dt
from copy import deepcopy

# Create some timeseries
df = pd.DataFrame(
    {
        "time": pd.date_range(dt.datetime(2021, 7, 1), freq="D", periods=62),
        "vals": np.random.randn(62),
    })
ts = TimeSeries.from_dataframe(df, time_col="time", value_cols=["vals"])

lst = [ts, ts.copy(), ts.copy()]
# Try to deepcopy the list
x = deepcopy(lst)
```

**Expected behavior**
Returns a deepcopy of the list

**System (please complete the following information):**
 - Python version: 3.10
 - darts version: 0.21.0

thomktz:
You're right. Also, the `deep` parameter in the current implementation of copy is not used:

```python 
class TimeSeries:
      [...]

      def __copy__(self, deep: bool = True):
           return self.copy()

      def __deepcopy__(self):
            return self.__class__(self._xa.copy())
```

Should probably be 

```python
     def __copy__(self, deep: bool=True):
           return self.__deepcopy__() if deep else self.copy()

``` 
But with the right parameters for `__deepcopy__`

hrzn:
Thanks for reporting!
Will be fixed here: https://github.com/unit8co/darts/pull/1326

PR #1326, opened 2022-10-29
fix deepcopy, remove traffic from lfs

hrzn:
* Fixes https://github.com/unit8co/darts/issues/1261
* Removes `traffic.csv` from git LFS (I had to do it in this same PR otherwise I couldn't commit)


dennisbader (approved):
LGTM, thanks!

PR #1327, opened 2022-10-29
Add dedicated LocalForecastingModel and renaming some model classes

dennisbader:
### Summary
- introduces a dedicted LocalForecastingModel class: no model inherits directly from ForecastingModel anymore -> we can decouple the logic from now on
- renames the local forecasting model base class like DualCovariatesForecastingModel -> FutureCovariatesLocalForecastingModel: I believe we talked once about the naming confusion for this one, and that it should actually be called FutureCovariates... Please correct me if I'm wrong.
  -  feel free to propose different class names, or whether I should revert some of the renaming

### DIFF
--- a/darts/models/forecasting/ensemble_model.py
+++ b/darts/models/forecasting/ensemble_model.py
@@ -10,6 +10,7 @@ from darts.logging import get_logger, raise_if, raise_if_not
 from darts.models.forecasting.forecasting_model import (
     ForecastingModel,
     GlobalForecastingModel,
+    LocalForecastingModel,
 )
 from darts.timeseries import TimeSeries
 
@@ -28,9 +29,7 @@ class EnsembleModel(GlobalForecastingModel):
         List of forecasting models whose predictions to ensemble
     """
 
-    def __init__(
-        self, models: Union[List[ForecastingModel], List[GlobalForecastingModel]]
-    ):
+    def __init__(self, models: List[ForecastingModel]):
         raise_if_not(
             isinstance(models, list) and models,
             "Cannot instantiate EnsembleModel with an empty list of models",
@@ -38,9 +37,7 @@ class EnsembleModel(GlobalForecastingModel):
         )
 
         is_local_ensemble = all(
-            isinstance(model, ForecastingModel)
-            and not isinstance(model, GlobalForecastingModel)
-            for model in models
+            isinstance(model, LocalForecastingModel) for model in models
         )
         self.is_global_ensemble = all(
             isinstance(model, GlobalForecastingModel) for model in models
@@ -48,7 +45,7 @@ class EnsembleModel(GlobalForecastingModel):
 
         raise_if_not(
             is_local_ensemble or self.is_global_ensemble,
-            "All models must either be GlobalForecastingModel instances, or none of them should be.",
+            "All models must be of the same type: either GlobalForecastingModel, or LocalForecastingModel.",
             logger,
         )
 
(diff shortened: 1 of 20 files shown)


darts/models/forecasting/ensemble_model.py
```diff
@@ -29,7 +29,7 @@ class EnsembleModel(GlobalForecastingModel):
     """
 
     def __init__(
-        self, models: Union[List[ForecastingModel], List[GlobalForecastingModel]]
+        self, models: Union[List[LocalForecastingModel], List[GlobalForecastingModel]]
```
hrzn, line 32: That's one evidence that this is cleaner, if we needed one :)

darts/models/forecasting/forecasting_model.py
```diff
@@ -1068,6 +1068,20 @@ def load(path: Union[str, BinaryIO]) -> "ForecastingModel":
         return model
 
 
+class LocalForecastingModel(ForecastingModel, ABC):
+    """The base class for "local" forecasting models, handling only single univariate time series.
+
+    Local Forecasting Models (LFM) are models that can be trained on a single univariate target series only. In Darts,
+    most models in this category tend to be simpler statistical models (such as ETS or FFT). LFMs usually train on
+    the entire target series you supplied when calling :func:`fit()` at once. They can also predict in one go with
```
hrzn, line 1076:
```suggestion
    the entire target series supplied when calling :func:`fit()` at once. They can also predict in one go with
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -1371,12 +1385,18 @@ def _get_encoders_n(self, n) -> int:
...
-    Among other things, it lets Darts forecasting models wrap around statsmodels models
-    having a `future_covariates` parameter, which corresponds to future-known covariates.
+class FutureCovariatesLocalForecastingModel(LocalForecastingModel, ABC):
+    """The base class for future covariates "local" forecasting models, handling single uni- or multivariate target
+    and optional future covariates time series.
 
-    All implementations have to implement the `_fit()` and `_predict()` methods defined below.
+    Future Covariates Local Forecasting Models (FC-LFM) are models that can be trained on a single uni- or multivariate
+    target and optional future covariates series. In Darts, most models in this category tend to be simpler statistical
+    models (such as ARIMA). FC-LFMs usually train on the entire target and future covariates series you supplied when
```
hrzn, line 1394:
```suggestion
    models (such as ARIMA). FC-LFMs usually train on the entire target and future covariates series supplied when
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -1535,12 +1555,22 @@ def _predict_wrapper(
...
+):
+    """The base class for transferable future covariates "local" forecasting models, handling single uni- or
+    multivariate target and optional future covariates time series. Additionally, at prediction time, it can be
+    applied to new data unrelated to the original series used for fitting the model.
+
+    Transferable Future Covariates Local Forecasting Models (TFC-LFM) are models that can be trained on a single uni-
+    or multivariate target and optional future covariates series. Additionally, at prediction time, it can be applied
+    to new data unrelated to the original series used for fitting the model. Currently in Darts, all models in this
+    category wrap to statsmodel models such as VARIMA. TFC-LFMs usually train on the entire target and future covariates
+    series you supplied when calling :func:`fit()` at once. They can also predict in one go with :func:`predict()`
```
hrzn, line 1569:
```suggestion
    series supplied when calling :func:`fit()` at once. They can also predict in one go with :func:`predict()`
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -22,7 +19,7 @@
 class RegressionEnsembleModel(EnsembleModel):
     def __init__(
         self,
-        forecasting_models: Union[List[ForecastingModel], List[GlobalForecastingModel]],
+        forecasting_models: List[ForecastingModel],
```
hrzn, line 22: Note: that's not exactly the same thing as `Union[List[LocalForecastingModel], List[GlobalForecastingModel]]`

hrzn (approved):
Looks good, I only added a few tiny comments/suggestions.
Thanks for taking the time to make this cleaner.

darts/models/forecasting/regression_ensemble_model.py
dennisbader: Oh yeah, good catch!

PR #1329, opened 2022-10-31
Feat/model deeptime

### LINKED ISSUE #1152: [new model] DeepTIMe
https://arxiv.org/pdf/2207.06046.pdf

madtoinou:
Fixes #1152.

### Summary

Implement the DeepTIMe model from https://arxiv.org/pdf/2207.06046.pdf, based on the original repository https://github.com/salesforce/DeepTime and the article pseudo-code.

Also implement some basics tests, inspired by the tests for N-Beats.

### Other Information

In the original article, distinct optimizers are defined for the three groups of parameters: Ridge Regression regularization term, the biais/norm of the Implicit Neural Representation (INR) network and the weights of the INR. This was accomplished by overriding the configure_optimizer method and partially breaking the logic behind the lr_scheduler_cls and lr_scheduler_kwargs arguments. To make the model easier to use out of the box, the default arguments correspond to the original article parameters (including for the optimizer).

All the module necessary for this architecture were included in the same file to limit the fragmentation of the code. The Ridge Regression and the INR modules could however be extracted if others models require them.

### DIFF
--- a/darts/models/forecasting/deeptime.py
+++ b/darts/models/forecasting/deeptime.py
@@ -343,28 +343,32 @@ class _DeepTimeModule(PLPastCovariatesModule):
         # define a scheduler for each optimizer
         lr_sched_kws = {k: v for k, v in self.lr_scheduler_kwargs.items()}
 
+        total_epochs = lr_sched_kws["total_epochs"]
         warmup_epochs = lr_sched_kws["warmup_epochs"]
-        T_max = lr_sched_kws["T_max"]
         eta_min = lr_sched_kws["eta_min"]
         scheduler_fns = []
 
-        def no_scheduler(T_cur):
+        def no_scheduler(current_epoch):
             return 1
 
-        def cosine_annealing(T_cur):
+        def cosine_annealing(current_epoch):
             return (
                 eta_min
                 + 0.5
                 * (eta_max - eta_min)
                 * (
                     1.0
-                    + np.cos((T_cur - warmup_epochs) / (T_max - warmup_epochs) * np.pi)
+                    + np.cos(
+                        (current_epoch - warmup_epochs)
+                        / (total_epochs - warmup_epochs)
+                        * np.pi
+                    )
                 )
             ) / lr
 
-        def cosine_annealing_with_linear_warmup(T_cur):
-            if T_cur < warmup_epochs:
-                return T_cur / warmup_epochs
+        def cosine_annealing_with_linear_warmup(current_epoch):
+            if current_epoch < warmup_epochs:
+                return current_epoch / warmup_epochs
             else:
                 return (
                     eta_min
@@ -373,7 +377,9 @@ class _DeepTimeModule(PLPastCovariatesModule):
                     * (
                         1.0
                         + np.cos(
-                            (T_cur - warmup_epochs) / (T_max - warmup_epochs) * np.pi
+                            (current_epoch - warmup_epochs)
+                            / (total_epochs - warmup_epochs)
+                            * np.pi
                         )
                     )
                 ) / lr
@@ -563,9 +569,15 @@ class DeepTimeModel(PastCovariatesTorchModel):
                 "weight_decay": 0.0,
             }
 
+            raise_if_not(
+                self.n_epochs > 0,
+                "`self.n_epochs` should be greater than 0, it is used to declare the learning rate scheduler.",
+                logger,
+            )
+
             self.pl_module_params["lr_scheduler_kwargs"] = {
                 "warmup_epochs": 5,
-                "T_max": self.n_epochs,
+                "total_epochs": self.n_epochs,
                 "eta_min": 0.0,
                 "scheduler_names": [
                     "cosine_annealing",
@@ -583,8 +595,8 @@ class DeepTimeModel(PastCovariatesTorchModel):
 
             raise_if_not(
                 "lr_scheduler_kwargs" in self.pl_module_params.keys(),
-                "`lr_scheduler_kwargs` should contain the following arguments: `Tmax` and "
-                "`warmup_epochs` in order to create the optimizer.",
+                "`lr_scheduler_kwargs` should contain the following arguments: `eta_min`, "
+                "`warmup_epochs` and `scheduler_names` in order to create the optimizer.",
                 logger,
             )
 
@@ -592,7 +604,6 @@ class DeepTimeModel(PastCovariatesTorchModel):
                 "weight_decay",
                 "lambda_lr",
                 "lr",
-                "T_max",
                 "warmup_epochs",
                 "eta_min",
                 "scheduler_names",
--- a/darts/models/forecasting/deeptime.py
+++ b/darts/models/forecasting/deeptime.py
@@ -34,12 +34,31 @@ ACTIVATIONS = [
 
 
 class GaussianFourierFeatureTransform(nn.Module):
-    """
-    https://github.com/ndahlquist/pytorch-fourier-feature-networks
-    Given an input of size [..., time, dim], returns a tensor of size [..., n_fourier_feats, time].
-    """
-
     def __init__(self, input_dim: int, n_fourier_feats: int, scales: List[float]):
+        """
+        Implementation of the Gaussian Fourier features mapping.
+        https://arxiv.org/abs/2006.10739
+        https://github.com/ndahlquist/pytorch-fourier-feature-networks
+
+        Parameters
+        ----------
+        input_dim
+            The dimensionality of the input time series.
+        n_fourier_feats
+            Number of Fourier components to sample to represent to time-serie in the frequency domain
+        scales
+            Scaling factors applied to the normal distribution sampled for Fourier components' magnitude
+
+        Inputs
+        ------
+        x of shape `(1, input_chunk_length+output_chunk_length, 1)`
+            Tensor containing the [0,1] normalised time representation.
+
+        Outputs
+        -------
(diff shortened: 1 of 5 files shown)


darts/tests/models/forecasting/test_deeptime.py
```diff
@@ -0,0 +1,186 @@
...
+from darts.utils import timeseries_generation as tg
+
+logger = get_logger(__name__)
+
+try:
+    from darts.models.forecasting.deeptime import DeepTimeModel
+
+    TORCH_AVAILABLE = True
+except ImportError:
+    logger.warning("Torch not available. TCN tests will be skipped.")
```
eliane-maalouf, line 17:
```suggestion
    logger.warning("Torch not available. DeepTime tests will be skipped.")
```

darts/tests/models/forecasting/test_deeptime.py
```diff
@@ -0,0 +1,186 @@
...
+    class DeepTimeModelTestCase(DartsBaseTestClass):
+        def setUp(self):
+            self.temp_work_dir = tempfile.mkdtemp(prefix="darts")
+
+        def tearDown(self):
+            shutil.rmtree(self.temp_work_dir)
+
+        def test_creation(self):
+            with self.assertRaises(ValueError):
+                # if a list is passed to the `layer_widths` argument, it must have a length equal to `num_stacks`
```
eliane-maalouf, line 32:
```suggestion
                
```

eliane-maalouf (commented):
few comments on quick initial things I noticed

darts/tests/models/forecasting/test_deeptime.py
```diff
@@ -0,0 +1,186 @@
...
+            # Test basic fit and predict
+            model = DeepTimeModel(
+                input_chunk_length=1,
+                output_chunk_length=1,
+                n_epochs=10,
+                inr_num_layers=2,
+                inr_layers_width=20,
+                n_fourier_feats=64,
+                scales=[0.01, 0.1, 1, 5, 10, 20, 50, 100],
+                random_state=42,
```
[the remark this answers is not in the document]
madtoinou: thank you, completely forgot them.

darts/models/forecasting/deeptime.py
```diff
@@ -0,0 +1,661 @@
...
+        # taken from nbeats, could be simplified
+        y = y.view(
+            y.shape[0], self.output_chunk_length, self.input_dim, self.nr_params
+        )[:, :, : self.output_dim, :]
+        return y
+
+    def get_coords(self, lookback_len: int, horizon_len: int) -> Tensor:
+        """Return time axis encoded as float values between 0 and 1"""
+        coords = torch.linspace(0, 1, lookback_len + horizon_len)
+        # would it be clearer with the unsqueeze operator?
```
hrzn, line 305: To me yes, a bit.

darts/models/forecasting/deeptime.py
```diff
@@ -0,0 +1,661 @@
...
+        lookback_reprs = time_reprs[:, : -self.output_chunk_length]
+        horizon_reprs = time_reprs[:, -self.output_chunk_length :]
+        # learn weights from the lookback
+        w, b = self.adaptive_weights(lookback_reprs, x)
+        # apply weights to the horizon
+        y = torch.einsum("... d o, ... t d -> ... t o", [w, horizon_reprs]) + b
+
+        # taken from nbeats, could be simplified
+        y = y.view(
+            y.shape[0], self.output_chunk_length, self.input_dim, self.nr_params
```
hrzn, line 298: Reshaping to `nr_params` as last dim will likely fail when `nr_params > 1` (e.g. if you set `likelihood=GaussianLikelihood()`. Would there be a natural way to have multiple final outputs of the network? It can perhaps a matter of setting `output_dim * nr_params` instead of `output_dim` somewhere earlier.

darts/tests/models/forecasting/test_deeptime.py
```diff
@@ -0,0 +1,187 @@
+import shutil
```
hrzn, line 1: Thanks for these tests! Could you also add some tests to `test_global_forecasting_models.py` (just add the model class in there along the others).

darts/tests/models/forecasting/test_deeptime.py
```diff
@@ -0,0 +1,187 @@
...
+                )
+
+        def test_logtensorboard(self):
+            ts = tg.constant_timeseries(length=50, value=10)
+
+            # Test basic fit and predict
+            model = DeepTimeModel(
+                input_chunk_length=1,
+                output_chunk_length=1,
+                n_epochs=10,
```
hrzn, line 147: maybe use `n_epochs=1` wherever the goodness of fit doesn't matter in a test.

hrzn (commented):
Looks very good, nice job @madtoinou !
After glancing at the paper, I also think it would be a nice addition to Darts.
I haven't looked into all minute details of the processing being done but I trust you :) I've got a few small comments. Perhaps the main one concerns `nr_params` which we should try to exploit before we merge.

darts/models/forecasting/deeptime.py
```diff
@@ -0,0 +1,661 @@
...
+        )
+
+        raise_if_not(
+            n_fourier_feats % (2 * len(scales)) == 0,
+            f"n_fourier_feats: {n_fourier_feats} must be divisible by 2 * len(scales) = {2 * len(scales)}",
+            logger,
+        )
+
+        raise_if_not(
+            self.n_epochs > 0,
```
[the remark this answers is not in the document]
madtoinou: No, I looked in all the parents classes (`PastCovariatesTorchModel`, `TorchForecastingModel` and `GlobalForecastingModel`) and the `self.n_epochs` attribute is instantiated in `TorchForecastingModel` without any check. Should I move it there or should the user to allowed to instantiate a model without a number of epochs?

darts/models/forecasting/deeptime.py
```diff
@@ -0,0 +1,661 @@
...
+
+class RidgeRegressor(nn.Module):
+    def __init__(self, lambda_init: float = 0.0):
+        """Implementation of the closed form Ridge Regression with a regularization coefficient."""
+        super().__init__()
+        self._lambda = nn.Parameter(torch.as_tensor(lambda_init, dtype=torch.float))
+
+    def forward(self, reprs: Tensor, x: Tensor, reg_coeff: float = None) -> Tensor:
+        if reg_coeff is None:
+            reg_coeff = self.reg_coeff()
```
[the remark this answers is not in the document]
madtoinou: I tried several approach to initialize this attribute in the constructor but when the model is trained a RuntimeError is raised by PyTorch during back-propagation : 
```
RuntimeError: Trying to backward through the graph a second time (or directly access saved
tensors after they have already been freed). Saved intermediate values of the graph are freed
when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward
through the graph a second time or if you need to access saved tensors after calling backward.
```

darts/models/forecasting/deeptime.py
madtoinou: I can see two possible approaches to increase dimension of the output:
1. One RidgeRegressor per nr_params is applied to the output of the INR network, each solving the equation AX = B_i where each B_i correspond to a nr_params (same encoding of the time axis A for all the nr_params)
2. the INR network can generate one time representation per `nr_params`, by multiplying the last layer by `nr_params` as you suggested. Then, either the same RidgeRegressor is applied to each X matrix or the option 1 can be reused.

By same RidgeRegressor, I mean that the lambda parameter (regularization, learned during the training) is identical. I will start experimenting with option 2 combined with a unique RidgeRegressor and remain open to suggestions.

darts/models/forecasting/deeptime.py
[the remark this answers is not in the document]
hrzn: I think this check could be move to `TorchForecastingModel` if possible. There should always be a number of epochs (even if that's the default one).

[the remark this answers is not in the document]
madtoinou: I anticipated your answer, the check was moved there already!

hrzn:
I find that the results do not seem fantastic in the probabilistic setting. E.g. when running the following code:
```python
from darts.models import DeepTimeModel
from darts.utils.likelihood_models import GaussianLikelihood, LaplaceLikelihood

scaler = Scaler()
train, val = scaler.fit_transform(series[:-36]), scaler.transform(series[-36:])

model = DeepTimeModel(input_chunk_length=24,
                      output_chunk_length=12,
                      likelihood=GaussianLikelihood())

model.fit(train, epochs=100)

pred = model.predict(series=train, n=36, num_samples=300)
train.plot()
pred.plot()
```

I get this - the variance seems almost zero.
image

I'm wondering whether this might be due to our treatment of the distributions parameters, which perhaps happens too early in the processing (when creating the time representations), which could (maybe?) cause degenerate results. Could we maybe find a way to "tile" tensors somewhere else later in the forward pass? WDYT @madtoinou ?

madtoinou:
This is indeed a bit disappointing, I should have spend more time looking at the variance of the resulting distribution.

There is not much room for tiling downstream: after the INR (fully connected network), there is only the ridge regression trying to solve the equation AX = B where A is the time representation transpose time itself, and B is the time representation transposed multiplied by the observations. I don't see how we could tweak this part.

I am going to experiment with using different Fourier features for each distribution parameter (before the INR), it should add bit of heterogeneity but I am not sure that it could solve the issue.

PR #1331, opened 2022-11-01
Remove IPython as a dependency

### LINKED ISSUE #1286: Remove IPython as a dependency
Currently IPython is listed as a dependency, but [only a single method is used](https://github.com/unit8co/darts/blob/0b727722bfebb3b1be9e9d6af101a052180814d9/darts/utils/utils.py#L90-L100). IPython adds a lot of transitive dependencies, some of which vary by platform (mac vs linux) which makes annoying to use in production environments.

Remove IPython from the list of dependencies and make it an optional import in the `utils.py` file linked above:
```python
try:
    from IPython import get_ipython
except ModuleNotFoundError:
    get_ipython = None
```
Then add a simple check at the top of the `_is_notebook` function:
```python
if get_ipython is None:
    return False
```

None

**Additional context**
I can open a PR with the change if this seems reasonable.

erik-hasse:
Fixes #1286.

### Summary
Removes IPython as a core dependency, which makes darts easier to install and maintain in production environments.


hrzn (approved):
Thanks!

PR #1347, opened 2022-11-10
Feat/normalized deviation metric

### LINKED ISSUE #877: Add metric: normalized deviation
See appendix A of https://arxiv.org/abs/2002.02887

madtoinou:
Fixes #877 .

### Summary
Implement the normalized deviation metric (ND) based on the description in the Annexe A of [F. Yu et al. "Temporal Regularized Matrix Factorization for High-dimensional Time Series Prediction"](https://papers.nips.cc/paper/2016/hash/85422afb467e9456013a2a51d4dff702-Abstract.html) and added corresponding tests.

### Other Information
Since all the other metrics had exactly the same signature, I used the same one even if some arguments are not used/relevant for ND.

### DIFF
--- a/darts/metrics/__init__.py
+++ b/darts/metrics/__init__.py
@@ -11,6 +11,7 @@ from .metrics import (
     marre,
     mase,
     mse,
+    nd,
     ope,
     r2_score,
     rho_risk,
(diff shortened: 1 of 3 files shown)


hrzn (commented):
As discussed, I'm not sure this method makes much sense given that in its current form it's proportional to the MAPE... Shall we park it and revisit if the need arises?

madtoinou:
I agree with you, we can close with PR until someone express a need for it or find an error in the implementation.

Issue #1337, opened 2022-11-05
[BUG] verbose=False does not affect in predict of RNNModel


vpozdnyakov:
**Describe the bug**
When I set `verbose=False` in `predict` method of RNNModel it still shows the progress bar.

**To Reproduce**
```python
from darts.timeseries import TimeSeries
from darts.models import RNNModel
model = RNNModel(
    input_chunk_length=100,
    output_chunk_length=1,
    n_epochs=1,
)
ts = TimeSeries.from_values(np.random.randn(1000))
model.fit(ts, verbose=True)
pred = model.predict(n=1, series=ts, verbose=False)
```

<img src='https://user-images.githubusercontent.com/12996098/200109957-6e9995fa-b8de-428e-8ae2-479d64dd22ca.png' width=700>

**Expected behavior**
Progress bar is not displayed.

**System:**
 - Python version: 3.7.15
 - darts version: 0.22.0

vpozdnyakov:
The rough solution is to drop the trainer after fitting:
```python
from darts.timeseries import TimeSeries
from darts.models import RNNModel
model = RNNModel(
    input_chunk_length=100,
    output_chunk_length=1,
    n_epochs=1,
)
ts = TimeSeries.from_values(np.random.randn(1000))
model.fit(ts, verbose=True)
model.trainer = None
pred = model.predict(n=1, series=ts, verbose=False)
```

hrzn:
This has been fixed in the latest release

PR #1349, opened 2022-11-11
corrected the misleading sentences in example 02 about covariates usage

### LINKED ISSUE #472: [BUG] Confusing example 02
Hi,
thank you very much for this nice open source package.

I would like to report what I think are confusing statements in the last part of example 02, **Training with Covariates**.

The section builds a `BlockRNNModel` which uses `past_covariates` because, as said in a previous section, this method does not support `future_covariates`, only `RNNModel` does.
However, the example goes on saying:

> Now, to train the model with covariates, it is as simple as providing the covariates (in form of a list matching the target series) as `future_covariates` argument to the `fit()` function. The argument is named `future_covariates` to remind us that the model can use future values of these covariates in order to make a prediction.

which is **totally out of context**, since we are not using `future_covariates`.
The final part stresses that we are using `past_covariates` which are anyhow very predictable for the future:

>  If these past covariates happen to also be known into the future, then these models are also able to produce forecasts for n > output_chunk_length (as shown above for BlockRNNModel) in an auto-regressive way.

Indeed, I tried to compare `BlockRNNModel` (`model_cov`) with `RNNModel` and `future_covariates` (`model_futcov`), and as you can see in the picture, the first does perform slightly better:

image

I would kindly suggest to fix the statement in the cell "*Now, to train the model with covariates, ...*", or better to include an example for both `BlockRNNModel` and `RNNModel` to clarify the topic.

Thanks for your attention!

madtoinou:
Fixes #472 .

### Summary
Replaced 'future_covariate' with 'past_covariates' in the explanation associated with the BlockRNNModel (since it is actually able to use 'past_covariates' only) and added an example with RNNModel as it relies on the 'future_covariates' to show how simple and similar the two approaches are.

### Other Information
Had to use the argument verbose=False for the fitting of the two models since the historical_forecasting reuses the same trainer for prediction as the one in training, effectively bypassing the verbosity set for the prediction/historical_forecasting as described in #1337. I will try to tackle this bug in a separate PR.


hrzn (approved):
Very cool, thanks!

PR #1350, opened 2022-11-11
Fix/nan in resample

### LINKED ISSUE #1237: [BUG] when resample a TS, first sample is nan in some situations.
**Describe the bug**
When I use resample, the first element of the resampled time series can be nan.

**To Reproduce**
`from darts import TimeSeries
from darts.utils.timeseries_generation import gaussian_timeseries
import pandas as pd
# starting at 23h works fine
ts = gaussian_timeseries(0, 1, start=pd.Timestamp(2020,1,1, 23, 0, 0), length=1000, freq='15T')
ts.resample(freq= '1h')

# starting at 23:30h gives a nan in the first element of the ts, as it takes as starting point 23h
ts = gaussian_timeseries(0, 1, start=pd.Timestamp(2020,1,1, 23, 30, 0), length=1000, freq='15T')
ts.resample(freq= '1h')
`

**Expected behavior**
Expected behavior would be to sample without nan in the first value. Also, here should we resample from 23:30, (00:30, 1:30, ...) or 00:00 (01:00, )?

madtoinou:
Fixes #1237.

### Summary
Added kwargs to `TimeSeries.resample` to be able to pass arguments to the `xarray.resample` method and give the possibility to the user to shift the start of the resampling to avoid getting `nan` at the beginning of the resampled `TimeSeries`.

### Other Information
Write an example in the docstring to cover the most likely usage, it might be too long...


darts/timeseries.py
```diff
@@ -2878,13 +2879,46 @@ def resample(self, freq: str, method: str = "pad") -> "TimeSeries":
             'pad': propagate last valid observation forward to next valid
 
             'backfill': use NEXT valid observation to fill.
+        kwargs
+            some keyword arguments for the `xarray.resample` method, notably `loffset` or `base` to indicate where
+            to start the resampling and avoid nan at the first value of the resampled TimeSeries
+            (https://docs.xarray.dev/en/stable/generated/xarray.DataArray.resample.html)
```
hrzn, line 2885:
```suggestion
            For more informations, see the `xarray resample() documentation
            <https://docs.xarray.dev/en/stable/generated/xarray.DataArray.resample.html>`_.
```

darts/timeseries.py
```diff
@@ -2878,13 +2879,46 @@ def resample(self, freq: str, method: str = "pad") -> "TimeSeries":
...
+
+        Examples
```
hrzn, line 2887: I love this, thanks :)

hrzn (approved):
Looks great :) just got a minor suggestion on link formatting.

PR #1351, opened 2022-11-11
Quantiles_df method speed optimization with wrap class of Quantile_timeseries

tranquilitysmile:
Trying to optimization of estimate Quantiles_df method, with wraping Quantile_timeseries to pd.dataframe

For saving older method, added toggle "fast_mode", by default  using older method with False

With this optimization, process speed up to ~80-450%+ of time(depends on size of dataframe). 

Example on 5 minutes resolution data, forecasted one week in to future.  
On big datasets time for estimation extremely increased 

photo_2022-11-11_19-08-02


darts/timeseries.py
```diff
@@ -1559,8 +1594,10 @@ def quantiles_df(self, quantiles: Tuple[float] = (0.1, 0.5, 0.9)) -> pd.DataFram
         pandas.DataFrame
             The Pandas DataFrame containing the quantiles for each component.
         """
-        # TODO: there might be a slightly more efficient way to do it for several quantiles at once with xarray...
-        return pd.concat([self.quantile_df(quantile) for quantile in quantiles], axis=1)
+        if fast_mode == True:
+            return pd.concat([self.quantile_timeseries_to_df(quantile) for quantile in quantiles], axis=1)
```
hrzn, line 1598: If I understand well, `quantile_timeseries_to_df()` is exactly the same method as `quantile_timeseries()`, save for the dataframe translation at the end. Couldn't we simply do:

```suggestion
            return pd.concat([self.quantile_timeseries(quantile).pd_dataframe() for quantile in quantiles], axis=1)
```

Or am I missing something?

hrzn (commented):
Thanks for giving this a shot, it looks interesting. We have to get rid of `quantile_timeseries_to_df()` as it's basically a copy of `quantile_timeseries()` (I made a proposal - let me know).

Two more things:
* Could you check whether the results are the same (modulo some numeric rounding)? If yes, I don't see any reason for not removing `fast_mode` (and basically always have it to True). If there are differences, we have another problem :) Either way, I think we should remove `fast_mode` and always use whichever correct method is the fastest.
* Could you add some unit tests for this method in `test_timeseries.py` ? I think we're missing them currently... that would be a perfect occasion.

hrzn:
> Thanks for giving this a shot, it looks interesting. We have to get rid of `quantile_timeseries_to_df()` as it's basically a copy of `quantile_timeseries()` (I made a proposal - let me know).
> 
> Two more things:
> 
> * Could you check whether the results are the same (modulo some numeric rounding)? If yes, I don't see any reason for not removing `fast_mode` (and basically always have it to True). If there are differences, we have another problem :) Either way, I think we should remove `fast_mode` and always use whichever correct method is the fastest.
> * Could you add some unit tests for this method in `test_timeseries.py` ? I think we're missing them currently... that would be a perfect occasion.

@tranquilitysmile could you also fix the linting issues? In case of doubt, check [here](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md#code-formatting-and-linting).

tranquilitysmile:
Hi! 

> Could you check whether the results are the same (modulo some numeric rounding)? If yes, I don't see any reason for not removing fast_mode (and basically always have it to True). If there are differences, we have another problem :) Either way, I think we should remove fast_mode and always use whichever correct method is the fastest.

I couldn't find any difference in numeric rounding in result dataframes. Thats way i follow you advice and remove fast_mode toggle, and method quantile_timeseries_to_df(). 

> If I understand well, quantile_timeseries_to_df() is exactly the same method as quantile_timeseries(), save for the dataframe translation at the end. Couldn't we simply do

One small thing is different, it's naming of resulting column in quantile_timeseries(), which passes the static name "{comp}_quantiles", i modify it to return "{comp}_{quantile}" to avoid  identical name of resulting columns in quantiles_df()

> @tranquilitysmile could you also fix the linting issues? In case of doubt, check [here](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md#code-formatting-and-linting).

if I did everything right the problem should be solved

> Could you add some unit tests for this method in test_timeseries.py ? I think we're missing them currently... that would be a perfect occasion.

I don't have experience writing unit tests right now, I'll try to figure it out soon

hrzn (commented):
Looks good @tranquilitysmile ! Do you think you could still add a unit test?

tranquilitysmile:
Add unit test for the method quantiles_df, please check the correctness

PR #1356, opened 2022-11-12
Fix/verbose argument effect for prediction with [Block]RNN models

madtoinou:
Fixes ##1337.

### Summary
First, the `verbose` argument provided by the user was not reaching the `TorchForecastingModel.predict` method and was overwritten with the default value `None`:  the `verbose` argument was added to the `_predict_wrapper` and `predict` methods. I tried using a `kwargs` argument to "absorb" this unused argument but it was interfering with the `RegressionModel._predict_and_sample` method that actually uses a `kwargs` when calling `predict`.

Furthermore, the trainer instantiated with the `fit` method had to be overwritten in the `TorchForecastingModel.predict_from_dataset` to change its attributes related to the `verbose` argument( '["enable_model_summary"]' and '["enable_progress_bar"]').

### Other Information
I am running tests to make sure that re-instantiating the trainer during prediction does not cause issues when saving/loading/retraining the model, I'll keep you updated with the results.


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1173,8 +1173,7 @@ def predict_from_dataset(
         # Set mc_dropout rate
         self.model.set_mc_dropout(mc_dropout)
 
-        # setup trainer. will only be re-instantiated if both `trainer` and `self.trainer` are `None`
-        trainer = trainer if trainer is not None else self.trainer
```
hrzn, line 1177: It'd be great if @dennisbader could take a look at this

darts/models/forecasting/arima.py
```diff
@@ -91,6 +91,7 @@ def _predict(
         historic_future_covariates: Optional[TimeSeries] = None,
         future_covariates: Optional[TimeSeries] = None,
         num_samples: int = 1,
+        verbose: bool = None,
```
hrzn, line 94: I find it awkward to initialize a `bool` to `None`. Couldn't we set it to `False`?

madtoinou: I set it to `None` because `_setup_trainer` was doing so, in order to differentiate the cases where the user actually provide a value. I somehow also wanted to indicate that this argument is actually not used in 90% of these `predict` methods since the verbosity in handled elsewhere. I would not mind use `False` as a default value.

hrzn: I think False is clearer; IMO it'd actually be helpful for users to know what is the behavior if this one is not provided. WDYT?

darts/models/forecasting/torch_forecasting_model.py
dennisbader: I'm discussing with @madtoinou about general improvements to how we use the PTL trainer. 

I am working on some changes which I will probably merge into this branch once ready.

darts/models/forecasting/arima.py
dennisbader: I set it to `None` for TorchForecastingModels to know when the user actively changed the verbosity

Reason was that user can activate/deactivate the verbosity at model creation (pl_trainer_kwargs such as `enable_progress_bar`, ...).
If we didn't use `None`, then the prameters from model creation would just be bypassed.

Looking at it now, I think it's probably okay to actually ignore these params from model creation, and only rely on a boolean `verbose` for TorchForecastingModels. Let me know what you think @hrzn and @madtoinou

dennisbader: We would have to do it for both.
As otherwise, if the user wants actually no verbosity, then he would have to recreate the model with deactivated `enable_progress_bar`, ... (by default it is activated)

hrzn: right, yes, makes sense

madtoinou: Thanks to both of you for the feedback, I'll change all the default value to False. @dennisbader, are you taking care of the PL-related verbosity argument overwriting or should I do it before you merge your change into this branch?

hrzn (approved):

PR #1359, opened 2022-11-14
Fix/ Allow empty TimeSeries

### LINKED ISSUE #1279: Empty time series allowed
Instead of crashing when we call a TimeSeries[a:b] and there are no points between a and b, we get an empty timeseries.

madtoinou:
Fixes #1279.

### Summary
Remove the assert preventing to create an empty `TimeSeries` in the constructor and replaced it by a warning to limit the risk of unexpected behavior for the users. Edited the corresponding test to catch the `ValueError` about missing dimensions in input `xarray.DataArray` instead of its emptiness.

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -292,6 +292,10 @@ class TimeSeries:
         # Store static covariates and hierarchy in attributes (potentially storing None)
         self._xa = _xarray_with_attrs(self._xa, static_covariates, hierarchy)
 
+        # at the end since other exceptions can prevent the creation of the time serie
+        if not xa.size > 0:
+            logger.warning("TimeSeries is empty.")
+
     """
     Factory Methods
     ===============
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -86,7 +86,6 @@ class TimeSeries:
             "TimeSeries.from_times_and_values(), etc...).",
             logger,
         )
-        raise_if_not(xa.size > 0, "The time series array must not be empty.", logger)
         raise_if_not(
             len(xa.shape) == 3,
             f"TimeSeries require DataArray of dimensionality 3 ({DIMS}).",
--- a/darts/tests/test_logging.py
+++ b/darts/tests/test_logging.py
@@ -63,7 +63,7 @@ class LoggingTestCase(unittest.TestCase):
             (
                 "darts.timeseries",
                 "ERROR",
-                "ValueError: The time series array must not be empty.",
+                "ValueError: TimeSeries require DataArray of dimensionality 3 (('time', 'component', 'sample')).",
             )
         )
 


darts/timeseries.py
```diff
@@ -293,6 +292,10 @@ def __init__(self, xa: xr.DataArray):
         # Store static covariates and hierarchy in attributes (potentially storing None)
         self._xa = _xarray_with_attrs(self._xa, static_covariates, hierarchy)
 
+        # at the end since other exceptions can prevent the creation of the time serie
+        if not xa.size > 0:
+            logger.warning("TimeSeries is empty.")
```
hrzn, line 297: Actually, I think we maybe shouldn't warn. I think the main reason for allowing empty series is to allow more general kinds of "time series arithmetic", which I would guess happen typically when we mass-process lots of series, and it might be a pain to have lots of warnings.

hrzn (commented):
Looks good to me, but:
* I would actually remove the warning (let me know if you think it's a terrible idea)
* We need to slightly adapt the introductory docstring at the top of TimeSeries saying one of the guarantees is that it's non-empty (maybe also in the user guide - to be checked)
* Could you try and see what happens when you print the series, or when you plot it? I expect plot() to crash... we should perhaps support creating an empty figure.

darts/timeseries.py
madtoinou: I was just worried that some users could end up with weird issues (during training for example) and eventually struggle to identify the source of the issue but I did not take the time to check what kind of errors happen when these methods are called with an empty TimeSeries (I could eventually add exceptions/warning in these methods instead of the constructor?). It's true that when handling massive amount of TimeSeries, the warnings could become quite annoying.

madtoinou:
>     * I would actually remove the warning (let me know if you think it's a terrible idea)
Done, I was probably being too pessimistic about the users capabilities to debug their code.
>     * We need to slightly adapt the introductory docstring at the top of TimeSeries saying one of the guarantees is that it's non-empty (maybe also in the user guide - to be checked)
I looked at the User Guide, could not find any mention of the non-emptiness. 
>     * Could you try and see what happens when you print the series, or when you plot it? I expect plot() to crash... we should perhaps support creating an empty figure.
It was indeed crashing, I added support for plotting empty TimeSeries and made it so that the x-axis is labeled correctly (using the `time_index` attribute) and the x-axis range is between 0 and 5, making it obvious that the TimeSeries is empty. We could also add limits to the y-axis to make it even more obvious.

darts/timeseries.py
```diff
@@ -3191,6 +3189,18 @@ def plot(
...
+            # empty TimeSeries
+            elif central_series.shape[0] == 0:
+                p = plt.plot(
+                    [],
+                    [],
+                    *args,
+                    **kwargs,
+                )
+                # make the emptiness of the TimeSerie obvious
+                plt.xlim(-0.2, 5.2)
```
hrzn, line 3201: Looks good - maybe just check that this doesn't affect plotting other series on the same figure later on.

darts/timeseries.py
```diff
@@ -3191,6 +3189,18 @@ def plot(
...
+                plt.xlabel(self.time_index.name)
+                # TODO: also change the limits of the y axis?
```
hrzn, line 3203: I would say that's not needed.

darts/timeseries.py
madtoinou: It was indeed interfering with the others plots, I removed this line. Should I also remove the empty TimeSeries from the labels or is it fine if there is a legend of an non-existing/appearing curve in the plot?

hrzn: It's fine I think.

hrzn:
> > ```
> > * I would actually remove the warning (let me know if you think it's a terrible idea)
> > ```
> 
> Done, I was probably being too pessimistic about the users capabilities to debug their code.
> 
> > ```
> > * We need to slightly adapt the introductory docstring at the top of TimeSeries saying one of the guarantees is that it's non-empty (maybe also in the user guide - to be checked)
> > ```
> 
> I looked at the User Guide, could not find any mention of the non-emptiness.
> 
> > ```
> > * Could you try and see what happens when you print the series, or when you plot it? I expect plot() to crash... we should perhaps support creating an empty figure.
> > ```
> 
> It was indeed crashing, I added support for plotting empty TimeSeries and made it so that the x-axis is labeled correctly (using the `time_index` attribute) and the x-axis range is between 0 and 5, making it obvious that the TimeSeries is empty. We could also add limits to the y-axis to make it even more obvious.

For the non-emptiness reference, I was referring to [this line](https://github.com/unit8co/darts/blob/master/darts/timeseries.py#L25).

madtoinou:
> For the non-emptiness reference, I was referring to [this line](https://github.com/unit8co/darts/blob/master/darts/timeseries.py#L25).

Oh I already removed this line based on your previous comment, my previous comment was just to say that I could not find any reference about the non-emptiness in the user-guide :)

hrzn:
> > For the non-emptiness reference, I was referring to [this line](https://github.com/unit8co/darts/blob/master/darts/timeseries.py#L25).
> 
> Oh I already removed this line based on your previous comment, my comment was just to say that I could not find any reference about the non-emptiness in the user-guide :)

My bad, I missed it. Thanks!

hrzn (approved):
💯

connesy:
Maybe I'm misunderstanding the purpose of this PR, but as far as I can tell this doesn't fix the original [issue](https://github.com/unit8co/darts/issues/1279):

```python
import darts
import pandas as pd

df = pd.DataFrame({"a": [1, 2, 3, 4, 5]}, index=pd.date_range(start="2022-01-01", end="2022-01-05"))
timeseries = darts.TimeSeries.from_dataframe(df)

print(timestamp[5:6])

...
ValueError: Neither `start` nor `end` can be NaT
```

<details open>
<summary>Full traceback</summary>

```python
File ~/miniconda3/envs/test2/lib/python3.10/site-packages/darts/timeseries.py:4717, in TimeSeries.__getitem__(self, key)
   4713     xa_ = self._xa.isel({self._time_dim: key})
   4714     _set_freq_in_xa(
   4715         xa_
   4716     )  # indexing may discard the freq so we restore it...
-> 4717     return self.__class__(xa_)
   4718 elif isinstance(key.start, pd.Timestamp) or isinstance(
   4719     key.stop, pd.Timestamp
   4720 ):
   4721     _check_dt()

File ~/miniconda3/envs/test2/lib/python3.10/site-packages/darts/timeseries.py:191, in TimeSeries.__init__(self, xa)
    184     self._xa.get_index(self._time_dim).freq = self._freq
    186     # We have to check manually if the index is complete. Another way could be to rely
    187     # on `inferred_freq` being present, but this fails for series of length < 3.
    189     is_index_complete = (
    190         len(
--> 191             pd.date_range(
    192                 self._time_index.min(), self._time_index.max(), freq=self._freq
    193             ).difference(self._time_index)
    194         )
    195         == 0
    196     )
    198     raise_if_not(
    199         is_index_complete,
    200         "Not all timestamps seem to be present in the time index. Does "
   (...)
    204         logger,
    205     )
    206 else:

File ~/miniconda3/envs/test2/lib/python3.10/site-packages/pandas/core/indexes/datetimes.py:1125, in date_range(start, end, periods, freq, tz, normalize, name, closed, inclusive, **kwargs)
   1122 if freq is None and com.any_none(periods, start, end):
   1123     freq = "D"
-> 1125 dtarr = DatetimeArray._generate_range(
   1126     start=start,
   1127     end=end,
   1128     periods=periods,
   1129     freq=freq,
   1130     tz=tz,
   1131     normalize=normalize,
   1132     inclusive=inclusive,
   1133     **kwargs,
   1134 )
   1135 return DatetimeIndex._simple_new(dtarr, name=name)

File ~/miniconda3/envs/test2/lib/python3.10/site-packages/pandas/core/arrays/datetimes.py:367, in DatetimeArray._generate_range(cls, start, end, periods, freq, tz, normalize, ambiguous, nonexistent, inclusive)
    364     end = Timestamp(end)
    366 if start is NaT or end is NaT:
--> 367     raise ValueError("Neither `start` nor `end` can be NaT")
    369 left_inclusive, right_inclusive = validate_inclusive(inclusive)
    370 start, end, _normalized = _maybe_normalize_endpoints(start, end, normalize)
ValueError: Neither `start` nor `end` can be NaT
```
</details>

Same thing happens with
```python
print(timeseries[pd.Timestamp("2022-01-06"):pd.Timestamp("2022-01-08")])

...
ValueError: Neither `start` nor `end` can be NaT
```

madtoinou:
Hi!

I got a bit too excited and forgot to check if the fix was also working for TimeSeries with DatetimeIndex... Thank you for reporting the bug, I am working on a fix and open a PR by the end of the week.

PR #1364, opened 2022-11-15
Add tests to new datasets

### LINKED ISSUE #1322: Add tests to new datasets

FEJTWOW:
Fixes #1322.

### Summary
Add tests to new datasets


darts/tests/datasets/test_dataset_loaders.py
```diff
@@ -145,10 +153,19 @@ def test_pre_process_fn(self):
...
         ts = ele_multi_series_dataset.load().pd_dataframe()
-        ms = UberTLCDataset()._to_multi_series(ts)
-        self.assertEqual(len(ms), 5)
-        self.assertEqual(len(ms[0]), 140256)
 
         ms = ElectricityDataset()._to_multi_series(ts)
         self.assertEqual(len(ms), 5)
         self.assertEqual(len(ms[0]), 105216)
+
+        multi_series_datasets = [
```
hrzn, line 161: Could we maybe also move `ElectricityDataset` in there? And that's probably a dummy question, but will the length always be 5?

hrzn: Thanks for the explanation, I forgot about this small sample file. I think it's a good idea, we can leave this small sample approach for all datasets (we don't want the tests to take too long).

hrzn (approved):

PR #1371, opened 2022-11-19
Fix/ptl trainer handling

### LINKED ISSUE #1363: Try avoid saving PTL trainer (and redefine it in fit() and predict() calls)

dennisbader:
Fixes #1363 

### Summary (Edited)
Various improvements to TorchForecastingModel
- now each time fit/predict is called, the models create a new trainer object (or use the user supplied trainer). 
  - with this we avoid getting unexpected behavior after loading a model, errors where trainer is not attach to the model, ...
  - with this, parameters such as `verbose` in fit/predict always get added to the trainer
- remove saving the PyTorch module twice
  - we don't save `model.model` and `model.trainer` anymore by adapting what gets pickled
    - only saves the "empty" base TorchForecastingModel as pickle
    - saves only the necessary parameters for LightningModule (weights, hyperparameters, ...) instead of pickling 
- allows to map the model to another location upon loading with `map_location` and `model.to_cpu()`
  - **As we don't have a GPU in test env, couldn't add unittest for saving on GPU and loading to CPU**
- Edit: tested the following for all TorchForecastingModels:
  - train & save model on GPU (colab), load to CPU (colab) and train & predict
  - train & save model on GPU (colab), load to CPU (local, different computer) and train & predict
  -> Loading to CPU and predicting gives identical results for colab and local
  -> Loading to CPU and predicting gives results with negligible difference compared on GPU (diff <= +/- 1e-16)
#### Additional Info
@madtoinou, regarding our discussions, we can't avoid to store the Trainer object as an attribute in TorchForecastingModel, as otherwise the reference gets lost (LightnignModule just points to trainer via a property).


madtoinou (approved):

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -877,7 +885,7 @@ def fit_from_dataset(
         train_num_epochs = epochs if epochs > 0 else self.n_epochs
 
         # setup trainer
-        self._setup_trainer(trainer, verbose, train_num_epochs)
+        self.trainer = self._setup_trainer(trainer, verbose, train_num_epochs)
```
hrzn, line 888: +1, that's cleaner than having those methods mutate `self` state

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1378,28 +1383,47 @@ def load_from_checkpoint(
...
         file_path = os.path.join(checkpoint_dir, file_name)
         logger.info(f"loading {file_name}")
 
-        model.model = model.model.__class__.load_from_checkpoint(file_path)
+        model.model = model._load_from_checkpoint(file_path, **kwargs)
         model.load_ckpt_path = file_path
         return model
 
+    def _load_from_checkpoint(self, file_path, **kwargs):
+        """Loads a checkpoint for the underlying :class:`PLForecastingModule` (PLM) model.
```
hrzn, line 1408: Nice

darts/models/forecasting/forecasting_model.py
```diff
@@ -1085,6 +1085,14 @@ def load(path: Union[str, BinaryIO]) -> "ForecastingModel":
 
         return model
 
+    def __getstate__(self):
```
hrzn, line 1088: Is it needed to define these methods here? They'll anyway be overwritten in TFM no?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -78,6 +78,7 @@
 CHECKPOINTS_FOLDER = "checkpoints"
 RUNS_FOLDER = "runs"
 INIT_MODEL_NAME = "_model.pth.tar"
+TFM_ATTRS_NO_PICKLE = {"model": None, "trainer": None}
```
hrzn, line 81: Nice! Maybe 1-2 comments here to explain that pickling a TFM will not pickle these attributes would be helpful.

hrzn (approved):
Looks great! I only added very minor comments (and one question regarding the need to redefine dunder methods in ForecastingModel.

darts/models/forecasting/forecasting_model.py
dennisbader: No, not needed. I thought it would make it a bit cleaner but I will remove it :)

PR #1376, opened 2022-11-21
Docs/example multivatiate ts

### LINKED ISSUE #851: [BUG]- Tutorials request on multivariate time series (multiple target columns) using BlockRNN model (RNN,LSTM,GRU) and VARIMA
Hi Team,

Can you share a tutorial link to do multivariate (multiple target columns) using statsmodel,ml and deep learning models

Like below link i want to do multivariate time series
https://www.machinelearningplus.com/time-series/vector-autoregression-examples-python/

Regards
Rahul

madtoinou:
Fixes #851.

### Summary

Added a section about training models (`VARIMA` and `RNNModel` with `GRU`) with multivariate `TimeSeries` using the `ElectrictyDataset`.

Added an utility method to the `TimeSeries` class in order to retain only a subset of `components` (previously, it was only possible to drop a subset of columns). The documentation of the `label` argument of `plot` was improved (since it is not passed as is to `xarray.plot` and `plt.plot`) and the support for a list of label was added in order to give more control to the user.

### Other Information

The training using several multivariate `TimeSeries` in a list is still not covered but there is a sentence at the beginning of the multivariate series section to mention that this feature exists.

Not sure if it is intentional but in my opinion, if the `label` argument of `plot` is a string, the legend should still use the name of component instead of using their index (using the `label` value as a prefix). I did not modify this behavior but it might be interesting to change it.


darts/timeseries.py
```diff
@@ -2796,6 +2796,32 @@ def drop_columns(self, col_names: Union[List[str], str]) -> "TimeSeries":
...
+
+        raise_if_not(
+            all([(x in self.columns.to_list()) for x in col_names]),
+            "Some column names in col_names don't exist in the time series.",
+            logger,
+        )
+
+        new_xa = self._xa.sel({"component": col_names})
+        return self.__class__(new_xa)
+
```
eliane-maalouf, line 2824: hello, is this function equivalent to selecting the component names from the series as in the following code : 
```
from darts import TimeSeries

univ_series = AirPassengersDataset().load()
multi_series = univ_series.stack(univ_series + 200)
stoch_series = multi_series.concatenate(multi_series, axis = 2) # stochastic, with sample = 2

stoch_series['#Passengers'] # should give a series of 1 component and sample = 2
```

madtoinou: Nice trick! I did not know that it was possible to use the slicing operator to extract subset of the columns/components. It certainly makes this function useless, I'll remove it in my next commit.

eliane-maalouf: check first before deleting if it makes more sense to do it like you are doing it, your method might be efficient memory wise, as it might not be copying the data. I guess the Xarray.sel() is not copying but then if a new TimeSeries is created I am not sure what would happen. you might want to check if the data is being copied or just a view is provided on the underlying numpy.

madtoinou: After looking at the source code of the `__getitem__` operator, it appears that it is also using the `xarray.sel`method to create and return a new `TimeSeries`. The two differences are that the `__getitem__` method makes sure to drop any existing hierarchy whereas my method `keep_columns` verify that the columns asked are indeed in the original `TimeSeries`. Unless we want to explicitly have such a function, it would probably be better to remove my function and use the section I wrote to also showcase the fact that the slicing operator can also be applied to components.

hrzn: +1 for dropping this one, I think `__getitem__` is sufficient. It does return a new TimeSeries containing its own copy of the data, but as should ALL TimeSeries methods :) This is required to keep our TimeSeries immutable and avoid lots and lots of future pain.

darts/timeseries.py
```diff
@@ -3199,6 +3228,18 @@ def plot(
                 )
             )
 
+        if isinstance(label, list):
```
hrzn, line 3231: Checking for an actual list might be a little restrictive here. For instance it'd make sense to accept tuples as well. Perhaps check for `Sequence` type instead (it's usually a good fit if you rely on the order and the length being well defined).

darts/timeseries.py
```diff
@@ -3134,6 +3160,7 @@ def plot(
         low_quantile: Optional[float] = 0.05,
         high_quantile: Optional[float] = 0.95,
         default_formatting: bool = True,
+        label: Optional[Union[str, List[str]]] = "",
```
hrzn, line 3163: Good idea

hrzn (commented):
Looks good!

madtoinou:
<p>No, these are comments I wrote myself. I just wanted to incorporate something about the fact that each model differ and that they handle multivariate timeseries differently. But it might not be clear enough?</p><p> </p><p>Good idea, completly forgot to talk about the covariate.</p><p> </p><p>I will try to play a bit with the parameters but I was trying to keep the duration of the training under one minute. I'll see what I can do, the seasonality of this dataset should help a bit.</p>

---

 <a href='https://app.reviewnb.com/unit8co/darts/pull/1376/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

hrzn (approved):
👍 thanks!

PR #1382, opened 2022-11-24
Docs/grid search notebook

### LINKED ISSUE #1365: grid_search notebook example
add notebook example explaining the use of grid_search

eliane-maalouf:
Fixes #1365

### Summary
basic notebook example for grid search illustration. 

### Other Information
- corrects for pow, add, sum, mul and truediv behavior with np.float32 and np.float64 
- corrects some doc


darts/timeseries.py
```diff
@@ -3885,7 +3885,7 @@ def __len__(self):
         return len(self._xa)
 
     def __add__(self, other):
-        if isinstance(other, (int, float, np.integer)):
+        if isinstance(other, (int, float, np.integer, np.float32, np.float64)):
```
hrzn, line 3888: We could maybe have
```suggestion
        if isinstance(other, (int, float, np.number)):
```

hrzn (commented):
My advice in general: try to keep things as simple as you can, and maybe focus only on showing how grid search works here rather than, say, trying to understand something about DLINEAR vs NLINEAR, or how different models compare, which could deserve its own notebook. I would recommend you select only one "interesting enough" model (e.g. one with at least one mandatory hyper-param) and show the whole example on this one. If that results in a short & concise notebook, even better :)


> `However, for models with hyper-parameters that are not easily discretizable (e.g., learning rates in optimization algorithms)`

It's true that this is a reason to prefer continuous optimization methods, however IMO it's far less important as a reason than the exponential time complexity of grid search.


> `DOES NOT LOOK LIKE IT'S THE CASE, BUT THEN I DON'T UNDERSTAND WHAT IS THE USEFULNESS OF DLINEAR`

I guess it might be the case for some datasets maybe? That could be an interesting thing to explore, but I would say orthogonal to demoing gridsearch.

> `df_day_avg = original_df.groupby(original_df.index.astype(str).str.split(" ").str[0]).mean().reset_index()`

Isn't there a way to use e.g. `TimeSeries.resample()` instead? It's cleaner if we can avoid resorting to extracting the underlying DF I think.


> `WEIRD BEHAVIOR`

Could it be because you use predict() instead of historical_forecasts() (used internally by grid search)?

eliane-maalouf:
@hrzn  thanks for the feedback. I think I will close this PR for now, rework the notebook and resubmit it in another PR along another example notebook (window transformations). WDYT?

hrzn:
> @hrzn thanks for the feedback. I think I will close this PR for now, rework the notebook and resubmit it in another PR along another example notebook (window transformations). WDYT?

Sounds good to me!

PR #1388, opened 2022-11-25
feat: collect dynamically all the jupyter-notebooks in examples

### LINKED ISSUE #689: change github workflow notebook example tests from matrix to something else.
Currently jupyter notebooks are tested with matrix strategy in `.github/workflows/merge.yml` and `../develop.yml` as seen in image below.

This could be changed to something non-matrix based that dynamically runs all files in examples folder.

image

madtoinou:
Fixes #689.

### Summary

I tried to find approaches to list all the files present in the examples folder and iterate on the resulting list to run test using gradlew, however I think that it would negatively impact the readability of the github workflow since all the notebooks would be combined in a single action. I suggest this approach as a draft, that has the advantage of iterating on all the examples while being able to read their individual status.

### Other Information

Thanks to the caching mechanism I have the impression that the container running the notebooks is technically being reused across the actions (cf. issue #594). In my opinion, it makes the matrix strategy reasonably efficient.


.github/actions/run-examples/run-examples.py
```diff
@@ -0,0 +1,32 @@
+import os
+import sys
+from typing import List
+
+
+# TODO: add a group argument to be able to select subset of notebooks
+def collect_files(directory: str, file_extension: str) -> List(str):
+    matching_files = []
```
hrzn, line 8: For this kind of things I like this kind of more functional looking code:
```python
return filter(lambda f: f.endswith(file_extension), os.listdir(directory))
```

madtoinou:
I think that the current bottleneck in the example notebook tests is the fact that we have to manually add them to the workflow. As it is, the container is correctly being reused thanks to the cache and do not slow things down.

If we want to test all the examples notebook, for the sake of workflow output readability, it would probably be worth it to update these scripts. Let me know what you think, I am still in the middle of figure out what is the best way to do it in for the notebooks.

Issue #675, opened 2021-12-07
Add support for interpretable outputs in TFTModel


dennisbader:
We should think about adding interpretable outputs for TFTModel as done in the [original paper](https://arxiv.org/pdf/1912.09363.pdf).

Mainly:
- variable/feature importance from VariableSelectionNetworks (paper page 18)
- attention from InterpretableMultiHeadAttention (paper page 21)

doaa-altarawy:
Yes, please, will be very helpful, it's one of the advantages of TFT.

MagMueller:
As a quick and dirty workaround to have some insights into the variable selection weights until the feature is available one can:
1. ```tft.predict(n=Your horizon)```
2. while insert print in `darts.models.forecasting.tft_module.py`
print(encoder_sparse_weights.mean(axis=1)) - for past variables selection (including the past of target)
print(decoder_sparse_weights.mean(axis=1)) - for future variables selection
3. To plot the results - save them in a txt - maybe edit them to be readable by numpy
4. then do someting like:
```
def plot_cov_selection(filename, title = "Variable importance"):
    variable_selection = np.loadtxt(filename)
    cov_percentage = variable_selection.mean(axis=0).round(3) * 100
    # your covariates names, for example created with datetime_attribute_timeseries
    cov_names = covs.columns 
    
    # instert target value for past covariates
    if len(cov_names) != len(cov_percentage):
        cov_names = ["Past of Target"] + [cov_names.values[i] for i in range(len(cov_names))] 
        print(cov_names)
    plt.bar(cov_names, cov_percentage)
    plt.title(title)
    plt.xlabel("Variable", fontsize=12 )
    plt.xticks(rotation=45, ha='right', )
    plt.ylabel("Variable importance in %")
    plt.show()

plot_cov_selection('value_selection_encoder.txt', title = "Variable importance")
plot_cov_selection('value_selection_decoder.txt', title = "Variable importance")
```
To get something like

![variable_importance_past](https://user-images.githubusercontent.com/67061560/193576135-06d0e766-2f83-4706-9f0b-01af774b7e5f.png)
![a2c4b711-196c-4053-b413-87663fc23da3](https://user-images.githubusercontent.com/67061560/193576649-8066c70b-0c89-439a-bbc6-f6ffe90d259b.png)

MagMueller:
@doaa-altarawy 
Similar for the attention weights:
1. In tft_model.py in forward pass: 
[...]

PR #1392, opened 2022-11-27
Add TFTExplainer

Cattes:
Implements explainability for the TFT model as discussed in #675 

### Summary
I have added a `TFTExplainer` class similar to the `ShapExplainer` class to get the explainability insights for the `TFT` model.

The class contains the `encoder_importance` and `decoder_importance` of the trained `TFT` Model (grouped together with a `plot` option in the `get_variable_selection_weight` method). 

The `TFTExplainer.explain()` method calls the `predict` method of the trained `TFT` Model to get the attention over time.

I have also provided a `plot_attention_heads` method to plot the attention over time either as the average attention, as a heatmap or a detailed plot of all available attention heads.

I have added a section on how to use the class to the `13-TFT-example.ipynb` notebook.

### Other Information
I have oriented myself to the suggestions in the Issue made by @hrzn 
```python
from darts.explainability import TFTExplainer

my_model = TFTModel(...)
my_mode.fit(...)

explainer = TFTExplainer(my_model, ...)

# get the explainability results
results = explainer.explain()

# plot variable selection weights
explainer.plot_variable_selection(results)

# plot the attention over time (three plot options)
explainer.plot_attention(results, plot_type="all")
explainer.plot_attention(results, plot_type="time")
explainer.plot_attention(results, plot_type="heatmap")

# get feature importance values
encoder_importance = results.get_encoder_importance()
decoder_importance = results.get_decoder_importance()
static_covariates_importance = results.get_static_covariates_importance()
# get attention `TimeSeries`
attention = results.get_attention()
``` 

The inital code on how to get the details from the TFT class was provided by @MagMueller in the Issue.

Edit (@dennisbader, 27-07-2023):
- refactored `ForecastingModelExplainer` and `ExplainabilityResult` to simplify/unify implementation of new explainers
- explain() now calls model.predict() with the passed foreground/background series. The attention and feature importances are actually dependent on the input to predict/forward (and not fixed after training).
- plot methods are now relying on the `explain()` output
- added static covariates importance
- updated the attention plots to have xaxis 0 point where the prediction starts (start of output chunk)
- added colorbar to attention heat map
- fixed `TFTModel` attention mask when `full_attention=True`
- fixed an issue with encoded covariates not being properly stored in the `TorchForecastingModel` in case of training on a single series
- adapted tests to check on combinations of input series (univariate, multivariate, multiple multivariate) and covariates, encoders, and model creation parameter `add_relative_index`


hrzn:
Thanks a lot for this PR @Cattes ! Please bear with us... we are a bit slow to review at the moment (busy preparing the release of 0.23), but we'll get at it, and we feel very enthusiastic about adding TFT explainability!

darts/timeseries.py
```diff
@@ -3620,6 +3620,7 @@ def plot(
         high_quantile: Optional[float] = 0.95,
         default_formatting: bool = True,
         label: Optional[Union[str, Sequence[str]]] = "",
+        plot_all_components: Optional[bool] = False,
```
hrzn, line 3623: Could we maybe replace this by `max_nr_components: int = 10`, which would configure the maximum number of components to plot? We can also say that `-1` means all components.

darts/explainability/tft_explainer.py
```diff
@@ -0,0 +1,249 @@
+from typing import Dict, List, Literal, Optional
```
hrzn, line 1: Could you add a docstring header in the file, so that it gets rendered along with the rest of the doc?
You can also put some high level explanations about what it does here, and point to the key methods. Even better: provide code snippets showing how to use it.
You can have a look at `shap_explainer.py` for an example of such a header.

darts/explainability/tft_explainer.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        super().__init__(model)
+
+        if not model._fit_called:
+            raise ValueError("The model needs to be trained before explaining it.")
+
+        self._model = model
+
+    @property
+    def encoder_importance(self):
```
hrzn, line 36: Could you add docstrings explaining what this and `decoder_importance` are returning? They can be quite useful I think.

darts/explainability/tft_explainer.py
```diff
@@ -0,0 +1,249 @@
...
+                self.decoder_importance,
+                title="Decoder variable importance",
+            )
+
+        return {
+            "encoder_importance": self.encoder_importance,
+            "decoder_importance": self.decoder_importance,
+        }
+
+    def explain(self, **kwargs) -> ExplainabilityResult:
```
hrzn, line 80: How about taking some of the `predict()` parameters explicitly? At least `series`, `past_covariates`, `future_covariates` and `n` would make sense IMO. It will produce more comprehensible API documentation.

darts/explainability/tft_explainer.py
```diff
@@ -0,0 +1,249 @@
...
+        attention_heads = (
+            self._model.model._attn_out_weights.squeeze().sum(axis=1).detach()
+        )
+
+        # return the explainer result to be used in other methods
+        return ExplainabilityResult(
+            {
+                0: {
+                    "attention_heads": TimeSeries.from_dataframe(
+                        pd.DataFrame(attention_heads).transpose()
```
hrzn, line 115: If `attention_heads` is a NumPy array you could probably also do something simpler such as `TimeSeries.from_values(attention_heads.T)`.

darts/explainability/tft_explainer.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # get the weights and the attention head from the trained model for the prediction
+        attention_heads = (
+            self._model.model._attn_out_weights.squeeze().sum(axis=1).detach()
+        )
+
+        # return the explainer result to be used in other methods
+        return ExplainabilityResult(
+            {
+                0: {
```
hrzn, line 113: Is this always relating to horizon 0 only? How about the cases where `predict()` is called with `n > 1` above?

darts/explainability/tft_explainer.py
```diff
@@ -0,0 +1,249 @@
...
+        if plot_type == "all":
+            fig = plt.figure()
+            attention_heads.plot(
+                label="Attention Head",
+                plot_all_components=True,
+                figure=fig,
+            )
+            # move legend to the right side of the figure
+            plt.legend(bbox_to_anchor=(0.95, 1), loc="upper left")
+            plt.xlabel("Time steps in past")
```
hrzn, line 140: I don't find it immediately clear that the higher these time steps the further ago the corresponding lags are.
How about maybe relabeling as `"Time steps in the past (# lags)"` ? This also applies to other places where this is used as an x-label.

hrzn (commented):
This looks very good @Cattes , I think it'll be a nice addition to Darts. Thanks a lot!
I have some comments, mainly related to the docstrings, but also one question about how we handle the horizons in the returned `ExplainabilityResult` - are you sure we should always consider horizon 0 only there? What about the case where the actual forecast horizon is larger?
When `n > output_chunk_length`, `forward()` will be called multiple times auto-regressively, so we have to be careful there. Maybe it'd make sense to disregard `n` in `explain()`, and return some result for each horizon in the `output_chunk_length`? I might also be missing something.

darts/explainability/tft_explainer.py
hrzn: On another front, `Literal` is not available in Python 3.7, you can use this instead:
```python
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

darts/explainability/tft_explainer.py
Cattes: I had to set the `0` here to be compatible with the `ForecastingModelExplainer` base class. To get the `attention_heads` the `predict` method of the TFT class has to be called or the attention_heads will not show the correct values. I am not sure why yet. Placing this logic into the explain() method as the `ExplainabilityResult` felt like a sensible choice. 
We could deviate from the `ForecastingModelExplainer` class or add a note to the docstring that the `0` is irrelevant in this context.

hrzn: So if I follow well, here the explanation is for all forecasted horizons at once, right?
I would then propose the following. We can adapt the class `ExplainabilityResult` in order to make it a little bit more flexible:
* It could be used with one explanation per horizon (as now), or
* with one single explanation for all horizons (as required in this case for the TFT).

To accommodate the second case, we could make it possible to build `ExplainabilityResult` with only a `Dict[str, TimeSeries]` (in addition to `Dict[integer, Dict[str, TimeSeries]]`), so we avoid specifying the horizon. We can also adapt `ExplainabilityResult.get_explanation()` to make specifying the `horizon` optional, and not supported if the underlying explanation is not split by horizon.

WDYT? I would find this cleaner than "hacking" the class by using a fake horizon 0.

darts/timeseries.py
```diff
@@ -3688,26 +3691,33 @@ def plot(
...
+            n_components_to_plot = self.n_components
+        elif self.n_components > max_nr_components:
             logger.warning(
-                "Number of components is larger than 10 ({}). Plotting only the first 10 components.".format(
-                    self.n_components
-                )
+                f"Number of components is larger than {max_nr_components} ({self.n_components}). "
+                f"Plotting only the first {max_nr_components} components."
+                f"You can overwrite this in the using the `plot_all_components` argument in plot()"
+                f"Beware that plotting all components may take a long time."
```
hrzn, line 3702:
```suggestion
                f"Beware that plotting a large number of components may cause performance issues."
```

darts/explainability/tft_explainer.py
Cattes: I changed the import as you suggested but I am now receiving a linting error.

Do you know why I get the linting error at my newly added code: https://github.com/Cattes/darts/blob/feature/675_tft_explainer/darts/explainability/tft_explainer.py#L14
while the linter is not complaining about it at other pieces of code that are older like in `utils/data/tabularization.py` for example?
https://github.com/Cattes/darts/blob/feature/675_tft_explainer/darts/utils/data/tabularization.py#L6

hrzn: Yes, this is because your new import (line 14) is correct, but you need to remove `Literal` from line 1.

darts/explainability/tft_explainer.py
hrzn: @Cattes any thoughts on this? ^

hrzn:
@Cattes  are you still willing to wrap up this PR? I think it looks great, we just have to fix this issue with the `ExplainabilityResult`.

darts/explainability/tft_explainer.py
Cattes: I am not sure if that is relevant here at all. 

I do not understand why predict has to be called to get the proper attention heads of the time series. The learned autoregressive connections should depend on how `predict` is called. But if `predict` is not called at all the `attention_heads` saved in `self._model.model._attn_out_weights` do not have the right format. I assume they are still in a state of training and the ` predict()` call changes that. 

If that is the case I would rather remove the `**kwargs` completely from the `explain` method here and call predict once with `self._model.model.output_chunk_length` to get the correct attention heads.

darts/explainability/tft_explainer.py
Cattes: I think its a good idea to change the class to handle the TFT explainer. I didn't want to do it before discussing it. Having the hack with horizon=0 was just to conform with the given api. It was not an intuitive solution. 
I have added `Dict[str, TimeSeries]` to the valid type for class initialization and made the `horizon` optional. 
I also added a few more validations to deal with the different input types explicitly.

Cattes:
> @Cattes are you still willing to wrap up this PR? I think it looks great, we just have to fix this issue with the `ExplainabilityResult`.

Hi @hrzn 
sorry for the delay.
I have added the docstrings to the properties that were still missing.
I have also changed the `ExplainabilityResult` to better reflect the `TFTExplainer` class.

One open point is still the `explain` method and if we should pass any arguments to `predict` at all or not? I don't understand the TFT class good enough to judge that.

darts/explainability/tft_explainer.py
hrzn: Yes I agree with you we need to call `predict()` here. However `predict()` takes a lot more arguments than just `n`. It takes `series` (the series to predict), as well as covariates arguments and other arguments: see [the API doc](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.tft_model.html#darts.models.forecasting.tft_model.TFTModel.predict). 
I think we should probably change the signature of `explain()` to something like
```python
def explain(self, series, past_covariates, future_covariates, **kwargs) -> ExplainabilityResult
```
This in the docstring you can list `series`, `past_covariates` and `future_covariates`, and explain that those are passed down to `predict()`. You can also say that `n` will always be set to `output_chunk_length` (unless I'm wrong I think that's always what's needed), and that `**kwargs` can contain extra arguments for the predict method and link to the API documentation of `TFTModel.predict()`.
I hope it makes sense.

hrzn:
> Hi @hrzn sorry for the delay. I have added the docstrings to the properties that were still missing. I have also changed the `ExplainabilityResult` to better reflect the `TFTExplainer` class.
> 
> One open point is still the `explain` method and if we should pass any arguments to `predict` at all or not? I don't understand the TFT class good enough to judge that.

Thanks a lot @Cattes, `ExplainabilityResult` looks good IMO. I've left a comment regarding `explain()` and `predict()` to clarify what I meant.

darts/explainability/tft_explainer.py
Cattes: I think calling `predict()` is just a technicality to get to the correct attention weights. I don't think the way we call `predict` matters at all for the result, its just important that it was called (for whatever reason). 
If I understand it correctly the attention weights are learned during training and are not impacted by the data used in the `predict` call.

They don't have a similar logic behind them like shapley values but are learned during the training and are a fixed part of the trained model. 

Maybe I am wrong, but if I am right I would rather remove all parameter passed to `explain()` and have the `predict()` call happen without the user needing to know about it at all.

darts/timeseries.py
Cattes: I have changed to code to use `max_nr_components` as suggested

darts/explainability/tft_explainer.py
Cattes: Thank you. I have removed the line from the code and it's now working. 
I had the docstrings uncommited locally and overlooked the removal of the `Literal` from the `typing` module import when I pushed my code.

darts/explainability/tft_explainer.py
Cattes: Thank you for the suggestion. I have changed to code accordingly

darts/explainability/tft_explainer.py
Cattes: I have changed the axis label to `"Time steps in the past (# lags)"`

darts/explainability/tft_explainer.py
Cattes: I have added docstrings to the module and properties. I wasn't 100% sure on the details of the model so if you could have a look at it that would be great? If everything is fine you can resolve this conversation.

rahuldixit18: Hey, I was trying to call this function and ran into an error. it says no attribute 'encoder_sparse_weights'. I went to the tft_model.py and uncommented this code chunk : 
return self.to_network_output(
             prediction=self.transform_output(out, target_scale=x["target_scale"]),
             attention=attn_out_weights,
             static_variables=static_covariate_var,
             encoder_variables=encoder_sparse_weights,
             decoder_variables=decoder_sparse_weights,
             decoder_lengths=decoder_lengths,
             encoder_lengths=encoder_lengths,
         )

It now says TFTModule has no attribute called to_network_output.

Can I get some help regarding how to call the explainer and use it in my code?

tuomijal:
This would be great! Any progress on this? Thank you for the work!

dennisbader:
Hi @Cattes, thanks for the TFTExplainer and really sorry about how long this has been stale from our side. 
I'll be able to pick it up/review this week.
Let me know if you're still interested in working on it.

dennisbader:
@Cattes, sounds good, I'll be able to review on Friday.
Just out of interest, does it in the current state support the following:
- static covariates importance
- attention on future input (when `full_attention=True` the attention is also applied on the future covariates)

Cattes:
> Hi @Cattes, I finally have some time to work on this. If you want I can take over the development from here, as there are a couple of things that changed in meantime.

Hi @dennisbader 
Yes, if there were some larger changes that need to be incorporated, feel free to take over. From the basic functionality, only the question of the `explain()` method's arguments is still open: https://github.com/unit8co/darts/pull/1392#discussion_r1118591955

> Just out of interest, does it in the current state support the following:
> * static covariates importance
> * attention on future input (when full_attention=True the attention is also applied on the future covariates)

`test_tft_explainer.py` only contains one model at the moment, so changing it to loop through a list of models with other configurations would be good if the explainability components of the model have to be calculated/extracted differently depending on the model parameters.

```python
         TFTModel(
            input_chunk_length=input_chunk_length_ice,
            output_chunk_length=forecast_horizon_ice,
            hidden_size=32,
            lstm_layers=1,
            batch_size=16,
            n_epochs=10,
            dropout=0.1,
            add_encoders={"cyclic": {"future": ["month"]}},
            add_relative_index=False,
            optimizer_kwargs={"lr": 1e-3},
            random_state=42,
        )
```

dennisbader:
Hi @Cattes I finished the adaptions for TFTExplainer now. The refactor got quite big, because I took the time to refactor the `ForecastingModelExplainer` and `ExplainabilityResult` backbone to make it easier to implement new explainers in the future.

I also caught one or two bugs on the way that I fixed with it.

I updated the PR description with the points I added/adapted.

Let me know if you want to go over the changes and if you're okay with the changes :)

Thanks for this great PR and sorry again for the time it took us to get this through 🚀 
This will be part of the next release which comes in one to two weeks 💯

darts/tests/explainability/test_tft_explainer.py
```diff
@@ -0,0 +1,482 @@
...
+            encoders, covariates, and addition of relative index."""
+            series_option: str
+            cov_option: set
+            add_relative_idx: bool
+            use_encoders: bool
+
+            series_options = [
+                "univariate",
+                "multivariate",
+                # "multiple",
```
madtoinou, line 84: either remove or indicate that its tested in dedicated tests

darts/timeseries.py
```diff
@@ -3719,10 +3723,14 @@ def plot(
         default_formatting
             Whether or not to use the darts default scheme.
         label
+            A prefix that will appear in front of each component of the TimeSeries or a list of string of
+            length the number of components in the plotted TimeSeries (default "").
             Can either be a string or list of strings. If a string and the series only has a single component, it is
             used as the label for that component. If a string and the series has multiple components, it is used as
             a prefix for each component name. If a list of strings with length equal to the number of components in
             the series, the labels will be mapped to the components in order.
```
madtoinou, line 3731: to remove, the new docstring is more concise?

madtoinou (commented):
Nice refactoring of the explainability and really cool feature (awaited by a lot of users)!

Some minors comments, this is almost ready for merge 🚀

darts/explainability/explainability.py
```diff
@@ -48,149 +48,64 @@ def __init__(
...
             )
-
-        if model._is_probabilistic:
-            logger.warning(
-                "The model is probabilistic, but num_samples=1 will be used for explainability."
-            )
-
         self.model = model
+        # default forecasting horizon
+        self.n: Optional[int] = getattr(self.model, "output_chunk_length")
```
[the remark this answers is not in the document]
dennisbader: good catch

darts/models/forecasting/tft_model.py
```diff
@@ -406,18 +411,30 @@ def get_attention_mask_full(
...
+            decoder_mask = torch.zeros(
+                (decoder_length, decoder_length), dtype=torch.bool, device=device
+            )
+        else:
+            # attend only to past steps relative to forecasting step in the future
+            # indices to which is attended
+            attend_step = torch.arange(decoder_length, device=device)
+            # indices for which is predicted
+            predict_step = torch.arange(0, decoder_length, device=device)[:, None]
+            # do not attend to steps to self or after prediction
```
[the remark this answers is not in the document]
dennisbader: self is basically the self attention at the current prediction step. It will also ignore the prediction steps of next horizons (ahead of it's own predict step, self).

So we actually do not attend to steps **related** to the individual prediction and any future predictions.

madtoinou (approved):

dennisbader (approved):

Cattes:
Thank you @dennisbader and @madtoinou for finishing up the PR! Sorry I could not have a look at it before the merge because I was on holidays.

Issue #1393, opened 2022-11-28
[BUG] Decrease in performance for historical_forecasts


gdevos010:
**Describe the bug**
@hrzn I was working with the NLinear model this weekend on the master branch and noticed a drop in performance when using `historical_forecasts` compared to when I first did the performance analysis for the models (predict still performs the same). I've been able to narrow it down to these two commits, but can't seem to run any of the commits in between them since I get a mat multi error. Also, this error is not probabilistic (Quantile regression) related. I can create a simple script to reproduce this error later this week, but Im not doing anything fancy. 

### https://github.com/unit8co/darts/pull/1292/commits/b753fb41395e749576505a6d424fc2fa7b1e989a
![Backtest_Air Passengers NLinear (QR)](https://user-images.githubusercontent.com/15316026/204174175-111dce46-9353-442b-885c-7a4c582e120d.png)

### https://github.com/unit8co/darts/pull/1292/commits/237f4b4d099fd4d1a9f335e60a365f5ca22a7fb3
![Backtest_Air Passengers NLinear (QR)](https://user-images.githubusercontent.com/15316026/204174197-446e3a9a-a0fc-4f2f-92c6-b80625f99e9b.png)

**To Reproduce**
Steps to reproduce the behavior, preferably code snippet.

hrzn:
Hi @gdevos010, thanks for reporting. This is really strange, because `historical_forecasts()` relies on `predict()`. Are you sure that `predict()` is returning the same thing? Are you retraining in `historical_forecasts()`? Between the two commits you mention, I have added support for future and static covariates. Are you using covariates? How do you conclude that the error is not related to quantile regression? If you have a script to reproduce the issue that would be helpful.

gdevos010:
@hrzn It is future covariates related. I forgot I had future covariates in my encoders. Here are some experiments run on master. I haven't had time to make a simple script yet.

### Without Future covariates
```python
  encoders = {
      "datetime_attribute": {"past": ["month", "year"]},
      "position": {"past": ["relative"]},
      "transformer": Scaler(StandardScaler(), n_jobs=-1),
  }
```

![Forecast_Air Passengers NLinear ()](https://user-images.githubusercontent.com/15316026/204397894-5c52fc99-d7af-47d2-89bc-c4e577fd56aa.png)
![Backtest_Air Passengers NLinear ()](https://user-images.githubusercontent.com/15316026/204397880-4cd3f2df-e867-4de7-84e1-6c796febf44d.png)

[...]

PR #1401, opened 2022-11-30
fix issues with nlinear and dlinear future covariates

hrzn:
Fixes https://github.com/unit8co/darts/issues/1393

@gdevos010 could you have a look at this one?


gdevos010:
Looks good to me!

#### Without future covariates 
image
#### With future covariates 
image

gdevos010 (approved):

Issue #1134, opened 2022-08-09
XGboost regression model (and probabilistic forecasting faisability)


dumjax:
* Make sure to support probabilistic forecasts supported by xgboost
* it seems XGBoost also supports multi-outputs natively since version 1.6.2 - we should make sure we leverage this.

yepher:
Do you have any idea when XGBoost will become available in Darts?

hrzn:
No estimate yet, although it's fairly high on our backlog. Would you be interested in contributing this model @yepher ?

yepher:
I am not familiar with the internals of Darts. But could be a fun project. I will take a look to see if it is something I could accomplish.

yepher:
Here in [gradient_boosted_model.py#L7](https://github.com/unit8co/darts/blob/master/darts/models/forecasting/gradient_boosted_model.py#L7) the code says:

`...To enable LightGBM support in Darts, follow the detailed install instructions for LightGBM in the README...`

But when I look in the [README.m](https://github.com/unit8co/darts/blob/master/README.md) I do not see any mention of how to `install LightGGM`

beckernick:
Would love to see XGBoost in darts! On the topic of multi-output regression, the [documentation](https://xgboost.readthedocs.io/en/stable/tutorials/multioutput.html) notes that multi-output has "limited support from objectives and metrics".

If there are certain objective functions that are important for darts but not currently available, we'd love to learn more. cc @trivialfs @dantegd

trivialfis:
Hi, I help maintain the XGBoost project and developed the initial support for multi-output. Please let me know if I can be useful here. Would love the see XGBoost being part of darts.

hrzn:
Hi @beckernick @trivialfis, I have started some work to include XGBoost in Darts. I haven't yet played with multi-output regression but this is not super duper important: we are wrapping several models in a MultiOutputRegressor by default so things will work as intended even on multivariate series or for multi-step forecasting.

One thing that I think would be very valuable, though, would be to have a way to do quantile regression, e.g., by having an `objective` allowing for quantile loss (sometimes called pinball loss). I haven't found anything like it in the documentation. Do you know if xgboost has a feature along those lines that I maybe missed?
Thanks!

hrzn:
[...]

PR #1405, opened 2022-12-02
Feat/xgboost

hrzn:
An XGBoost implementation, supporting both `poisson` and `quantile` likelihoods. For now it still relies on `MultiOutputRegressor` for simplicity (this way it reuses what we have in place for other regression models); we can try and leverage multi-output capabilities of xgboost at some point.

Solves https://github.com/unit8co/darts/issues/1134


docs/userguide/forecasting_overview.md
```diff
@@ -280,7 +280,7 @@ Monte Carlo Dropout can be combined with other likelihood estimation in Darts, w
...
+Some regression models can also be configured to produce probabilistic forecasts too. At the time of writing, [LinearRegressionModel](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.linear_regression_model.html) and [LightGBMModel](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.lgbm.html) support a `likelihood` argument. When set to `"poisson"` the model will fit a Poisson distribution, and when set to `"quantile"` the model will use the pinball loss to perform quantile regression (the quantiles themselves can be specified using the `quantiles` argument).
```
eliane-maalouf, line 284: I guess this list would also contain the XGBModel

darts/models/__init__.py
```diff
@@ -42,7 +42,7 @@
     )
 
 try:
-    from darts.models.forecasting.gradient_boosted_model import LightGBMModel
+    from darts.models.forecasting.lgbm import LightGBMModel
```
eliane-maalouf, line 45: good idea to change the file name

darts/models/forecasting/xgboost.py
```diff
@@ -0,0 +1,246 @@
...
+        self._model_container = None
+        self.quantiles = None
+        self.likelihood = likelihood
+        self._rng = None
+
+        # parse likelihood
+        available_likelihoods = ["poisson", "quantile"]  # to be extended
+        if likelihood is not None:
+            self._check_likelihood(likelihood, available_likelihoods)
+            if likelihood in {"poisson"}:
```
eliane-maalouf, line 121: is there a reason for testing if likelihood in the set here instead of just testing for the string as is the case for "quantile" ?

darts/models/forecasting/xgboost.py
```diff
@@ -0,0 +1,246 @@
...
+            TimeSeries or Sequence[TimeSeries] object containing the target values.
+        past_covariates
+            Optionally, a series or sequence of series specifying past-observed covariates
+        future_covariates
+            Optionally, a series or sequence of series specifying future-known covariates
+        val_series
+            TimeSeries or Sequence[TimeSeries] object containing the target values for evaluation dataset
+        val_past_covariates
+            Optionally, a series or sequence of series specifying past-observed covariates for evaluation dataset
+        val_future_covariates : Union[TimeSeries, Sequence[TimeSeries]]
```
eliane-maalouf, line 169:
```suggestion
        val_future_covariates :
```

darts/models/forecasting/xgboost.py
```diff
@@ -0,0 +1,246 @@
...
+            return self._predict_poisson(x, num_samples, **kwargs)
+        else:
+            return super()._predict_and_sample(x, num_samples, **kwargs)
+
+    def _is_probabilistic(self) -> bool:
+        return self.likelihood is not None
+
+    @property
+    def min_train_series_length(self) -> int:
+        # LightGBM requires a minimum of 2 train samples, therefore the min_train_series_length should be one more than
```
eliane-maalouf, line 239:
```suggestion
        #  XGBModel  requires a minimum of 2 train samples, therefore the min_train_series_length should be one more than
```

docs/userguide/forecasting_overview.md
hrzn: Good catch, thanks!

PR #1408, opened 2022-12-04
refactors MovingAverage filter 

eliane-maalouf:
- makes MovingAverage filter to use window_transform() for TimeSeries
- adds static covariates and hierarchy to the transformed series (output of window_transform())

### DIFF
--- a/darts/models/filtering/moving_average.py
+++ b/darts/models/filtering/moving_average.py
@@ -9,7 +9,7 @@ from darts.timeseries import TimeSeries
 
 class MovingAverage(FilteringModel):
     """
-    A simple moving average filter. Works on deterministic series (made of 1 sample).
+    A simple moving average filter. Works on deterministic and stochastic series.
     """
 
     def __init__(self, window: int, centered: bool = True):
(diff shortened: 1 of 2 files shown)


darts/models/filtering/moving_average.py
```diff
@@ -9,7 +9,7 @@
 
 class MovingAverage(FilteringModel):
     """
-    A simple moving average filter. Works on deterministic series (made of 1 sample).
```
hrzn, line 12: 🥳

darts/models/filtering/moving_average.py
```diff
@@ -44,14 +41,14 @@ def filter(self, series: TimeSeries):
...
-            hierarchy=series.hierarchy,
+        transformation = {
+            "function": "mean",
+            "mode": "rolling",
+            "window": self.window,
+            "center": self.centered,
+            "min_periods": 1,
+        }
+
+        return series.window_transform(
```
hrzn, line 52: Much cleaner 👍

darts/timeseries.py
```diff
@@ -3525,6 +3540,8 @@ def _get_kwargs(transformation, forecasting_safe):
                 len(new_index), -1, n_samples
             ),
             columns=new_columns,
+            static_covariates=self.static_covariates,
```
hrzn, line 3543: good catch

hrzn (approved):
💯 thanks!

PR #1409, opened 2022-12-04
Refactor/data_transformers

### LINKED ISSUE #1407: Automatic `_*_iterator`s and `component_mask`ing for Data Transformers

In order to implement a new data transformer, users currently need to override the `_*_iterator` method (where `*` is `transform`, `fit`, and/or `inverse_transform`) if they wish to pass class attributes and/or fitted parameters to their implemented
`ts_transform`/`ts_fit`/`ts_inverse_transform` methods. Moreover, if users want their transform to be able to 'mask out' particular components using the `component_mask` keyword argument, they need to manually call `_reshape_in` and `_reshape_out` inside of their implemented `ts_*` method.

For a 'solid' example of what I mean, check out the [`BoxCox` transformer code](https://github.com/unit8co/darts/blob/master/darts/dataprocessing/transformers/boxcox.py) - notice how ` _fit_iterator`, ` _transform_iterator`, and `_inverse_transform_iterator` all need to be overridden to pass fitted parameters to their respective methods. Similarly, each method needs to `pop` the `component_mask` from `kwargs`, and then call `_reshape_in` to apply these masks.

In my view, these two steps should be made 'automatic' by default (I'll explain precisely what I mean by this shortly), for two reasons:
1. For most transformations, a lot of boilerplate code is required to achieve basic functionality (e.g. passing fixed or fitted parameters to `ts_transform`)
2. For new users wishing to implement their own transform, it's quite unintuitive having to override a private method (i.e. `_*_iterator`) to achieve something as basic as passing fitted parameters to their `ts_transform` method. 

By default, all fixed parameters (i.e. attributes defined in the child-most class) and fitted parameters (if the transformer is fittable) should be passed to a user's implemented `ts_*` methods. In my view, the easiest way to achieve this is to have `ts_transform`, `ts_fit`, and `ts_inverse_transform` all accept an additional `params` argument, where `params` is a `dict` that contains up to two keys:
1. `fixed`, which is another dictionary that stores the child-most class's attributes (i.e. the fixed parameters of the transformer).
2. `fitted`, which stores the fitted parameters returned by `ts_fit`; if the transformer doesn't inherit from `FittableDataTransformer` or `ts_fit` has been called, this key shouldn't be present.

As a simple illustration of what I mean by this:
```python
from darts.dataprocessing.transformers.fittable_data_transformer import FittableDataTransformer
from darts.dataprocessing.transformers.invertible_data_transformer import InvertibleDataTransformer

class MyTransformer(FittableDataTransformer, InvertibleDataTransformer):
    def __init__(self, my_fixed_param):
            self._my_fixed_param = my_fixed_param
   
   def ts_fit(series, params):
         fixed_param =  params['fixed']['_my_fixed_param']
         return (fitted_param_1, fitted_param_2)

    def ts_transform(series, params):
           fixed_param = params['fixed']['_my_fixed_param']
           fitted_param_1, fitted_param_2 = params['fitted']
           # Transform code here
          return transformed_ts

    def ts_inverse_transform(series, params):
           fixed_param = params['fixed']['_my_fixed_param']
           fitted_param_1, fitted_param_2 = params['fitted']
           # Inverse Transform code here
          return inverse_transformed_ts
```
Moreover, if `component_mask` is supplied to any of these methods, then the `series` provided to that method will *already have the relevant components removed*, so the user doesn't need to worry about masking the `series` themselves. Similarly, after `ts_transform`/`ts_inverse_transform` has returned the transformed components, these should automatically be 'added back' to the original timeseries with the unmasked components. If the user doesn't want this 'automatic masking' behaviour to be applied and, instead, have `component_masks` given to them as a `kwarg` (i.e. current behaviour), there should obviously be an option allowing that.

**Additional context**
I have been working on this change, so I'll post a PR soon.

Any comments and/or suggestions are more than welcome.

Cheers,
Matt.

mabilton:
Fixes #1407.

### Summary

This PR makes two major changes to how data transformers work:
1. Fixed parameters and fitted parameters (if any) are now 'automatically' passed to `ts_transform`, `ts_inverse_transform`, and `ts_fit` by default, without the user having to re-implement `_transform_iterator`, `_inverse_transform_iterator`, or `_fit_iterator`. To accommodate this, each of these `ts_*` methods now accepts a `params` (dictionary) argument, where `params['fixed']` stores the fixed parameters of the transformation (defined to be all those attributes defined in the child-most class *before calling `super().__init__`*) and `params['fitted']` stores the fitted parameters (i.e. what `ts_fit` returned).
2. `component_mask` key word arguments will be automatically applied to timeseries inputs given to `ts_*` and automatically 'unapplied' to timeseries outputs returned by these methods, which means that users don't have to worry about 'manually' dealing with these arguments inside of their implemented `ts_*` methods.  If the user does not wish for `component_masks` to be automatically applied, they may specify `mask_components=False` when calling `super().__init__`; this will cause any `component_mask` key word argument to be passed via `kwargs` to the called method (i.e. current behaviour).

To see how these changes can help simplify the work involved in implementing a new transformation, compare the [current implementation of `BoxCox`](https://github.com/unit8co/darts/blob/master/darts/dataprocessing/transformers/boxcox.py) and with the `BoxCox` implementation in this PR. 

### Other Information

Some other minor changes that come with this PR:
1. I split up the `_reshape_in` method into two new methods (`apply_component_mask` and `stack_samples`), as well as `_reshape_out` into two new methods (`unapply_component_mask` and `unstack_samples`). There are two reasons for this change:
     - `_reshape_in` and `_reshape_out` were responsible for applying two distinctly different changes to the data: masking component columns and stacking the samples of each component along a single axis. From a user interaction and maintainability perspective, I think it's much cleaner to have these two pieces of functionality separated from one another. The names `_reshape_in` and `_reshape_out` are, in my opinion, also a bit vague.
     - In the original implementation of `_reshape_in`/`_reshape_out`, the 'stacking step' was performed using a `for` loop; I've changed this so that only `np.swapaxes` and `np.reshape` operations are used , which theoretically should speed things up a bit.
     - I've made these new methods explicitly public, so that it's explicitly clear to new users looking to write their own transformations that it's "okay" to call these functions. 
2. I've refactored all of the currently implemented transformations so that they conform with these changes, with the exception of `StaticCovariatesTransformer`, which directly overrides the `fit`, `inverse_transform` and `transform` methods anyways. Similarly, I also had to make some minor adjustments to existing tests.
3. I've implemented new tests for the refactored `BaseDataTransformer` , `FittableDataTransformer`, and `InvertableDataTransformer` classes.
4. Some transformations, such as `BoxCox`, allow for different fixed parameter values to be distributed over different parallel jobs. To facillitate this, I've added a `parallel_params` argument to the `*Transformer` classes, which allows the user to specify which parameters should take different values for different parallel jobs. 

There are two drawbacks to what I've done here:
1. Some of these changes are slightly breaking; with that being said, I doubt many users currently have their own private code bases with their own custom-implemented `darts` transformations (although I could be wrong).
2. One 'gotcha' with this proposed interface is  that the user must only call `super().__init__` *after initialising the fixed parameters of their transformation*. For example, the following will allow the user to access `'_my_param'` in `params['fixed']`:
   ```python
   class MyTransform(BaseDataTransformer):
       def __init__(self):
              # Correct: Define fixed parameter *before* calling `super().__init__`
              self._my_param = 1
              super().__init__()
   ```
   whereas the following will **not**:
   ```python
   class MyTransform(BaseDataTransformer):
       def __init__(self):
              # Incorrect: Define fixed parameter *after* calling `super().__init__`
              super().__init__()
              self._my_param = 1
   ```

Any thoughts/comments on these changes are more than welcome.

Cheers,
Matt.

### DIFF
--- a/darts/dataprocessing/transformers/base_data_transformer.py
+++ b/darts/dataprocessing/transformers/base_data_transformer.py
@@ -4,12 +4,22 @@ Data Transformer Base Class
 """
 
 from abc import ABC, abstractmethod
-from typing import Iterator, List, Optional, Sequence, Tuple, Union
+from typing import (
+    Any,
+    Generator,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 
 import numpy as np
 
 from darts import TimeSeries
-from darts.logging import get_logger, raise_if_not
+from darts.logging import get_logger, raise_if_not, raise_log
 from darts.utils import _build_tqdm_iterator, _parallel_apply
 
 logger = get_logger(__name__)
@@ -17,7 +27,11 @@ logger = get_logger(__name__)
 
 class BaseDataTransformer(ABC):
     def __init__(
-        self, name: str = "BaseDataTransformer", n_jobs: int = 1, verbose: bool = False
+        self,
+        name: str = "BaseDataTransformer",
+        n_jobs: int = 1,
+        verbose: bool = False,
+        parallel_params: Union[bool, Sequence[str]] = False,
     ):
         """Abstract class for data transformers.
 
@@ -47,7 +61,27 @@ class BaseDataTransformer(ABC):
             required amount of time.
         verbose
             Optionally, whether to print operations progress
+        parallel_params
+            Optionally, specifies which fixed parameters (i.e. the attributes initialized in the child-most
+            class's `__init__`) take on different values for different parallel jobs. Fixed parameters specified
+            by `parallel_params` are assumed to be a `Sequence` of values that should be used for that parameter
+            in each parallel job; the length of this `Sequence` should equal the number of parallel jobs. If
+            `parallel_params=True`, every fixed parameter will take on a different value for each
+            parallel job. If `parallel_params=False`, every fixed parameter will take on the same value for
+            each parallel job. If `parallel_params` is a `Sequence` of fixed attribute names, only those
+            attribute names specified will take on different values between different parallel jobs.
         """
+        # Assume `super().__init__` called at *very end* of
+        # child-most class's `__init__`, so `vars(self)` contains
+        # only attributes of this child-most class:
+        self._fixed_params = vars(self).copy()
+        # If `parallel_params = True`, but not a list of key values:
+        if parallel_params and not isinstance(parallel_params, Sequence):
+            parallel_params = self._fixed_params.keys()
+        # If `parallel_params = False`:
+        elif not parallel_params:
+            parallel_params = tuple()
+        self._parallel_params = parallel_params
         self._name = name
         self._verbose = verbose
         self._n_jobs = n_jobs
@@ -81,7 +115,7 @@ class BaseDataTransformer(ABC):
 
     @staticmethod
     @abstractmethod
-    def ts_transform(series: TimeSeries) -> TimeSeries:
+    def ts_transform(series: TimeSeries, params: Mapping[str, Any]) -> TimeSeries:
         """The function that will be applied to each series when :func:`transform()` is called.
 
         The function must take as first argument a ``TimeSeries`` object, and return the transformed ``TimeSeries``
@@ -141,7 +175,8 @@ class BaseDataTransformer(ABC):
                 return zip(series, (i for i in range(len(series))))
 
         """
-        return zip(series)
+        params = self._get_params(n_timeseries=len(series))
+        return zip(series, params)
 
     def transform(
         self, series: Union[TimeSeries, Sequence[TimeSeries]], *args, **kwargs
@@ -187,6 +222,49 @@ class BaseDataTransformer(ABC):
         return (
             transformed_data[0] if isinstance(series, TimeSeries) else transformed_data
         )
+
+    def _get_params(
+        self, n_timeseries: int
+    ) -> Generator[Mapping[str, Any], None, None]:
+        """
+        Creates generator of dictionaries containing fixed parameter values
+        (i.e. attributes defined in the child-most class). Those fixed parameters
+        specified by `parallel_params` are given different values over each of the
+        parallel jobs. Called by `_transform_iterator` and `_inverse_transform_iterator`,
+        if `Transformer` does *not* inherit from `FittableTransformer`.
+        """
+        self._check_fixed_params(n_timeseries)
+
+        def params_generator(n_timeseries, fixed_params, parallel_params):
+            fixed_params_copy = fixed_params.copy()
+            for i in range(n_timeseries):
+                for key in parallel_params:
+                    fixed_params_copy[key] = fixed_params[key][i]
+                if fixed_params_copy:
+                    params = {"fixed": fixed_params_copy}
+                else:
+                    params = None
+                yield params
+            return None
+
+        return params_generator(n_timeseries, self._fixed_params, self._parallel_params)
+
+    def _check_fixed_params(self, n_timeseries: int) -> None:
+        """
+        Raises `ValueError` if `self._parallel_params` specifies a `key` in
+        `self._fixed_params` that should be distributed, but
+        `len(self._fixed_params[key])` does not equal `n_timeseries`.
+        """
+        for key in self._parallel_params:
+            if len(self._fixed_params[key]) != n_timeseries:
+                key = key[1:] if key[0] == "_" else key
+                msg = (
+                    f"{n_timeseries} TimeSeries were provided "
+                    f"but only {len(self._fixed_params[key])} {key} values "
+                    f"were specified upon initialising {self.name}."
+                )
+                raise_log(ValueError(msg))
+        return None
 
     @staticmethod
     def _reshape_in(
--- a/darts/dataprocessing/transformers/base_data_transformer.py
+++ b/darts/dataprocessing/transformers/base_data_transformer.py
@@ -17,9 +17,10 @@ from typing import (
 )
 
 import numpy as np
+import xarray as xr
 
 from darts import TimeSeries
-from darts.logging import get_logger, raise_if_not, raise_log
+from darts.logging import get_logger, raise_if, raise_if_not, raise_log
 from darts.utils import _build_tqdm_iterator, _parallel_apply
 
 logger = get_logger(__name__)
@@ -32,6 +33,7 @@ class BaseDataTransformer(ABC):
         n_jobs: int = 1,
         verbose: bool = False,
         parallel_params: Union[bool, Sequence[str]] = False,
+        mask_components: bool = True,
     ):
         """Abstract class for data transformers.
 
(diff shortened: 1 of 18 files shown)


mabilton:
So I see that the following line in cell `[40]` within the `00-quickstart.ipynb` is throwing an error:
```python
# scale back:
pred_air = scaler.inverse_transform(pred_air)
```
From what I can see, `scaler` is fitted to two different timeseries in cell `[31]`:
```python
scaler = Scaler()
train_air_scaled, train_milk_scaled = scaler.fit_transform([train_air, train_milk])
```
What's the intended behaviour for a data transformation which is trained on two series but is given only a single series to inverse transform? I was clearly under the misunderstanding that an error should be thrown in such cases.

hrzn:
Thanks a lot @mabilton ! We just need a bit more time to get to it and review :)

darts/dataprocessing/transformers/diff.py
```diff
@@ -226,14 +210,18 @@ def ts_inverse_transform(
...
     def _get_component_mask(kwargs, dropna):
         component_mask = kwargs.get("component_mask", None)
         raise_if(
             dropna and (component_mask is not None),
-            "Cannot specify `component_mask` with `dropna = True`.",
+            (
+                "Cannot specify `component_mask` with `dropna = True`, "
+                "since differenced and undifferenced components will be "
+                "of different lengths."
+            ),
```
madtoinou, line 225: Maybe reference the logger?

darts/dataprocessing/transformers/fittable_data_transformer.py
```diff
@@ -50,14 +69,20 @@ def __init__(
         amount of data. Using instance methods would imply copying the instance's data through multiple processes, which
         can easily introduce a bottleneck and nullify parallelisation benefits.
         """
-        super().__init__(name=name, n_jobs=n_jobs, verbose=verbose)
+        super().__init__(
```
madtoinou, line 72: shouldn't the `super().__init__()` be called after instantiating the other attributes or is it valid only for the transforms that directly inherit from `BaseDataTransformer`?

madtoinou (commented):
LGTM, I wrote some minor comments about details.

Thank you for this PR that should solve many of the issues related to data transformers!

mabilton:
Hey @madtoinou - thanks for all the useful feedback. Just letting you know that I'm pretty busy in my personal life at the moment, so it'll probably take a day or two to implement your suggestions. Apologies for the delay.

Cheers,
Matt.

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -81,7 +127,7 @@ def set_n_jobs(self, value: int):
 
     @staticmethod
     @abstractmethod
-    def ts_transform(series: TimeSeries) -> TimeSeries:
+    def ts_transform(series: TimeSeries, params: Mapping[str, Any]) -> TimeSeries:
```
hrzn, line 130: You could add `params` to the docstring.

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -184,115 +242,238 @@ def transform(
...
+        -------
+        stacked
+            `np.ndarray` of shape `(n_timesteps * n_samples, n_components)`, where the `i`th column is formed
+            by concatenating all of the samples of the `i`th component in `vals`.
+        """
+        if isinstance(vals, TimeSeries):
+            vals = vals.all_values()
+        shape = vals.shape
+        new_shape = (shape[0] * shape[2], shape[1])
+        stacked = np.swapaxes(vals, 1, 2).reshape(new_shape)
```
hrzn, line 428: That's better indeed :)

darts/dataprocessing/transformers/fittable_data_transformer.py
hrzn: My understanding is that this is necessary in the childmost classes only (where the instance variables are defined).

darts/dataprocessing/transformers/window_transformer.py
```diff
@@ -88,49 +78,38 @@ def __init__(
             <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.rolling.html>`_
             and `pandas.DataFrame.expanding().apply() documentation
             <https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.expanding.html>`_.
-
         treat_na
             Specifies how to treat missing values that were added by the window transformations
             at the beginning of the resulting TimeSeries. By default, Darts will leave NaNs in the resulting TimeSeries.
             This parameter can be one of the following:
-
```
hrzn, line 96: I think removing these blank lines before a list in rst format might have adverse effect once this is rendered in HTML. Could you perhaps check rendering the doc in HTML (`./gradlew buildDocs`) or - if you want to be on the safe side - simply keep these blank lines.

darts/tests/dataprocessing/transformers/test_fittable_data_transformer.py
```diff
@@ -0,0 +1,272 @@
...
+import numpy as np
+
+from darts import TimeSeries
+from darts.dataprocessing.transformers.fittable_data_transformer import (
+    FittableDataTransformer,
+)
+from darts.utils.timeseries_generation import constant_timeseries
+
+
+class BaseDataTransformerTestCase(unittest.TestCase):
```
hrzn, line 14: Nice unit tests, thanks!

hrzn (approved):
That looks _really_ good to me. We can almost merge it as such IMO :) Only got a few pretty minor comments. Nice job @mabilton! And sorry for the super late review...

mabilton:
Hey @hrzn , @madtoinou.

I've (finally) found some time to implement your suggestions in addition to the following changes:
1. I've added more detail to the docstrings of the transformer classes/methods.
2. I've added a 'global_fit' option to the `FittableDataTransformer`. When `global_fit = True`, the entire `Sequence` of `TimeSeries` passed to the `fit` method will be passed to `ts_fit`, thereby allowing the user to fit a single set of parameters to all of the (disjoint) `TimeSeries` provided by the user, as opposed to independently fitting a different set of parameters to each `TimeSeries`. At this point in time, only the `BoxCox`, `Scaler`, and `StaticCovariatesTransformer` (can) use this option. Indeed, my motivation for introducing this `global_fit` option was so I could refactor the `StaticCovariatesTransformer`, which I'll talk about now.
3. In the last iteration of this PR, the `StaticCovariatesTransformer` was basically left untouched and was implemented in a way that did not utilise the refactored `fit`, `transform`, and `inverse_transform` methods. Since fitting the `StaticCovariatesTransformer` requires first collating the static covariates of all of the provided `TimeSeries`, I had to first had introduce a 'global fitting' option before refactoring `StaticCovariatesTransformer` (as mentioned in point 2). With that done, I could then refactor `StaticCovariatesTransformer` so that only a `ts_transform`, `ts_inverse_transform`, and a `ts_fit` method needed to be implemented (as opposed to directly overriding the `transform`, `inverse_transform`, and `fit` methods).

As an aside, it appears that `statsforecast==1.5.0` (which was only released a few days ago) introduces some dependency conflicts, and resulted in some tests failing. It looks like this is a problem on the `statsforecast` side, so to get around this for the time being, I've specified `statsforecast>=1.4,<1.5` in `requirements/core.txt`.

Hopefully that all makes sense, and please let me know what you guys think. Thanks in advance for any help.

Cheers,
Matt.

hrzn (approved):
LGTM, nice job @mabilton !

PR #1415, opened 2022-12-07
Fix/split at when split_point is not in series

### LINKED ISSUE #1312: [BUG] Unexpected split_before/split_after result when split_point is between two time indices
**Describe the bug**
If `TimeSeries.split_after` is performed using a timepoint which is located between two time-indices of the `TimeSeries` as `split_point` parameter, `split_after` splits the `TimeSeries` after the next date after the split_point. 

For example:
Let's say I have a `TimeSeries` with the following timepoints as indices:
`[2020-01-01, 2020-01-03, 2020-01-05]`
Using `split_after` with `2020-01-02`  as `split_point` I would expect the following split:
`[2020-01-01], [2020-01-03, 2020-01-05]`
However the actual split is the following:
`[2020-01-01, 2020-01-03], [2020-01-05]`
This is unexpected, as a date that is after `split_point` is allocated to the first result.

The analogous problem happens with `split_before`.

**To Reproduce**
```python
import pandas as pd
import numpy as np
from darts import TimeSeries

dates = pd.date_range(pd.Timestamp(2020,1,1), pd.Timestamp(2020,1,6), freq='2D')
vals = np.random.rand(3)
ts = TimeSeries.from_times_and_values(dates, vals)

before, after = ts.split_after(pd.Timestamp(2020,1,2))
print(before.time_index)

# prints: DatetimeIndex(['2020-01-01', '2020-01-03'], dtype='datetime64[ns]', name='time', freq='2D')
```

**Expected behavior**
After using `split_after`, all timepoints in the 1st result should be `<= split_point`, regardless of whether `split_point` is contained as index in the series or inbetween two indices.

 - Python version: 3.10
 - darts version: 0.21.0

DavidKleindienst:
Fixes #1312 

### Summary

This PR changes the behavior of `TimeSeries._split_at` - and therefore also `split_before` and `split_after` - when `split_point` is a `Timestamp` which is not an index of the `TimeSeries` but lies between two indeces.
 It also adds an appropriate test to the ´TimeSeriesTestCase.test_split` unit test.

Let's say we have a `TimeSeries` indexed by dates `2020-01-01`, `2020-01-03` and `2020-01-05`: 

**Current behavior**:
`TimeSeries._split_at(2020-01-02, after=True)` produces two `TimeSeries` with indeces `[2020-01-01, 2020-01-03]` for the first, and `2020-01-05` for the second.
`TimeSeries._split_at(2020-01-02, after=False)` tries to produce two `TimeSeries` with indeces `[ ]` for the first, and `[2020-01-01, 2020-01-03, 2020-01-05]` for the second, but errors because the first one is empty.

**Proposed behavior**:
`TimeSeries._split_at(2020-01-02)` produces two `TimeSeries` with indeces `2020-01-01` for the first, and `[2020-01-03, 2020-01-05]` irrespective of whether `after` is `True` or `False`.

I believe this is the behavior most people would expect when splitting a `TimeSeries` at a point that lies between two dates.

This PR does not change the behavior of `TimeSeries` splits when `split_point` is not a Timestamp, or `split_point` is a Timestamp contained in the series.


hrzn (approved):
Thanks @DavidKleindienst , this makes sense to me

Issue #178, opened 2020-08-21
Accept Prophet with `growth='logistic'`


felipedram:
**Proplem**
To use Prophet with `growth='logistic'`, the dataframe provided must have a column named 'cap'.

**Describe proposed solution**
Accept 'cap' and 'floor' as arrays provided as parametes on the model initialization.

**Describe potential alternatives**
Do not consider TimeSeries to be always multivariate and rename the columns in order.

TheMP:
Hi @felipedram, thanks for raising the issue - it's actually part of the bigger design thing to tackle mentioned in #180, we will try to find a solution that works for everyone.

LeoTafti:
Hi @felipedram , sorry for the delay.

This is a good point. Unfortunately I do not see an easy and clean way of fixing this at the moment.

- Adding `cap` and `floor` parameters that would only be used when `growth='logistic'` is also given doesn't seem very clean to me
- `fit()` in `Prophet` calls `UnivariateForecastingModel.fit()`, which itself check `series._assert_univariate()`. I don't think we can prevent this from happening without either treating `growth='logistic'` as a special case or making `Prophet` a `MultivariateForecastingModel` (but this isn't an easy task either)

I'm not sure that accepting `growth='logistic'` justifies the tradeoffs listed above, but if you have another suggestion I'd love to hear it.

madtoinou:
Implemented by #1419.

PR #1419, opened 2022-12-09
Feat/prophet logistic growth (#628)

### LINKED ISSUE #628: Logistic growth with Prophet
See: https://github.com/unit8co/darts/issues/178

Maybe we could add an addition method “model.add_cap_and_floor()”, similar to model.add_seasonality().

DavidKleindienst:
Fixes #628 .

### Summary

Adds support for logistic growth to `Prophet` and a corresponding unit-test.

This PR adds the function `Prophet.set_capacity` to allow setting `cap` and `floor` parameters needed.
`Prophet.set_capacity` takes two arguments (`cap` and `floor`), which may be either numbers or Callables. 
`floor` defaults to 0, to replicate `FBProphet's` default behaviour. 
Numbers should be used as arguments when modeling a constant carrying capacity, Callables can be supplied when variable carrying capacities are needed. 
A Callable supplied as argument to `Prophet.set_capacity` must accept a `DatetimeIndex` or `RangeIndex` as input, and return a Sequence of numbers with the same length. Each number in this Sequence indicates the carrying capacity at the corresponding index.

During `Prophet.fit` and `Prophet.predict` cap and floor columns with the appropriate values are added to `fit_df` and `predict_df`, respectively.
If `Prophet` was instantiated with `growth = 'logistic`, `Prophet.fit` now raises a `ValueError` if carrying capacities have not been set.
If `Prophet.set_capacity` is called when `growth` is not logistic, a warning is logged and the `cap` and `floor` values set are ignored.

### DIFF
--- a/darts/models/forecasting/prophet_model.py
+++ b/darts/models/forecasting/prophet_model.py
@@ -299,8 +299,12 @@ class Prophet(FutureCovariatesLocalForecastingModel):
 
     def set_capacity(
         self,
-        cap: Union[float, Callable[[Sequence[pd.Timestamp]], Sequence[float]]],
-        floor: Union[float, Callable[[Sequence[pd.Timestamp]], Sequence[float]]] = 0,
+        cap: Union[
+            float, Callable[[Union[pd.DatetimeIndex, pd.RangeIndex]], Sequence[float]]
+        ],
+        floor: Union[
+            float, Callable[[Union[pd.DatetimeIndex, pd.RangeIndex]], Sequence[float]]
+        ] = 0,
     ) -> None:
         """Set carrying capacities for predicting with logistic growth.
         These capacities are only used when `Prophet` was instantiated with `growth = 'logistic'`
@@ -309,8 +313,8 @@ class Prophet(FutureCovariatesLocalForecastingModel):
 
         The `cap` and `floor` parameters may be:
         - a number, for constant carrying capacities
-        - a function taking a Sequence of Timestamps and returning a corresponding a Sequence of numbers,
-          where each number indicates the carrying capacity at this timepoint.
+        - a function taking a DatetimeIndex or RangeIndex and returning a corresponding a Sequence of numbers,
+          where each number indicates the carrying capacity at this index.
 
         Parameters
         ----------
--- a/darts/models/forecasting/prophet_model.py
+++ b/darts/models/forecasting/prophet_model.py
@@ -5,7 +5,7 @@ Facebook Prophet
 
 import logging
 import re
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Callable, Sequence
 
 import numpy as np
 import pandas as pd
@@ -98,6 +98,12 @@ class Prophet(FutureCovariatesLocalForecastingModel):
         self._execute_and_suppress_output = execute_and_suppress_output
         self._model_builder = prophet.Prophet
 
+        self.is_logistic = False
+        if "growth" in prophet_kwargs and prophet_kwargs["growth"] == "logistic":
+            self.is_logistic = True
+        self._cap = None
+        self._floor = None
+
     def __str__(self):
         return "Prophet"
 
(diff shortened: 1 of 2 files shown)


darts/models/forecasting/prophet_model.py
```diff
@@ -98,6 +98,12 @@ def __init__(
         self._execute_and_suppress_output = execute_and_suppress_output
         self._model_builder = prophet.Prophet
 
+        self.is_logistic = False
+        if "growth" in prophet_kwargs and prophet_kwargs["growth"] == "logistic":
```
hrzn, line 102: It could have been enough to write
```python
self.is_logistic = "growth" in prophet_kwargs and prophet_kwargs["growth"] == "logistic"
```

darts/models/forecasting/prophet_model.py
```diff
@@ -110,6 +116,14 @@ def _fit(self, series: TimeSeries, future_covariates: Optional[TimeSeries] = Non
         fit_df = pd.DataFrame(
             data={"ds": series.time_index, "y": series.univariate_values()}
         )
+        if self.is_logistic:
+            raise_if(
```
hrzn, line 120: Could we receive `cap` and `floor` as (optional) arguments of `__init__` instead? I think it would be cleaner as it would allow checking for their presence directly whenever a model is instantiated with logistic growth. I also think it'd make sense to see them as model hyperparameters and so specifying them in the constructor is natural I think. Finally, it would remove the need for users to have to know about `set_capacity()`. But let me know if you think I missed something.

darts/models/forecasting/prophet_model.py
```diff
@@ -265,6 +297,40 @@ def add_seasonality(
...
+        floor: Union[
+            float, Callable[[Union[pd.DatetimeIndex, pd.RangeIndex]], Sequence[float]]
+        ] = 0,
+    ) -> None:
+        """Set carrying capacities for predicting with logistic growth.
+        These capacities are only used when `Prophet` was instantiated with `growth = 'logistic'`
+        See <https://facebook.github.io/prophet/docs/saturating_forecasts.html> for more information
+        on logistic forecasts.
+
+        The `cap` and `floor` parameters may be:
```
hrzn, line 314: You should add a blank line for the bullet-point list to render correctly in HTML (you can try building & rendering the doc with `./gradlew buildDocs` to check the HTML files directly).

hrzn (commented):
Many thanks @DavidKleindienst and I'm really sorry it took us so long to review!

I think it looks really good. I would just prefer if we could specify `cap` and `floor` in the constructor instead of relying on an extra `set_capacity()` method. Let me know your thoughts.

darts/models/forecasting/prophet_model.py
DavidKleindienst: There were some concerns about adding them as optional arguments in #178, which is I think why this issue #628 suggested using a separate function. 
But I agree, I also think it would be cleaner to have `cap` and `floor` as arguments to `__init__`.
Shall I change it?

darts/models/forecasting/prophet_model.py
```diff
@@ -92,12 +98,50 @@ def __init__(
...
             For information about the parameters see:
             `The Prophet source code <https://github.com/facebook/prophet/blob/master/python/prophet/forecaster.py>`_.
         """
 
+        """Set carrying capacities for predicting with logistic growth.
+        These capacities are only used when `Prophet` was instantiated with `growth = 'logistic'`
+        See <https://facebook.github.io/prophet/docs/saturating_forecasts.html> for more information
+        on logistic forecasts.
+
+        The `cap` and `floor` parameters may be:
```
hrzn, line 132:
```suggestion
        The `cap` and `floor` parameters may be:
        
```

hrzn (approved):

hrzn:
LGTM, thanks @DavidKleindienst !

DavidKleindienst:
Fixed the linting problem and removed a the previous docstring that accidentally remained there. 
The remaining failing tests seem to be unrelated.
Thanks @hrzn for reviewing!

PR #1435, opened 2022-12-18
Fix: `append_values` and `prepend_values` correctly extends `RangeIndex`-typed `TimeSeries`

### LINKED ISSUE #1434: [BUG] `append_values` incorrectly extends `time_index` of `pd.RangeIndex`-typed `TimeSeries`
**Describe the bug**
When `append_values` is used to add new values to a `pd.RangeIndex`-typed `TimeSeries` , the `time_index` of the entire series is 'reset' so that each value is labeled with its index position. This is particularly unexpected if the `TimeSeries` in question does not have a frequency of `1` and/or does not start with a `time_index` of `0`.

**To Reproduce**
Consider the following example:
```python
import numpy as np

series = linear_timeseries(start=1, length=5, freq=2)
print('Before `append_values`:')
print(list(series.time_index))
new_values = np.ones((1,))
print('After `append_values`:')
print(list(series.append_values(new_values).time_index))
```
This yields:
```
Before `append_values`:
[1, 3, 5, 7, 9]
After `append_values`:
[0, 1, 2, 3, 4, 5]
```
Notice how the `time_index` now starts at `0` instead of `1` **and** has a frequency of `1` instead of `2`.

**Expected behavior**
Instead of resetting the `time_index` of `pd.RangeIndex`-typed `TimeSeries`, `append_values` should 'extend' the `time_index`. More explicitly, for the previous example, one should expect the output:
```
Before `append_values`:
[1, 3, 5, 7, 9]
After `append_values`:
[1, 3, 5, 7, 9, 11]
```
Indeed, this is the [behaviour that's advertised by `append_values`' docstring](https://unit8co.github.io/darts/generated_api/darts.timeseries.html#darts.timeseries.TimeSeries.append_values) and, additionally, is how it behaves when dealing with `pd.DatetimeIndex`-typed `TimeSeries`. To see this, consider the following example:
```python
import numpy as np
import pandas as pd

series = linear_timeseries(start=pd.Timestamp('1/1/2000'), length=2, freq='2d')
print('Before `append_values`:')
print(list(series.time_index))
new_values = np.ones((1,))
print('After `append_values`:')
print(list(series.append_values(new_values).time_index))
```
This prints:
```
Before `append_values`:
[Timestamp('2000-01-01 00:00:00', freq='2D'), Timestamp('2000-01-03 00:00:00', freq='2D')]
After `append_values`:
[Timestamp('2000-01-01 00:00:00', freq='2D'), Timestamp('2000-01-03 00:00:00', freq='2D'), Timestamp('2000-01-05 00:00:00', freq='2D')]
```
Notice how the `append_values` has simply added the date `2000-01-05` to the `time_index` of `series`.

In cases where the user really wants 'reset' the `time_index` of a `TimeSeries` after appending values, this should probably be achieved by implementing a separate `TimeSeries.reset_index()` method.

 - Python version: 3.10.6
 - darts version: 0.22.0

**Additional context**
N/A

mabilton:
Fixes #1434.

### Summary

Previously, `append_values` and `prepend_values` did not correctly handle extending `RangeIndex`-typed `TimeSeries`. For example, the following code snippet:
```python
import numpy as np

series = linear_timeseries(start=1, length=5, freq=2)
print('Before `append_values`:')
print(list(series.time_index))
new_values = np.ones((1,))
print('After `append_values`:')
print(list(series.append_values(new_values).time_index))
```
would print:
```
Before `append_values`:
[1, 3, 5, 7, 9]
After `append_values`:
[0, 1, 2, 3, 4, 5]
```
This has been fixed in this PR so that this code now prints:
```
Before `append_values`:
[1, 3, 5, 7, 9]
After `append_values`:
[1, 3, 5, 7, 9, 11]
```

Similarly, the code snippet:
```python
import numpy as np

series = linear_timeseries(start=1, length=5, freq=2)
print('Before `prepend_values`:')
print(list(series.time_index))
new_values = np.ones((1,))
print('After `prepend_values`:')
print(list(series.prepend_values(new_values).time_index))
```
would previously print:
```
Before `append_values`:
[1, 3, 5, 7, 9]
After `append_values`:
[0, 1, 2, 3, 4, 5]
```
Now, this same code snippet prints:
```
Before `prepend_values`:
[1, 3, 5, 7, 9]
After `prepend_values`:
[-1, 1, 3, 5, 7, 9]
```

### Other Information

Tests have also been updated to explicitly check for this behaviour.

### DIFF
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -621,15 +621,53 @@ class TimeSeriesTestCase(DartsBaseTestClass):
 
     def test_append(self):
         TimeSeriesTestCase.helper_test_append(self, self.series1)
+        # Check `append` deals with `RangeIndex` series correctly:
+        series_1 = linear_timeseries(start=1, length=5, freq=2)
+        series_2 = linear_timeseries(start=11, length=2, freq=2)
+        appended = series_1.append(series_2)
+        expected_vals = np.concatenate(
+            [series_1.all_values(), series_2.all_values()], axis=0
+        )
+        expected_idx = pd.RangeIndex(start=1, stop=15, step=2)
+        self.assertTrue(np.allclose(appended.all_values(), expected_vals))
+        self.assertTrue(appended.time_index.equals(expected_idx))
 
     def test_append_values(self):
         TimeSeriesTestCase.helper_test_append_values(self, self.series1)
+        # Check `append_values` deals with `RangeIndex` series correctly:
+        series = linear_timeseries(start=1, length=5, freq=2)
+        appended = series.append_values(np.ones((2, 1, 1)))
+        expected_vals = np.concatenate(
+            [series.all_values(), np.ones((2, 1, 1))], axis=0
+        )
+        expected_idx = pd.RangeIndex(start=1, stop=15, step=2)
+        self.assertTrue(np.allclose(appended.all_values(), expected_vals))
+        self.assertTrue(appended.time_index.equals(expected_idx))
 
     def test_prepend(self):
         TimeSeriesTestCase.helper_test_prepend(self, self.series1)
+        # Check `prepend` deals with `RangeIndex` series correctly:
+        series_1 = linear_timeseries(start=1, length=5, freq=2)
+        series_2 = linear_timeseries(start=11, length=2, freq=2)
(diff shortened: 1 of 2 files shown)


hrzn (approved):
Great, thanks a lot 👍

PR #1449, opened 2022-12-28
Fix Literal import for python37

### LINKED ISSUE #1444: [BUG] Darts no longer works on python versions lower than 3.8.
**Describe the bug**
Darts TimeSeries no longer works on python versions lower than 3.8. 

In the new release, typing.Literal was added to the imports in timeseries.py .. But Literal was added to typing.py in 3.8. 

**To Reproduce**
In an environment with python less than 3.8, try from darts import TimeSeries

 - Python version: 3.7.6
 - darts version: 0.23.0

**Additional context**
To use Literal in older versions we could do this:

```
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

https://stackoverflow.com/questions/61206437/importerror-cannot-import-name-literal-from-typing

dennisbader:
Fixes #1444.

### Summary
- fixes Literal import for python3.7

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -37,7 +37,7 @@ import re
 from collections import defaultdict
 from copy import deepcopy
 from inspect import signature
-from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


hrzn (approved):
Thanks!

hrzn:
This could warrant a new patch release...

PR #1465, opened 2023-01-04
fix historical forecasts retraining of TFMs

### LINKED ISSUE #1461: `historical_forecasts()` retrains with brand-new model instances
`historical_forecasts()` retrains the same model instance on the new training section instead of using a brand-new model instance.

**Additional context**
The example code below shows that `historical_forecasts()` gives the same results as calling the `fit()` method multiple times on the same model instance. However, I would expect the outputs of `historical_forecasts()` to be the same as fitting a brand-new model instance, with the same parameters, on each training section.
```
from darts.models.forecasting.block_rnn_model import BlockRNNModel
import pandas as pd

# A function for instantiating a rnn model with the initial weights fixed
def instantiate_BlockRNNModel():
    return BlockRNNModel(
        input_chunk_length=12,
        output_chunk_length=6,
        model='RNN',
        hidden_dim=10,
        n_rnn_layers=1,
        hidden_fc_sizes=None,
        dropout=0,
        n_epochs=5,
        batch_size=16,
        optimizer_kwargs={"lr": 1e-3},
        random_state=42,
        save_checkpoints=True
    )

# Prepare the data
transformer = Scaler()
series_transformed = transformer.fit_transform(series)

# 1. `historical_forecasts()`, it is a 3-pass historical_forecast
model = instantiate_BlockRNNModel()
historical_forecasts_result = model.historical_forecasts(
    series_transformed,
    start=len(series_transformed)-3*6,
    train_length=None,
    forecast_horizon=6,
    stride=6,
    retrain=True,
    verbose=False,
    last_points_only=False
)
historical_forecasts_result = [forecast.values() for forecast in historical_forecasts_result]

# 2. As far as I know, `historical_forecasts()` is similar to calling `fit()` on the same model with 3 training sections
train_set_1, val_set_1 = series_transformed.split_after(pd.Timestamp('19590601'))
train_set_2, val_set_2 = series_transformed.split_after(pd.Timestamp('19591201'))
train_set_3, val_set_3 = series_transformed.split_after(pd.Timestamp('19600601'))

individual_forecasts_by_same_model = []
same_model = instantiate_BlockRNNModel()
same_model.fit(train_set_1, verbose=True)
individual_forecasts_by_same_model.append(same_model.predict(n=6).values())
same_model.fit(train_set_2, verbose=True)
individual_forecasts_by_same_model.append(same_model.predict(n=6).values())
same_model.fit(train_set_3, verbose=True)
individual_forecasts_by_same_model.append(same_model.predict(n=6).values())

# 3. Fitting three individual model instances with 3 training sections
# From the previous two parts, `historical_forecasts_result` is the same as `individual_forecasts_by_same_model`
# However, I would expect the results given by the `historical_forecasts()` to be the same as 
# fitting a brand-new model on each training section, like the following `individual_forecasts_by_different_models`.
train_set_1, val_set_1 = series_transformed.split_after(pd.Timestamp('19590601'))
train_set_2, val_set_2 = series_transformed.split_after(pd.Timestamp('19591201'))
train_set_3, val_set_3 = series_transformed.split_after(pd.Timestamp('19600601'))

individual_model_1 = instantiate_BlockRNNModel()
individual_model_2 = instantiate_BlockRNNModel()
individual_model_3 = instantiate_BlockRNNModel()
individual_model_1.fit(train_set_1, verbose=True)
individual_model_2.fit(train_set_2, verbose=True)
individual_model_3.fit(train_set_3, verbose=True)
individual_forecasts_by_different_models = [
    model.predict(n=6).values() 
    for model
    in [individual_model_1, individual_model_2, individual_model_3]
]
```

Result
```
historical_forecasts_result:
[array([[0.40726016],
        [0.14743934],
        [0.42281431],
        [0.30751979],
        [0.3935865 ],
        [0.23179494]]),
 array([[0.40049573],
        [0.38148109],
        [0.45629141],
        [0.3826234 ],
        [0.38148096],
        [0.40922707]]),
 array([[0.53207854],
        [0.50775845],
        [0.5872545 ],
        [0.43763365],
        [0.43169829],
        [0.55752873]])]

individual_forecasts_by_same_model:
[array([[0.40726016],
        [0.14743934],
        [0.42281431],
        [0.30751979],
        [0.3935865 ],
        [0.23179494]]),
 array([[0.40049573],
        [0.38148109],
        [0.45629141],
        [0.3826234 ],
        [0.38148096],
        [0.40922707]]),
 array([[0.53207854],
        [0.50775845],
        [0.5872545 ],
        [0.43763365],
        [0.43169829],
        [0.55752873]])]

individual_forecasts_by_different_models
[array([[0.40726016],
        [0.14743934],
        [0.42281431],
        [0.30751979],
        [0.3935865 ],
        [0.23179494]]),
 array([[0.402741  ],
        [0.15947664],
        [0.42768211],
        [0.34189958],
        [0.42966399],
        [0.25139858]]),
 array([[0.46655868],
        [0.2012079 ],
        [0.50961595],
        [0.35336985],
        [0.44060068],
        [0.28922872]])]
```

Save the model before any training, and load the untrained model before retraining.
Alternatively, reset the model to its initial weights before retraining.

dennisbader:
Fixes #1461.

### Summary
Currently `historical_forecasts` trains the same model object multiple times. For TorchForecastingModels this an issue as instead of training a fresh model from scratch for each iteration, we are rather "fine-tuning" the existing model.

- trains a model from scratch when `retrain=True` 

**Important things to consider**
Below are some changes to the current behavior. IMO these make sense, but let me know if think of better approaches:
- currently, calling `historical_forecasts` (HF) on an already fitted model with `retrain=True` will raise an error if input dimensions/covariates/... do not match the input from initial training -> this will not be raised with this PR as we train new model from scratch
- currently, the model instance `self` gets updated (with fit calls) in each iteration of HF using retrain=True -> with this PR, the instance gets only updated in the first iteration if it has not been fit before. (?) Do we even want to "mess" with `self` at all?


darts/models/forecasting/forecasting_model.py
```diff
@@ -871,13 +865,17 @@ def historical_forecasts(
                     if future_covariates_
                     else None,
                 ):
-                    self._fit_wrapper(
+                    # avoid fitting the same model multiple times
+                    model = self if not self._fit_called else self.untrained_model()
```
hrzn, line 869: I'm thinking we should always call `untrained_model()` even for the first pass here, WDYT?

hrzn (commented):
Nice, thanks!!
I would say we shouldn't mess at all anymore with `self` :)
There's an error in the quickstart notebook now, could you have a look?

solalatus:
Extremely good that you clarified this, Kudos! 
Question: would it not be good also to have a "finetune" behavior?
I am right now trying to do a finetuning style transfer learning approach, and this would come in handy.

solalatus:
I recon, this would entail some decisions about what to do with the optimizer state, the LR_schedule,...

I am struggling right now to do a reset of these without touching the model weights. I can give some input if someone is willing to try a less hacky way then me.

dennisbader:
As discussed, leaving the model object unchanged in historical_forecasts() brought some necessary changes with it. See for example the quickstart notebook.

dennisbader:
> Extremely good that you clarified this, Kudos! Question: would it not be good also to have a "finetune" behavior? I am right now trying to do a finetuning style transfer learning approach, and this would come in handy.

Hi @solalatus. I believe the fine tuning would make the historical_forecasts too complex. Also this would only be supported by our TorchForecastingModel (TFM). I would rather opt for making it easier to fine-tune our TFMs in general.

IMO it's best keep this separate, but let me know if I'm missing something

solalatus:
Absolutely agree. I was thinking about some different function then the standard historical forecast. In fact, as I tried to think it over in [this "issue"](https://github.com/unit8co/darts/issues/1471) a reinit feature and the normal fit would suffice.

I hacked together a **really** ugly solution you can see in [this gist](https://gist.github.com/solalatus/e6329879c5c9479ce60af9f6b0e22bb4), any feedback and thoughts are much appreciated!

hrzn (approved):
LGTM, thanks @dennisbader

PR #1469, opened 2023-01-05
Bug fix - correct for multiple lengths of series in input sequence when adding static covariates to regression models

### LINKED ISSUE #1460: [BUG] Static covariates error when training global models on lists of TimeSeries
**Describe the bug**
A ValueError is thrown when attempting to train global models (I've tried LightGBMModel and LinearRegressionModel) on lists of TimeSeries objects that all have the same static covariates assigned. 

The static covariate labels are strings and I've tried both with and without using `StaticCovariatesTransformer()` to convert the strings to float labels.

**To Reproduce**
All TimeSeries objects have the same structure but vary in length. For example:
image

AND

image

Fit the model:
```python
model = LightGBMModel(lags=[-1] , output_chunk_length=1)
model.fit(all_train_scaled, future_covariates=None)
```

Error:
```python
/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.9/site-packages/darts/utils/data/tabularization.py in _add_static_covariates(model, series, features)
    219 
    220         # concatenate static covariates to features
--> 221         return np.concatenate([features, static_covs], axis=1)

/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.9/site-packages/numpy/core/overrides.py in concatenate(*args, **kwargs)

ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 16571 and the array at index 1 has size 16570
```

**Expected behavior**
The `static_covs` array should match the dimensions of the `features` array so that they can be concatenated.

 - Python version: 3.9.5
 - darts version 0.23.0
 - lightgbm version 3.3.2

**Additional context**
DARTS has been a huge boost for me. Thanks for all the effort, especially the effort put into the docs.

eliane-maalouf:
Fixes #1460 

### Summary

In the original version for the code adding static covariates to regression models, it was assumed that the series are of same length.
This PR corrects this assumption and allows to support  cases where the series in the input sequence have different sizes. 

### Other Information
None

### DIFF
--- a/darts/tests/models/forecasting/test_regression_models.py
+++ b/darts/tests/models/forecasting/test_regression_models.py
@@ -655,17 +655,15 @@ class RegressionModelsTestCase(DartsBaseTestClass):
         target_series = series2seq(target_series)
         past_covs = series2seq(past_covs)
         future_covs = series2seq(future_covs)
-
         scovs_set = []
-        scovs_set = scovs_set.extend(
-            s.static_covariates.columns.values if s.has_static_covariates else [""]
-            for s in target_series
-        )
+        for s in target_series:
+            if s.has_static_covariates:
+                scovs_set.extend(s.static_covariates.columns.to_list())
 
         all_target_width = target_series[0].n_components
         scovs_width = (
-            len(set(scovs_set)) if scovs_set is not None else 0 * all_target_width
-        )
+            len(set(scovs_set)) if scovs_set is not None else 0
+        ) * all_target_width
         all_past_width = past_covs[0].n_components if past_covs is not None else 0
         all_future_width = future_covs[0].n_components if future_covs is not None else 0
 
--- a/darts/tests/models/forecasting/test_regression_models.py
+++ b/darts/tests/models/forecasting/test_regression_models.py
@@ -656,14 +656,16 @@ class RegressionModelsTestCase(DartsBaseTestClass):
         past_covs = series2seq(past_covs)
         future_covs = series2seq(future_covs)
 
-        max_scovs_width = max(
-            s.static_covariates_values(copy=False).reshape(1, -1).shape[1]
-            if s.has_static_covariates
-            else 0
+        scovs_set = []
+        scovs_set = scovs_set.extend(
+            s.static_covariates.columns.values if s.has_static_covariates else [""]
             for s in target_series
         )
 
         all_target_width = target_series[0].n_components
+        scovs_width = (
+            len(set(scovs_set)) if scovs_set is not None else 0 * all_target_width
+        )
         all_past_width = past_covs[0].n_components if past_covs is not None else 0
         all_future_width = future_covs[0].n_components if future_covs is not None else 0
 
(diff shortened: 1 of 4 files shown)


darts/utils/data/tabularization.py
```diff
@@ -161,30 +161,115 @@ def _create_lagged_data(
...
+            abs(v)
+            for v in [min_target_lag, min_past_cov_lag, min_future_cov_lag]
+            if v is not None
+        ]
+    )
+
     # collect static covariates info
-    map = {"covs_width": [], "values": []}
-    for ts in series:
+    map = {"covs_width": [], "values": [], "reps": []}
```
hrzn, line 229: This is not a good name as it overwrites the builtin `map`. Could you maybe rename to `map_` ?

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -649,71 +743,176 @@ def test_static_covs_addition(self):
...
+        expected_height, expected_width = self.helper_get_expected_shape_static_covs(
+            reg_model,
+            target_series,
+            past_covs,
+            future_covs,
+            target_lag,
+            past_covs_lag,
+            future_covs_lag,
+        )
+        features = reg_model._create_lagged_data(
```
hrzn, line 912: In these tests you test `_create_lagged_data` directly, which is good. Could we also maybe add another "higher level" test going through all the fit/predict process with one model in a difficult case (series of varying lengths and widths with static covs) and testing that the final outputs are as expected?

hrzn (commented):
LGTM. Just had a small comment about a variable name, and possibly adding one more test. Thanks for coming up with a (non-trivial) fix quickly @eliane-maalouf !

darts/utils/data/tabularization.py
eliane-maalouf: oops!

eliane-maalouf:
I made a modification to the function to control for the order of the static covariates, in case they change among the series in the input. I thought it would be easier to implement the control in code instead of asking the user to make sure a certain order is present in the static covariates.

darts/tests/models/forecasting/test_regression_models.py
eliane-maalouf: ok will do

hrzn:
> I made a modification to the function to control for the order of the static covariates, in case they change among the series in the input. I thought it would be easier to implement the control in code instead of asking the user to make sure a certain order is present in the static covariates.

👍

darts/utils/data/tabularization.py
```diff
@@ -197,24 +281,27 @@ def _add_static_covariates(model, series, features):
...
-                np.concatenate((map["values"][i], pad_zeros), axis=1)
-                if map["covs_width"][i] > 0
-                else pad_zeros
-            )
-            static_covs.append(np.tile(scovs, reps=(reps, 1)))
+        for i in range(len(target_series)):
+            df = scovs_map["df"][i]
+
+            if not df.empty:
+                df = df.reindex(col_names, axis=1, fill_value=0.0)
```
hrzn, line 298: Please make sure this is not affecting the original DF inside the user-provided TimeSeries (otherwise copy it).

hrzn (approved):
LGTM, thanks @eliane-maalouf

PR #1476, opened 2023-01-09
Improvement/statsforecastets: make sf_ets probabilistic + add future_covariate support for sf_ets + add AutoTheta

Beerstabr:
This is the first time I make contribution to a library like this, so please forgive me if I forgot to do something obvious!

Main changes:
- StatsForecastETS is now probabilistic. 
- StatsForecasETS now uses future_covariates through first performing OLS and then perform ETS on the in-sample residuals. 
- Moved StatsForecastAutoARIMA and StatsForecastETS to the same file and extracted some (2) functions.
- Added AutoTheta from the StatsForecast library.
- Removed sf_auto_arima.py and sf_ets.py since the code is now already included in sf_models.py.

Let me know if you'd like something implemented in a different way. Or if you'd like me to comment a bit more on the changes I've made.

I could also have included AutoCES from StatsForecast. Let me know if you want me to include that one too.


hrzn:
Thanks @Beerstabr, that looks very good :) will have a look in detail soon.
Regarding AutoCES, it would be a welcome addition to Darts, but could also be part of another PR later.

darts/models/forecasting/sf_models.py
```diff
@@ -0,0 +1,357 @@
...
+    samples = np.expand_dims(samples, axis=1)
+    return samples
+
+
+def unpack_sf_dict(
+    forecast_dict: dict,
+):
+    """Unpack the dictionary that is returned by the StatsForecast 'predict()' method."""
+    mu = forecast_dict["mean"]
+    std = forecast_dict["hi-68.27"] - mu
```
hrzn, line 356: You could perhaps have the value `68.27` defined as a constant at the top of the file and reused/imported in the different models, so we avoid any risk of copying it wrong as it's used in different places.

darts/models/forecasting/sf_models.py
```diff
@@ -0,0 +1,357 @@
...
+        forecast_dict = self.model.predict(
+            h=n,
+            level=(68.27,),  # ask one std for the confidence interval
+        )
+
+        mu_ets, std = unpack_sf_dict(forecast_dict)
+
+        if future_covariates is not None:
+            mu_linreg = self._linreg.predict(n, future_covariates=future_covariates)
+            mu_linreg_values = mu_linreg.values().reshape(
```
hrzn, line 132:
```suggestion
            mu_linreg_values = mu_linreg.values(copy=False).reshape(
```

darts/models/forecasting/sf_models.py
```diff
@@ -0,0 +1,357 @@
...
+                n,
+            )
+            mu = mu_ets + mu_linreg_values
```
hrzn, line 135: Nice!

hrzn (commented):
Many thanks @Beerstabr , that looks really good and we will definitely be able to include that in the next release of Darts :) I like the way future covariates are now handled.

I just have two fairly important things to ask:
* Could you please add unit tests for anything new that you've added (1-2 tests to make sure future covariates work as expected with ETS, and adding AutoTheta to the suite of tests)
* Could you split the models in separate files (I know i've initially suggested to regroup... but I have second thoughts because of the impact on the documentation...)

I also have a few other very minor comments, but nothing important.

Beerstabr:
@hrzn thanks for having a look!

Will work on your request and comments. They all seem totally reasonable.

darts/models/forecasting/sf_ets.py
```diff
@@ -25,6 +31,12 @@ def __init__(self, *ets_args, add_encoders: Optional[dict] = None, **ets_kwargs)
         This model accepts the same arguments as the `statsforecast ETS
         <https://nixtla.github.io/statsforecast/models.html#ets>`_. package.
 
+        In addition to the StatsForecast implementation, this model can handle future covariates. It does so by first
```
hrzn, line 34: Perfect, thanks 👍

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -51,6 +52,11 @@
     (ARIMA(12, 2, 1), 5.2),
     (ARIMA(1, 1, 1), 24),
     (StatsForecastAutoARIMA(season_length=12), 4.6),
+    (
+        StatsForecastAutoTheta(season_length=12, decomposition_type="multiplicative"),
```
hrzn, line 56: Could you also add the CES model here?

requirements/core.txt
```diff
@@ -13,7 +13,7 @@ requests>=2.22.0
 scikit-learn>=1.0.1
 scipy>=1.3.2
 shap>=0.40.0
-statsforecast>=1.0.0
+statsforecast>=1.4.0
```
hrzn, line 16: Are you sure this is the lowest possible version that supports everything?
I'm double checking because the higher our lower bounds, the more stringent a condition it puts on users who may have constraints on their Pypi settings.

hrzn (commented):
Looks really good, only a few minor comments/questions left. Thanks @Beerstabr !

requirements/core.txt
Beerstabr: I'm going to look into it for a bit more. I think it might depend on if we want AutoCES to be probabilistic as well. But let me check it and I'll come back to you with the options.

Beerstabr: @hrzn, I'm sure now that it is as follows:

 - ETS (from statsforecast 1.3.2 onwards is called AutoETS) is available from 1.0.0, but only made probabilistic in 1.4.0
 - AutoTheta became available in 1.3.0 and was probabilistic from the start
 - AutoCES became available in 1.1.0 and is not yet probabilistic but will be -soon- in the next pip release (1.4.1/1.5.0?)

So if we don't choose 1.4.0 as a requirement, then we will have to drop the probabilistic nature of StatsForecastETS. 
And we might want to rename StatsForecastETS to StatsForecastAutoETS, because it is also renamed by statsforecast. 

What would you like me to put in there as a requirement? And do you want me to rename the ETS model?

hrzn: OK thanks a lot for investigating @Beerstabr .
We can keep 1.4.0 as the lower bound (and even update it once AutoCES is made probabilistic). I think that's OK because anyway all releases of statsforecast are quite recent.
I also think we should rename `StatsForecastETS` to `StatsForecastAutoETS` if they renamed it on their side.
Thanks!

hrzn:
@Beerstabr the only last thing left is to renamed the model to `StatsForecastAutoETS`. Let me know if you're fine with still doing this. Thanks again!

Beerstabr:
@hrzn yes, sorry forgot about it. Will change the name tonight!

hrzn (approved):
Thanks a lot @Beerstabr ! This will go in the next release.

Issue #1466, opened 2023-01-05
Fix `historical_forecasts()` with future covariates


hrzn:
To reproduce:
```python
import os

from darts.models import TFTModel
from darts.datasets import AirPassengersDataset
from darts.utils.timeseries_generation import linear_timeseries

output_chunk_length = 6
series = AirPassengersDataset().load()
past_covs = linear_timeseries(start=series.start_time(), length=len(series), freq=series.freq)
future_covs = linear_timeseries(start=series.start_time(), length=len(series) + output_chunk_length, freq=series.freq)

save_path = os.path.join(os.getcwd(), "nbeats_1455.pt")
model = TFTModel(input_chunk_length=12, output_chunk_length=output_chunk_length)
model.fit(
    series,
    past_covariates=past_covs,
    future_covariates=future_covs,
    verbose=True,
    epochs=3
)
model.save(save_path)

model_loaded = TFTModel.load(save_path)
model_loaded.historical_forecasts(
    series=series,
    past_covariates=past_covs,
    future_covariates=future_covs,
    # start=0.15,
    forecast_horizon=6,
    stride=6,
    retrain=False,
    verbose=False,
)
```

PR #1481, opened 2023-01-11
Fix/multi forecast fut cov

dumjax:
Fixes when only fut cov.


hrzn:
Fixes https://github.com/unit8co/darts/issues/1466

hrzn (approved):
Thanks!

PR #1485, opened 2023-01-13
Fix: Syntax Highlighting in Notebook-Generated Sphinx Pages

### LINKED ISSUE #1468: Restore code highlighting in the examples and quick start
For some reason the highlighting seems to be gone from the notebook-generated pages.

mabilton:
Fixes #1468.

### Summary

It appears that  a bug associated with `ipython==8.7.0` caused syntax highlighting in the notebook-generated example pages to break - see https://github.com/spatialaudio/nbsphinx/issues/687. Explicitly specifying `ipython==8.8.0` in `requirements/release.txt` seems to solve this problem (at least when I compile the docs locally with `Python 3.10.6`).

### DIFF
--- a/requirements/release.txt
+++ b/requirements/release.txt
@@ -1,5 +1,6 @@
 bump2version==1.0.1
 docutils==0.17.1
+ipython==8.8.0
 ipykernel==5.3.4
 ipywidgets==7.5.1
 jinja2==3.0.3


mabilton:
Looking at the online documentation, it appears that syntax highlighting in the examples has since been restored - I guess it must have automatically updated to `ipython==8.8.0`?  To confirm that the `ipython` version was the source of the problem, I specified `ipython==8.7.0` in `requirements/release.txt` - recompiling the docs with this specified `ipython` version does, indeed, reproduce the highlighting bug. With that in mind, I imagine its probably useful to specify `ipython==8.8.0` as a requirement, just to ensure this doesn't happen again.

hrzn (approved):
Many thanks for checking this @mabilton

PR #1486, opened 2023-01-14
Remove unused return type

### LINKED ISSUE #1484: [BUG] TImeSeries.from_group_dataframe incorrect return type
`TimeSeries.from_group_dataframe` is annotated with return type `Union["TImeSeries", List["TimeSeries"]`, however it always returns a `list[TimeSeries]`: https://github.com/unit8co/darts/blob/0.23.1/darts/timeseries.py#L801-L813.

This makes it annoying to use in typed code, as one has to cast the result to `List[TimeSeries]` even though this is guarantied.

**System:**
 - Python version: 3.10
 - darts version: 0.23.1

connesy:
Fixes #1484.

### Summary

Replace the `Union` return type of `TimeSeries.from_group_dataframe` with a `List["TimeSeries"]`, since the function is guarantied to return a list.

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -702,7 +702,7 @@ class TimeSeries:
         fill_missing_dates: Optional[bool] = False,
         freq: Optional[str] = None,
         fillna_value: Optional[float] = None,
-    ) -> Union["TimeSeries", List["TimeSeries"]]:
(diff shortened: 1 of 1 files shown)


dennisbader (approved):
LGTM, thanks!

PR #1496, opened 2023-01-18
Update forecasting_model.py

musabbirsaeed:
I have added a try-except block in the _evaluate_combination(param_combination) at the gridsearch method

Fixes #.

### DIFF
--- a/darts/models/forecasting/forecasting_model.py
+++ b/darts/models/forecasting/forecasting_model.py
@@ -1226,49 +1226,54 @@ class ForecastingModel(ABC, metaclass=ModelMeta):
         )
 
         def _evaluate_combination(param_combination) -> float:
-            param_combination_dict = dict(
-                list(zip(parameters.keys(), param_combination))
-            )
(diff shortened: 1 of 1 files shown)


darts/models/forecasting/forecasting_model.py
```diff
@@ -1226,49 +1226,54 @@ def gridsearch(
...
+                        past_covariates,
+                        future_covariates,
+                        num_samples=1,
+                        verbose=verbose,
+                    )
+                    error = metric(val_series, pred)
 
-            return float(error)
+                return float(error)
+            except Exception as e:
```
hrzn, line 1273: I think it's a good idea, however this `try` block is quite huge and it'd be good to narrow it down if possible. I assume the errors we're trying to catch would happen in fit/predict calls, right? In that case, could we only have try/catch block(s) around the relevant calls?

hrzn:
@musabbirsaeed are you still willing to see this through?

PR #1497, opened 2023-01-19
fix NoneType error for encoder train index generation

dennisbader:
### Summary
- fixes a bug when using a future covariates encoder with a maximum lag value of 0.


hrzn (approved):
Thanks!

Issue #1498, opened 2023-01-19
[BUG] Slicing operator does not work as indented for RangeIndex ts


madtoinou:
**Describe the bug**
According to the documentation of the slice method, the returned ts is "starting later than `start_ts` and ending before `end_ts`" but if the RangeIndex step is greater than one, the beginning of the returned TimeSeries can be smaller than start_ts. Same observation when the start_ts is outside of the sliced TimeSeries (greater than the last value of the time_index), resulting in a non-empty TimeSeries.

**To Reproduce**

1. when the start value is in between the original TimeSeries RangeIndex:

```python
idx = pd.RangeIndex(start=0, stop=60, step=2)
values = np.arange(0,30)
ts = darts.TimeSeries.from_times_and_values(idx, values)
print(ts.slice(11, 31).time_index)
>>> RangeIndex(start=10, stop=30, step=2, name='time')
```

2. when the start value is outside the original TimeSeries RangeIndex:
```python
ts = darts.TimeSeries.from_values(values=np.arange(5))
print(ts.slice(12,16).time_index)
>>> RangeIndex(start=4, stop=5, step=1, name='time')
```

**Expected behavior**
1. The RangeIndex of the returned TimeSeries should start at 12.
2. The RangeIndex of the returned TimeSeries should be empty.

**System :**
 - Python version: 3.9.15
 - darts version: 0.23.1

**Additional context**
I noticed this bug when trying to have a consistent behavior between the __getitem__ operator and the slicing operation, trying to make sure that they were consistent with the documentation of [xarray](https://docs.xarray.dev/en/stable/user-guide/indexing.html#indexing) and [pandas](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html) (which are already consistent with each other).

PR #1499, opened 2023-01-19
Fix/allow empty datetimeindexed ts

### LINKED ISSUE #1279: Empty time series allowed

madtoinou:
Fixes #1279, #1498.

### Summary
The DatetimeIndex completeness check in the TimeSeries constructor is performed only for non-empty TimeSeries, effectively allowing empty Datetime indexed TimeSeries (for #1279).

Additionally, when trying to compare the behavior of the slice method with the slicing operator, noticed a inconsistency for the TimeSeries using RangeIndex: the beginning of the returned TimeSeries was smaller than the `start` argument when the requested start index is greater than the last value in the TimeSeries index or between two of its values (when the index has a step greater than 1). This behavior is contradictory with the docstring of the method, which indicate that the returned TimeSeries beginning should be greater or equal to this `start` argument.

### Other Information
Incrementing the `effective_start_ts` in the slicing method also solved the issue of unexpected non-empty TimeSeries when requesting start and end index outside of the range of the original TimeSeries.


darts/tests/test_timeseries.py
```diff
@@ -121,9 +130,23 @@ def test_integer_indexing(self):
...
+        with self.assertRaises(IndexError):
+            series[100]
+
         # slicing should act the same irrespective of the initial time stamp
         np.testing.assert_equal(series[10:20].values().flatten(), values[10:20])
 
+        # slicing outside of the range should return an empty ts
+        self.assertEqual(len(series[105:110]), 0)
+        # multiply the slice start and end values by 2 to compare the same values
+        self.assertEqual(series[105:110], series.slice(210, 220))
```
hrzn, line 143: Isn't this anyway testing that the two TimeSeries are empty?

hrzn (approved):
LGTM, thanks @madtoinou !

darts/tests/test_timeseries.py
madtoinou: It is, but wanted to properly cover this scenario in case the behavior changes later...

Issue #1495, opened 2023-01-18
Issue with training from checkpoint torch models


rijkvandermeulen:
epochs_trained attribute is not being saved/loaded from the checkpoint after 0.23.0 release. I think this causes a problem when resuming training as there's now a discrepancy with the trainer 'max_epochs` argument.

Can be reproduced by:
- Instantiate and fit a torch model
- Load trained model from checkpoint
- Call fit again with epochs < the # of epochs the model loaded from checkpoint already trained

madtoinou:
Hi,

I am trying to implement a function to wrap the fine-tuning of a "pre"-trained model, but in the meantime, you could use the approach described in this [issue](https://github.com/unit8co/darts/issues/1090) as follow:
```python
import os
import numpy as np
import pandas as pd
import pytorch_lightning as pl
from darts import TimeSeries
from darts.models import NHiTSModel
from darts.models.forecasting.torch_forecasting_model import _get_checkpoint_folder, _get_checkpoint_fname

train_ts = TimeSeries.from_series(pd.Series(np.random.random(100)), freq="1H")

epochs = 20
additional_epochs = 19

# first training
model = NHiTSModel(2,2, n_epochs=epochs, save_checkpoints=True)
model.fit(train_ts, verbose=True)

# retrieving the checkpoint path and load the previous model
ckpt_dir = _get_checkpoint_folder(model.work_dir, model.model_name)
file_name = _get_checkpoint_fname(model.work_dir, model.model_name, best=False)
ckpt_path = os.path.join(ckpt_dir, file_name)

model_retrain = NHiTSModel.load_from_checkpoint(model.model_name, work_dir=model.work_dir, best=False)
trainer_params = model_retrain.trainer_params
# update the training attribute
trainer_params['max_epochs'] = trainer_params['max_epochs'] + additional_epochs

trainer = pl.Trainer(resume_from_checkpoint=ckpt_path, **trainer_params)
# train the model from the checkpoint for the `additional epochs`
model_retrain.fit(train_ts, trainer=trainer)
```

Issue #1471, opened 2023-01-06
Trainer / optimizer / logger reinit feature for Torch models


solalatus:
**Is your feature request related to a current problem? Please describe.**
I am trying to load a previously trained NHiTS model and "finetune" it on a new dataset. It entails changing the tensorboard logger, the resetting and changing the LR for the optimizer, removing LR scheduler, but KEEPING THE WEIGHTS, and continuing with a small learning rate some more epochs on new data.

It was a considerable struggle to get it to work, and I did a super hacky and ugly code. 

https://gist.github.com/solalatus/e6329879c5c9479ce60af9f6b0e22bb4

**Describe proposed solution**
Provide reinit_training() or something alike function that allows for the modification of the trainer / hyperparams of Torch based models, keeping the weights intact, but allowing new training runs eg. for finetuning on new data with different optimizers and learning rates.

**Describe potential alternatives**
A clear and concise description of any alternative solutions or existing features that might solve it.

**Additional context**
General context is transfer learning WITH FINETUNING, so not the widspreadly documented zero shot scenario: I have a big model trained on a large dataset, I want to finetune it on my small new dataset but with very small learning rate so as not to suffer catastrophical forgetting.

PR #1501, opened 2023-01-20
Feat/improved training from ckpt

### LINKED ISSUE #1090: [BUG] TFTModel.load_from_checkpoint and .fit() is returning an error.
**Describe the bug**
First of all we train a model with `TFTModel` with 30 epochs. Then, we aim to do transfer learning by re-training the previous model loading it from last checkpoint. Then, we execute the `.fit(..,epochs=additional_n_epochs)` but an error occurs:

```
File "<string>", line 1, in <module>
File ".../python3.9/site-packages/darts/utils/torch.py", line 70, in decorator
  return decorated(self, *args, **kwargs)
File ".../python3.9/site-packages/darts/models/forecasting/torch_forecasting_model.py", line 771, in fit
  return self.fit_from_dataset(
File ".../python3.9/site-packages/darts/utils/torch.py", line 70, in decorator
  return decorated(self, *args, **kwargs)
File ".../python3.9/site-packages/darts/models/forecasting/torch_forecasting_model.py", line 930, in fit_from_dataset
  self._train(train_loader, val_loader)
File ".../python3.9/site-packages/darts/models/forecasting/torch_forecasting_model.py", line 952, in _train
  self.trainer.fit(
File ".../python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 770, in fit
  self._call_and_handle_interrupt(
File ".../python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 723, in _call_and_handle_interrupt
  return trainer_fn(*args, **kwargs)
File ".../python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 811, in _fit_impl
  results = self._run(model, ckpt_path=self.ckpt_path)
File ".../python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1232, in _run
  self._checkpoint_connector.restore_training_state()
File ".../python3.9/site-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 199, in restore_training_state
  self.restore_loops()
File ".../python3.9/site-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 293, in restore_loops
  raise MisconfigurationException(
pytorch_lightning.utilities.exceptions.MisconfigurationException: You restored a checkpoint with current_epoch=29, but you have set Trainer(max_epochs=5).
```

**To Reproduce**
```
additional_n_epochs=5
my_model = TFTModel.load_from_checkpoint(mymodelname, work_dir=mymodeldir, best=False)
my_model.fit(...,epochs=additional_n_epochs)
```

**Expected behavior**
We aim to get a training process departing from the epoch of last checkpoint and continue until the total number of epochs is: `my_model.n_epochs + additional_n_epochs` .

 - Python version: 3.9
 - darts version 0.18.0

### LINKED ISSUE #1109: Transfer learning using a different learning rate
Hello,

I hope to change the learning rate when fitting a  pretrained LSTM model in a new data set. But I failed to modify it. Is that possible?Could you help me with this problem? It is a bit urgent. Thanks!

madtoinou:
Fixes #1109, #1090, #1495 and #1471.

### Summary
Implement `setup_finetuning`; wrapping the loading of a checkpoint and the setup of the various elements related to the training of a model (trainer, optimizer, lr_scheduler) and returning a model instance which can then directly be trained using the `fit` method. For scenario where the training of a model is resumed from a checkpoint due to an error/crash, `load_from_checkpoint` remain the most efficient approach.

### Other Information
The number of additional training epochs can be provided either using directly the `additional_epochs` argument or using the `trainer_params` dict (if both are provided, a sanity check is performed).
The fine-tuned model checkpoints can either be saved in the same folder as the "original one" or in a different folder (recommended to avoid overwriting the loaded checkpoint).
The fine-tuning can be chained (the _model.ckpt.tar file is created when calling `setup_finetuning`, after updating the trainer parameters) to give granularity to the user and avoid unexpected behaviors.

A big thank to @solalatus for providing a gist with all the attributes to update.

I tried to find methods that would allow to directly update the `model.model_params` attribute but it seems to be performed by PytorchLightning.

Example of use:
```python
import numpy as np
import pandas as pd
from darts import TimeSeries
from darts.models import NHiTSModel
train_ts = TimeSeries.from_series(pd.Series(np.random.random(1000)), freq="1H")

epochs = 5
additional_epochs = 5

# first training
model_name = "original_model"
model = NHiTSModel(2,2, n_epochs=epochs, save_checkpoints=True, model_name=model_name, force_reset=True)
model.fit(train_ts)
print("first training", model.epochs_trained)

# loading the last checkpoint of the original model
model_finetune = NHiTSModel.setup_finetuning(
                                            additional_epochs=additional_epochs,
                                            old_model_name = model_name,
                                            new_model_name = model_name+"_ft",
                                            trainer_params = {"log_every_n_steps":1,
                                                              "enable_model_summary":False},
                                            )
# fine-tune the model for 5 epochs (model.trained_epochs = 10)
model_finetune.fit(train_ts)
print("first finetuning", model_finetune.epochs_trained)

# fine-tune the last checkpoint of the fine-tuned model/checkpoint
model_finetune = NHiTSModel.setup_finetuning(
                                            additional_epochs=5,
                                            old_model_name = model_name+"_ft",
                                            new_model_name = model_name+"_ft2",
                                            trainer_params = {"log_every_n_steps":1,
                                                              "enable_model_summary":False},
                                            )
# fine-tune the model for an additional 5 epochs (model.trained_epochs = 15)
model_finetune.fit(train_ts)
print("second finetuning", model_finetune.epochs_trained)

# loading the last checkpoint of the original model, save checkpoint in place (overwriting the last-epoch checkpoint)
# use the trainer_params to provide the new total number of epochs
model_finetune = NHiTSModel.setup_finetuning(
                                            old_model_name = model_name,
                                            save_inplace=True,
                                            trainer_params = {"log_every_n_steps":1,
                                                              "enable_model_summary":False,
                                                              "max_epochs":30},
                                            )
# fine-tune the model
model_finetune.fit(train_ts)
print("inplace finetuning", model_finetune.epochs_trained)
```


solalatus:
Supercool! Thanks @madtoinou for making the effort! It looks way better then the "raw" version.
I will be able to test it instead of the "raw" in a couple of days. Very cool indeed!

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -916,6 +916,167 @@ def fit_from_dataset(
         self._train(train_loader, val_loader)
         return self
 
+    @staticmethod
+    def setup_finetuning(
+        old_model_name: str,
+        new_model_name: str = None,
+        additional_epochs: int = 0,
```
hrzn, line 923: Do we need this parameter, given that `fit()` already accepts `epochs` ?
I would find it cleaner to rely exclusively on `fit()`'s parameter. If there's a problem with it, could we maybe fix it there (i.e. handle the trainer correctly in `fit()` to handle `epoch`)?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -916,6 +916,167 @@ def fit_from_dataset(
         self._train(train_loader, val_loader)
         return self
 
+    @staticmethod
+    def setup_finetuning(
```
hrzn, line 920: How about extending `load_from_checkpoint()` instead? By making it prepare the model for further training when the extra parameters are passed?
If we keep it separate, I'd rename this one to something like `load_for_training()`. To make it clear that (i) it's loading a model, and (ii) this is for further training ("finetuning" is a little bit more specialised than this and could be understood as training some layers only)

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -916,6 +916,167 @@ def fit_from_dataset(
...
+        file_name: str = None,
+        best: bool = False,
+        save_inplace: bool = False,
+        force_reset: bool = False,
+        optimizer_cls: torch.optim.Optimizer = torch.optim.Adam,
+        optimizer_kwargs: Optional[Dict] = None,
+        lr_scheduler_cls: Optional[torch.optim.lr_scheduler._LRScheduler] = None,
+        lr_scheduler_kwargs: Optional[Dict] = None,
+        **kwargs,
+    ):
```
hrzn, line 935: There's no docstring

hrzn (commented):
That looks quite good but I'm a bit hesitant about introducing a new method. Could you maybe simply improve `load_from_checkpoint()` and fix the `epoch` issue in `fit()` ?

darts/models/forecasting/torch_forecasting_model.py
madtoinou: Extending `load_from_checkpoint()` should be possible, I created another method because I wanted them to cover a different scenario: `load_from_checkpoint()` for when the training have to be resumed (no parameter change) and `setup_finetuning()` if the training is explicitly performed for additional steps, eventually with another scheduler/optimizer/trainer parameters.
I agree with you, the name is a bit confusing, I will adapt it.

darts/models/forecasting/torch_forecasting_model.py
madtoinou: There is an issue with the `fit()` parameter (#1495), I think that @rijkvandermeulen is already working on a fix. I will remove the `epochs` argument from this method and wait for the patch to be merged.

madtoinou:
Took your comments into account;
-  all the functionalities had been moved to `load_from_checkpoint`.
- `additional_epochs` argument has been removed, it's still possible to modify the number of training epochs using the `max_epochs` value of `pl_trainer_kwargs`, it will be overwritten by the value provided by `fit()`
- added the docstring

solalatus:
@madtoinou when I try to use your branch and load a model, I get a `FileNotFoundError`.

The context is: I trained the model on a machine with different home folder name now I want to load it in a new machine with new folder name (which I give in as `work_dir,` param), and it still looks for the model at the old place irrespective of `work_dir`, but only in [here](https://github.com/unit8co/darts/blob/feat/improved-training-from-ckpt/darts/models/forecasting/torch_forecasting_model.py#L1272) 

I tried to trace through, and one thing is suspicious:

Are you sure that this line [here](https://github.com/unit8co/darts/blob/feat/improved-training-from-ckpt/darts/models/forecasting/torch_forecasting_model.py#L1571) should be `model.work_dir` insteab of just `work_dir`?

Am I messing something up? 

Please advise!

solalatus:
For me, this helped:

```
diff --git a/darts/models/forecasting/torch_forecasting_model.py b/darts/models/forecasting/torch_forecasting_model.py
index 0cd9602..5c57a95 100644
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -1565,6 +1565,9 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         model.pl_module_params["lr_scheduler_cls"] = model.model.lr_scheduler_cls
         model.pl_module_params["lr_scheduler_kwargs"] = model.model.lr_scheduler_kwargs
 
+        if work_dir:
+            model.work_dir=work_dir
+
         # save the initialized TorchForecastingModel to allow re-training of already re-trained model
         model.save(
             os.path.join(
```

Don't know if this is too crude, though!

solalatus:
And also - and not wanting to be horrible here - what about [this](https://github.com/unit8co/darts/blob/feat/improved-training-from-ckpt/darts/models/forecasting/torch_forecasting_model.py#L1526) line?

Shouldn't it be more like
```
            if "callbacks" in pl_trainer_kwargs.keys() and len(
                pl_trainer_kwargs["callbacks"] )> 0:
```

I mean the ( is not making sense for me in this case.

But again, I might be missing the point here...

solalatus:
Ok, to be more specific:

```
diff --git a/darts/models/forecasting/torch_forecasting_model.py b/darts/models/forecasting/torch_forecasting_model.py
index 0cd9602..86146fe 100644
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -1523,8 +1523,9 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
             model.trainer_params.update(pl_trainer_kwargs)
 
             # special parameters handling
-            if "callbacks" in pl_trainer_kwargs.keys() and len(
-                pl_trainer_kwargs["callbacks"] > 0
+            if (
+                "callbacks" in pl_trainer_kwargs.keys()
+                and len(pl_trainer_kwargs["callbacks"]) > 0
             ):
                 model.trainer_params["callbacks"] = [
                     checkpoint_callback
```

solalatus:
Ok, I am starting to become annoying. Sorry!

Some suggested refinement:

As it is now implemented [here](https://github.com/unit8co/darts/blob/feat/improved-training-from-ckpt/darts/models/forecasting/torch_forecasting_model.py#L1550) if one does not give any new `lr_scheduler_cls` to the `load_from_checkpoint`, then the loaded scheduler does not get changed. So far so good.

But in my use case, I want to finetune my model, so I definitely want to get rid of the scheduler it was loaded with. I give in an explicit `None`, but that does not help.

Suggestion: Adding the ability to enter `delete` as a string or so, and to be able to get rid of the scheduler altogether.

What do you think @madtoinou ? Any better solutions?

solalatus:
Additional observation: If I used a RAdam during training and I want to switch to SGD after reload, I get  a KeyError about momentum. Cause: currently the optimizer's param dict is not overwritten, but appended to, so whatever I put in load_from_checkpoint, I can not get rid of the momentum param. This is the same type of issue as the one in my previous comment. Maybe plainly overwriting things as default would be a better idea?

madtoinou:
After discussing offline with some others contributors, I decided to refactor the feature and instead of reloading all the attributes of the model's present in the checkpoint (which are relevant only when resuming training, not really for retraining/inference), the user will have to instantiate a new model and then load only the weights from the checkpoint (the method taking care of running some sanity check and initialization of the model without having to call `fit_from_dataset`). This should prevent the issue of "incoherent/non-updated" attribute (looking at you, learning rate scheduler and optimizer parameters) and the user will have full control on what is in the model.

It's almost done, I am currently writing the test and making sure that I am not overlooking a corner case. I'll try to update the PR in the next few days.

solalatus:
Cool! :tada: 

I tested the branch a lot, and I think yours is the good way. Too many combinations of crazy stuff can happen.

Looking forward to the new version! I will test it as soon as I can!

solalatus:
One remark: [This](https://github.com/unit8co/darts/issues/1503) may inform your design decisions.

There would be some benefit in some "setup for fit" like functionality over and beyond just fit itself.

madtoinou:
Yeah saw this other issue you opened, this could easily be included in the PR as well. If we are satisfied with the way I instantiate the model without calling `fit_from_dataset`, we could probably move this part of the code to an utility method that could be called in both scenarios.

Just pushed the latest approach, the method is now called `load_weights_from_checkpoint` (likely to change again):

```python

import numpy as np
import pandas as pd
from darts import TimeSeries

original_model_name = "original"
series = TimeSeries.from_series(pd.Series(np.random.random(100)), freq="1H")

# original model, Adam optimizer 
model_orig = RNNModel(
    12, "RNN", 10, 10,
    save_checkpoints=True,
    model_name=original_model_name,
    force_reset=True,
    pl_trainer_kwargs={"enable_model_summary":False}
)
model_orig.fit(series, verbose=False, epochs=5)

# instantiate a new model, with the same dimensions as the original, eventually with another optimizer
# or work_dir if the retraining is occurring in another path.
model_retr = RNNModel(12, "RNN", 10, 10)
# load the weights, indicate the work_dir if not using the default one
model_retr.load_weights_from_checkpoint(model_name=model_orig.model_name, best=False)
model_retr.fit(series, epochs=10)
``` 

The approach is still a bit hacky because I cannot retrieve the exact shape of the training dataset from the information available in the checkpoint to create a mock `train_sample` (used to define the input/output dim of the model". ATM, I am using an arbitrary large value and it order to make it cleaner, I think that we would need to save additional information in the checkpoints (or force the user to provide this information).

It should work at it is, your feedback would be very welcome. If the `output/input_chunk_length` and `shift` values that you are using are very large, use the `length` and `shift` arguments of `load_weights_form_checkpoint` to make sure that the mock dataset does not complain about the training data being too short to create the lags.

solalatus:
This looks cleaner indeed! I will try it out!

As a first reaction: saving some data is not at all a bad idea, as well as some housekeeping about input shape. On a somewhat related note, I ran into [this](https://github.com/unit8co/darts/issues/1521).

hrzn:
> The approach is still a bit hacky because I cannot retrieve the exact shape of the training dataset from the information available in the checkpoint to create a mock train_sample (used to define the input/output dim of the model". ATM, I am using an arbitrary large value and it order to make it cleaner, I think that we would need to save additional information in the checkpoints (or force the user to provide this information).

Couldn't we "simply" save the train samples dimensions in the checkpoints? AFAIR all we need there is the dimensionality (nr of components) of the input (target + stacked past covariates) and output (target). It might be cleaner to save these two values along with the checkpoint information if that's possible. Asking the user to provide those is not really an option IMO, it would force users to understand some details about the dimensionality of the models.

madtoinou:
Updated the PR; the `PLForecastingModule` is now saving the shape of the train sample so that it can be reconstructed when trying to instantiate the model without calling `fit_from_dataset`. This means that the feature is not back-compatible as the entry `train_sample_shape` won't be present in older checkpoints. Also, `_fit_called` is set to True to enable inference without retraining the laoded weights.

When comparing the hyper-parameters contained in the checkpoint with those of the model, noticed a few inconsistencies in the name of the hyper-parameters (often to make the API of the model constructor more self-explanatory) and I wonder if it would be a good opportunity to uniformize them a bit.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1429,6 +1419,105 @@ def _load_from_checkpoint(self, file_path, **kwargs):
...
+        if file_name is None:
+            file_name = _get_checkpoint_fname(work_dir, model_name, best=best)
+
+        checkpoint_dir = _get_checkpoint_folder(work_dir, model_name)
+
+        ckpt = torch.load(os.path.join(checkpoint_dir, file_name), **kwargs)
+        ckpt_hyper_params = ckpt["hyper_parameters"]
+
+        # verify that the arguments passed to the constructor match those of the checkpoint
+        for param_key, param_value in self.model_params.items():
```
dennisbader, line 1489: If we hard set `strict=True` we could let PyTorch handle any discrepancies later on when calling 
```
self.model.load_state_dict(ckpt["state_dict"], strict=True)
```

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -290,6 +291,74 @@ def test_train_from_10_n_epochs_20_fit_15_epochs(self):
...
+            model_rt = RNNModel(
+                12,
+                "RNN",
+                5,
+                1,
+                n_epochs=5,
+                work_dir=self.temp_work_dir,
+                model_name=retrained_model_name,
+            )
+            model_rt.load_weights_from_checkpoint(
```
dennisbader, line 322: It would be good to add a check here that `model_rt` (before fitting) produces identical forecasts as `model1`

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1429,6 +1419,105 @@ def _load_from_checkpoint(self, file_path, **kwargs):
         pl_module_cls = getattr(sys.modules[self._module_path], self._module_name)
         return pl_module_cls.load_from_checkpoint(file_path, **kwargs)
 
+    def load_weights_from_checkpoint(
+        self,
+        model_name: str = None,
+        work_dir: str = None,
+        file_name: str = None,
+        best: bool = True,
+        strict: bool = True,
```
dennisbader, line 1428: Should we allow `strict=False`? Is there a scenario in which this might be required?

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         self,
         input_chunk_length: int,
         output_chunk_length: int,
+        train_sample_shape: Tuple = None,
```
dennisbader, line 33: Missing parameter definition and
```suggestion
        train_sample_shape: Optional[Tuple] = None,
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1429,6 +1419,105 @@ def _load_from_checkpoint(self, file_path, **kwargs):
...
+
+        # use the name of the model being loaded with the saved weights
+        if model_name is None:
+            model_name = self.model_name
+
+        if work_dir is None:
+            work_dir = os.path.join(os.getcwd(), DEFAULT_DARTS_FOLDER)
+
+        # load PyTorch LightningModule from checkpoint
+        # if file_name is None, find most recent file in savepath that is a checkpoint
```
dennisbader, line 1479:
```suggestion
        # if file_name is None, find the path of the best or most recent checkpoint in savepath
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1429,6 +1419,105 @@ def _load_from_checkpoint(self, file_path, **kwargs):
...
+        work_dir: str = None,
+        file_name: str = None,
+        best: bool = True,
+        strict: bool = True,
+        **kwargs,
+    ):
+        """
+        Load only the weights from automatically saved checkpoints under '{work_dir}/darts_logs/{model_name}/
+        checkpoints/'. This method is used for models that were created with ``save_checkpoints=True`` and
+        that need to be re-trained or fine-tuned with different optimizer or learning rate scheduler.
```
dennisbader, line 1434: they don't necessarily need to be retrained , but can also just be used for prediction

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -290,6 +291,74 @@ def test_train_from_10_n_epochs_20_fit_15_epochs(self):
...
+            # original model, checkpoints are saved
+            model1 = RNNModel(
+                12,
+                "RNN",
+                5,
+                1,
+                n_epochs=5,
+                work_dir=self.temp_work_dir,
+                save_checkpoints=True,
+                model_name=original_model_name,
```
dennisbader, line 306: we should fix the `random_state` with these tests, as there is a possibility that they will fail at some point

dennisbader (commented):
Thanks a lot @madtoinou , this looks great! I think this is a good way to support fine tuning/retraining from now on 🚀 

I added some minor suggestions and that if possible, it would be nice if we added support for loading the weights only from manually saved models

darts/models/forecasting/torch_forecasting_model.py
madtoinou: True but for this kind of use (prediction only, without changing optimizer or learning rate scheduler), wouldn't load_from_checkpoint be more appropriate?

I added the sentence, "However, it can also be used to load weights for inference."

darts/models/forecasting/torch_forecasting_model.py
madtoinou: This is what I did initially but the error message are not informative at all, making it quite difficult for the user to realize that the problem comes from the definition of the model into which the weights are loaded.

madtoinou:
I modified the default path for the checkpoint to replace the ":" (which is not allowed in filename in Windows) with "_" to solve issue #1560 (since it's also load/save related). The resulting name is less straightfoward but should work on all platforms.

This change will prevent the automatic loading of the checkpoints obtained with darts version <= 0.23.1, they will have to be renamed or passed as argument to the loading functions.

The two new methods, `load_weights` and `load_weights_from_checkpoint`, won't be able to automatically load "old" checkpoints as they rely on a new entry in the checkpoint (`train_sample_shape`). It should however be possible to make the old checkpoints compatible by adding manually this entry, using the attribute `train_sample` currently present in loaded models.

madtoinou:
Since it is also related to the original purpose of this PR, I also included a fix for #1561.

Now, If a model is saved directly after loading it from a checkpoint, the original ".ckpt" checkpoint is duplicated so that this model can easily be loaded from the path given to `save()`.

I also added a bit of logic in the `load()` method; if no ".ckpt" can be found, the attribute `_fit_called` is set to `False` so that if the user try to call predict, the correct exception will be raised (instead of some errors about `set_predict_parameters`).

darts/models/forecasting/torch_forecasting_model.py
dennisbader: True, but on the other hand this requires all TorchForecastingModels and their corresponding PLForecastingModules to share the same model parameter names, which is not the case as you mention (and might be difficult to enforce in some cases).

So the torch error can still be raised, or maybe I'm missing something :)

dennisbader (commented):
Thanks again for all the updates and for fixing the other issues 👍 
Had some last minor suggestions, and a fix for loading the models with the correct dtype to load identical weights and produce identical forecasts between original and loaded model.

darts/models/forecasting/forecasting_model.py
```diff
@@ -1555,8 +1555,8 @@ def save(self, path: Optional[Union[str, BinaryIO]] = None, **pkl_kwargs) -> Non
         ----------
         path
             Path or file handle under which to save the model at its current state. If no path is specified, the model
-            is automatically saved under ``"{ModelClass}_{YYYY-mm-dd_HH:MM:SS}.pkl"``.
-            E.g., ``"RegressionModel_2020-01-01_12:00:00.pkl"``.
+            is automatically saved under ``"{ModelClass}_{YYYY-mm-dd_HH_MM_SS}.pkl"``.
```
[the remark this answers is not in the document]
madtoinou: Not sure to understand what you refer to by "deep learning model docs", is it the docstring above each model constructor? If so, they do not make mention of the `save()` `path` argument.

However, I noticed that the docstring of the `model_name` argument was not consistent with the code, namely the described default string is "YYYY-mm-dd_HH:MM:_torch_model_run_PID" whereas the code was generating "YYYY-mm-dd_HH.MM.SS._torch_model_run_PID" (`.` instead of `:`, already compatible with Windows). I suppose that we should probably use the same convention for `model_name` and default `path` in `save()`: `_` or `.` to separate the hours, minutes and second?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1273,9 +1279,24 @@ def save(self, path: Optional[str] = None) -> None:
...
+            if os.path.exists(self.load_ckpt_path):
+                shutil.copy(self.load_ckpt_path, path_ptl_ckpt)
+            else:
+                logger.warning(
+                    f"Model was not trained since the last loading and attempt to retrieve PyTorch "
+                    f"Lightning checkpoint {self.load_ckpt_path} was unsuccessful: model was saved "
+                    f"without its weights."
+                )
+        else:
+            logger.warning(
```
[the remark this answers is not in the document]
madtoinou: This should not occur as the first logic condition is on the `self.load_ckpt_path` attribute which should be `None` for model that are instantiated for the first time with `fit_from_dataset`.

darts/models/forecasting/torch_forecasting_model.py
madtoinou: Indeed, the torch error can still be raised if the discrepancy is in one of these parameter that do not have the same name in these two objects.

We could eventually have a dict in each model that that tries to map the name of the parameters in order to be able to run this sanity check thoroughly? Or try to catch torch error if `load_state_dict` fails and raise a meaningful message to the user, indicating that the weights mismatch can be caused by invalid parameters (or by a change of convention in torch...)?

darts/models/forecasting/torch_forecasting_model.py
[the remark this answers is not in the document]
madtoinou: The warning was removed, the `load()` method takes care of raising a warning if no weights could be loaded.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1429,6 +1455,154 @@ def _load_from_checkpoint(self, file_path, **kwargs):
...
+        # load only the weights from the state dict
+        self.model.load_state_dict(ckpt["state_dict"], strict=strict)
+        # update the fit_called attribute to allow for direct inference
+        self._fit_called = True
+
+    def load_weights(self, path: str, **kwargs):
+        """
+        Loads a model weights from a Pytorch Lightning checkpoint file ('.ckpt') path.
+
+        This method is used for the checkpoints created with ``model.save()``
```
dennisbader, line 1580: (see my comment below)
```suggestion
        Loads the weights from a manually saved model (saved with :meth:`save() <TorchForecastingModel.save()>`).
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1429,6 +1455,154 @@ def _load_from_checkpoint(self, file_path, **kwargs):
...
+
+        Parameters
+        ----------
+        path
```
dennisbader, line 1584: Sorry, I just saw this now. I think we should use the path that the user gave when manually saving the model, i.e. `model.save("my_model.pt")`, rather than the .ckpt path.

Then we just replace ".pt" with ".pt.ckpt" and get the checkpoint from there. Check here that the ckpt exists similar to how do it now in `TorchForecastingModel.load()`

dennisbader (commented):
Really great work @madtoinou , thanks for that!

Just one last change (that I missed earlier, sorry) and then it's ready to merged!

darts/models/forecasting/torch_forecasting_model.py
madtoinou: Good catch, it makes the interface much more consistent and intuitive.

The `load_weights` function now expects the `.pt` path and the `.ckpt` is added afterward (it's actually concatenated, not replaced).

dennisbader (approved):
Thanks for this great PR, everything looks good now 🚀 
Feel free to merge!

PR #1502, opened 2023-01-20
Fix/solve warnings in unittests

madtoinou:
### Summary
Various modifications to eliminate warning raised during tests:
- close matplotlib figures
- updated various deprecated arguments; freq for pd.Timestamp, CountryHolidays for holidays, line_terminator for pd.to_csv, n_init for slearn.Kmeans
- use pl_trainer_kwargs to set `log_every_n_steps` to 1 when model are trained for a very small number of steps
- avoid catastrophic precision loss due to nan in skew and kurtosis computation
- tweak kernel parameters of gaussian process filters

### Other Information
A few warning are caused by dependencies, notably shap (import of IPython.core.display and use of np.bool).

98 warnings (over the remaining 128) are caused by Maximum Likelihood optimization failing to converge (certainly models from the statsmodels library, in test_local_forecasting_models.py). I tried to increase maxiter argument but it did not solve the problem, suggestions are welcome!
The remaining warnings are for the majority caused by either "Non-stationary starting autoregressive parameters", "non-invertible starting MA parameters" (both in test_local_forecasting_models.py) or "test statistic outisde of the range of p-values available in the look-up table" (in test_granger_causality testcase)


hrzn (approved):
Very nice, thanks a lot @madtoinou !
We can already merge this one, even if more optimizations could be done later on.

Issue #1300, opened 2022-10-18
Make transformers show warning or exception when there is mismatch between number of series used for fit() and and transform()


hrzn:
see #1288.

- We should raise a warning or even an exception when we have a mismatch in number of series used for transformer.fit() and transformer.transform()
- we could also add a mappable or indices of the specific series we want to transform

madtoinou:
This should be solved by #1409, which implements a mappable to match the `fit()` and the `transform()` series and raises an error in case of mismatch [code snippet in the PR](https://github.com/unit8co/darts/pull/1409/files#diff-5249055987cc2108cf8d2e208cd50247c61be5bcbe6a8061e42f92d021cdef6bR375-R381)

PR #1508, opened 2023-01-23
Feat/improved data-transformers

### LINKED ISSUE #1288: Improvements for Scalers applied on multiple series
**Describe the bug**
If a `Scaler` is fitted on `n_fitted` sequences at once. It will always only returns `n_fitted`  Sequences but not the number of inputted sequences. 

**To Reproduce**
```
     scaler = Scaler()
    fitted_n = 2
    predicted_n = 3
    s = TimeSeries.from_times_and_values(pd.date_range("2022-01-01", "2022-01-10"), range(10))

    scaler.fit_transform([s]*fitted_n)

    ss_scaled = scaler.transform([s]*predicted_n)
    ss_scaled_inverted = scaler.inverse_transform(ss_scaled)
    ss_inverted = scaler.inverse_transform([s]*predicted_n)

    assert len(ss_scaled) == predicted_n # fails == fitted_n
    assert len(ss_scaled_inverted) == predicted_n # fails == fitted_n
    assert  len(ss_inverted) == predicted_n # fails == fitted_n
```

**Expected behavior**
Should scale all series independently and return the same number as inputed. 

 - Python version: 3.7
 - darts version: 0.21.

madtoinou:
Fixes #1288,  #1300.

### Summary
- `Scaler` verifies that the dimensions of the input to `transform()` and `inverse_transform()`  match the dimensions of the fitted parameters, raise an exception if the condition is not verified.
- Fixed a small typo in `BaseDataTransformer`, the fitted parameters were not properly stored in the attribute.

### Other Information
Would be happy to also implement the Mappable support in the same PR, I just need to think of a nice way to implement it without breaking the support of Sequence.

**[EDIT]** My first approach was to place the logic in `Fittable`-/`InvertibleDataTransformer` but it was requiring transversal modifications to cover all the types of transformation: I tried to add an attribute `input_dim_invariant` to bypass the input dimension check for the relevant Transformers (notably for `StaticCovariatesTransformer`) but it was too messy so I moved the logic to `Scaler`)

### DIFF
--- a/darts/dataprocessing/transformers/fittable_data_transformer.py
+++ b/darts/dataprocessing/transformers/fittable_data_transformer.py
@@ -218,7 +218,7 @@ class FittableDataTransformer(BaseDataTransformer):
         raise_if(
             len(data) != len(self._fitted_params),
             f"There is a mismatch between the number of TimeSeries used during training"
-            f" ({len(self._fitted_params)}) and transformation ({len(data)}).",
+            f" ({len(self._fitted_params)}) and to transform ({len(data)}).",
             logger,
         )
 
--- a/darts/dataprocessing/transformers/fittable_data_transformer.py
+++ b/darts/dataprocessing/transformers/fittable_data_transformer.py
@@ -7,7 +7,7 @@ from abc import abstractmethod
 from typing import Iterator, List, Sequence, Tuple, Union
 
 from darts import TimeSeries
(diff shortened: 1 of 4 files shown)


madtoinou:
Already covered by #1409, which is bringing a more in depth improvement of the data-transformers.

Issue #1491, opened 2023-01-18
[BUG] TimeSeries.from_group_dataframe without time_col but with pd.RangeIndex fails


2-5:
**Describe the bug**
`TimeSeries.from_group_dataframe` fails when called without specifying `time_col`, even if the dataframe has a `pd.RangeIndex` as documented.

**To Reproduce**

```python
import pandas as pd

from darts import TimeSeries

data = {
    "a": [5, 6, 7, 8],
    "b": [9, 9, 0, 0],
}
df = pd.DataFrame(data=data)

print(df)
print()
print(df.index)
print()

ts = TimeSeries.from_group_dataframe(df, "a")
print(ts)
```

**System (please complete the following information):**
 - Python version: 3.10
 - darts version 0.23.1

**Additional context**

```python
   a  b
0  5  9
1  6  9
2  7  0
3  8  0

RangeIndex(start=0, stop=4, step=1)

ValueError: If time_col is not specified, the DataFrame must be indexed either with a DatetimeIndex, or with a RangeIndex.
Traceback (most recent call last):
  File "test.py", line 16, in <module>
    ts = TimeSeries.from_group_dataframe(df, "a")
  File ".venv\lib\site-packages\darts\timeseries.py", line 802, in from_group_dataframe
    return [
  File ".venv\lib\site-packages\darts\timeseries.py", line 803, in <listcomp>
    TimeSeries.from_dataframe(
  File ".venv\lib\site-packages\darts\timeseries.py", line 656, in from_dataframe
    raise_if_not(
  File ".venv\lib\site-packages\darts\logging.py", line 78, in raise_if_not
    raise ValueError(message)
ValueError: If time_col is not specified, the DataFrame must be indexed either with a DatetimeIndex, or with a RangeIndex.
```

madtoinou:
Hi!

Thank you for reporting the bug, it seems to be caused by the output of `pd.groupby()` that replaces the `RangeIndex` of the original `DataFrame` by a `Int64Index` (which is deprecated since pandas 1.4.0).

The three solutions I can think of are:
- remove support for RangeIndex if the `time_col` is not provided, forcing the user to create a DatetimeIndex prior to calling `from_group_dataframe`. Even by doing so, the user will probably have to provide the `freq` argument to avoid a `ValueError` about "missing freq that could not be inferred" (unless the observations in each group follow a pattern)
[...]

PR #1527, opened 2023-01-31
Feat/timeseries integer index

### LINKED ISSUE #1226: Extend freq inference, and support for specifying freq in factory methods, to integer-indexed series
Currently our factory methods only accept an `Optional[str]`, but they could also accept an integer for the case of integer indexed series. Furthermore, the freq could also be inferred when possible (similar to what we do for Datetime indexed series).

### LINKED ISSUE #1528: Add support for valid integer index input to TimeSeries creation factory methods
We should add support for valid integer index (non-RangeIndex) input to TimeSeries creation factory methods.
A valid integer index is one that we can convert to a pandas RangeIndex.

dennisbader:
Addresses #1491 fixes #1226, fixes #1528.

### Summary
- adds support for Integer indexed input to `TimeSeries.from_*` factory methods (dataframes/series/xarrays/times_and_values, ...) if the index can be converted to a pandas RangeIndex. We basically have now the same support as we do for DatetimeIndex input (support for unsorted input, different frequencies (step size other than 1), fill_missing_dates, provided `freq`, ...)
  - since we already have full support of RangeIndex input, we can simply check for a valid integer Index input and then convert to a RangeIndexed xarray 
- added unit tests for new integer index support
- drastically speed up time series creation with `fill_missing_dates=True` for DatetimeIndex


darts/timeseries.py
```diff
@@ -603,63 +633,55 @@ def from_dataframe(
...
-                elif np.issubdtype(time_col_vals.dtype, np.datetime64):
+            if np.issubdtype(time_col_vals.dtype, np.integer):
+                # We have to check all integers appear only once to have a valid index
+                raise_if(
+                    time_col_vals.duplicated().any(),
+                    "The provided integer time index column contains duplicate values.",
+                )
+
+                # Temporarily use an Int64Index (soon to be NumericIndex) to sort the values,
+                # and replace by a RangeIndex in `TimeSeries.from_xarray()`
```
madtoinou, line 657: Are these comments still relevant? I have the impression that this is not using Int64Index (which are going to be deprecated by pandas soon).

madtoinou (approved):
Looks good!

darts/timeseries.py
```diff
@@ -4233,66 +4269,138 @@ def _fill_missing_dates(
...
+                ),
+                logger,
+            )
+
+        freq = observed_freqs.pop()
+
+        return cls._restore_xarray_from_frequency(sorted_xa, freq)
+
+    @staticmethod
+    def _sort_index(xa: xr.DataArray) -> xr.DataArray:
```
hrzn, line 4310: Two points related to this method:
* Could we maybe call it from the constructor as well? I think there we have such a case of "sort-and-copy or copy", where we could reuse this method.
* Do we always need the copy? Looking the places from where this method is called, I think some of them may not need the `.copy()` and we could avoid copying in a few places where it's not needed. I haven't checked in details but I think the copying might be needed only in the constructor (as all factory methods end up calling the constructor, if it's done there we know that the `TimeSeries` will carry its own copy anyway and we can avoid copying earlier in the factory methods). I would recommend maybe having a `copy` boolean argument here. Given how often TimeSeries are built everywhere, avoiding these copies as much as possible could make a difference.

hrzn (commented):
Looks good 👍 I only have a comment related to reducing the number of copies of the DataArrays in some cases.

darts/timeseries.py
```diff
@@ -4233,66 +4269,138 @@ def _fill_missing_dates(
...
         step_size = 3
-        n_dates = len(time_index)
-        # this creates n steps containing 3 timestamps each; used to infer frequency of time_index
+        n_dates = len(index)
+        # this creates n steps containing 3 timestamps each; used to infer frequency of index
         steps = np.column_stack(
-            [time_index[i : (n_dates - step_size + (i + 1))] for i in range(step_size)]
+            [index[i : (n_dates - step_size + (i + 1))] for i in range(step_size)]
         )
         observed_freqs = {pd.infer_freq(step) for step in steps}
```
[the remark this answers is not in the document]
dennisbader: Yes that's true. Maybe we could look unique time deltas only between time stamps `i+1` and `i`.
- find unique time deltas (similar to step size for integer index)
- infer the frequency for all these unique deltas (maybe twice per unique delta, once using (i-1, i, i+1) and the other with (i, i+1, i+2)
- continue as before

This will probably in most cases drastically reduce the examples `steps` we have to infer.
What do you think @hrzn ?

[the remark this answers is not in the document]
dennisbader: @hrzn. Okay, so I came up with an optimized approach here. This reduces time series creation time drastically. See below (left image shows creation time in 10^(n) seconds, right image shows time in seconds).
This new approach is around 300 times faster for larger series! (I didn't compute series length 10^7 for the old approach as it's too slow).

image

darts/timeseries.py
dennisbader: 👍 achieved approx. a 2% increase in time series creation time with larger series.

darts/timeseries.py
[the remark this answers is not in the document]
hrzn: Wow, that's super cool 👍 well done!

darts/timeseries.py
hrzn: Nice!

hrzn (approved):

Issue #1545, opened 2023-02-08
Add feature values and SHAP explanation object to ExplainabityResult


rijkvandermeulen:
Currently the ExplainabilityResult object only contains the explained forecast (i.e., the shapley values for each prediction). I think it would be useful to also store the corresponding feature values and potentially also the SHAP explanation object itself. This will give the end user more options to work with the forecasts that are explained.

hrzn:
Sounds like a good idea to me, adding to backlog.

dennisbader:
Fixed in #1546

PR #1546, opened 2023-02-08
Feature/add feature values to explainability result

rijkvandermeulen:
#1545 

### Summary
Currently the ExplainabilityResult object only contains the explained forecast (i.e., the shapley values for each prediction). I think it would be useful to also store the corresponding feature values and the SHAP explanation object itself. This will give the end user more options to work with the explainability results (e.g., creating plots, doing analysis etc). In any case, having this functionality added would help how we use darts a great deal so from a personal pov I'd love to see this added :)

### Other Information
Since I added two methods that very much resemble 'get_explanation' I've written a decorator to take care of the validation steps. Main goal here was to avoid duplicating the validation checks across multiple methods. If you think this makes the code necessarily complex and/or less readable let me know - then we can change it back.

@hrzn @dumjax 
WDYT? Do you see the added value of this? Any thoughts on how to make it better?


darts/explainability/shap_explainer.py
```diff
@@ -580,6 +601,7 @@ def shap_explanations(
                                 :, :, self.target_dim * (h - 1) + t_idx
                             ]
                         )
+                        tmp_t.data = shap_explanation_tmp.data
```
dennisbader, line 604: is this required?

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -267,19 +268,28 @@ def test_explain(self):
         with self.assertRaises(ValueError):
             # wrong horizon
             results.get_explanation(horizon=5, component="price")
+            results.get_feature_values(horizon=5, component="price")
```
dennisbader, line 271: Each of these checks should be within a dedicated `with self.assertRaises(ValueError)` scope.

Currently, as soon as one check failed it will skip the ones afterwards and exit the `with` scope.

Would you mind fixing this in general in this test file (I see some occurences also in the existing code, not related to your PR) :)

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -301,12 +313,17 @@ def test_explain(self):
...
         self.assertTrue(len(ts_res) == 1)
         self.assertTrue(ts_res.time_index[-1] == pd.Timestamp(2014, 6, 5))
+        feature_vals = results.get_feature_values(horizon=2, component="power")
+        self.assertTrue(len(feature_vals) == 1)
+        self.assertTrue(feature_vals.time_index[-1] == pd.Timestamp(2014, 6, 5))
 
         with self.assertRaises(ValueError):
             # wrong horizon
             results.get_explanation(horizon=5, component="price")
+            results.get_feature_values(horizon=5, component="price")
```
dennisbader, line 323: here again a dedicated `with` scope (as from my comment above)

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -552,3 +569,51 @@ def test_plot(self):
...
+            lags=4,
+            output_chunk_length=1,
+        )
+        model.fit(
+            series=self.target_ts,
+        )
+        shap_explain = ShapExplainer(model)
+        explanation_results = shap_explain.explain()
+        df = pd.merge(
+            self.target_ts.pd_dataframe(),
```
dennisbader, line 584: We could try to make this more robust if we also test `explain()` or `get_feature_values()` output against the actual shap output.

darts/explainability/explainability_result.py
```diff
@@ -16,6 +18,45 @@
 logger = get_logger(__name__)
 
 
+def check_input_validity_for_querying_explainability_result(func: Callable) -> Callable:
```
dennisbader, line 21: Could we simply make this a normal function and call it at the begging in each method?

Wrappers can swallow up the actual method/function arguments in the documentation (from IDE for example) without some additional code

darts/explainability/explainability_result.py
```diff
@@ -28,9 +69,21 @@ def __init__(
             Dict[integer, Dict[str, TimeSeries]],
             Sequence[Dict[integer, Dict[str, TimeSeries]]],
         ],
+        feature_values: Union[
```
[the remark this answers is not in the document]
rijkvandermeulen: Great suggestion; makes sense!

darts/explainability/shap_explainer.py
rijkvandermeulen: Yep, I believe it is. Reason being that in this particular if condition the shap.Explanation object is 'manually' constructed so the tmp_t dict doesn't contain the feature values (i.e., data) yet.

darts/tests/explainability/test_shap_explainer.py
rijkvandermeulen: haha silly mistake - thx for spotting it. Adjusted it across the whole test file

darts/tests/explainability/test_shap_explainer.py
rijkvandermeulen: Done in separate unit test

rijkvandermeulen:
@dennisbader thanks for the review! I believe I've addressed all your comments and suggestions. Let me know in case you've any additional remarks :)

darts/explainability/explainability_result.py
```diff
@@ -81,10 +103,80 @@ def get_explanation(
...
+                Dict[integer, Dict[str, shap.Explanation]],
+                Sequence[Dict[integer, Dict[str, shap.Explanation]]],
+            ]
+        ],
+    ):
+        super().__init__(explained_forecasts)
+        self.feature_values = feature_values
+        self.shap_explanation_object = shap_explanation_object
+
+    def get_feature_values(
```
dennisbader, line 137: We could further reduce the code here by adding a private method to ExplainabilityResult (see below):

```
    def _query_explainability_result(
        self,
        attr: Union[Dict[integer, Dict[str, TimeSeries]], Sequence[Dict[integer, Dict[str, TimeSeries]]]],
        horizon: int,
        component: Optional[str] = None,
    ) -> Any:
        """
        Returns one or several `TimeSeries` representing the feature values
        for a given horizon and component.

        Parameters
        ----------
        attr
            An explainability result attribute from which to extract the horizon and comonent.
        horizon
            The horizon for which to return the feature values.
        component
            The component for which to return the feature values. Does not
            need to be specified for univariate series.
        """
        self._validate_input_for_querying_explainability_result(horizon, component)
        if isinstance(attr, list):
            return [attr[i][horizon][component] for i in range(len(attr))]
        else:
            return attr[horizon][component]
```

Now we can simply call this method from all the getters (explanation, features, shap expl object):

```
def get_feature_values(...):
    return self._query_explainability_result(self.feature_values, horizon=horizon, component=component)

File: darts/explainability/explainability_result.py [hunk 8]
```diff
@@ -81,10 +103,80 @@ def get_explanation(
...
+        self,
+        explained_forecasts: Union[
+            Dict[integer, Dict[str, TimeSeries]],
+            Sequence[Dict[integer, Dict[str, TimeSeries]]],
+        ],
+        feature_values: Union[
+            Dict[integer, Dict[str, TimeSeries]],
+            Sequence[Dict[integer, Dict[str, TimeSeries]]],
+        ],
+        shap_explanation_object: Optional[
```
dennisbader, line 126: I guess the Optional can now be removed

darts/tests/explainability/test_shap_explainer.py
```diff
@@ -555,3 +578,77 @@ def test_plot(self):
...
+
+        feature_values = explanation_results.get_feature_values(
+            horizon=1, component="price"
+        )
+        comparison = explanation_results.get_shap_explanation_object(
+            horizon=1, component="price"
+        ).data
+
+        assert_array_equal(feature_values.values(), comparison)
+        self.assertTrue(
```
dennisbader, line 627: shouldn't this be assertEqual?

dennisbader (changes requested):
Looks great @rijkvandermeulen, thanks for the updates!

I only had one suggestion for refactoring :) After that we can merge this one! 🚀

rijkvandermeulen:
@dennisbader implemented your suggestions! :)

BTW, not very important and also not something for this PR but I was thinking about whether it would make sense to at some point refactor the attributes in ExplainabilityResult (which are now nested dictionaries) to dataclasses? Might potentially make it even more robust and easier to maintain.

dennisbader (approved):
Thanks a lot for this @rijkvandermeulen, looks great :) 🚀 

And I like your idea about refactoring the ExplainabilityResult attributes with dataclasses 👍
That could make our lives much easier also for extending the Explainability module.

Issue #951, opened 2022-05-10
[New Model] Forecasting with Moving Average


marduk2:
**Is your feature request related to a current problem? Please describe.**

Moving Average (MA) is used as a "baseline model" in many time-series forecasting use cases. With the current darts library, we only have access to MA as:

- a Filtering Model (issue: only applicable to past time series),
- or as a special case of the ARIMA models (issue: quite slow, if in the end all we want to obtain is a MA).

**Describe proposed solution**

Implement MA as a Forecasting Model class, so all the underlying functions (fit, predict, backtesting, gridsearch, etc.) can be used. Potentially the work already done for MA Filtering Model can be leveraged.

**Describe potential alternatives**

- An example or code snippet on how to use the existing MA Filtering Model for this purpose (viz. **predict**, not filter). This is somewhat what I expected to find in this example (https://unit8co.github.io/darts/examples/10-Kalman-filter-examples.html) upon reading its title ("Filtering and predicting using the darts filters").
- Speed up ARIMA models drastically if a pure MA is requested.

Thanks in any case for the great library, awesome work!

hrzn:
+1, adding to our backlog, thanks for the suggestion.
Also, please let us know if you would be interesting to make a contribution around this.

marduk2:
@hrzn my Python and GitHub chops need to improve a lot before I can confidently contribute, but if there are any non-technical improvements you might already have identified, glad to help there.

In the meantime we'll have a look at open issues in case we can shed light on some of them.

shaido987:
Since the model is first fit and then used to predict future values, the prediction of a moving average model would always be the mean of the last `window` number of values in the time series used for training (with a constant value as the prediction independent of the forecast horizon). This would be equivalent to using the `NaiveMean` on the last `window` of the time series. Is this logic correct?

hrzn:
> Since the model is first fit and then used to predict future values, the prediction of a moving average model would always be the mean of the last `window` number of values in the time series used for training (with a constant value as the prediction independent of the forecast horizon). This would be equivalent to using the `NaiveMean` on the last `window` of the time series. Is this logic correct?

Yes I think that's correct.

JanFidor:
@hrzn Hi, could I pick up this issue?

hrzn:
@JanFidor yes, that'd be great! On second thoughts thought I'm not quite sure this statement is correct:
[...]

PR #1557, opened 2023-02-10
Add NaiveMovingAverage baseline model

JanFidor:
Fixes [#951 ](https://github.com/unit8co/darts/issues/951)

### Summary
I did a bit of research, and couldn't find anything that would support auto-regressive rolling window, so I implemented it from the ground up. I've decided to use numpy as the backbone of my implementation, main reasons being:
1) Easy incorporation of multivariate series support with vector operations
2) Readable
3) It's very popular when it comes to data science and ml in python, so anyone deciding to contribute to / browse darts' code will most likely be familiar with it

There are a few shortcomings with it, mainly iterating by index and using it to mutate the ndarray, which is a concatenation of the forecasts and the starting sliding window. Personally it feels a little iffy when it comes to clean code, but this approach saves us from an if-else block, which would be even less readable. 

My first idea was to use a double linked list -> collections.deque for its popLeft() / append() methods and dynamically keeping a sliding window of forecasts. In the end I decided it wasn't a good idea to introduce this kind of algorithmic heavy approach out of the blue, so I shelved it and went with numpy


darts/models/forecasting/baselines.py
```diff
@@ -123,6 +123,55 @@ def predict(self, n: int, num_samples: int = 1, verbose: bool = False):
         return self._build_forecast_series(forecast)
 
 
+class NaiveMovingAverage(LocalForecastingModel):
+    def __init__(self, W: int = 1):
```
madtoinou, line 127: It would be great to have something more meaningful than "W" as argument, something such as "window", "window_size" or "input_chunk_length" to be consistent with other models.

darts/models/forecasting/baselines.py
```diff
@@ -123,6 +123,55 @@ def predict(self, n: int, num_samples: int = 1, verbose: bool = False):
...
+    def __str__(self):
+        return f"Naive moving average model, with W={self.W}"
+
+    def fit(self, series: TimeSeries):
+        super().fit(series)
+
+        raise_if_not(
+            len(series) >= self.W,
+            f"The time series requires at least W={self.W} points",
+            logger,
```
madtoinou, line 153: This test is unnecessary, it will be trigerred in `ForecastingModel.fit()` as you already overrode the `min_train_series_length` property.

darts/models/filtering/moving_average_filter.py
```diff
@@ -7,7 +7,7 @@
 from darts.timeseries import TimeSeries
 
 
-class MovingAverage(FilteringModel):
+class MovingAverageFilter(FilteringModel):
```
madtoinou, line 10: I am not sure to understand why it is necessary to rename this filtering model?

madtoinou (changes requested):
Thank you for contributing to darts! The PR looks great, just a few small things to address in my opinion.

The rolling-window implementation looks nice, however the `np.convolve` method could probably have been used in order to obtain a one-liner and avoid the for loop (but at the cost of readability). I don't know which one of the two options is closer to darts' mentality, I'll let @dennisbader  decide :)

darts/models/forecasting/baselines.py
JanFidor: Now that you mention it, I have to agree that it could use a better name, for consistency's sake I'll go with  "input_chunk_length"

darts/models/filtering/moving_average_filter.py
JanFidor: In the issue and the original [PR](https://github.com/unit8co/darts/pull/1201) one of the points was that "MovingAverage" might be a bit misleading for users who might mistake it for the new prediction model. When taking over the issue I thought about it and decided to do the same, but if you think it's not worth the trouble I'll change it back

madtoinou: My bad, I was not aware of this conversation, `MovingAverageFilter` it is then!

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -84,6 +85,7 @@
     (NaiveSeasonal(), 32),
     (NaiveMean(), 37),
     (NaiveDrift(), 39),
+    (NaiveMovingAverage(W=5), 34),
```
hrzn: Shouldn't this be `input_chunk_length`? How come this test is passing?

hrzn (commented):
Thank you @JanFidor , looks good! I just have a few pretty small comments.

darts/models/forecasting/baselines.py
```diff
@@ -123,6 +123,51 @@ def predict(self, n: int, num_samples: int = 1, verbose: bool = False):
...
+        super().__init__()
+        self.input_chunk_length = input_chunk_length
+        self.rolling_window = None
+
+    @property
+    def min_train_series_length(self):
+        return self.input_chunk_length
+
+    def __str__(self):
+        return f"Naive moving average model, with input_chunk_length={self.input_chunk_length}"
```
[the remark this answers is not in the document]
JanFidor: Yep, making it code-like looks much sharper and to the point. I looked around the repo and it seems like most models share this style, but there are some outliers including other NaiveModels. What do you think about opening a new issue for bringing other models in line with it?

darts/tests/models/forecasting/test_local_forecasting_models.py
JanFidor: It wasn't actually, at [f453990](https://github.com/unit8co/darts/pull/1557/commits/f4539901c257ef58d1c9e89ded46c2d90f1fcaaa) I fixed that, so I must've pushed it a bit after you pulled my previous changes

darts/models/forecasting/baselines.py
```diff
@@ -123,6 +123,53 @@ def predict(self, n: int, num_samples: int = 1, verbose: bool = False):
...
+    @property
+    def min_train_series_length(self):
+        return self.input_chunk_length
+
+    def __str__(self):
+        return f"NaiveMovingAverage({self.input_chunk_length})"
+
+    def fit(self, series: TimeSeries):
+        super().fit(series)
+        assert series.n_samples == 1, "This model expects deterministic time series"
```
madtoinou, line 150: Instead of this assert, it would be cleaner in my opinion to use the `raise_if` method combined with the `is_deterministic` attribute of the `TimeSeries`.

madtoinou (approved):

darts/models/forecasting/baselines.py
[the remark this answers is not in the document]
hrzn: That's a good idea 👍

darts/models/forecasting/baselines.py
```diff
@@ -147,8 +147,8 @@ def __str__(self):
 
     def fit(self, series: TimeSeries):
         super().fit(series)
-        assert series.n_samples == 1, "This model expects deterministic time series"
-
+        raise_if_not(series.is_deterministic, "This model expects deterministic time series")
+
```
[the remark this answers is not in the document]
JanFidor: Thanks for the review and sorry for overlooking this. One quick question, do we always want to pass logger to `raise_if_not` and `raise_if` ? I noticed that it isn't always present in other models, if that's the case I'd include them in my refactor at https://github.com/unit8co/darts/issues/1586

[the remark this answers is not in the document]
madtoinou: I am not 100% sure but I have the impression that it's better, allow to control the logging with more granularity.

hrzn (approved):
LGTM, thanks @JanFidor !

JanFidor:
Just wanted to say that I had an absolute blast contributing this PR, in big part thanks to your reviews, explanations and patience with the occasional workflow fail on black ; P . Thanks @madtoinou  @hrzn !

hrzn:
> Just wanted to say that I had an absolute blast contributing this PR, in big part thanks to your reviews, explanations and patience with the occasional workflow fail on black ; P . Thanks @madtoinou  @hrzn !

That's great to hear! And your contribution is greatly appreciated 👍 thanks!

Issue #1367, opened 2022-11-15
Add a pinball loss metric and deprecate rho_risk (?)


deltamacht:
I've been playing around with some probabilistic predictions with various models. I've done an LGBM quantile regression as well as a TFT model with probabilistic outputs. In both cases I computed the quantiles [0.05, 0.5, 0.95]. When I view the output I find that the 0.05 and 0.5 quantiles are pretty similar between the two models, whereas the LGBM's 0.95 quantile is much, much higher than that of the TFT (and is generally a lot higher than the actual data). However, when I computed the rho_risk for quantiles 0.05, 0.5, and 0.95, I find that the two models have similar rho risk for 0.5 and 0.95, but much different rho risk for 0.05 (so, the opposite of what I would have expected looking at the plots).

This led me down the path of trying to figure out how this could be, and looking at the rho_risk code itself. While it seems like the implementation of rho_risk is according to the description in the paper that went with the PR for this feature, I don't really understand why it's supposed to work. 

It's a simple bit of code, so I'll include it here for brevity:
  
```
z_true = z_true.sum(axis=0)
z_hat = z_hat.sum(axis=0)  # aggregate all individual sample realizations

z_hat_rho = np.quantile(z_hat, q=rho)  # get the quantile from aggregated samples

pred_above = np.where(z_hat_rho >= z_true, 1, 0)
pred_below = np.where(z_hat_rho < z_true, 1, 0)

rho_loss = 2 * (z_true - z_hat_rho) * (rho * pred_below - (1 - rho) * pred_above)
return rho_loss / z_true
```

Let's imagine an artificial time zeries `z_true`, which alternates between -1 and 1 on subsequent time steps. As such, it's `z_true.sum(axis=0)` is 0. Now let's imagine a `z_hat` which alternates from 1 and -1 on subsequent time steps (either 1 sample or you could sample from a gaussian of small width to make it stochastic; let's say you sample 10 samples from a gaussian of width 0.01 or something). I would expect `z_hat_rho` for the 0.5 quantile to be 0, and thus `z_true - z_hat_rho` would also be 0, and thus the rho_loss would be 0 which corresponds to a perfect score. However, for this hypothetical series where the predictions are out of phase with the targets by one time step (being -1 when the target is 1, and being 1 when the target is -1), the target is never even within the sampled distribution, so how could it get such a perfect score?

[...]

PR #1559, opened 2023-02-11
Add quantile loss metric

JanFidor:
Fixes [#1367 ](https://github.com/unit8co/darts/issues/1367)

### Summary
Add quantile loss metric with tests

I found an already existing implementation of quantile loss here: https://github.com/unit8co/darts/blob/e1c8d34b43f11d51e9b3fa34780ea8394fafb277/darts/utils/likelihood_models.py#L1121
and used it for refactoring. Substituting that implementation with the new metric function would definitely be in line with DRY, but for now it doesn't support CUDA, while the QuantileRegression does. I'm not certain whether it actually needs it or if it was just an implementation quirk, so I left it as is for now. I could add support for multiple tau values as a parameter and returning their mean loss.
### Other Information
While we're at the topic of the QuantileRegression , while browsing it I stumbled upon this line: https://github.com/unit8co/darts/blob/e1c8d34b43f11d51e9b3fa34780ea8394fafb277/darts/utils/likelihood_models.py#L1126
I'll be honest I've been looking at it for quite a while now and I just can't wrap my head around why it takes that sum before returning the mean. It would be penalizing the amount of quantiles the model optimizes for, which would kind of work as a regularization technique. Also if we have quantiles q and 1 - q we optimize for, they negate each other and give us 2 instances with quantile 0.5, so we could try to skip those unnecessary operations, I'm not certain if the speed-up would be meaningful though. Please let me know what you think


darts/metrics/__init__.py
```diff
@@ -14,6 +14,7 @@
     ope,
     r2_score,
     rho_risk,
+    quantile_loss,
```
madtoinou, line 17: Hi,

I think that this line is causing issues with darts linter verification. Can you try to follow the instructions found [here](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md) to install the dev tools so that the tests can pass.

darts/tests/metrics/test_metrics.py
```diff
@@ -4,6 +4,7 @@
 from darts import TimeSeries
 from darts.metrics import metrics
 from darts.tests.base_test_class import DartsBaseTestClass
+from ...metrics.metrics import quantile_loss
```
madtoinou, line 7: This should probably be imported from darts.metrics, or metrics directly since it's imported two lines above.

darts/metrics/__init__.py
JanFidor: Hi, thanks for the review, I've tried to install the pre-commit hooks for a while, but I either get 
```
RuntimeError: The Poetry configuration is invalid:
        - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$
```
error when running it or have to use `pre-commit autoupdate`  which changes pre-commit-config. I can't find anything new on github or SO, if it's something you dealt with before I'd really appreciate the help, if not no biggie and I'll just  try to remember to run black and isort manually before PRs

JanFidor:
In the last commit I used simple mean across all values instead of the previously used reduction param, because it wasn't passing
https://github.com/unit8co/darts/blob/e1c8d34b43f11d51e9b3fa34780ea8394fafb277/darts/tests/metrics/test_metrics.py#L112

darts/metrics/metrics.py
```diff
@@ -1212,3 +1212,81 @@ def rho_risk(
 
     rho_loss = 2 * (z_true - z_hat_rho) * (rho * pred_below - (1 - rho) * pred_above)
     return rho_loss / z_true
+
+
+# rho-risk (quantile risk)
```
madtoinou, line 1217: this comment should be removed, probably a leftover from the rho_risk copy-paste

madtoinou (approved):
LGTM, the reduction argument is actually supposed to be used when reducing the metrics for multivariate ts (thanks to the decorator) and should not be used inside the metric itself.

darts/metrics/metrics.py
JanFidor: Ahhh, you caught me red-handed I was absolutely certain I hid all the evidence of my little copy and paste shenanigan ;P

JanFidor:
> LGTM.
> 
> I don't have enough experience to comment the line you highlighted in [darts/darts/utils/likelihood_models.py](https://github.com/unit8co/darts/blob/e1c8d34b43f11d51e9b3fa34780ea8394fafb277/darts/utils/likelihood_models.py#L1126), I'll try to think more about it and circle back to you.

I mentioned it because I managed to stumble into it and it's somewhat connected to the PR. It's nothing crucial for now, so it might be a good idea to put together a separate issue for it to keep the PR small and focused on just the metric, let me know what you think

darts/metrics/metrics.py
```diff
@@ -1212,3 +1212,81 @@ def rho_risk(
...
+    *,
+    reduction: Callable[[np.ndarray], float] = np.mean,
+    inter_reduction: Callable[[np.ndarray], Union[float, np.ndarray]] = lambda x: x,
+    n_jobs: int = 1,
+    verbose: bool = False
+) -> float:
+    """
+    Given a time series of actual values :math:`y` of length :math:`T` and a time series of stochastic predictions
+    (containing N samples) :math:`y'` of shape :math:`T  x N`, quantile loss is a metric that quantifies the
+    accuracy of a specific quantile :math:`q` from the predicted value distribution.
```
hrzn, line 1234: You could call it \tau instead of q, as this is the name of the variable.

darts/metrics/metrics.py
```diff
@@ -1212,3 +1212,81 @@ def rho_risk(
...
+
+    Returns
+    -------
+    float
+        The quantile loss metric
+    """
+
+    raise_if_not(
+        pred_series.is_stochastic,
+        "quantile (quantile) loss should only be computed for stochastic predicted TimeSeries.",
```
hrzn, line 1271:
```suggestion
        "quantile (pinball) loss should only be computed for stochastic predicted TimeSeries.",
```

hrzn (commented):
LGTM, only small comments.

hrzn:
@JanFidor thanks for the PR!
Related to your comment about the other implementation in [likelihood_models.py](https://github.com/unit8co/darts/blob/master/darts/utils/likelihood_models.py). This other implementation is in PyTorch (and supporting CUDA) because it's directly used as a loss function for training the neural networks models (when the `likelihood` is set to `"quantile"`).
In the case of the metrics, it's better to keep a Numpy implementation only, as you did.

Regarding the operation
https://github.com/unit8co/darts/blob/e1c8d34b43f11d51e9b3fa34780ea8394fafb277/darts/utils/likelihood_models.py#L1126
I would have to dive into it a little more, but I think this is intended. We are first summing the losses over the last dimensions, so that the final loss for each sample is the sum of the losses for each quantile; note that when training the gradients will flow backward on this sum and be redistributed to the individual "quantile heads" of the neural network, in order to tune them to predict the individual quantiles correctly. The last `mean()` call is simply to take the mean over the batch dimension. So at first sight nothing there looks suspicious to me, but I might be missing something of course.

JanFidor:
Thanks for the explanation, I wasn't looking at the big picture with back-prop, it makes sense now

JanFidor:
Hmmm, I think macos failure might be unrelated to the code as ubuntu passes, please let me know if there's something I should change or if the CI's flaky

hrzn (approved):

hrzn (approved):
💯

JanFidor:
Thanks for the reviews @madtoinou @hrzn !

PR #1583, opened 2023-02-20
Fix NLinear normalization to support past covariates

Eliotdoesprogramming:
### Normalization description from [original paper](https://arxiv.org/pdf/2205.13504.pdf)

NLinear: To boost the performance of Linear when there is a distribution shift in the dataset, NLinear first subtracts the input by the last value of the sequence. Then, the input goes through a linear layer, and the subtracted part is added back before making the final prediction. The subtraction and addition in NLinear are a simple normalization for the input sequence.

### Summary

current implementation of normalization follows the implementation [here](https://github.com/cure-lab/LTSF-Linear/blob/main/models/NLinear.py)

this implementation works when the number of covariates being predicted as our target variable is the same as the number of covariates in our input (prev comment in implementation is incorrect, will work when n_params > 1 if n_params = target covariates)

since self.n_params == the amount of covariates we are predicting for AND we know that they are ordered first in our tensor
`input_tensor = [batch,timesteps, n_params+past_covariates]` we can slice the tensor to only include the covariates in our target tensor like so `last_seq[:,:,:n_params]`

### Other Information

New to doing open source work, please let me know if theres more that I need to do!! This was something I found when working on one of my own projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed an issue with `TorchForecastingModel.load_from_checkpoint()` not properly loading the loss function and metrics. [#1749](https://github.com/unit8co/darts/pull/1749) by [Antoine Madrona](https://github.com/madtoinou).
 - Fixed a bug when loading the weights of a `TorchForecastingModel` trained with encoders or a Likelihood. [#1744](https://github.com/unit8co/darts/pull/1744) by [Antoine Madrona](https://github.com/madtoinou).
 - Fixed a bug when using selected `target_components` with `ShapExplainer. [#1803](https://github.com/unit8co/darts/pull/#1803) by [Dennis Bader](https://github.com/dennisbader).
+- Fixed `TimeSeries.__getitem__()` for series with a RangeIndex with start != 0 and freq != 1. [#1868](https://github.com/unit8co/darts/pull/#1868) by [Dennis Bader](https://github.com/dennisbader).
 
 **Removed**
 - Removed support for Python 3.7 [#1864](https://github.com/unit8co/darts/pull/#1864) by [Dennis Bader](https://github.com/dennisbader).
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -867,7 +867,7 @@ class TimeSeriesTestCase(DartsBaseTestClass):
             # Cannot divide by 0.
             self.series1 / 0
 
-    def test_getitem(self):
+    def test_getitem_datetime_index(self):
         seriesA: TimeSeries = self.series1.drop_after(pd.Timestamp("20130105"))
         self.assertEqual(self.series1[pd.date_range("20130101", " 20130104")], seriesA)
         self.assertEqual(self.series1[:4], seriesA)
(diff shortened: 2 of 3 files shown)


madtoinou:
Hi,

The unittest are failing because of linting issue. Could you please follow the [contributing guidelines](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md) to install the pre-commit hooks and apply them to your PR?

Eliotdoesprogramming:
I linted manually with `./gradlew lint`. Using a M1 Macbook and having Poetry issues with `./gradlew test_all`

```txt
RuntimeError: The Poetry configuration is invalid:
            - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'
```

darts/models/forecasting/nlinear.py
```diff
@@ -129,7 +129,7 @@ def forward(
             x = self.layer(x)  # (batch, out_dim, out_len * nr_params)
 
             if self.normalize:
-                x = x + seq_last
+                x = x + seq_last[:, :, : self.nr_params]  # (batch, out_dim, nr_params)
```
dennisbader, line 132: Is this different from the current implementation?

From line 126 `seq_last` will have shape (batch, out_dim, 1).  So even if we take `seq_last[:, :, :self.nr_params]` here, it will still have shape (batch, out_dim,1).

Maybe I'm missing something?

dennisbader (commented):
Hi @Eliotdoesprogramming, thanks a lot for the PR and the patience.

I left a comment as I'm not sure that the implementation will change the current behavior.
What are your thoughts on that?

darts/models/forecasting/nlinear.py
Eliotdoesprogramming: Its specifically for the case when you have covariates. for example if you have 2 time series your predicting for and 3 covariates:

the linear layer will output (B , T, 2 ) but seq last will be of shape (B, T, 5) which will error because the C dim does not match in shape. This fix is specifically to account for that.

dennisbader: Could you have a look at line 126 which defines `seq_last` (see below)? It will always be of shape (B, T, 1) no matter what input we have:

```
seq_last = x[:, :, -1:].detach()  # (batch, out_dim, 1)
```

Eliotdoesprogramming: Your right for the case when self.shared_weights is true. It looks like when `self.shared_weights` is true **the code simply disregards the covariates**. 

line 122:
            ```# discard covariates, to ensure that in_dim == out_dim
            x = x[:, :, : self.output_dim]```

I didn't realize that the behavior was different for the shared weights. 


looking back, I misunderstood nr_params at the time (didnt realize it was for likelihood forecasts). I encountered the issue in the case when nr_params == 1

For when `self.shared_weights` is false, the on line 150 was the issue I encountered.

in this case, `seq_last` is of shape `Batch, 1 , in_dim` but the output from the linear layer `x` is of shape `Batch, Timesteps, out_dim * nr_params` 

I'm realizing now that in this case the correct behavior would be to reshape `x` to `(Batch, Timestep, n_params, out_dim` and seq_last as `(Batch, 1, 1, :out_dim)` and do subtraction that way.

If the correct behavior for when `shared_weights` is false is to disregard past covariates, I'd be happy to update my PR to disallow past covariates, as well as update the `readme.md` to reflect that.

sfc-gh-jhu (commented):
I got the same issues and I used the same fix. So I would say this would have the ability to do normalization when either past of future covariates are present while weights are not shared, and the forecast target output dim is usually smaller than the input dim.

darts/models/forecasting/nlinear.py
felixdivo: In #1850, I implemented it such that the normalization is only applied to the targets, which is a problem if `shared_weights  == False`. It is not the same problem, but connected.

felixdivo:
@sfc-gh-jhu

> I got the same issues and I used the same fix. So I would say this would have the ability to do normalization when either past of future covariates are present while weights are not shared, and the forecast target output dim is usually smaller than the input dim.

I did that in #1850.

felixdivo:
#1850 was closed to deduplicate efforts and put focus here. Since these issues prohibit the use of NLinear and DLinear in many scenarios, is anyone willing to wrap this up after a few stale months? 🙂

Eliotdoesprogramming:
Sorry originally thought I had closed this one due to some confusion with the necessity of my changes. I'll see if I can put in some work in this week on providing a more complete fix.

felixdivo:
That would be lovely @Eliotdoesprogramming!

I also think that there are problems with static covariates, but let's do that in a aseparate effort (I'll write an issue with a reproducer for it once this issue here solved, I'm not 100% sure whether there even is a problem).

PR #1584, opened 2023-02-21
Fix/Allow TimeSeries.with_hierarchy to accept strings as parents

### LINKED ISSUE #1581: [BUG] Unhelpful error message on TimeSeries.with_hierarchy
**Describe the bug**
When encoding the hierarchy the time series hierarchy we need to construct a hierarchy dict of the format `{child_node: [parent_node1, parent_node2, ...]}`. 
However, if there is only one parent node in the whole TimeSeries it is easy to make the mistake of constructing a dict `{child_node: parent_node}` instead of `{child_node: [parent_node}`. 
Doing so leads to the following rather unhelpful error message:
`ValueError: The values of the hierarchy must only contain component names matching those of the series.`

**To Reproduce**
```python
dates = pd.date_range("2020-01-01", "2020-12-31", freq="D")
nr_dates = len(dates)
t1 = TimeSeries.from_times_and_values(dates, 3 * np.ones(nr_dates), columns=["T1"])
t2 = TimeSeries.from_times_and_values(dates, 5 * np.ones(nr_dates), columns=["T2"])
t3 = TimeSeries.from_times_and_values(dates, np.ones(nr_dates), columns=["T3"])
tsum = TimeSeries.from_times_and_values(dates, 9 * np.ones(nr_dates), columns=["T_sum"])

ts = concatenate([t1, t2, t3, tsum], axis="component")
hierarchy = {"T1": "T_sum", "T2": "T_sum", "T3": "T_sum"}
ts = ts.with_hierarchy(hierarchy)
```

**Expected behavior**
Either it should simply detect that one component was specified rather than a list and internally convert it to a list, or a helpful error message should be issued

 - Python version: 3.10.6
 - darts version: master

**Additional context**
I think there's two possibilities to solve this
- Detect strings and convert them to a list of one string
- Raise an error message that instructs the user to use `["string"]` instead of `"string"` when constructing the hierarchy dict

I'm happy to submit a PR if you let me know which of these options is preferred

DavidKleindienst:
Fixes #1581.

### Summary

This PR allows TimeSeries to accept hierarchy of the format `{child_component: parent_component}`, removing the need to specify parent components as list if there is only one parent

### Other Information

The check for string is implemented as `isinstance(parent, string) and parent in ts.components` for backwards compatibility:
Currently, it is possible to use single letters as component names (e.g. "a", "b", "c") and specify `{"a": "bc"}`  to mean "a" is a child of  "b" and "c". This behavior is kept unless "bc" is also a component. 
Let me know if this makes sense, or it's better to remove this check and interpret "bc" always as ["bc"] even if no "bc" component exists.


dennisbader (changes requested):
Thank you @DavidKleindienst for this PR, and sorry for the long wait!

Looks good already, just had some minor suggestions to avoid mutating the user's hierarchy object:)

DavidKleindienst:
Thanks for reviewing @dennisbader 
I've added the changes you suggested

dennisbader (approved):

dennisbader:
Thanks! Ready to merge 🚀

Issue #1580, opened 2023-02-20
Allow for usage of native way of dealing with categorical features RegressionModels


rijkvandermeulen:
**Summary**
@hrzn in line with the earlier discussion we had on gitter (see link below), I'd like to propose the following functionality: for RegressionModels that have a native way of dealing with categorical features (especially gradient boosted trees models such as LightGBM, Catboost) enable the usage of this parameter throught the Darts API. Ideally this would work for all kinds of covariates (past, fut and static).

**Link to gitter discussion for more detailed explanation**
https://matrix.to/#/!uumevxjBaNJhovFYgj:gitter.im/$w1tobiqP8nxHN1hQmRKJU9_inojDwfZkFHSSt4B0tlY?via=gitter.im&via=matrix.org&via=matrix.thegolem.cz

Do you still see the added value of this? If so, I can also give this a go.

hrzn:
Hi @rijkvandermeulen , we definitely still see a lot of value in this :) Many thanks for the PR! We'll take a look asap.

PR #1585, opened 2023-02-21
Functionality to let LightGBM effectively handle categorical features

rijkvandermeulen:
Fixes https://github.com/unit8co/darts/issues/1580.

### Summary
- This PR allows the user to specify (when fitting the model) which of the covariates should be treated as categorical features by the underlying `LightGBMRegressor`. Supports categorical past-, fut-, and static covariates.
- A similar approach can be taken for the other RegressionModels that use gradient boosting on decision trees (XGBoost and CatBoost). They would require a bit more effort though. CatBoost, for example, does not allow passing floating point numbers as categorical features. Hence, we cannot simply pass X as an ndarray as we currently do. We would have to have a step in between to preprocess the training_samples in a format that the model accepts. Would you like to this as part of this PR or create a separate ticket?

### Why have this?
For categorical features (especially with high cardinality), the LightGBM native way of handling categorical features works better than using one-hot (or any other type of) encoding. In my experience, this can make quite a big impact in practice --> better performance and faster training. For more info see: https://lightgbm.readthedocs.io/en/latest/Features.html#optimal-split-for-categorical-features

### Practicalities

- This PR is still **WIP**. Functionality wise it seems to work, but it requires a bit of cleaning up, documenting, more extensive unit tests etc before we merge. Before going into this, I wanted to first check whether you guys agree  with the approach or maybe have some other ideas?
- An alternative to passing the categorical_covariates when fitting the model (as is currently implemented) we could also make these attributes of the `TimeSeries `class (e.g., something like `categorical_components` and `categorical_static_covariates`). In fact, I played around with this a bit, but I feel like it would be a less robust and user-friendly solution.

@hrzn @dennisbader could you have a look and share your thoughts? :)


darts/models/forecasting/lgbm.py
```diff
@@ -152,6 +155,15 @@ def fit(
             Optionally, a series or sequence of series specifying past-observed covariates for evaluation dataset
         val_future_covariates : Union[TimeSeries, Sequence[TimeSeries]]
             Optionally, a series or sequence of series specifying future-known covariates for evaluation dataset
+        categorical_past_covariates
```
dennisbader, line 158: Can we move this to model construction (`__init__`)?
In `fit()` this will add more complexity and we would have different APIs for the models that support categorical and the ones that do not.
- in general we aim at making the API as unified as possible (we already have some complexity with validation series and others)

With the categorical variable names already known at fitting time, we can extract them from the input (`series`, `past_covariates`, `future_covariates`) on the fly

darts/models/forecasting/lgbm.py
```diff
@@ -163,6 +175,53 @@ def fit(
             Additional kwargs passed to `lightgbm.LGBRegressor.fit()`
         """
 
+        if categorical_past_covariates:
```
dennisbader, line 178: There is a bit of code duplication here, this could be avaoided by iterating through the three covariate types/names in a zipped fashion

darts/models/forecasting/lgbm.py
```diff
@@ -195,11 +257,66 @@ def fit(
...
+
+        training_samples, training_labels = self._create_lagged_data(
+            target_series,
+            past_covariates,
+            future_covariates,
+            max_samples_per_ts,
+        )
+
+        # Check if there are covariates that should be treated as categorical
+        categorical_past_covariates = kwargs.pop("categorical_past_covariates", None)
```
dennisbader, line 289: Assume we move the categorical variable definition to model creation. We can already process the values and store the `categorical_covariates` in `self`.

Then we have access to these variables already and we can simply call your `_get_categorical_features()` on `series`, `past_covariates`, and `future_covariates.

darts/models/forecasting/regression_model.py
```diff
@@ -812,6 +812,61 @@ def _predict_and_sample(
...
+                for component in past_covs_ts.components
+            ]
+            + [
+                f"fut_cov_{component}_{lag}"
+                for lag in self.lags.get("future", [])
+                for component in fut_covs_ts.components
+            ]
+            + (
+                list(target_ts.static_covariates.columns)
+                if isinstance(target_ts.static_covariates, pd.DataFrame)
```
dennisbader, line 855:
```suggestion
                if target_ts.has_static_covariates
```

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -2093,6 +2094,201 @@ def test_catboost_model_with_eval_set(self, lgb_fit_patch):
         assert lgb_fit_patch.call_args[1]["eval_set"] is not None
         assert lgb_fit_patch.call_args[1]["early_stopping_rounds"] == 2
 
+    def test_fit_predict_with_categorical_features(self):
```
dennisbader, line 2097: Could we maybe expand this test a bit so that we can test that the categorical variables actually improve the forecast?

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -2093,6 +2094,201 @@ def test_catboost_model_with_eval_set(self, lgb_fit_patch):
...
+        """Test a RegressionModel with categorical features"""
+        target_df = pd.DataFrame(
```
dennisbader, line 2099: there is a bit of code duplication for target_df, past_cov_df, and their time series creation throughout the tests.
Can we try reduce this?

We can for example save the time series in the Test Class itself and then just access them in the tests.

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -2093,6 +2094,201 @@ def test_catboost_model_with_eval_set(self, lgb_fit_patch):
...
+                ),
+                past_covariates=TimeSeries.from_dataframe(
+                    df=past_cov_df,
+                    time_col="date",
+                    value_cols=["past_cov_cat_1", "past_cov_cat_2", "past_cov_regular"],
+                ),
+                categorical_past_covariates=["past_cov_cat_1", "past_cov_cat_2"],
+                categorical_static_covariates=["static_cat_1", "static_cat_2"],
+            )
+            model.predict(1)
```
dennisbader, line 2132: Can we somehow test whether the LightGBM model actually received and used these categorical variables?

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -2093,6 +2094,201 @@ def test_catboost_model_with_eval_set(self, lgb_fit_patch):
...
+
+    def test_error_raised_invalid_categorical_past_cov(self):
```
dennisbader, line 2134: The general logic of the tests are almost identical. Can we try to write one helper test case carrying this logic, and then call this from `test_error_raised_invalid_categorical_past_cov`, `test_error_raised_invalid_categorical_static_cov`, (and also we need to test for future covs)

dennisbader (changes requested):
Hey @rijkvandermeulen and thanks for giving this a go! :) Looks like a good start already 🚀 

I haven't fully reviewed, but want to address some points already:
- I would opt for moving the categorical variable definition out of `fit()` and into the model constructor (`__init__()`).
- there is a bit of code duplication which we could easily reduce, I added comments where I see this

darts/tests/models/forecasting/test_regression_models.py
rijkvandermeulen: I've now placed all validation checks in a single unit test (`test_fit_with_categorical_features_raises_error`)

darts/tests/models/forecasting/test_regression_models.py
rijkvandermeulen: The LightGBMRegressor doesn't seem to keep track of categorical_features it used in one of its attributes. Hence, best I could think of was mocking it's fit call and check that the value of the categorical_feature argument is in line with what we expect. WDYT? (see `test_lgbm_categorical_features_passed_to_fit_correctly`)

darts/tests/models/forecasting/test_regression_models.py
rijkvandermeulen: Good point. I've created the property `inputs_for_tests_categorical_covariates` for this. Does that make sense to you?

darts/tests/models/forecasting/test_regression_models.py
rijkvandermeulen: Yes! See `test_quality_forecast_with_categorical_covariates`

darts/models/forecasting/lgbm.py
rijkvandermeulen: Gave this a go for past and future covs. For static covs the check is slightly different so I'm handling that one still separately. Is this what you had in mind?

darts/models/forecasting/lgbm.py
rijkvandermeulen: Makes sense. Done.

rijkvandermeulen:
@dennisbader  Thanks for the review and the valuable suggestions! I think I've covered all of your remarks; could you do a second round of review? :)

rijkvandermeulen:
BTW, pipeline seems to be broken by an unit test (`test_stationarity_tests`) unrelated to this PR.

dennisbader:
> @dennisbader Thanks for the review and the valuable suggestions! I think I've covered all of your remarks; could you do a second round of review? :)

Hi @rijkvandermeulen and thanks for updates! I'll review next week.

dennisbader (changes requested):
Thank you @rijkvandermeulen for the updates, looks really good! 🚀 

We discussed the categorical covariates support internally, and think it would be nice to add a new class inheriting from `RegressionModel` which implements and governs the categorical support.

With this we could easily add CatBoost (including casting float to int in the model maybe) and XGBoost (as soon fully supporting categoricals).

Thanks as well for the unit tests 👍  I proposed using a different test for testing the model performance to keep it more light-weight (I added the code in the comments). The proposal relies on one of our existing examples, see [section 6 in this example](https://github.com/unit8co/darts/blob/master/examples/15-static-covariates.ipynb). This one only checks if categorical static covariates improve the predictions. In theory, if it works for static covariates, it should translate well to future/past covariates.

Let me know if you have any questions about adding the new Class-based categorical support approach or others, I'll glady help.

rijkvandermeulen:
Hey @dennisbader,

Great suggestion of creating a new "base" class, thanks! I gave this a go - looking forward to hearing what you think :)

Some points for discussion still might be:

- In order to make this work for CatBoost we have to cast categorical covs from float to int. Maybe I'm missing something, but I don't think this is very trivial with numpy (guess we would have to work with strucutred arrays or the like). So for now I took the approach of converting X to a pandas df and do the conversion there. It works, but I'm not sure whether it's the best approach. Any better ideas?
- Related question; I'm not sure whether the current `RegressionModelWithCatecoricalCovariates` in its current state is generic enough. For example, I'm not using XGBoost much, but judging from their API reference docs it seems that we would have to cast the categorical covs to `categorical` type. How do you see this? Maybe make a default implementation of `_cast_float_to_int`  (probably not the best naming anymore then ;)) in `RegressionModelWithCatecoricalCovariates` and have this method potentially overwritten by each subclass (e.g., XGBoost once we implement this later on once stable)?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ## [Unreleased](https://github.com/unit8co/darts/tree/master)
 [Full Changelog](https://github.com/unit8co/darts/compare/0.23.1...master)
+- `LightGBM` model now supports native categorical feature handling as described
```
rijkvandermeulen, line 9: TODO; add CatBoost to changelog entry

darts/models/forecasting/regression_model.py
```diff
@@ -1032,3 +1032,358 @@ def __init__(self):
 
     def __str__(self):
         return f"_QuantileModelContainer(quantiles={list(self.keys())})"
+
+
+class RegressionModelWithCategoricalCovariates(RegressionModel):
```
rijkvandermeulen, line 1037: What's the best place for this new class? Within regression_model.py (as it is now) or in the models/forecasting directory?

rijkvandermeulen:
Hey @dennisbader, I was wondering when you'd expect to have the time to take another look at the latest iteration of this PR? Thank you!

dennisbader:
Ho @rijkvandermeulen, and sorry for the long waiting. From mid April on I have dedicated time for Darts, which will make reviewing much quicker :). 

I'll review the changes this weekend. Plan is to release the new Darts version around the end of March and we definitely want to include this one!

darts/models/forecasting/regression_model.py
dennisbader: looks good in this file 👍

darts/models/forecasting/regression_model.py
```diff
@@ -1032,3 +1032,358 @@ def __init__(self):
...
+                else []
+            )
+        )
+
+        if not categorical_covariates:
+            return [], []
+        else:
+            target_ts = series if isinstance(series, TimeSeries) else series[0]
+            past_covs_ts = past_covariates[0] if past_covariates else None
+            fut_covs_ts = future_covariates[0] if future_covariates else None
```
dennisbader, line 1291: this will actually create a new timeseries with a single time step if past/future covariates are a single TimeSeries. 
Could you add the following function to darts/utils/utils.py.
We can use this function in the future as well.

```
def get_single_series(
    ts: Optional[Union[TimeSeries, Sequence[TimeSeries]]]
) -> Optional[TimeSeries]:
    """Returns a single (first) TimeSeries or `None` from `ts`. Returns `ts` if  `ts` is a TimeSeries, `ts[0]` if
    `ts` is a Sequence of TimeSeries. Otherwise, returns `None`.

    Parameters
    ----------
    ts
        None, a single TimeSeries, or a sequence of TimeSeries.

    Returns
    -------
        `ts` if  `ts` is a TimeSeries, `ts[0]` if `ts` is a Sequence of TimeSeries. Otherwise, returns `None`

    """
    if isinstance(ts, TimeSeries) or ts is None:
        return ts
    else:
        return ts[0]
```

```suggestion
            target_ts = get_single_series(series)
            past_covs_ts = get_single_series(past_covariates)
            fut_covs_ts = get_single_series(future_covariates)
```

darts/models/forecasting/regression_model.py
```diff
@@ -1032,3 +1032,358 @@ def __init__(self):
...
+        self._categorical_col_indices = cat_col_indices
+
+        training_samples_df = self._cast_float_to_int(training_samples, cat_col_indices)
+
+        kwargs[self.categorical_fit_param_name] = cat_col_indices
+        self.model.fit(
+            training_samples_df,
+            training_labels,
+            **kwargs,
+        )
```
dennisbader, line 1379: I think we should avoid for now having the conversion from numpy array to pandas DataFrame.
This can get quite slow for large arrays. 

Let's exclude Catboost categorical support for now in this PR until we come up with a robust solution. Thanks to your work, adding support at a later time will be very easy! 👍 🚀  :)

Also, now we can make use of the parent method again

```suggestion
        cat_col_indices, _ = self._get_categorical_features(
            target_series,
            past_covariates,
            future_covariates,
        )
        # Updated each time the model is fit such that the indices can be easily retrieved in the predict method later
        self._categorical_col_indices = cat_col_indices
        kwargs[self.categorical_fit_param_name] = cat_col_indices
        super()._fit_model(
            target_series=target_series,
            past_covariates=past_covariates,
            future_covariates=future_covariates,
            max_samples_per_ts=max_samples_per_ts,
            **kwargs
        )
```

darts/models/forecasting/regression_model.py
```diff
@@ -1032,3 +1032,358 @@ def __init__(self):
...
+
+    def _predict_and_sample(
```
dennisbader, line 1381: let's remove this for now as described in my other comment

darts/models/forecasting/regression_model.py
```diff
@@ -1032,3 +1032,358 @@ def __init__(self):
...
+                for i, col in enumerate(feature_list)
+                for cat in categorical_covariates
+                if cat and cat in col
+            ]
+            col_names = [feature_list[i] for i in indices]
+
+            return indices, col_names
+
+    @staticmethod
+    def _cast_float_to_int(data: np.ndarray, col_indices: List[int]) -> pd.DataFrame:
```
dennisbader, line 1329: let's remove this for now as described in my other comment

darts/models/forecasting/catboost_model.py
dennisbader, line 1: let's remove covariates support for Catboost for now as described in my other comment

dennisbader (changes requested):
Thanks again for all the updates @rijkvandermeulen 🚀 The new covariates class for RegressionModels looks great! We're very close to merge this one, and should be the last iteration.

Apart from some minor suggestion, I believe we should exclude categorical support for Catboost from this PR. Reason behind it is that we should avoid having to convert from numpy array to pandas DataFrame. This can get quite slow for large arrays. 

For now let's put this into our backlog until we come up with a robust solution (we're also thinking about how to improve categorical support in the TimeSeries itself). 

Thanks to your work, adding support at a later time will be very easy! 👍 🚀 :)

dennisbader (changes requested):
Hi @rijkvandermeulen, and thanks again for the updates.

Could you apply the suggestion that I mentioned last time (I added a comment again).
Also it seems there are some merge conflicts, could you resolve them?

Thanks a lot! 🚀

darts/models/forecasting/regression_model.py
```diff
@@ -1032,3 +1036,309 @@ def __init__(self):
...
+            past_covariates,
+            future_covariates,
+        )
+
+        kwargs[self._categorical_fit_param_name] = cat_col_indices
+        self.model.fit(
+            training_samples,
+            training_labels,
+            **kwargs,
+        )
```
dennisbader, line 1344:
```suggestion
        cat_col_indices, _ = self._get_categorical_features(
            target_series,
            past_covariates,
            future_covariates,
        )
        # Updated each time the model is fit such that the indices can be easily retrieved in the predict method later
        self._categorical_col_indices = cat_col_indices
        kwargs[self.categorical_fit_param_name] = cat_col_indices
        super()._fit_model(
            target_series=target_series,
            past_covariates=past_covariates,
            future_covariates=future_covariates,
            max_samples_per_ts=max_samples_per_ts,
            **kwargs
        )
```

rijkvandermeulen:
@dennisbader apologies; must have missed that comment earlier. I've ommitted your suggestion and also resolved the merge conflicts.

darts/models/forecasting/regression_model.py
```diff
@@ -1029,3 +1033,311 @@ def _quantile_sampling(self, model_output: np.ndarray) -> np.ndarray:
...
+        training_samples, training_labels = self._create_lagged_data(
+            target_series,
+            past_covariates,
+            future_covariates,
+            max_samples_per_ts,
+        )
+
+        # if training_labels is of shape (n_samples, 1) flatten it to shape (n_samples,)
+        if len(training_labels.shape) == 2 and training_labels.shape[1] == 1:
+            training_labels = training_labels.ravel()
```
dennisbader, line 1328: Sorry, the whole thing here can be removed. Otherwise we create the lagged data twice. 
Let me apply this directly.

```suggestion
```

darts/models/forecasting/regression_model.py
```diff
@@ -1029,3 +1033,301 @@ def _quantile_sampling(self, model_output: np.ndarray) -> np.ndarray:
...
+        future_covariates,
+        max_samples_per_ts,
+        **kwargs,
+    ):
+        """
+        Custom fit function for `RegressionModelWithCategoricalCovariates` models, adding logic to let the model
+        handle categorical features directly.
+        """
+
+
```
dennisbader, line 1319:
```suggestion
```

dennisbader, line 1319:
```suggestion
```

dennisbader (approved):
LGTM! Thanks a lot @rijkvandermeulen 🚀

Issue #1586, opened 2023-02-22
Refactor forecasting model files (`__str__()` methods, and assertions)


JanFidor:
Original discussion with @hrzn in https://github.com/unit8co/darts/pull/1557#discussion_r1108282358

Refactor would include:
1.  Using a similar __str__() template for models
3. Using raise_if_not() in plase of assert

PR #1590, opened 2023-02-22
Refactor forecasting models

JanFidor:
Fixes https://github.com/unit8co/darts/issues/1586

### Summary
For now I only pushed changes to baseline models

@hrzn Potential changes I wanted to get your feedback on before pushing:
1.  Refactor all `__str__` for all models to be their class name with meaningful params captured in parenthesis  
ex.ample: ` "Auto-ARIMA-Statsforecasts"` -> `"StatsForecastAutoARIMA"`
2. Add logger to every `raise_if` and `raise_if_not` call which don't yet have it
3. Use fstring formatting wherever possible


dennisbader:
Actually, it could be super nice if we updated the representation of our models (in the `__repr__` method) to something similar to how sklearn does it.

They represent models like this:
- default model without changing hparams:
  ```
  from sklearn.linear_models import LinearRegression
  model = LinearRegression()
  print(model)
  
  >>> LinearRegression()
  ```
- model with changed hyperparameters:
  ```
  from sklearn.linear_models import LinearRegression
  model = LinearRegression(fit_intercept=False)
  print(model)
  
  >>> LinearRegression(fit_intercept=False)
  ```

We already save the model parameters that the user manually sets at model creation. Accessible after model creation with:
```
model.model_params
```

What do you think? Would you be willing to give this a go?

JanFidor:
Sure thing @dennisbader, btw while working on the other PR I noticed [this line](https://github.com/unit8co/darts/blob/c80049b0d8916ccc1de5324410ed01a21701e31a/darts/models/forecasting/forecasting_model.py#L755) which imho looks like a raise_if() in disguise, should I convert it (and other similar asserts if I find them) to raise_if() / raise_if_not()?

JanFidor:
@dennisbader what do you think about this implementation? If it's okay, I'll start deleting __str__ overrides in children models

dennisbader:
> Sure thing @dennisbader, btw while working on the other PR I noticed [this line](https://github.com/unit8co/darts/blob/c80049b0d8916ccc1de5324410ed01a21701e31a/darts/models/forecasting/forecasting_model.py#L755) which imho looks like a raise_if() in disguise, should I convert it (and other similar asserts if I find them) to raise_if() / raise_if_not()?

This is okay, raise_log is sometimes better to use, especially in for loops. 
When we use formatted strings (f"something {val}") in for loops with raise_if or raisa_if_not, it can slow things down as the string needs to be computed every time.
With raise_log, the string is only evaluated if the condition is met.

darts/models/forecasting/forecasting_model.py
```diff
@@ -2220,3 +2220,40 @@ def _supress_generate_predict_encoding(self) -> bool:
     @property
     def extreme_lags(self):
         return (-1, 1, None, 0, 0)
+
+    def __str__(self):
```
dennisbader, line 2224: 👍 Very neat, thanks!

Could you rename the method into `__repr__` and move it to the `ForecastingModel` base class?

dennisbader (changes requested):
Very neat, and looks great! Thanks

Had only a minor comment about renaming the method and moving it to the correct place.

You can go ahead and remove the model specific __str__ methods!

darts/models/forecasting/forecasting_model.py
hrzn: +1 for `__repr__`, however, it could also be beneficial to have `__str__` setup correctly as well I guess (the two are called in different circumstances). WDYT?

hrzn (commented):
Looks good!

darts/models/forecasting/forecasting_model.py
JanFidor: I never really used `__repr__` but if I understood it's point correctly (pretty much just a more informative __str__), then my intuition would be to keep `__str__` as is and for `__repr__` provide all the key_value pairs of parameters including the defaults. This way users can both get simple model labels for graph legends when comparing hyper-parameters and get the entire set of model's parameters without having to delve into documentation to check their default values. Would that make sense?

dennisbader: From what I understand the intention behind the magic methods is:

"""
The `__str__()` method is intended to return information in a readable format for humans to easily understand it. It is perfect for logging or displaying object information. While the `__repr__()` method's goal is to deliver a string representation of the object that may be used to replicate it, it is also designed to do the opposite. 
([source](https://www.educative.io/answers/what-is-the-difference-between-str-and-repr))
"""

So in our case, the `__str__` implementation in this PR can actually be used to create a new model object (it's doing what the `__repr__` is intended to be used for). 

From another [source](https://www.digitalocean.com/community/tutorials/python-str-repr-functions#what-s-the-difference-between-__str__-and-__repr__):
"""
If you don’t define a `__str__()` method for a class, **then the built-in object implementation calls the `__repr__()` method instead.**
"""

So I guess we would only have to define `__repr__` and not `__str__`. It is perfectly fine though like you did it, I was just curious myself what the best practices were for these two magic methods :)

dennisbader (approved):
Looks great, thanks a lot for this PR and the patience @JanFidor! 🚀

JanFidor:
Hi @dennisbader , again thanks for the review and help, I'm just happy to make a little impact when it comes to ML : )

PR #1594, opened 2023-02-24
fix: replacing lambda with named function to make model pickable

### LINKED ISSUE #1587: [BUG] Calling save() after fit() throw pickle error for FFT models
**Describe the bug**
save() method does not work after fit() method in some cases.
```AttributeError: Can't pickle local object 'FFT.fit.<locals>.<lambda>'```

**To Reproduce**

```from darts.models import FFT
from darts.datasets import EnergyDataset
from darts.dataprocessing.transformers import Scaler, MissingValuesFiller



df = EnergyDataset().load().pd_dataframe()
df_day_avg = df.groupby(df.index.astype(str).str.split(" ").str[0]).mean().reset_index()
filler = MissingValuesFiller()
scaler = Scaler()
series = scaler.fit_transform(
    filler.transform(
        TimeSeries.from_dataframe(
            df_day_avg, "time", ["generation hydro run-of-river and poundage"]
        )
    )
).astype(np.float32)



train, val = series.split_after(pd.Timestamp("20170901"))


model = FFT(nr_freqs_to_keep=2, trend='linear', trend_poly_degree=2)
model.fit(train)


model.save('test.pht')
```

**Expected behavior**

The model should be saved properly as test.pht

 - Python version: 3.7.13
 - darts version 0.23.1

**Additional context**

In another code a co-worker got a similar error : 
```AttributeError: Can't pickle local object 'EvaluationLoop.advance.<locals>.batch_to_device'```

It was trying model.save() method after fitting a TFTModel. I couldn't reproduce the error on my machine though ...

### PR DESCRIPTION
**Pull request by madtoinou**
Fixes #1587.

### Summary

`save()` was not working on FTT model if the `trend` argument was neither "poly" or "exp" because the function `lambda x : 0` is not pickable. Just created a class method to have a named function that can be pickled and updated the docstring.

### Other Information

I'll make sure all the other models `save()` method are working before converting this draft PR into a real one.

### DIFF
--- a/darts/models/forecasting/fft.py
+++ b/darts/models/forecasting/fft.py
@@ -269,17 +269,17 @@ class FFT(LocalForecastingModel):
             + ")"
         )
 
-    def _exp_trend(self, x):
+    def _exp_trend(self, x) -> Callable:
         """Helper function, used to make FFT model pickable."""
         return np.exp(self.trend_coefficients[1]) * np.exp(
             self.trend_coefficients[0] * x
         )
 
-    def _poly_trend(self, trend_coefficients):
+    def _poly_trend(self, trend_coefficients) -> Callable:
         """Helper function, for consistency with the other trends"""
         return np.poly1d(trend_coefficients)
 
-    def _null_trend(self, x):
+    def _null_trend(self, x) -> Callable:
         """Helper function, used to make FFT model pickable."""
         return 0
 
--- a/darts/models/forecasting/fft.py
+++ b/darts/models/forecasting/fft.py
@@ -238,7 +238,7 @@ class FFT(LocalForecastingModel):
             pd.Timestamp attributes that are relevant for the seasonality automatically.
         trend
             If set, indicates what kind of detrending will be applied before performing DFT.
-            Possible values: 'poly' or 'exp', for polynomial trend, or exponential trend, respectively.
+            Possible values: 'poly', 'exp' or None, for polynomial trend, exponential trend or no trend, respectively.
         trend_poly_degree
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by madtoinou** [comment]
The pickle issue was also occurring for `trend=exp` since it was also relying on a `lambda` function. Created helper function for all 3 cases. I had to store the `trend_coefficients` in the model itself in order to have a clean exponential detrending function. Let me know if there is a better way to implement such functionality.

File: darts/models/forecasting/fft.py [hunk 1]
```diff
@@ -269,6 +269,20 @@ def __str__(self):
...
+        """Helper function, used to make FFT model pickable."""
+        return np.exp(self.trend_coefficients[1]) * np.exp(
+            self.trend_coefficients[0] * x
+        )
+
+    def _poly_trend(self, trend_coefficients):
+        """Helper function, for consistency with the other trends"""
+        return np.poly1d(trend_coefficients)
+
+    def _null_trend(self, x):
```
dennisbader, line 282:
```
from typing import Callable
```
```suggestion
    def _exp_trend(self, x) -> Callable:
        """Helper function, used to make FFT model pickable."""
        return np.exp(self.trend_coefficients[1]) * np.exp(
            self.trend_coefficients[0] * x
        )

    def _poly_trend(self, trend_coefficients) -> Callable:
        """Helper function, for consistency with the other trends"""
        return np.poly1d(trend_coefficients)

    def _null_trend(self, x) -> Callable:
```

dennisbader (approved):
Looks good, thanks a lot! 
Just added the `Callable` for better interpretability

Issue #1283, opened 2022-10-11
Merge historical_forecast and backtests


dumjax:
We can get everything with almost 0 costs. I often need both and run twice the logic of historical_forecast (which is used in backtest) is not optimal.

JanFidor:
@dumjax  @hrzn  Could I pick up this issue?  I took a look at the code and I think that the best way to implement it would be to
add a new parameter to backtesting() for already exisiting historical_forecasts() with a default value None which would result in calculating them from scratch

hrzn:
@JanFidor I'd say yes, and your suggested approach sounds reasonable to me. Would be good to also run it by @dennisbader to get his opinion.

JanFidor:
Okay, it's rather straightforward, so I should manage to push a basic draft PR tomorrow before fully committing with tests and the other bells and whistles

PR #1597, opened 2023-02-24
add already existing forecasts param

JanFidor:
Fixes https://github.com/unit8co/darts/issues/1283

### Summary
For now it's just a rough implementation, once it's green lit by @dennisbader i'll add some tests


darts/models/forecasting/forecasting_model.py
```diff
@@ -907,6 +907,7 @@ def historical_forecasts(
     def backtest(
         self,
         series: Union[TimeSeries, Sequence[TimeSeries]],
+        historical_forecasts: Union[TimeSeries, Sequence[TimeSeries]] = None,
```
dennisbader, line 910: Could you add the parameter definition to the docstring?

dennisbader (changes requested):
Thanks @JanFidor for this!

Just one comment about the missing `historical_forecasts` defnition in the docs.
After that it's ready to be merged!

darts/models/forecasting/forecasting_model.py
JanFidor: On it

JanFidor:
Hi @dennisbader, I have 3 quick questions about the PR:
1) Should I add 1 or 2 tests  for the param in test_backesting.py?
2) Right now the forecasts param is second which imho makes sense when it comes to readability, but would be a breaking change when not using named params. Do you think it's worth it to stay with the more readable implementation?
3) I was thinking about potential edge cases while browsing the file and noticed [this line](https://github.com/unit8co/darts/blob/c80049b0d8916ccc1de5324410ed01a21701e31a/darts/models/forecasting/forecasting_model.py#L755) should I do something similar, but maybe instead of throwing an exception, log a warning and fallback on calling historical_forecasts() ?

JanFidor:
@dennisbader I added the docstring, so everything should be ready for merging, if you think any of my proposed changes from the previous comment would be useful to the PR just let me know and I'll get on them!

darts/models/forecasting/forecasting_model.py
```diff
@@ -950,6 +951,8 @@ def backtest(
         ----------
         series
             The (or a sequence of) target time series to use to successively train and evaluate the historical forecasts
+        historical_forecasts
+            Optionally, the (or a sequence of) historical forecasts time series to be evaluated.
```
dennisbader, line 955:
```suggestion
        historical_forecasts
            Optionally, the (or a sequence of) historical forecasts time series to be evaluated. Corresponds to 
            the output of :meth:`historical_forecasts() <ForecastingModel.historical_forecasts>`. If provided, will 
            skip historical forecasting and ignore parameters `num_samples`, `train_length`, `start`, 
            `forecast_horizon`, `stride`, `retrain`, `overlap_end`, and `last_points_only`.
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -950,6 +951,8 @@ def backtest(
         ----------
         series
             The (or a sequence of) target time series to use to successively train and evaluate the historical forecasts
+        historical_forecasts
```
dennisbader, line 954: The method description above should be updated as well. Most of what it describes is only applied if historical_forecasts are not provided. Could you make the changes?

dennisbader (changes requested):
Hey @JanFidor , thanks again for another nice PR :)

Had only some minor comments.

A small unittest would be nice (maybe to existing backtest tests) to check wether we get identical results with and without provided `historical_forecasts`.

darts/models/forecasting/forecasting_model.py
```diff
@@ -942,12 +943,12 @@ def backtest(
...
-        the end of the
-        training set forward by `stride` time steps. A metric (given by the `metric` function) is then evaluated
-        on the forecast and the actual values. Finally, the method returns a `reduction` (the mean by default)
-        of all these metric scores.
+        If `historical_forecasts` are provided, it uses them to calculate loss for the actual values otherwise
+        it repeatedly builds a training set from the beginning of `series`.
+        It trains the current model on the training set, emits a forecast of length equal to forecast_horizon,
+        and then moves the end of the training set forward by `stride` time steps.
+        A metric (given by the `metric` function) is then evaluated on the forecast and the actual values.
+        Finally, the method returns a `reduction` (the mean by default) of all these metric scores.
```
dennisbader, line 951:
```suggestion
        If `historical_forecasts` are provided, the metric (given by the `metric` function) is evaluated directly on 
        the forecast and the actual values. Otherwise, it repeatedly builds a training set from the beginning of 
        `series`. It trains the current model on the training set, emits a forecast of length equal to 
        `forecast_horizon`, and then moves the end of the training set forward by `stride` time steps. The metric is 
        then evaluated on the forecast and the actual values. Finally, the method returns a `reduction` (the mean by 
        default) of all these metric scores.
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -1035,21 +1041,22 @@ def backtest(
...
+                num_samples=num_samples,
+                train_length=train_length,
+                start=start,
+                forecast_horizon=forecast_horizon,
+                stride=stride,
+                retrain=retrain,
+                overlap_end=overlap_end,
+                last_points_only=last_points_only,
+                verbose=verbose,
+            )
```
dennisbader, line 1059:
```suggestion
        if historical_forecasts is None:
            forecasts = self.historical_forecasts(
                series=series,
                past_covariates=past_covariates,
                future_covariates=future_covariates,
                num_samples=num_samples,
                train_length=train_length,
                start=start,
                forecast_horizon=forecast_horizon,
                stride=stride,
                retrain=retrain,
                overlap_end=overlap_end,
                last_points_only=last_points_only,
                verbose=verbose,
            )
        else:
            forecasts = historical_forecasts
```

dennisbader (changes requested):
Hi @JanFidor and thanks for the updates and patience!

Only had minor suggestions mainly for rephrasing the docs.
After that we can merge this one  🚀

JanFidor:
Hi @dennisbader , thanks for the reviews and help with this PR! Special thanks for the suggested changes to the docs, writing them is still not my strong suit and something I'll have to keep working on ; )

dennisbader (approved):
LGTM, thanks @JanFidor 🚀

PR #1602, opened 2023-02-28
Fix/NaiveEnsembleModel prediction component name

### LINKED ISSUE #1601: [BUG] NaiveEnsembleModel predictions have wrong component names
**Describe the bug**
Predictions should have the same component names as the TimeSeries used to fit the model.
For `NaiveEnsembleModel.predict` that's not the case - The component of the returned `TimeSeries` is always called `"0"`

**To Reproduce**
```python
import pandas as pd
import numpy as np
from darts import TimeSeries
from darts.models import NaiveEnsembleModel, NaiveMean

dates = pd.date_range("2020-01-01", "2020-12-31", freq = "D")
nr_dates = len(dates)
ts=TimeSeries.from_times_and_values(dates, 3*np.ones(nr_dates), columns=["TS"])
model = NaiveEnsembleModel([NaiveMean()])
model.fit(ts)
prediction = model.predict(10)
assert ts.components == prediction.components
```
raises an AssertionError

**Expected behavior**
No assertion error should be raised

 - Python version: 3.10
 - darts version 0.23.1

DavidKleindienst:
Fixes #1601.

### Summary
Fixes the component naming of `NaiveEnsembleModel.predict`, so that the component of the prediction has the same name as the component of the TimeSeries used for fitting.


dennisbader (approved):
LGTM. Thanks for this PR and the patience @DavidKleindienst

PR #1609, opened 2023-03-02
implementing changes for exposing Lightning Tuner's LR finder

### LINKED ISSUE #1503: Lightning Tuner functions
The [Tuner functions](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.tuner.tuning.Tuner.html?highlight=lr_find#pytorch_lightning.tuner.tuning.Tuner) of Lightning would come in handy, and could be relatively easy to implement.

Currently it is possible to access them, but in a super hacky way in two steps:
1. Start a model fit, and interrupt it (This initializes the trainer, the datasets and all the necessary bells and whistles)
2. Then access it via some very awkward syntax:

`lr_finder = model_nhits.model.trainer.tuner.lr_find(model_nhits.model, train_dataloaders=model_nhits.model.trainer._data_connector._train_dataloader_source.instance)`

(see  [here](https://github.com/unit8co/darts/issues/1403#issuecomment-1332598925) )

The observation is, that the `fit()` and from it the `fit_from_dataset()` calls are essentially doing all the setup, and the only thing not needed would be one line [here](https://github.com/unit8co/darts/blob/master/darts/models/forecasting/torch_forecasting_model.py#L916), which is actually calling the training.

Suggestion:
- Adding a flag like `setup_only=True` to fit (default False), thus letting people use the fit function for setup
- Making the `lr_find()` and `scale_batch_size()` functions available from the Darts model, not having to use this awkward way of accessing it.

Alternative would be to refactor the fit in a way, that it has an internal `setup()` call, and then continues with training, then letting people call this `setup()` externally, but I think this is way more effort for basically no benefit.

**Additional context**
In case of more straining model trainings, both tuner options are pretty beneficial, the  `scale_batch_size()` for maximizing GPU utilization, the  `lr_find()` for more efficient LR policies, like [1cycle](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.OneCycleLR.html) that depend heavily on max LR measurement.

solalatus:
Fixes #1503.

### Summary

As proposed in the linked issue [here](https://github.com/unit8co/darts/issues/1503), this PR implements a lightweight setting for the `fit()` methods of the `TorchForecastingModel`s to just do a "setup" run, and return gracefully without being interrupted and without running the main training loop. This enabled the usage of [Lightning Tuner](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.tuner.tuning.Tuner.html) to find optimal learning rates. For this a new method is exposed, that wraps Lightning's method and provides sane defaults.

### Other Information

I have not done unittests for this functionality, since I am basically exposing some already present functionality. None the less I included a Notebook in the examples [here](https://github.com/solalatus/darts/blob/master/examples/18-lr_finder.ipynb) that demonstrates the usage of the feature.

Updates (from @dennisbader):
- decouple `fit_from_dataset` and `_train` setup from fitting the model. like this we can setup the model without having to fit the model, or having to add a parameter `setup_only`
- adds unit tests


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -915,11 +938,86 @@ def fit_from_dataset(
...
-        # Train model
-        self._train(train_loader, val_loader)
+        # Train model if not in "setup_only" mode
+        if not setup_only:
+            self._train(train_loader, val_loader)
         return self
 
+    def lr_find(
+        self,
+        min_lr: float = 1e-08,
```
dennisbader, line 951: We could also just accept the `lr_find`parameters as kwargs (apart from those that we will set such as model, dataloaders, update_attr). Then we just link to the PyTorch Lightning for the `lr_find` docs.

Like that we can avoid having to maintain possible changes to the lr_find API.

dennisbader (changes requested):
Hi @solalatus and thanks for this PR! Looks good already :)

I made a suggestion for adding a new private method that handles setting up the model for training, to avoid adapting the fit API and users having to call fit() with setup_only=True before lr_find.

darts/models/forecasting/torch_forecasting_model.py
solalatus: Cool. This makes things more transparent.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -775,6 +791,7 @@ def fit_from_dataset(
         verbose: Optional[bool] = None,
         epochs: int = 0,
         num_loader_workers: int = 0,
+        setup_only: bool = False,
```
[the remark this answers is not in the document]
solalatus: The dilemma I have here is, that there is considerable amount of checking logic in `fit()` as well as `fit_from_dataset()`, and I think, both should be executed in both cases, when people call `fit()` as well as `lr_find()`. IF this is true (which I am not 100% convinced about), I don't really know how to elegantly refactor into `_setup_train()` AND keep both `fit()` and `fit_from_dataset()`.
As said, I am not completely sure I am right here, so any advice is very much welcome!

dennisbader:
Hi @solalatus, just wanted  to check in how it's going with this PR. 
We plan to release a new version in the next 2 weeks, and would like to include this one :) (no rush though!)

solalatus:
@dennisbader thanks for checking. Sadly I am completely overwhelmed, so will not be able to make progress on this. Can anyone please take over and push it through the "finish line"?

dennisbader:
Yes, I can take over from here. Thanks for the work so far!

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -590,3 +593,44 @@ def test_invalid_metrics(self):
                     12, "RNN", 10, 10, n_epochs=1, torch_metrics=torch_metrics
                 )
                 model.fit(self.series)
+
+        @pytest.mark.slow
+        def test_lr_find(self):
+            train_series, val_series = self.series[:-40], self.series[-40:]
+            model = RNNModel(12, "RNN", 10, 10, random_state=42)
+            # find the learining rate
```
madtoinou, line 601:
```suggestion
            # find the learning rate
```

madtoinou (approved):
LGTM, I find this approach quite elegant!

dennisbader (approved):
Looks good now :) Thanks @solalatus for the contribution!

solalatus:
@dennisbader thanks for pushing it through! Cheers!

Issue #1595, opened 2023-02-24
[BUG] NaiveEnsembleModel does not work with NaiveSeasonal


turbotimon:
**Describe the bug**
NaiveEnsembleModel with a NaiveSeasonal with k>3 almost allways gives me an ValueError.
I dont' found a way to prevent this

**To Reproduce**

```python
from darts import TimeSeries
from darts.models.forecasting.baselines import NaiveMean, NaiveSeasonal, NaiveEnsembleModel
import numpy as np

y = TimeSeries.from_values(np.arange(100))

# NaiveSeasonal alone works fine (no error)
naive_seasonal = NaiveSeasonal(7)
naive_seasonal.fit(y[:50])
naive_seasonal.backtest(y);

# However inside an Ensemble, it gives me a ValueError
naive_ensemble= NaiveEnsembleModel([NaiveMean(), NaiveSeasonal(7)])
naive_ensemble.fit(y[:50])
naive_ensemble.backtest(y);
# ValueError: Train series only contains 3 elements but Naive seasonal model, with K=7 model requires at least 7 entries

# Even if the inputs are dividable by 7 its an ValueError
naive_ensemble= NaiveEnsembleModel([NaiveMean(), NaiveSeasonal(7)])
naive_ensemble.fit(y[:49])
naive_ensemble.backtest(y[:98]);
# ValueError: Train series only contains 3 elements but Naive seasonal model, with K=7 model requires at least 7 entries
```

**Expected behavior**
It should work as it works alone

**System (please complete the following information):**
 - Python 3.8.10
 - darts 0.23.1

**Additional context**
- As fare as i figured out, it goes down to the EnsembleModel Base Class
- It is may solvable by taking into account the `extreme_lags`  proberty of each model inside of the ensemble
- However, it is strange to me, that it is also happening when the inputs are dividable by k=7. So i don't even have a workaround at the time! In fact, i didn't found any length of y [0..100] or k>3 so that it works

turbotimon:
Ok, the workaround is to start k lags later: `naive_ensemble.backtest(y[7:]);`

But i still consider this as a bug as it works alone. At least the error message could be improved in such cases. It is also somehow not clear to me why it works with k=3.

JanFidor:
[...]

PR #1616, opened 2023-03-04
Fix/ensemble historical forecasts

JanFidor:
Fixes https://github.com/unit8co/darts/issues/1595

### Summary

Calling `historical_forecasts `or `backtest` on an ensemble model containg at least one model having `lags` horizon > 1 or `input_chunk_length` > 1 throws error.
Caused by using default ensemble's `extreme_lags` property, instead of calculating a new one using its models'  `extreme_lags` 

### Other Information
Added  `extreme_lags` property to theta models -> they were previously throwing errors. 
There might be a way of deleting `extreme_lags` altogether and only using `min_train_series_length`, but I wanted to fix the bug ASAP and think about potential refactorization later.


darts/models/forecasting/ensemble_model.py
```diff
@@ -197,5 +197,21 @@ def ensemble(
...
+    def extreme_lags(self):
+        extreme_lags_length = len(super().extreme_lags)
+
+        return [self._find_max_lag_or_none(i) for i in range(extreme_lags_length)]
+
+    def _find_max_lag_or_none(self, lag_id):
+        max_lag = None
+        for model in self.models:
+            curr_lag = model.extreme_lags[lag_id]
+            if max_lag is None or (
```
dennisbader, line 210: I'm not sure if this respects the actual min/max of all extreme lags.

For example lags_future_covariates for one regression models could be [-3, -2], and for a second one [-1, 1]. 
The min lag for both models should be -3, and max lag +1. 
However, the current implementation would give min/max lags of -3 and -2

dennisbader (changes requested):
Hi  @JanFidor and thanks for another PR 🚀  :) 

Looks good! Only had a comment where I believe we're not getting the correct overall extreme lags.

darts/models/forecasting/ensemble_model.py
JanFidor: Okay, I assumed that for a given lag all values must either be positive or negative, and abs() would be working then. I checked out the docs and I think that the new implementation fixes the problem

JanFidor:
Hi @dennisbader ! As always, I really appreciate both that you took the time to review the PR and the chance to help out with darts, even if it's just a little for now 🚀

darts/models/forecasting/ensemble_model.py
```diff
@@ -197,5 +197,22 @@ def ensemble(
...
         return max(model.min_train_series_length for model in self.models)
 
+    @property
+    def extreme_lags(self):
+        lag_aggregators = (min, max, min, min, max)
+        return [
+            self._find_max_lag_or_none(i, agg) for i, agg in enumerate(lag_aggregators)
+        ]
+
+    def _find_max_lag_or_none(self, lag_id, aggregator):
```
dennisbader, line 207: Could you add a unit test for this? It would be great to check that ensemble models now get the correct extreme lags (for example for the regression model example I mentioned in a previous comment?) No need to call backtest/historical_forecasts, just a simple check that we get the correct lags.

dennisbader (commented):
Hi @JanFidor and thanks for the updates 🚀 

I just had a suggestion about adding a unit test that checks that we get correct extreme lags based on some example models.

After that, it's ready to be merge :)

dennisbader (approved):
Looks good, thanks a lot @JanFidor  🚀 
I took the chance to update some parts so we can go ahead with the release soon.

JanFidor:
Thanks for the review and help with the merge conflicts @dennisbader  ! I think this PR might be my favorite one, so I'm very happy it will be released soon. A little unrelated, but Uni picked up significantly when it comes to the workload, so I'll unfortunately have to cut down on the new contributions for the time being, but I'll do my best to finish up an issue I picked up. I really enjoyed working with everyone here, had a blast writing the code and yeah, I just wanted to let you all know, so It doesn't seem like I disappeared out of the blue. I'll certainly be back with even more PRs, but it might be only after the term and exams end, so around the summer. Until then! cc @madtoinou

PR #1619, opened 2023-03-05
Fix/hist forecast non retrain error

### LINKED ISSUE #1617: [BUG] historical_forecasts always retrains model
**Describe the bug**
historical_forecasts always retrains the model ignoring the retrain argument.

**To Reproduce**
This was likely introduced in https://github.com/unit8co/darts/pull/1465. Adding `model = self.untrained_model()` no longer sets `self._fit_called`. When evaluating whether to retrain `not self._fit_called` is always true.

```python
import pandas as pd
import numpy as np

from darts import TimeSeries
from darts import models

series_air = AirPassengersDataset().load().astype(np.float32)
model = models.LinearRegressionModel(lags=4)

model.historical_forecasts(
    series_air,
    start=0.9,
    retrain=False, # <-- this is ignored
)
```

**Expected behavior**
forecasting models to respect the retrain parameter.

 - Python version: 3.9.16
 - darts version 0.23.1

dennisbader:
Fixes  #1617.

### Summary

General historical_forecasts improvements.
- raises an error when retrain is False and model has not been fit before
- now correctly uses a retrained model in next iterations when retrain is not training every step
- adds unit tests for different `retrain` parameter values


hrzn (approved):
👍

PR #1628, opened 2023-03-07
warn user if suspicion of wrong index in from_group_dataframe()

### LINKED ISSUE #1607: Raise warning in `from_group_dataframe()` for monotonically increasing index.
See #1606.

In `TimeSeries.from_group_dataframe()`:
- raise a warning when we observe a monotonically increasing index, and `time_col=None` to avoid downstream issues
- extend docs that the actual index will be respected if `time_col=None`, to make user aware

dennisbader:
Fixes #1607.


darts/timeseries.py
```diff
@@ -775,6 +776,12 @@ def from_group_dataframe(
         List[TimeSeries]
             A list containing a univariate or multivariate deterministic TimeSeries per group in the DataFrame.
         """
+        if time_col is None and df.index.is_monotonic_increasing:
+            logger.warning(
+                "UserWarning: `time_col` was not set and `df` has a monotonically increasing (time) index. You can "
+                "ignore this warning if the index represents the actual index of each individual time series group."
```
madtoinou, line 782: I would maybe add something between parentheses saying that a monotonically increasing time index in this context means that the groups have non-overlapping (time) index.

madtoinou (commented):
Minor comment about the phrasing, to make it even more obvious to the user that a monotonically increasing time index for the whole DataFrame is potentially not what they expect/want.

darts/timeseries.py
dennisbader: 👍

PR #1654, opened 2023-03-20
fix failing test when lightning version <2.0.0

dennisbader:
### Summary
- fixes failing unit test from #1651 when test env uses older pytorch lightning versions


madtoinou (approved):
LGTM 🚀

PR #1592, opened 2023-02-23
Fix/Make reconcilliation order independent

DavidKleindienst:
Fixes #1582.

### Summary
Fixes `get_projection_matrix` in `BottomUpReconciliator` and `TopDownReconcilliator` such that the result takes the order of components of the `TimeSeries` into account.
Adds a test to assert the results of `BottomUpReconciliator`, `TopDownReconciliator` and `MinTReconciliator` do not depend on component order

PR #1656, opened 2023-03-21
Fix/Failing test - make TimeSeries.with_hierarchy type hint python 3.7 compatible

DavidKleindienst:
Fixes the type hint introduced in #1592 to make it compatible with python 3.7.
Should fix the failing test on master

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -2857,7 +2857,7 @@ class TimeSeries:
             )
         )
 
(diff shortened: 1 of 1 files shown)


dennisbader (approved):
Thanks for the quick fix!

dennisbader:
There was an error in the merge workflow due to Codecov not uploading, this is an issue unrelated to this PR

PR #1658, opened 2023-03-21
avoid no checkpoint warning when loading from checkpoint

dennisbader:
### Summary
- currently, when loading a TorchForecastingModel from checkpoint, a warning is raised that weights could not be loaded (because we call `TorchForecastingModel.load()`). We can avoid this by simply calling torch.load(), as also done in `load_weights_from_checkpoint()`


madtoinou (dismissed):
I was not able to comment it directly since it did not change but the docstring for the `kwargs` argument should be updated,  and the entirety of the `kwargs` should be passed to `torch.load()` instead of just the `map_location`.

Otherwise, looking good to me, I feel bad for not thinking about it before :)

PR #1668, opened 2023-03-24
Feature/add_midas_transformer

### LINKED ISSUE #1570: Add transformer that converts higher frequency time series to lower frequency
In time series forecasting sometimes you come across a situation in which the frequency of your target doesn't match the frequency of your covariates. 

For example, your target might be quarterly while some of your covariates are monthly. 

A simple way of dealing with this is up- or downsampling your data. However, doing this means that you either lose some data (e.g. when you go from a monthly to a quarterly frequency) or it can lead to information leakage (e.g. when you go from quarterly to monthly). 

It might be nice to incorporate some other way of dealing with this in Darts. One way of addressing low frequency targets in combination with high frequency covariates is called [mixed-data sampling (MIDAS)](https://en.wikipedia.org/wiki/Mixed-data_sampling). In the case of quarterly targets and monthly covariates, it would mean creating three columns for each monthly covariate, each containing different lags and relating those to the quarterly targets.

Would this be a worthwhile addition to Darts or would it add too much complexity?
It could be part of data processing or it could somehow be incorporated in e.g. RegressionModel.

Beerstabr:
Fixes #1570 

### Summary

Added a transformer that converts a (high frequency) time series into a lower frequency time series using the mixed-data sampling (MIDAS) approach. For example, covert a monthly time series into a quarterly time series. The benefit of doing this is that you can now use e.g. a monthly time series as an input for a model that forecasts a quarterly targets series without losing (as much) information as you would when you downsample a time series from monthly to quarterly using e.g. pandas.DataFrame.resample.

Limitations:
- The frequency of the original time series must always be exactly the same multiple of the target frequency for MIDAS to be possible. For example, there are always four quarters in a year, but the number of days in a month is not always the same.
- MIDAS only works when you go from a high frequency to a lower frequency.

Possible improvements:
- Aggregating columns to avoid an intermittent time series or to limit the increase in components. For example, when you go from seconds to minutes you'd get 60 columns holding data for each of the seconds in a minute. You could also aggregate this to four columns each representing 15 seconds.


madtoinou:
@Beerstabr Some tests seem to fail because of `None` values, could you add the necessary logic to the MIDAS transformation so that these cases are properly handled?

Beerstabr:
Mmm, okay I will have a look!


Op ma 27 mrt. 2023 om 18:18 schreef madtoinou ***@***.***>

Beerstabr:
@madtoinou it should be fixed now. I wasn't dealing with the 'args' in the correct way, it had changed without me rerunning the tests.

dennisbader:
Hi @Beerstabr, and thanks for this PR! 🚀 Before I review I was thinking that we could make this an invertible transformer.

I would be intersted to see what happens when we have a higher frequency target and lower frequency covariates.
We'd use MIDAS to transform the target into a multivariate series of this lower frequency. Now we forecast that multivariate series with the covariates and in the end inverse transform to the higher frequency.

What do you think? :)

Beerstabr:
@dennisbader sounds good! I'll work on it next week!

dennisbader:
Hi @Beerstabr , just wanted to check in if there were any updates on this one? :)

Beerstabber:
Hi @dennisbader, I've been drowned in work lately, and I've not gotten to it. Really would like to finish it though. No updates for now. I have heard this kind of a technique has been implemented in a large retail firm with success so it should be worthwhile.

Beerstabr:
Hi @dennisbader, I've been drowned in work lately, and I've not gotten to it. Really would like to finish it though. No updates for now. I have heard this kind of a technique has been implemented in a large retail firm with success (giving a significant increase in accuracy) so it should be worthwhile.

dennisbader:
Hi @Beerstabr, thanks for the update. No worries, I totally understand. We can take it over from here.
Thanks again for all the work so far on this very neat feature! 🚀

madtoinou:
Hi @Beerstabr,

Thank you a lot of the work, I am going to take over this PR and try to finish it in time for the next release 🚀.

Beerstabr:
@madtoinou thanks! Let me know if something is unclear and I can help to explain it.

madtoinou:
Thank you, the code was clear enough.

In order to reverse the transformation, the number of component generated during the transform is extracted and used to divide the low frequency period to retrieve the high frequency period. This approach does not work for period above months because of the varying number of days so I added some rule-based logic (pandas `Timedelta` does not support units 'M', 'Y', and 'y' as they "do not represent unambiguous timedelta values durations", making me thing that there is no other solution).

Still requires a bit work to properly support multivariate `TimeSeries` but should not be too difficult.

dennisbader:
Closing this as #1820 was merged.

Thanks again for this @Beerstabr! 🚀

PR #1675, opened 2023-03-27
Fix/historical_forecasts callable retrain argument

### LINKED ISSUE #1655: [BUG] Historical forecast with retrain=Callable forecasts wrong dates
**Describe the bug**
When providing a `Callable` as `retrain` attribute, `CatBoostModel.historical_forecasts` creates a historical forecast for the wrong dates.
This wrong behaviour only happens when `output_chunk_length` is set to a value `>=3`. Setting `retrain=False` leads to correct results.
Interestingly, `retrain=False` and `retrain=lambda *args: False` produce different results, which is quite unexpected.

**To Reproduce**
```python
from darts.models import CatBoostModel
from darts import TimeSeries
import pandas as pd
import numpy as np

dates = pd.date_range("2021-02-01", "2021-04-19", freq="W-Mon")
series = TimeSeries.from_times_and_values(dates, np.sin(np.arange(len(dates))))
future_cov = TimeSeries.from_times_and_values(dates, np.cos(np.arange(len(dates))))

model = CatBoostModel(lags_future_covariates=(0, 1), output_chunk_length=10)
model.fit(series, future_covariates=future_cov)


hist_fc = model.historical_forecasts(series, future_covariates=future_cov, start=pd.Timestamp("2021-02-11"), retrain=lambda *args: False)
print(hist_fc.start_time())
# >> 2021-04-12 00:00:00

hist_fc2 = model.historical_forecasts(series, future_covariates=future_cov, start=pd.Timestamp("2021-02-11"), retrain=False)
print(hist_fc2.start_time())
# >> 2021-02-15 00:00:00
```
**Expected behavior**
`model.historical_forecasts` should honor the `start` parameter and start forecasting at the first timepoint after `start`, also when `retrain` is set to a callable. Further, historical_forecasts with `retrain=False` and `retrain=lambda *args: False` should return the same results (except maybe for some randomness in the forecast)

 - Python version: 3.10
 - darts version: 0.23.1 & master

**Additional context**

### LINKED ISSUE #1670: I can't see the importances of my variables
Hi,

I just implemented a lightGBM. I would like to see the importances of my variables, but when I ask for feature importances I get an array of numbers, so i can't see which importance belongs to each variable.
image
And if i ask for the names of the variables so I can see the importances of each one, I get these names
image

So in the end, I'm still unable to view the importances of my variables.

In the other hand, I would like to know the criteria on which the importances are based. We don't expect to be equal the importances based on the criterion "split" than the ones based on the criterion "gain".
How can I solve these issues? Is there anything I can help with?

Thank you very much for your great job and for the attention.

madtoinou:
Fixes #1655.

### Summary

The retrain argument of `historical_forecast()` was not properly handled when it was a Callable, in order to support it appropriately the following changes has been made:
- the time index that can be used in "training" and "prediction" mode are computed separately, and if retrain is a Callable, the longest is used to build the iterator
- the retrain function is systematically called before initiating the training round, using the "dataset that would be used for training"
- the retrain function has a fixed signature with the following positional arguments: `[counter, pred_time, train_series, past_covariates, future_covariates]`

And the explanation behind the fact that any value <= 3 would mask the problem is linked to the fact that `model.min_train_series_length` takes the maximum between this constant and the `output_chunk_length`. Since this value could be longer than the `model.min_predict_series_length`, the Callable retrain was slicing the predictable time index too aggressively.

### Other Information
I added some tests to cover the models with:
- "lags + future covariates, output_chunk_length = 1"
- "lags + future covariates, output_chunk_length > 3"
- "future covariates, output_chunk_length > 3"

But I am not entirely sure of the expected forecast's length for the last 2 scenarios, I need to run some additional tests.

Update 10-04-2023 (@dennisbader):
- move around `start` and `train_length` logic to account for all inputs
- improved error messages for start param sanity checks
- raises warning when valid start param but not in predictable/trainable index
- improvements to the way we handle train length, raises warning when too large
- option to disable warnings
- sets min samples required for torch models to 1 (only regression models require minimum of 2 samples)
- update docs for historical_forecasting, backtest, gridsearch, residuals


darts/models/forecasting/forecasting_model.py
```diff
@@ -834,34 +907,36 @@ def historical_forecasts(
...
-                        min(
-                            pred_time + train_series.freq * forecast_horizon,
-                            series_.end_time(),
+                # retrain_func processes the series that would be used for training
+                if (
+                    retrain_func(
+                        counter=_counter,
+                        pred_time=pred_time,
+                        train_series=train_series,
+                        past_covariates=past_covariates_.drop_after(pred_time)
```
dennisbader, line 927: the same here. If we really want to drop, then we should give the correct past/future covariate time spans given by the model lags.
Otherwise, I think we can simply pass the entire past/future covariates, as the user already has the pred time, and train series up until the forecasting point.
(it can take some time to create time series with drop_after, especially in a loop like historical forecasts)

darts/models/forecasting/forecasting_model.py
```diff
@@ -885,6 +960,14 @@ def historical_forecasts(
                     # use retrained model if `retrain` is not training every step
                     model = model if model is not None else self
 
+                    # slice the series for prediction without retraining
+                    if min_timestamp_predict > series_.time_index[0]:
+                        train_series = series_.drop_before(
+                            min_timestamp_predict - 1 * series_.freq
+                        ).drop_after(pred_time)
+                    else:
+                        train_series = series_.drop_after(pred_time)
```
dennisbader, line 969: no need to drop before, right?
```suggestion
                    train_series = series_.drop_after(pred_time)
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -834,34 +907,36 @@ def historical_forecasts(
...
+                                series_.end_time(),
+                            )
                         )
+                        if future_covariates_
+                        else None,
                     )
-                    if future_covariates_
-                    and ("future_covariates" in retrain_func_signature)
-                    else None,
+                    and pred_time in historical_forecasts_time_index_train
```
dennisbader, line 939: `if` statements are executed from left to right. 
So we have to check like 

```
if pred_time in historical_forecasts_time_index_train and retrain_func(...):
```

An example: the following will raise an error because we cannot perform `>` on NoneTypes:
```
>>> x = None
>>> if x > 1 and x is not None:
...     pass
... else:
...     print(1)
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

But this one doesn't raise an error because we first check if `x` is not None:
```
>>> if x is not None and x > 1:
...     pass
... else:
...     print(1)
... 
1
...

File: darts/tests/models/forecasting/test_historical_forecasts.py [hunk 4]
```diff
@@ -47,14 +47,28 @@
...
             LinearRegressionModel,
             {"lags": 4, "lags_past_covariates": 6, "output_chunk_length": 4},
             (6, 4),
         ),
+        # target + future covariates
+        (LinearRegressionModel, {"lags": 4, "lags_future_covariates": [0, 1]}, (4, 2)),
+        # target + fut cov + output_chunk_length > 3
+        (
+            LinearRegressionModel,
+            {"lags": 2, "lags_future_covariates": [0, 1], "output_chunk_length": 6},
```
dennisbader, line 63: This hasn't been adjusted yet after our discussion, right?

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -788,20 +848,36 @@ def helper_hist_forecasts(retrain_val):
...
-            def retrain_f_valid(pred_time, train_series):
+            def retrain_f_invalid_args(
+                counter, train_series, past_covariates, future_covariates
+            ):
+                if len(train_series) % 2 == 0:
+                    return True
+                else:
+                    return False
+
+            def retrain_f_valid(
```
dennisbader, line 864: let's also catch a function that returns a non-boolean

dennisbader (changes requested):
Hi @madtoinou and thanks for getting the retrain callable to work properly 🚀 🥳 

I left some minor suggestions, mainly about adapting the error messages a bit, trying to avoid creating new time series as much as we can, and fixes to some potential issues.

Btw, @dumjax and I were working on the last historical_forecasts bug fixes, that should will come in parallel with this PR. Then we're finally ready to put everything together for the release, and concentrate on optimizing the method for the next release :) 

Thanks for all the great work so far!

darts/models/forecasting/forecasting_model.py
madtoinou: The two conditions were interchangeable and I picked this order because it was to cover the scenario where the `historical_forecasts_time_index_predict` starts before the `historical_forecasts_time_index_train` with a `retrain_func` returning `True` for the first `Timestamps`.

But if for some reason, the `retrain_func` returns unexpected values for this value of `pred_time`, then yes, your order would be more robust.

darts/tests/models/forecasting/test_historical_forecasts.py
madtoinou: Not yet, I need to run some experiments with `Timestamps`

darts/models/forecasting/forecasting_model.py
dennisbader: The thing is we don't need to call retrain_func if the pred_time is not in the train indices. The way I suggest will not call retrain_func if it is not in the train index.

```
>>> pred_time = 0
>>> train_index = [1,2,3]
>>> def f():
...     print("called")
...     return True
... 
>>> if pred_time in train_index and f():
...     print("blabla")
... 
>>> if f() and pred_time in train_index:
...     print("blabla")
... 
called

File: darts/models/forecasting/forecasting_model.py (hunk 2 above)
Comment (dennisbader) [reply darts/models/forecasting/forecasting_model.py:972]: In that case could we try to do this only once per time series, and not in every fit/predict iteration?

File: darts/models/forecasting/forecasting_model.py (hunk 1 above)
Comment (madtoinou) [reply darts/models/forecasting/forecasting_model.py:930]: I agree with the idea of passing the entire past/future covariates and have the user leverage the `predict_time` argument of `retrain_func` to process everything.

File: darts/models/forecasting/forecasting_model.py [hunk 6]
```diff
@@ -730,37 +762,78 @@ def historical_forecasts(
             past_covariates_ = past_covariates[idx] if past_covariates else None
             future_covariates_ = future_covariates[idx] if future_covariates else None
 
-            # Determines the time index the historical forecasts will be made on
-            historical_forecasts_time_index = (
+            # Prediction
```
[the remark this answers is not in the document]
madtoinou: It's indeed handled by the `GlobalForecastingModel` `fit()` method when necessary.

darts/models/forecasting/forecasting_model.py
madtoinou: If we alternate between "training" and "predict" mode, the `drop _before` needs to be called at each iteration...

madtoinou:
After a lot of experimentation with the timestamps of the output of `historical_forecasts`, I identified some bugs:
- for models using positive `lags_future_covariates` only:  the first predictable timestamp is the first value of the input `series`, requires a dummy support ts for the `predict` method.
- the timestamp shift (`min_timestamp_predict`) should be affected by the lags in the past only
- the unittest with the "manual" boundaries were not allowing for enough granularity: the `output_chunk_length` is relevant only when `retrain=True` and cannot be used interchangeably with the largest positive`lags_future_covariate`
- the sanity check for `historical_forecasts` were not supporting models using only positive `lags_future_covariates` when `retrain=False`

dennisbader (commented):
Thanks @madtoinou , that looks really good :)

I think we're in the final round of review and can merge soon!  🚀 

Just had a couple of minor suggesetions.

darts/models/forecasting/forecasting_model.py
madtoinou: You're correct, since we overwrite `train_series` at each iteration, it should not interfere with the training mode and the predict mode is indeed capable of extracting the required inputs.

darts/models/forecasting/forecasting_model.py
```diff
@@ -889,6 +984,25 @@ def historical_forecasts(
                     # use retrained model if `retrain` is not training every step
                     model = model if model is not None else self
 
+                    # slice the series for prediction without retraining
```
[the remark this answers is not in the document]
madtoinou: Reformulating the two cases for eventual other reviewers:
- the model is not trained for the first "predictable" timestamp -> user should either call `fit()` before `historical_forecasts`, modify the `retrain` argument so that it returns `True` at least once before/at this timestamp or change the `start` value to the suggested timestamp so that the model training can be triggered before the prediction round.
- `retrain` is False for the first "trainable" timestamp and the model was not trained before `historical_forecasts` -> user must modify the `retrain` argument.

dennisbader (approved):
Looks good 🚀 💯  Great job @madtoinou !

PR #1676, opened 2023-03-27
[Feature] Support a user defined function name in the window transformation output

JQGoh:
### Summary
 As mentioned in the darts' [gitter chat](https://matrix.to/#/!uumevxjBaNJhovFYgj:gitter.im/$5_mjnsoUNwm3NSisIe_Oggp3Q2191nyQmbmKhQp4rqA?via=gitter.im&via=matrix.org&via=matrix.thegolem.cz), this PR supports a user-defined function name as part of the window transformation output name. 
 
 Without this feature, if we provide multiple user-defined functions for the window operations, we have multiple output names such as `rolling_udf_5_target`, `rolling_udf_5_target_1` and etc. This poses the challenge in tracking the derived feature names.

### Other Information
 * Also include an additional test case for window transformation that considers a user-defined function  such as `count_above_mean`.

### DIFF
--- a/darts/tests/dataprocessing/transformers/test_window_transformations.py
+++ b/darts/tests/dataprocessing/transformers/test_window_transformations.py
@@ -148,6 +148,23 @@ class TimeSeriesWindowTransformTestCase(unittest.TestCase):
             ],
         )
 
+        # test customized function name that overwrites the pandas builtin transformation
+        transforms = {
+            "function": "sum",
+            "mode": "rolling",
+            "window": 1,
+            "function_name": "customized_name",
+        }
+        transformed_ts = self.series_univ_det.window_transform(transforms=transforms)
+        self.assertEqual(
+            transformed_ts.components.to_list(),
+            [
+                f"{transforms['mode']}_{transforms['function_name']}_{str(transforms['window'])}_{comp}"
+                for comp in self.series_univ_det.components
+            ],
+        )
+        del transforms["function_name"]
+
         # multivariate deterministic input
         # transform one component
         transforms.update({"components": "0"})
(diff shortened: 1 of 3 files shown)


JQGoh:
@madtoinou FYI

JQGoh:
I realize that I did not validate that the `function_name` is a string type. Will do so after the code review.

darts/tests/dataprocessing/transformers/test_window_transformations.py
```diff
@@ -494,12 +544,18 @@ def test_transformers_pipeline(self):
         expected_transformed_series = TimeSeries.from_times_and_values(
             times1,
             [100, 15, 30, 45, 60, 75, 90, 105, 120, 135],
-            columns=["rolling_sum_3_2_0"],
+            columns=["rolling_TotalOperation_3_2_0"],
```
madtoinou, line 547: I would leave the name "sum", it describes the transform better in my opinion.

darts/tests/dataprocessing/transformers/test_window_transformations.py
```diff
@@ -494,12 +544,18 @@ def test_transformers_pipeline(self):
...
         )
 
         # adds NaNs
         window_transformations = [
-            {"function": "sum", "mode": "rolling", "window": 3, "min_periods": 2}
+            {
+                "function": "sum",
+                "function_name": "TotalOperation",
```
madtoinou, line 554: I would not use the "function_name" feature in this test, just to make sure that the default behavior when `fn` is not "apply" works correctly.

madtoinou (commented):
Looking good to me 🚀

I would just leave the "sum" transform name in the test, "TotalOperation" is less understandable in my opinion and we need to cover the case where the function name is not provided and `fn` is not "apply".

It's probably not necessary to check that `function_name` is a string, the documentation is clear and I don't see which other type the users might be tempted to use...

darts/tests/dataprocessing/transformers/test_window_transformations.py
JQGoh: Sounds good to me. My original intention is to test that for initialization via `WindowTransformer`, we could also get a customized name correctly and not want to initialize a new test to simply check the name in a repeated manner. I will remove this test

madtoinou (approved):
Thank you for addressing the comments so quickly 🚀

JQGoh:
@dennisbader May I check that any chance that this PR could be in time as part of the new release? Thanks

dennisbader (approved):
Looks good, thanks @JQGoh  🚀 And yes, this will be part of the 0.24.0 release :)

PR #1679, opened 2023-03-29
Feat/lagged features names

### LINKED ISSUE #1670: I can't see the importances of my variables

madtoinou:
Fixes #1670.

### Summary

Added helper functions in the tabularization and regression model in order to generate labels for the lagged features (and static covariates if applicable), which are stored in the `RegressionModel.lagged_features_name_` attribute (`List[List[str]`). This enable the usage of the `feature_importances` attribute, available for some sklearn models. It was not possible before because the `create_lagged_data` method returns arrays and lose the name of the columns.

If the model was fit on a single `TimeSeries`, the attribute contains only one `List[str]`, if trained on a `Sequence` of `TimeSeries`, the attribute contains several `List[str]`.

The Lists containing the lagged features names are always nested, the API could eventually be simplified:
- if the components names are identical across all the `TimeSeries` used during training, retain the first one only
- wrapping the access to this attribute in a function, asking the user to provide a series (and mapping the training TimeSeries to their lagged features names)

### Additional information
Added the corresponding tests.


madtoinou:
Updated the PR:
- if argument is a single `TimeSeries`, its components' name are used
- if argument is a `Sequence[TimeSeries]` and all the ts have exactly identical components names, these names are used
- if argument is a `Sequence[TimeSeries]` and any of the ts contains a component with a different name, generic names are generated ('target', 'past_cov', 'future_cov').

Each "variate" (target, past_covariates and future_covariates) is processed independently: it's possible to have a mixture of generic names and "original names".

madtoinou:
Updated to use the same naming conventions as the explainability module: `{comp_name}_lag{lag_idx}` for unique component names, `comp{i}_{variate_type}_lag{lag_idx}` for generic names.

dennisbader (approved):
Looks good, thanks a lot @madtoinou 🚀 

Ready for release! 🥳

PR #1681, opened 2023-03-31
fixes lgb warning when not using any categorical features

dennisbader:
### Summary

- removes warning when no categorical features are used with new categorical support for LightGBM models (introduced by #1585)
- fixes deprecation warning for joblib Parallel with sklearn.utils.fixes.delayed

### DIFF
--- a/darts/tests/models/forecasting/test_regression_models.py
+++ b/darts/tests/models/forecasting/test_regression_models.py
@@ -2338,8 +2338,12 @@ class RegressionModelsTestCase(DartsBaseTestClass):
 
         # Check that mocked super.fit() method was called with correct categorical_feature argument
         args, kwargs = lgb_fit_patch.call_args
+        (
+            cat_praam_name,
+            cat_param_default,
+        ) = self.lgbm_w_categorical_covariates._categorical_fit_param
         self.assertEqual(
-            kwargs[self.lgbm_w_categorical_covariates._categorical_fit_param_name],
+            kwargs[cat_praam_name],
             [2, 3, 5],
         )
 
--- a/darts/tests/models/forecasting/test_regression_models.py
+++ b/darts/tests/models/forecasting/test_regression_models.py
@@ -2339,11 +2339,11 @@ class RegressionModelsTestCase(DartsBaseTestClass):
         # Check that mocked super.fit() method was called with correct categorical_feature argument
         args, kwargs = lgb_fit_patch.call_args
         (
-            cat_praam_name,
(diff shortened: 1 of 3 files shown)


darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -2338,8 +2338,12 @@ def test_lgbm_categorical_features_passed_to_fit_correctly(self, lgb_fit_patch):
 
         # Check that mocked super.fit() method was called with correct categorical_feature argument
         args, kwargs = lgb_fit_patch.call_args
+        (
+            cat_praam_name,
```
madtoinou, line 2342:
```suggestion
            cat_param_name,
```

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -2338,8 +2338,12 @@ def test_lgbm_categorical_features_passed_to_fit_correctly(self, lgb_fit_patch):
...
+            cat_param_default,
+        ) = self.lgbm_w_categorical_covariates._categorical_fit_param
         self.assertEqual(
-            kwargs[self.lgbm_w_categorical_covariates._categorical_fit_param_name],
+            kwargs[cat_praam_name],
```
madtoinou, line 2346:
```suggestion
            kwargs[cat_param_name],
```

darts/utils/multioutput.py
```diff
@@ -1,9 +1,8 @@
-from joblib import Parallel
 from sklearn.base import is_classifier
 from sklearn.multioutput import MultiOutputRegressor as sk_MultiOutputRegressor
 from sklearn.multioutput import _fit_estimator
-from sklearn.utils.fixes import delayed
 from sklearn.utils.multiclass import check_classification_targets
+from sklearn.utils.parallel import Parallel, delayed
```
madtoinou, line 5: Hope that it does not break with python 3.7 which support is being dropped left and right... On scikit-learn github repository, they indicate "scikit-learn 1.0 and later require Python 3.7 or newer. scikit-learn 1.1 and later require Python 3.8 or newer.". Could not find in the log when did the parallel module change...

madtoinou (approved):
Noticed a small typo, a big concerned that the new Parallel import will break depending on the Python version but the merge github action will tell us if it does.

darts/utils/multioutput.py
dennisbader: Yes that's true, let's check later on in merge workflow.

PR #1685, opened 2023-04-01
fix/historical_fore_fut_cov_notarget_lags

dumjax:
Special case where the lags are just fut covariates


darts/models/forecasting/forecasting_model.py
```diff
@@ -440,20 +440,19 @@ def _get_historical_forecastable_time_index(
...
-        if (min_target_lag is not None) and (series is not None):
-            intersect_ = generate_index(
-                start=series.start_time()
-                - (min_target_lag - max_target_lag) * series.freq
-                if is_training
-                else series.start_time() - min_target_lag * series.freq,
-                end=series.end_time(),
-                freq=series.freq,
-            )
+        # longest possible time index for target
```
dennisbader, line 446: Thanks :) Could you add some unit tests which check the issue?
It's always good when fixing a bug to add a unit test, because we apparently haven't checked it before

darts/models/forecasting/forecasting_model.py
```diff
@@ -440,20 +440,19 @@ def _get_historical_forecastable_time_index(
...
+        intersect_ = generate_index(
+            start=series.start_time() - (min_target_lag - max_target_lag) * series.freq
+            if is_training
+            else series.start_time() - min_target_lag * series.freq,
+            end=series.end_time(),
+            freq=series.freq,
+        )
 
-        # past covariates longest possible time index
+        # longest possible time index for past covariates
```
dennisbader, line 455: I'm not sure the past covariates generated indexes cover all use cases:
- What happens with RegressionModel(lags_past_covariates=[-100], output_chunk_length=1)?
 The user supplied `past_covariates` for `fit()`can actually be much shorter than the target `series` (101 steps shorter, I believe).
With the current logic and the shorter `past_covariates`, the intersection would result in a much shorter historic_forecastable_time_index (because of the `end=past_covariates.end_time()`).
The same applies to future_covariates, which can in theory also be used as past_covariates only (lags_future_covariates=[-100]).

I think for future covariates it is correct, so similarly including the max_past_cov lag should probably work

dennisbader (changes requested):
Thanks @dumjax for fixing this! 🚀 

I believe the way we currently handle past covariates doesn't cover all uses cases. I added a comment. Could you have a look?

darts/models/forecasting/forecasting_model.py
```diff
@@ -279,52 +279,54 @@ def extreme_lags(
...
+        (-10, 6, None, None, None)
         >>> model = NBEATSModel(input_chunk_length=10, output_chunk_length=7, future_covariates_lags=[4, 6])
         >>> model.fit(train_series, future_covariates)
         >>> model.extreme_lags
-        (-10, 7, None, 4, 6)
+        (-10, 6, None, 4, 6)
         """
 
-        return (-1, 1, None, None, None)
+        return -1, 1, None, None, None, None
```
dennisbader, line 329: 0, I believe, right?
```suggestion
        return -1, 0, None, None, None, None
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -279,52 +279,54 @@ def extreme_lags(
...
+            - future covariates: then the fifth and sixth elements should be `None`.
 
         Should be overridden by models that use past or future covariates, and/or for model that have minimum target
-        lag and maximum target lags potentially different from -1 and 1.
+        lag and maximum target lags potentially different from -1 and 0.
 
         Notes
         -----
-        maximum target lag (second value) cannot be `None` and is always larger than 1.
+        maximum target lag (second value) cannot be `None` and is always larger or equal than 0.
```
dennisbader, line 304:
```suggestion
        maximum target lag (second value) cannot be `None` and is always larger than or equal to 0.
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -279,52 +279,54 @@ def extreme_lags(
...
         >>> model = LinearRegressionModel(lags=3, output_chunk_length=2)
         >>> model.fit(train_series)
         >>> model.extreme_lags
-        (-3, 2, None, None, None)
+        (-3, 1, None, None, None)
         >>> model = LinearRegressionModel(lags=[3, 5], past_covariates_lags = 4, output_chunk_length=7)
         >>> model.fit(train_series, past_covariates=past_covariates)
         >>> model.extreme_lags
-        (-5, 7, -4, None, None)
+        (-5, 6, -4, None, None)
```
dennisbader, line 314:
```suggestion
        (-5, 6, -4, -4, None)
```

dennisbader (commented):
Thanks @dumjax , looks great! 🚀 

Just some minor comments :)

dennisbader (approved):
🚀

PR #1689, opened 2023-04-03
Fix/training from checkpoint

### LINKED ISSUE #1495: Issue with training from checkpoint torch models

rijkvandermeulen:
Fixes #1495 

### Summary
epochs_trained attribute is not being saved/loaded from the checkpoint after 0.23.0 release. I think this causes a problem when resuming training as there's now a discrepancy with the trainer 'max_epochs` argument. This PR changes the epochs_trained property to account for also considering the epochs already trained by the 'loaded' model.


rijkvandermeulen:
@madtoinou as discussed this morning; here's the PR for #1495. I've changed the approach slightly (by modifying the epochs_trained property). Does that make sense?

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -471,6 +471,31 @@ def test_load_weights(self):
...
+            model.fit(self.series)
+            self.assertEqual(10, model.epochs_trained)
+
+            model_from_checkpoint = RNNModel.load_from_checkpoint(
+                "test_rnn", work_dir=self.temp_work_dir, best=False
+            )
+            # epochs_trained is still 10 when loading from checkpoint
+            self.assertEqual(10, model_from_checkpoint.epochs_trained)
+
+            model_from_checkpoint.fit(self.series, epochs=15)
```
madtoinou, line 496: I think that in order to check if the issue is properly fixed, the `epochs` argument should be equal to 5 (smaller than the value of `epochs_trained` in the loaded checkpoint, which was causing issue when loaded by the PytorchLightning trainer).

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1001,9 +1001,12 @@ def _setup_for_train(
                 collate_fn=self._batch_collate_fn,
             )
         )
-
-        # if user wants to train the model for more epochs, ignore the n_epochs parameter
-        train_num_epochs = epochs if epochs > 0 else self.n_epochs
+        # Fetch max_epochs to pass to the trainer
+        train_num_epochs = (
+            max(self.epochs_trained, epochs)
+            if epochs
```
madtoinou, line 1007: if epochs is smaller than self.epochs_trained, it means that the model won't be further trained, right? From my understanding, this PR was supposed to make so that calling `fit()` on a already trained model would "incrementally retrain" it for the number of epochs indicated.

madtoinou (changes requested):
I have the impression that the PR is not solving the problem described in the associated issue(s). Correct me if I am wrong but I think that the intended behavior is that regardless of the number of `epochs_trained`, when `fit()` is called on a model, it will be trained for this number of additional epochs (and not "until this number of epochs is reached", so that the user doesn't have to remember for how many epochs the model was trained.

On a side note, I think that this PR also fixes #1090 (which describe the exact same issue as #1495).

darts/models/forecasting/torch_forecasting_model.py
rijkvandermeulen: > I have the impression that the PR is not solving the problem described in the associated issue(s). Correct me if I am wrong but I think that the intended behavior is that regardless of the number of `epochs_trained`, when `fit()` is called on a model, it will be trained for this number of additional epochs (and not "until this number of epochs is reached", so that the user doesn't have to remember for how many epochs the model was trained.
> 
> On a side note, I think that this PR also fixes #1090 (which describe the exact same issue as #1495).

Well, this PR:
- fixes the exsting issue with resuming training from checkpoints;
- but leaves the rest of the logic (regarding how we handle the 'additional' number of epochs passed in the fit call) unchanged

I'm all in favour of making this change though. It seems to me this is indeed more intuitive. I'll incorporate it in this PR

rijkvandermeulen: @madtoinou updated the PR

dennisbader:
We discussed this and came to the conclusion, that we will stick with the current implementation.
Reasons:
- when using fit() with a model loaded with `load_from_checkpoint()` the training will always resume from the checkpoint epoch. The main usage for this is to resume training in case it was interrupted. The user will only have to load the model from checkpoint before calling fit, and can leave the rest of the script unchanged (i.e. the same number of epochs as in the interrupted run) 
- for fine tuning / retraining we recommend loading the model with `load_weights_from_checkpoint()` or `load_weights()`. The user can then use any number of epochs, new optimizer, learning rate scheduler, etc.
- the user can also use `load()` for a manually saved model. For training, the user can give any number of epochs, and it will use/continue on the state of the model when it was saved 

I will close this for now. Let me know if you disagree and we can discuss further.

PR #1692, opened 2023-04-04
Feat/probabilistic ensemble

### LINKED ISSUE #1682: Is it possible to generate quantiles from an ensemble method/model?
Is it possible to generate quantile forecasts given the machinery in-place via:

```
from darts.models import RegressionEnsembleModel
```

madtoinou:
Fixes #1682.

### Summary
The `num_sample` argument is properly passed to `regression_model.predict()` in `RegressionEnsembleModel.ensemble()`.

In order to obtain probabilistic predictions, all the models of the `RegressionEnsembleModel` need to be probabilistic, including its `regression_model`.

It's however still possible to obtain deterministic predictions if the `regression_model` is not probabilistic but the others are or by using `num_sample=1`.

### Other Information
The definition of a probabilistic `EnsembleModel` should be better defined; at the moment, it requires that all its model are probabilistic however for a `RegressionEnsembleModel`, we might also want the regression model itself to be probabilistic? Or on the opposite, ss a probabilistic `regression_model` enough to make a `RegressionEnsembleModel` probabilistic even if all the other models are deterministic?

### DIFF
--- a/darts/models/forecasting/baselines.py
+++ b/darts/models/forecasting/baselines.py
@@ -204,6 +204,7 @@ class NaiveEnsembleModel(EnsembleModel):
         self,
         predictions: Union[TimeSeries, Sequence[TimeSeries]],
         series: Optional[Sequence[TimeSeries]] = None,
+        num_samples: int = 1,
     ) -> Union[TimeSeries, Sequence[TimeSeries]]:
         def take_average(prediction: TimeSeries) -> TimeSeries:
             series = prediction.pd_dataframe(copy=False).sum(axis=1) / len(self.models)
--- a/darts/models/forecasting/ensemble_model.py
+++ b/darts/models/forecasting/ensemble_model.py
@@ -167,13 +167,14 @@ class EnsembleModel(GlobalForecastingModel):
             future_covariates=future_covariates,
             num_samples=num_samples,
         )
-        return self.ensemble(predictions, series=series)
+        return self.ensemble(predictions, series=series, num_samples=num_samples)
 
     @abstractmethod
     def ensemble(
         self,
         predictions: Union[TimeSeries, Sequence[TimeSeries]],
         series: Optional[Sequence[TimeSeries]] = None,
+        num_samples: int = 1,
     ) -> Union[TimeSeries, Sequence[TimeSeries]]:
         """
         Defines how to ensemble the individual models' predictions to produce a single prediction.
--- a/darts/models/forecasting/regression_ensemble_model.py
+++ b/darts/models/forecasting/regression_ensemble_model.py
@@ -157,6 +157,7 @@ class RegressionEnsembleModel(EnsembleModel):
         self,
         predictions: Union[TimeSeries, Sequence[TimeSeries]],
         series: Optional[Sequence[TimeSeries]] = None,
+        num_samples: int = 1,
     ) -> Union[TimeSeries, Sequence[TimeSeries]]:
 
         is_single_series = isinstance(series, TimeSeries) or series is None
@@ -165,7 +166,12 @@ class RegressionEnsembleModel(EnsembleModel):
 
         ensembled = [
             self.regression_model.predict(
-                n=len(prediction), series=serie, future_covariates=prediction
+                n=len(prediction),
+                series=serie,
+                future_covariates=prediction,
+                num_samples=num_samples
+                if self.regression_model._is_probabilistic()
+                else 1,
             )
             for serie, prediction in zip(series, predictions)
         ]


madtoinou:
Simplified the definition of probabilistic `RegressionEnsembleModel` : its regression model (the "last/aggregation layer") must be probabilistic, regardless of the others models characteristics.

Having probabilistic models is allowed, they will be trained as such but their prediction will contain only one sample when passed to the regression model at fitting and prediction time (the first one, due to the way `RegressionModel._create_lagged_data()` get rid of the `n_samples` dimension).

Attaching two codes snippets to demonstrate how to get probabilistic forecast with `EnsembleModel`:

```python
from darts.models import LinearRegressionModel, LightGBMModel, RegressionEnsembleModel
from darts.utils.timeseries_generation import sine_timeseries, gaussian_timeseries

# creating synthetic data
start = 10
end = 400
sin_series = sine_timeseries(start=start, end=end, value_amplitude=10)
gaus_series = gaussian_timeseries(mean=10, start=start, end=end)

tmp = sin_series + gaus_series
train, val = tmp.split_after(0.8)

quantiles = [0.25, 0.5, 0.75]

# probabilistic ensembling model
ensemble_lin_reg = LinearRegressionModel(quantiles=quantiles,
                                        lags_future_covariates=[0],
                                        likelihood="quantile")

# probabilistic models
lgbm_model = LightGBMModel(quantiles=quantiles, lags=4, likelihood="quantile")
linreg_model = LinearRegressionModel(quantiles=quantiles, lags=4, likelihood="quantile")

ensemble = RegressionEnsembleModel([lgbm_model, linreg_model],
                                   regression_train_n_points=140,
                                   regression_model=ensemble_lin_reg
                                   )

ensemble.fit(train)
pred = ensemble.predict(len(val), num_samples=1000)

val.plot(label="test")
pred.plot(label="prediction")
plt.show()
```
all_probabilistic

```python
# deterministic models
lgbm_model = LightGBMModel(lags=4)
linreg_model = LinearRegressionModel(lags=4)

ensemble = RegressionEnsembleModel([lgbm_model, linreg_model],
                                   regression_train_n_points=140,
                                   regression_model=ensemble_lin_reg
                                   )

ensemble.fit(train)
pred = ensemble.predict(len(val), num_samples=1000)

val.plot(label="test")
pred.plot(label="prediction")
plt.show()
```
aggreg_probabilistic

madtoinou:
Extended the feature: if the `forecasting_models` are probabilistic, they can be sampled to generate the training dataset for the regression model (instead of just taking the first sample). This is controlled with the `regression_train_num_samples` argument.

These samples are then reduced component-wise (one for each forecasting model) using either the `mean` or a quantile (using the `regression_train_samples_reduction` argument) before being used to train the regression model.

Also, `NaiveEnsemble` can generate probabilistic forecasts if all its `models` are probabilistic; samples are averaged across the models/components and the n_samples dimension remains untouched.

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -19,6 +19,9 @@
 
 logger = get_logger(__name__)
 
+# arbitrary threshold to raise warning for probabilistic forecasting models
+SAMPLES_WARNING_THRESHOLD = 1e12
```
dennisbader, line 23: Isn't this very high?
I would already start at 1e6, what do you think?

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,8 +88,56 @@ def __init__(
...
+            logger,
+        )
+
+        if (
+            regression_train_num_samples
+            * regression_train_n_points
+            * len(forecasting_models)
+            > SAMPLES_WARNING_THRESHOLD
+        ):
+            logger.warning(
```
dennisbader, line 130: Would be informative to add the actual number of generated samples in the message

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -148,10 +219,44 @@ def fit(
...
+        # support deterministic forecasting model with probabilistic ensembling
+        predictions = self._make_multiple_predictions(
+            n=n,
+            series=series,
+            past_covariates=past_covariates,
+            future_covariates=future_covariates,
+            num_samples=self.regression_train_num_samples,
+        )
+
+        if predictions[0].n_samples > 1:
```
dennisbader, line 250: I believe we can move `predict()` and `_predictions_reduction()` back to the base `EnsembleModel`.
- For the naive ensemble, we can set the `regression_train_samples_reduction=None` in the `__init__()` body. Then when calling super().predict() we set num_samples=1.

darts/models/forecasting/baselines.py
```diff
@@ -196,15 +196,49 @@ def fit(
 
         return self
 
+    def predict(
```
dennisbader, line 199: as mentioned in the other comment, we can generalize predict() and move it back to the EnsembleModel

darts/models/forecasting/baselines.py
```diff
@@ -196,15 +196,49 @@ def fit(
...
-            series.name = prediction.components[0]
-            return TimeSeries.from_series(series)
+            # average across the components, keep n_samples, rename components
+            # NOTE: could use `with_columns_renamed()` instead
+            return TimeSeries.from_times_and_values(
+                times=prediction.time_index,
+                values=prediction.mean(axis=1).all_values(),
+                freq=prediction.freq,
+                columns=[prediction.components[0]],
+            )
```
dennisbader, line 241:
```suggestion
            return prediction.mean(axis=1).with_columns_renamed(
                "components_mean", prediction.components[0]
            )
```

darts/tests/models/forecasting/test_ensemble_models.py
```diff
@@ -122,17 +121,54 @@ def test_predict_ensemble_local_models(self):
...
+            5, num_samples=num_samples
+        ) + 0.5 * model_alone_2.predict(5, num_samples=num_samples)
+
+        self.assertEqual(
+            forecast_mean.values().shape, pred_proba_many_sample.values().shape
+        )
+        self.assertEqual(forecast_mean.n_samples, pred_proba_many_sample.n_samples)
+        self.assertTrue(
+            np.array_equal(
+                pred_proba_many_sample.mean().values(), forecast_mean.mean().values()
```
dennisbader, line 169: could we check all values instead of the mean?

dennisbader (changes requested):
Very nice @madtoinou, thanks a lot 🚀 👍 

There are one or two points where we could simplify a bit.

Regarding the non-reduction of the samples, I thought more about stacking the points vertically to generate more rows, instead of adding the dimensions as columns.

Maybe we can discuss this again?

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -363,3 +363,188 @@ def test_call_backtest_regression_ensemble_local_models(self):
...
+            regression_train_num_samples=500,
+            regression_train_samples_reduction=0.5,
+        )
+
+        train, val = self.ts_sum1.split_after(0.9)
+        ensemble_model_none.fit(train)
+        ensemble_model_median.fit(train)
+        ensemble_model_0_5_quantile.fit(train)
+
+        pred_none_training = ensemble_model_none.predict(len(val))
```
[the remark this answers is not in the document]
madtoinou: It should not since the regression model is deterministic

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -179,3 +287,26 @@ def extreme_lags(
...
+    def _is_probabilistic(self) -> bool:
+        """
+        A RegressionEnsembleModel is probabilistic if its regression
+        model is probabilistic (ensembling layer)
+        """
+        return self.regression_model._is_probabilistic()
+
+    def _predictions_reduction(self, predictions: TimeSeries) -> TimeSeries:
+        """"""
+        if self.regression_train_samples_reduction is None:
```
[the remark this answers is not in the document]
madtoinou: After discussion, the vertical stacking will eventually be implemented in a separate PR.

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -47,6 +52,20 @@ def __init__(
...
             Any regression model with ``predict()`` and ``fit()`` methods (e.g. from scikit-learn)
             Default: ``darts.model.LinearRegressionModel(fit_intercept=False)``
+
+            .. note::
+                if `regression_model` is probabilistic, the `RegressionEnsembleModel` will also be probabilistic.
+            ..
+        regression_train_num_samples
+            Number of times a prediction is sampled from each forecasting models to train the regression
+            model (samples are averaged). Should be left set to 1 for deterministic models. Default: 1.
+        regression_train_samples_reduction
```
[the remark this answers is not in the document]
madtoinou: In backtest, `reduction` is applied to an array directly. In order to preserve the time index of the TimeSeries while allowing numpy callable, we would have to create a new TimeSeries with `from_times_and_values` after applying the callable to the TimeSeries.all_values() array.

WDYT?

darts/models/forecasting/regression_ensemble_model.py
madtoinou: Moved `predict()` and  `_predictions_reduction()` back to `EnsembleModel`, which now accept two new arguments in it constructor (`train_num_samples` and `train_samples_reduction`). They might be reusable during implemention of new ensemble training techniques such as boosting but are ignored in "single level" models such as `NaiveEnsemble`.

`predict()` relies on the value of the `reduction_method` attribute to know if it should use `num_samples` directly (to have a probabilistic forecast with `NaiveEnsemble`) or `regression_train_n_samples` so that the `future_covariates` passed to the `RegressionEnsembleModel` regressor have the dimensions seen during the training.

darts/models/forecasting/ensemble_model.py
```diff
@@ -26,10 +27,24 @@ class EnsembleModel(GlobalForecastingModel):
     ----------
     models
         List of forecasting models whose predictions to ensemble
+
+        .. note::
+                if all the models are probabilistic, the `EnsembleModel` will also be probabilistic.
```
dennisbader, line 32: This is true for naive ensemble but not for RegressionEnsembleModel, right?

darts/models/forecasting/ensemble_model.py
```diff
@@ -26,10 +27,24 @@ class EnsembleModel(GlobalForecastingModel):
...
+        ..
+    train_num_samples
+        Number of prediction samples from each forecasting model for multi-level ensembles. The n_samples
+        dimension will be reduced using the `train_samples_reduction` method.
+    train_samples_reduction
+        If `forecasting models` are probabilistic and `train_num_samples` > 1, method used to
```
dennisbader, line 38:
```suggestion
        If `models` are probabilistic and `train_num_samples` > 1, method used to
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
             f"{regression_model.lags}",
         )
 
+        raise_if(
+            regression_train_num_samples > 1 and not self._models_are_probabilistic(),
```
dennisbader, line 96: Couldn't we accept a mix of probabilistic and deterministic models?
In the end we reduce the probabilistic base models anyway to a deterministic prediction so I think it's fair to use both kinds.
For all deterministic models we'd set num_samples dynamically to 1
WDYT?

darts/models/forecasting/regression_ensemble_model.py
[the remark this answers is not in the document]
dennisbader: Good point, we can leave it for now like it is

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
             f"{regression_model.lags}",
         )
 
+        raise_if(
```
dennisbader, line 95: Can we move all those tests to the base class?

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
...
+            logger,
+        )
+
+        # check the reduction method
+        supported_reduction = ["mean", "median"]
+        if isinstance(regression_train_samples_reduction, float):
+            # this is already checked by `ts.quantile()`, maybe too redundant
+            raise_if(
+                regression_train_samples_reduction > 1.0
+                or regression_train_samples_reduction < 0,
```
dennisbader, line 108:
```suggestion
            raise_if_not(
                0.0 < regression_train_samples_reduction < 1.0,
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
...
+                f"`regression_train_samples_reduction` should be comprised between "
+                f"0 and 1 ({regression_train_samples_reduction}).",
```
dennisbader, line 110:
```suggestion
                f"if a float, `regression_train_samples_reduction` must be between "
                f"0 and 1, received ({regression_train_samples_reduction})",
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
...
+                logger,
+            )
+        elif isinstance(regression_train_samples_reduction, str):
+            raise_if(
+                regression_train_samples_reduction not in supported_reduction,
+                f"`regression_train_samples_reduction` should be one of {supported_reduction}, "
+                f"received ({regression_train_samples_reduction})",
```
dennisbader, line 117:
```suggestion
                f"if a string, `regression_train_samples_reduction` must be one of {supported_reduction}, "
                f"received ({regression_train_samples_reduction})",
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
...
+                logger,
+            )
+        else:
+            logger.exception(
```
dennisbader, line 121: use `raise_log(ValueError(...))` instead

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -69,6 +92,52 @@ def __init__(
...
+                f"`regression_train_samples_reduction` type not supported "
+                f"({regression_train_samples_reduction}). Must be `float` "
+                f" or one of {supported_reduction}."
+            )
+
+        if (
+            regression_train_num_samples
+            * regression_train_n_points
+            * len(forecasting_models)
+            > SAMPLES_WARNING_THRESHOLD
```
dennisbader, line 131: This was only required for the vertical stacking of samples and can be removed now

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -19,6 +19,9 @@
 
 logger = get_logger(__name__)
 
+# arbitrary threshold to raise warning for probabilistic forecasting models
```
dennisbader, line 22: can be removed as part of the vertical stacking which we dropped now

dennisbader (changes requested):
Nice, thanks a lot ! :) 
We're close! Mainly added a couple of minor suggestions. I think we could allow a mix of deterministic and probabilistic `forecasting_models` for `RegressionEnsembleModel` as internally we reduce the probabilistic ones anyway to a deterministic forecast. WDYT?

darts/models/forecasting/ensemble_model.py
madtoinou: Correct, the docstring is different in `RegressionEnsembleModel`. This note could probably be removed since `EnsembleModel` cannot be instantiated anyway.

darts/models/forecasting/regression_ensemble_model.py
madtoinou: I agree with you, if reduction occurs, then the mix should be accepted

darts/models/forecasting/regression_ensemble_model.py
madtoinou: Sure, there will be a small discrepancies since the name of the argument/attributes are slightly different between the two classes (`regression_` is used as a prefix in `RegressionEnsembleModel`)

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -363,3 +363,228 @@ def test_call_backtest_regression_ensemble_local_models(self):
...
+        self.assertTrue(ensemble_allproba._is_probabilistic())
+        ensemble_allproba.fit(self.ts_random_walk[:100])
+        # probabilistic forecasting is supported
+        pred = ensemble_allproba.predict(5, num_samples=10)
+        self.assertEqual(pred.n_samples, 10)
+
+        # forecasting models are a mix of probabilistic and deterministic, probabilistic regressor
+        ensemble_mixproba = RegressionEnsembleModel(
+            forecasting_models=[
+                self.get_probabilistic_global_model([-1, -3], quantiles),
```
dennisbader, line 398: can you check also mixed proba model with train_num_samples > 1 and a reduction?

dennisbader (approved):
Great work @madtoinou, thanks a lot! 🚀 

One more test about the mixed proba model and we're good to go!

darts/tests/models/forecasting/test_regression_ensemble_model.py
madtoinou: It fails because it cannot stack `TimeSeries` on the component axis if they don't have the same number of samples. We can either add an exception to prevent this scenario or apply the reduction inside `_make_multiple_prediction`. WDYT?

madtoinou: had to move the reduction to `_make_multiple_prediction()` so that they could be stacked properly

dennisbader (approved):
Perfect 💯 Great work @madtoinou!

PR #1693, opened 2023-04-04
Fix/start value for historical_forecasts in examples

aarondt1:
- Update start value from 0.6 to 0.75 to match the text above the cell.
- Fixing typos

The text above the cell reads: 

> We can also backtest the models using covariates. Say for instance we are interested in evaluating the running accuracy with a horizon of 12 months, starting at 75% of the air series

This commit resolves the issue where the start value was incorrectly set to 0.6 instead of the intended 0.75.

### DIFF
--- a/examples/01-multi-time-series-and-covariates.ipynb
+++ b/examples/01-multi-time-series-and-covariates.ipynb
@@ -758,13 +758,14 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
     "\n",
     "\n",
     "### Backtesting with Covariates\n",
-    "We can also backtest the models using covariates. Say for instance we are interested in evaluating the running accuracy with a horizon of 12 months, starting at 75% of the air series:"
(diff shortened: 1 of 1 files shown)


dennisbader (changes requested):
Thanks for this 🚀  

Could you rather change the text to 60% so that we don't have to rerun the example notebook?

aarondt1:
> Thanks for this 🚀
> 
> Could you rather change the text to 60% so that we don't have to rerun the example notebook?

Sure thing!

examples/01-multi-time-series-and-covariates.ipynb
```diff
@@ -758,13 +758,14 @@
    ]
   },
   {
+   "attachments": {},
```
dennisbader, line 761:
```suggestion
```

examples/01-multi-time-series-and-covariates.ipynb
```diff
@@ -848,6 +849,7 @@
...
```
dennisbader, line 852:
```suggestion
```

dennisbader (approved):
LGTM, thanks!

PR #1705, opened 2023-04-12
Update timeseries.py to idiomatically return fig when calling ts.plot()

jerry-pca:
Idiomatically return fig

Fixes #.


dennisbader:
closing this as stale

PR #1719, opened 2023-04-20
Change TimeSeries return type to Self

### LINKED ISSUE #1361: Annotate TimeSeries method returns with Self type
Many TimeSeries-methods are annotated with a return type `-> "TimeSeries"` even though they use `self` or `self.__class__`, e.g. `head`:
```python
# timeseries.py
class TimeSeries:
    ...
    def head(
        self, size: Optional[int] = 5, axis: Optional[Union[int, str]] = 0
    ) -> "TimeSeries":
```
This makes it difficult to subclass `TimeSeries`, since that annotation prevails:
```python
import inspect
class Child(TimeSeries):
    pass

print(inspect.get_annotations(Child.head))
{'size': typing.Optional[int], 'axis': typing.Union[int, str, NoneType], 'return': 'TimeSeries'}
```
even though `Child.head` actually returns an instance of `Child`.

If one would change `head` (and the other `TimeSeries` methods) to be annotated with `typing.Self`, then this would not be an issue.

connesy:
Fixes #1361 

### Summary

When a `TimeSeries` method returns an instance of its own class, the annotated return type is changed from `"TimeSeries"` to `typing.Self` (or `typing_extensions.Self` if python_version <= 3.10).

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -2308,9 +2308,7 @@ class TimeSeries:
             idx = pd.RangeIndex(effective_start_ts, effective_end_ts, step=self.freq)
             return self[idx]
 
-    def slice_n_points_after(
-        self, start_ts: Union[pd.Timestamp, int], n: int
-    ) -> Self:
+    def slice_n_points_after(self, start_ts: Union[pd.Timestamp, int], n: int) -> Self:
         """
         Return a new TimeSeries, starting a `start_ts` (inclusive) and having at most `n` points.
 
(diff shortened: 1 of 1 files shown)


dennisbader:
Hi @connesy and thanks for this PR. There seems to be a linting issue from black with `darts/timeseries.py`. Could you try formatting the file  and pushing the changes?

connesy:
@dennisbader Should be fixed now.

PR #1744, opened 2023-05-04
Fix/load weights from ckpt with encoder or likelihood

### LINKED ISSUE #1725: [BUG] TorchForecastingModel load_weights_from_checkpoint fails to load
**Describe the bug**
When attempting to finetune a TFT model, I've run into a coulple issues regarding the `load_weights_from_checkpoint` and was unsure if I'm misusing one/both of these, or if there is any workarounds. 

1. If you include a likelihood model (e.g. QuantileRegression), when calling `load_weights_from_checkpoint` it complains of the checkpoint and instantiated value being different. In the below case it is fine since the likelihood could be set to None to achieve the same effect, but was wondering there is a better way to go about this?

2. Another issue is when using `add_encoders`, we get a `ValueError: The provided 'past_covariates' must have equal dimensionality as the 'past_covariates' used for training the model.` This doesn't occur if we don't include anything in encoders field (or complains of future_covariates if we include these as "future").

**To Reproduce**
```python
import torch
import pandas as pd
import numpy as np

from darts import TimeSeries
from darts import models

# create fake dataset
series_air = AirPassengersDataset().load().astype(np.float32)
past_covs = series_air
future_covs = series_air
target_series = series_air

tft_accelerator = (
    {"accelerator": "gpu", "devices": -1, "auto_select_gpus": True}
    if torch.cuda.is_available()
    else {"accelerator": "cpu"}
)
def create_model(model_name: str):
    return models.TFTModel(
        input_chunk_length=4,
        output_chunk_length=4,
        add_encoders={                        # <--- error when included
            "cyclic": {"past": ["month"]},
            "datetime_attribute": {
                "past": ["hour"],
            },
            "transformer": Scaler(),
        },
        # likelihood=likelihood_models.QuantileRegression(), # <--- error if uncommented
        n_epochs=10,
        model_name=model_name,
        work_dir="models",
        force_reset=True,
        save_checkpoints=True,
        pl_trainer_kwargs={**tft_accelerator},
    )

model = create_model("TFTModel")
model.fit(
    target_series, 
    past_covariates=past_covs,
    future_covariates=future_covs,
)

model = create_model("TFTModel_finetune")
model.load_weights_from_checkpoint(
    model_name="TFTModel",
    work_dir="models",
    best=False,
)

model.historical_forecasts(
    target_series, 
    past_covariates=past_covs,
    future_covariates=future_covs,
    start=0.9,
    retrain=False,
)
```

**Expected behavior**
Both likelihood models, and add_encoders should be allowed when using `load_weights_from_checkpoint`. 

 - Python version: 3.9.16
 - darts version 0.24.0

madtoinou:
Fixes #1725

### Summary

- Add the new argument `load_encoders` (default: `True`) to `load_weights()` and `load_weights_from_checkpoints()` so that the model can run inference directly after loading the weights with the encoders. 
  - If `load_encoders=True`, the model loading the weights must be instantiated without encoder or with the same encoders, otherwise an exception is raised.
  - If `load_encoders=False` and the weights were trained with encoders, the model loading the weights must already be instantiated with encoders (identical or different than the ones used during the first training, must be careful about the dimensions) otherwise an exception is raised.
- Add `__eq__` operator to the `LikelihoodModel` (checking their numerical parameters only) and `DataTransformer` to prevent weights loading failure occurring during the hyper-parameters sanity check.

### Other Information

I don't know if it makes sense but we could eventually exclude the likelihood attribute from the hyper-parameter check when loading the weights in case some user want to fine-tune/retrain the model with a different likelihood.


madtoinou:
Extended the PR: 
- the new encoders are instantiated in `load_weights_from_checkpoint()` to allow for direct inference if they don't need to be fitted (no `transformer` for example). 
- some sanity check about the number of components generated by the encoders to generate an informative exception if the new encoders don't match the checkpoint ones (in case user want to use different encoders during fine-tuning).

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -582,3 +582,13 @@ def __str__(self):
 
     def __repr__(self):
         return self.__str__()
+
+    def __eq__(self, other) -> bool:
```
dennisbader, line 586: Could you add some comments here, it's difficult to understand what the equality here means.

for example, when I expected the below test to run fine, but it fails:
```

sc1 = Scaler()
sc2 = Scaler()
sc1.fit_transform(series)
sc2.fit_transform(series)

assert sc1 == sc2
```

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +219,215 @@ def test_manual_save_and_load(self):
...
+            os.mkdir(checkpoint_path_manual)
+            checkpoint_file_name = "checkpoint_0.pth.tar"
+            model_path_manual = os.path.join(
+                checkpoint_path_manual, checkpoint_file_name
+            )
+
+            # define encoders sets
+            encoders_past = {
+                "datetime_attribute": {"past": ["day"]},
+                "transformer": Scaler(),
```
dennisbader, line 263: can the user also load weights with `load_encoders=False` and the encoders use a different transformer than used when fitting the model?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1762,9 +1767,14 @@ def load_weights_from_checkpoint(
         # checkpoints generated by PL, prefix is defined in TorchForecastingModel __init__()
         if file_name[:5] == "last-" or file_name[:5] == "best-":
             checkpoint_dir = _get_checkpoint_folder(work_dir, model_name)
+            model_ckpt_dir = _get_runs_folder(work_dir, model_name)
+            model_ckpt_file_name = INIT_MODEL_NAME
         # manual save
         else:
             checkpoint_dir = os.getcwd()
```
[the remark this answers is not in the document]
madtoinou: No, for automatic checkpoints, the PL checkpoints are nested in a "checkpoints" folder whereas the TFM checkpoint ("_model.pth.tar") is at the root:
```
examples/darts_logs
├── dlin
│   ├── _model.pth.tar
│   └── checkpoints
│       └── last-epoch=19.ckpt
└── rnn
    ├── _model.pth.tar
    └── checkpoints
        └── last-epoch=19.ckpt
```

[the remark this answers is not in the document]
madtoinou: You're right!

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1762,9 +1767,14 @@ def load_weights_from_checkpoint(
         # checkpoints generated by PL, prefix is defined in TorchForecastingModel __init__()
         if file_name[:5] == "last-" or file_name[:5] == "best-":
             checkpoint_dir = _get_checkpoint_folder(work_dir, model_name)
+            model_ckpt_dir = _get_runs_folder(work_dir, model_name)
```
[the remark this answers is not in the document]
madtoinou: renamed it to `tfm_save_[...]`, WDYT?

darts/utils/likelihood_models.py
```diff
@@ -171,6 +171,13 @@ def num_parameters(self) -> int:
         """
         pass
 
+    @abstractmethod
+    def __eq__(self, other) -> bool:
+        """
+        Define (in)equality between two likelihood objects
+        """
+        pass
+
```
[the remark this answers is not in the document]
madtoinou: Attributes inheriting from torch.nn.Module need to be excluded because for example `nn.Softmax() == nn.Softmax()` returns `False`. Let me know what you think.

darts/dataprocessing/transformers/base_data_transformer.py
madtoinou: The equality is now comparing the `_fitted_params` entries: either directly np.ndarrays or all the np.ndarray attributes of the object (sklearn scalers)

darts/models/forecasting/torch_forecasting_model.py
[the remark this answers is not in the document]
dennisbader: sounds good 👍

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +223,337 @@ def test_manual_save_and_load(self):
...
+            # model with undeclared encoders
+            model_no_enc = create_DLinearModel("no_encoder", add_encoders=None)
+            # weights were trained with encoders, new model must be instantiated with encoders
+            with self.assertRaises(ValueError):
+                model_no_enc.load_weights_from_checkpoint(
+                    auto_name,
+                    work_dir=self.temp_work_dir,
+                    best=False,
+                    load_encoders=False,
+                    map_location="cpu",
```
dennisbader, line 320: you don't need to map to cpu here because we're always on cpu.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +223,337 @@ def test_manual_save_and_load(self):
...
+                )
+            # overwritte undeclared encoders
+            model_no_enc.load_weights_from_checkpoint(
+                auto_name,
+                work_dir=self.temp_work_dir,
+                best=False,
+                load_encoders=True,
+                map_location="cpu",
+            )
+            model_no_enc.to_cpu()
```
dennisbader, line 330: this is not necessary as we're always on cpu in test env.
I added this only to one test that we can check the method in general, but it's not required in all tests.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +223,337 @@ def test_manual_save_and_load(self):
...
+            model_other_enc_noload = create_DLinearModel(
+                "other_encoder_noload", add_encoders=encoders_other_past
+            )
+            model_other_enc_noload.load_weights(
+                model_path_manual, load_encoders=False, map_location="cpu"
+            )
+            model_other_enc_noload.to_cpu()
+            self.assertEqual(model_other_enc_noload.add_encoders, encoders_other_past)
+            # new encoders were instantiated
+            self.assertNotEqual(model_other_enc_noload.encoders, None)
```
dennisbader, line 383: better to check if it is a SequentialEncoder

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +223,337 @@ def test_manual_save_and_load(self):
...
+            # model with same encoders but no scaler (non-fittable)
+            model_new_enc_noscaler_noload = create_DLinearModel(
+                "same_encoder_noscaler", add_encoders=encoders_past_noscaler
+            )
+            model_new_enc_noscaler_noload.load_weights(
+                model_path_manual, load_encoders=False, map_location="cpu"
+            )
+            model_new_enc_noscaler_noload.to_cpu()
+            self.assertNotEqual(
+                model_new_enc_noscaler_noload.add_encoders, encoders_past
```
dennisbader, line 401: this test is redundant with the next test

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +223,337 @@ def test_manual_save_and_load(self):
...
+            # predictions are identical when using the same likelihood
+            self.assertTrue(np.array_equal(pred_auto.values(), pred_manual.values()))
+
+            # model with identical likelihood
+            model_same_likelihood = create_DLinearModel(
+                "same_likelihood", likelihood=GaussianLikelihood(prior_mu=0.5)
+            )
+            model_same_likelihood.load_weights(model_path_manual, map_location="cpu")
+            model_same_likelihood.to_cpu()
+            model_same_likelihood.predict(n=4, series=self.series)
```
dennisbader, line 525: We can check that `load_weights_from_checkpoint` and `load_weights` have equal predictions.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +223,337 @@ def test_manual_save_and_load(self):
...
+            # cannot check predictions since this model is not fitted, random state is different
+
+            # model with no likelihood
+            model_no_likelihood = create_DLinearModel("no_likelihood", likelihood=None)
+            # state_dict does not have the proper dimensions, expection raised by torch.load
+            with self.assertRaises(RuntimeError):
```
dennisbader, line 531: Shouldn't this raise an Error from Darts side? likelihood=None != likelihood=GaussianLikelihood() (during the parameter checks)

darts/dataprocessing/transformers/fittable_data_transformer.py
dennisbader, line 1: The more I think about this, I'm not sure that we should add the `__eq__` here and istead just test that the transformer in `add_encoders` is from the same class. What do you think?

darts/dataprocessing/transformers/base_data_transformer.py
dennisbader, line 1: (same here as my comment on fittable transformers)

dennisbader (changes requested):
Very nice, thanks :) 🚀 

Soon ready to be merged!
Only had some minor suggestions, and that it might be better to avoid the `__eq__()` and rather do a simple check when comparing `add_encoders` for the same transformer class.

darts/tests/models/forecasting/test_torch_forecasting_model.py
madtoinou: The Likelihood object do not have the same `random_state` (also mentioned in #1779), making the predictions different. Should I implement a method to set it before updating this test?

darts/tests/models/forecasting/test_torch_forecasting_model.py
madtoinou: Yup, the check in `load_weights_from_checkpoint()` was skipping the `None` attributes. Solved it!

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1892,6 +1923,94 @@ def _is_probabilistic(self) -> bool:
...
+            )
+
+            new_add_encoders: Dict = copy.deepcopy(tfm_save.add_encoders)
+            new_encoders: SequentialEncoder = copy.deepcopy(tfm_save.encoders)
+        else:
+            raise_if(
+                len(tfm_save.add_encoders) > 0 and self.add_encoders is None,
+                "Model was created without encoders and encoders were not loaded, but "
+                "the weights were trained using encoders. Either set `load_encoders` to "
+                "`True` or add a matching `add_encoders` dict at model creation.",
```
dennisbader, line 1976: maybe give here the `tfm_save.add_encoders` so that user knows what the required encoders are?

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +224,349 @@ def test_manual_save_and_load(self):
...
+                )
+            # overwritte undeclared encoders
+            model_no_enc.load_weights_from_checkpoint(
+                auto_name,
+                work_dir=self.temp_work_dir,
+                best=False,
+                load_encoders=True,
+            )
+            # equality between data transformers is not well defined
+            self.assertEqual(
```
dennisbader, line 330: the comparison of encoders and transformers can be put into a helper function as it occurs 3 times

darts/tests/models/forecasting/test_torch_forecasting_model.py
dennisbader: see my suggestion below

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -216,6 +224,349 @@ def test_manual_save_and_load(self):
...
+            # predictions are identical when using the same likelihood
+            self.assertTrue(np.array_equal(pred_auto.values(), pred_manual.values()))
+
+            # model with identical likelihood
+            model_same_likelihood = create_DLinearModel(
+                "same_likelihood", likelihood=GaussianLikelihood(prior_mu=0.5)
+            )
+            model_same_likelihood.load_weights(model_path_manual)
+            model_same_likelihood.predict(n=4, series=self.series)
+            # cannot check predictions since this model is not fitted, random state is different
```
dennisbader, line 545: I meant that a model from `load_weights_from_checkpoint` should have identical predictions as a model from `load_weights`.

Something like below:

```suggestion
            model_manual_same_likelihood = create_DLinearModel(
                "same_likelihood", likelihood=GaussianLikelihood(prior_mu=0.5)
            )
            model_manual_same_likelihood.load_weights(model_path_manual)
            preds_manual_from_weights = model_manual_same_likelihood.predict(n=4, series=self.series)

            # check that weights from checkpoint give identical predictions as weights from manual save
            model_auto_same_likelihood = create_DLinearModel(
                "same_likelihood", likelihood=GaussianLikelihood(prior_mu=0.5)
            )
            model_auto_same_likelihood.load_weights_from_checkpoint(
                auto_name,
                work_dir=self.temp_work_dir,
                best=False,
            )
            preds_auto_from_weights = model_auto_same_likelihood.predict(n=4, series=self.series)
            self.assertTrue(preds_manual_from_weights == preds_auto_from_weights)
```

dennisbader (changes requested):
Thanks a lot, it's very nice 🚀 
A last iteration with some minor suggestions and then we're good to go 👍

dennisbader (approved):
Great job @madtoinou 💯  Thanks a lot 🚀

PR #1745, opened 2023-05-05
Feat/local and global models in EnsembleModel

### LINKED ISSUE #737: EnsembleModel: Require all models to be global models only if the model is being used on multiple series (otherwise, allow a mix of local and global models)

### LINKED ISSUE #1038: Ensemble for global forecasting models
I want to use two consecutive global forecasting model - I need fast inference times-  back to back where the first predictions are treated as future covariates of the second model.

A global ensemble model, that takes a list of global forecasting model and takes another global forecasting model as the regressor, but it should have save and load functionality and interfaces for interacting with each model parameters.

An alternative could be that global forecasting model that accepts a future covariates accept a model with the time series, and if supplied it uses that global model to first predict the required length of future covariates and consume them behind the scenes.

**Additional context**
We need a feature to enable us to use [stacking, boosting] concepts with global forecasting models as future covariates.

### LINKED ISSUE #1492: Allow RegressionEnsembleModel to contain both Local and Global models
From gitter thread: 

"I have multiple univariate target TimeSeries. I want to train models on different "levels" - e.g. some local models on the univariate TimeSeries and some global models on a higher level, combining multiple TimeSeries. Then I'd like to use learned ensembling to combine those models' predictions in a smart way."

The EnsembleModel constructor should accept combination of Local and Global model. This will require some checks to make sure they are compatible (fit method already take care of raising exception for the training on multiple series and past-covariates functionnalities).

Cannot think of another approach...

madtoinou:
Fixes #1492, fixes #737, fixes #1038.

### Summary

`EnsembleModel.models` now accept a mix of `Local` and `GlobalForecastingModel`. If the ensemble contains at least one local model, it only supports single `TimeSeries` training/inference. 

If past/future covariates are provided, they will be passed only the models supporting them (not enforcing same covariates for all the models). I don't know if this is desirable or if all the models should support the exact same covariates in order to belong to the same `EnsembleModel`?

### Other Information

I wonder if we should also allow pre-trained `GlobalForecastingModel` in `EnsembleModel` with a flag to bypass the exception currently raised (and make sure the user is aware of the implications). This could be quite powerful but would open the door for data leakage that could be difficult to detect:
- global model pre-trained with the same target series but too many timestamps : detectable since the ts is stored
- global model pre-trained with multiple series, including the target series (to eventually bypass the limitation of the training with a single ts in the `RegressionEnsembleModel`) : undetectable


darts/models/forecasting/ensemble_model.py
```diff
@@ -22,31 +23,41 @@ class EnsembleModel(GlobalForecastingModel):
...
+            all(
+                [
+                    local_model or global_model
+                    for local_model, global_model in zip(
+                        is_local_model, is_global_model
+                    )
+                ]
+            ),
+            "Cannot instantiate EnsembleModel with forecasting models that are neither local, "
+            "nor global. Also, please make sure that all the models in `models` are instantiated.",
```
dennisbader, line 60:
```suggestion
            "All models must be of type `GlobalForecastingModel`, or `LocalForecastingModel`. "
            "Also, make sure that all models in `models` are instantiated.",
```

darts/models/forecasting/ensemble_model.py
```diff
@@ -229,3 +257,26 @@ def find_max_lag_or_none(lag_id, aggregator) -> Optional[int]:
 
     def _is_probabilistic(self) -> bool:
         return all([model._is_probabilistic() for model in self.models])
+
+    def _full_past_covariates_support(self) -> bool:
```
dennisbader, line 261: I think we should also adapt properties `supports_past_covariates` and `support_future_covariates` for ensemble models:

support X covariates if any of the underlying models supports X covariates.
WDYT?

darts/models/forecasting/ensemble_model.py
```diff
@@ -229,3 +257,26 @@ def find_max_lag_or_none(lag_id, aggregator) -> Optional[int]:
...
+
+    def _full_future_covariates_support(self) -> bool:
+        return all([model.supports_future_covariates for model in self.models])
+
+    def _logger_info_covariates_handling(
+        self,
+        past_covariates: Union[TimeSeries, Sequence[TimeSeries]],
+        future_covariates: Union[TimeSeries, Sequence[TimeSeries]],
+    ):
+        if past_covariates is not None and not self._full_past_covariates_support():
```
dennisbader, line 272: Should we just improve the documentation of the models instead of logging the info?
I'm always a bit careful with verbosity in fit/predict due to historical_forecasting spamming with prints :D 
Or at least avoid that it gets printed infinitely.

darts/models/forecasting/theta.py
```diff
@@ -40,7 +40,7 @@ def __init__(
 
         `season_mode` must be a ``SeasonalityMode`` Enum member.
 
-        You can access the Enum with ``from darts import SeasonalityMode``.
+        You can access the Enum with ``from darts.utils.utils import SeasonalityMode``.
```
dennisbader, line 43: 👍

darts/tests/models/forecasting/test_ensemble_models.py
dennisbader, line 1: Can we test the new covariates support as well?

dennisbader (changes requested):
Also a very nice addition, thanks :) 🚀 

Just had some minor comments.

Regarding the pre-trained models, I'm with you on this one and think we shouldn't go with this for now.

darts/models/forecasting/ensemble_model.py
madtoinou: Added the docstring about past/future covariates usage to the `NaiveEnsemble` and `RegressionEnsemble` to make it more visible in the doc.
I also moved the logger messages to the constructor to avoid spam in historical forecast (left it as `info` to avoid an irrelevant warning for users not using past/future covariates)

darts/tests/models/forecasting/test_ensemble_models.py
```diff
@@ -183,8 +189,35 @@ def test_call_predict_global_models_multivariate_input_with_covariates(self):
...
+        # ensemble is neither local, nor global
+        self.assertFalse(naive_ensemble.is_local_ensemble)
+        self.assertFalse(naive_ensemble.is_global_ensemble)
+
+        # ensemble contains one local model, no support for multiple ts fit
+        with self.assertRaises(ValueError):
+            naive_ensemble.fit([self.series1, self.series2])
+
+    @unittest.skipUnless(TORCH_AVAILABLE, "requires torch")
+    def test_mixed_models_with_covariates(self):
```
dennisbader, line 203: Using regression models with different covariates raise an error. 
e.g. take one regression model with lags_past_covariates, and another one with lags_future_covariates.
Now pass both past_covariates and future_covariates to the ensemble model's fit/predict. 

```
from darts.models import NaiveEnsembleModel


m_pc = LinearRegressionModel(lags=1, lags_past_covariates=[-1])
m_fc = LinearRegressionModel(lags=1, lags_future_covariates=[-1])
model = NaiveEnsembleModel(models=[m_pc, m_fc])
model.fit(series=series, past_covariates=series, future_covariates=series)
model.predict(n=1, series=series, past_covariates=series, future_covariates=series)
```
Maybe we need to update RegressionModel's properties `supports_past/future_covariates` to be based on what lags the user gave at model creation?

dennisbader (changes requested):
Nice, we're almost there 🚀 
Mainly minor suggestions, and a bug I found using regression models with different covariates support.

darts/models/forecasting/ensemble_model.py
```diff
@@ -71,34 +97,42 @@ def fit(
...
-            not self.is_global_ensemble and not isinstance(series, TimeSeries),
-            "The models are of type LocalForecastingModel, which does not support training on multiple series.",
+            not self.is_global_ensemble and not is_single_series,
+            "The models contain at least one LocalForecastingModel, which does not support training on multiple "
+            "series.",
             logger,
         )
         raise_if(
-            not self.is_global_ensemble and past_covariates is not None,
+            self.is_local_ensemble and past_covariates is not None,
```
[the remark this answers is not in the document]
madtoinou: Correct, the message is slightly more specific but could probable be removed

dennisbader (approved):
Everything looks great now! 💯  Thanks a lot @madtoinou 🚀

PR #1748, opened 2023-05-08
Add normalization to BlockRNNModel

### LINKED ISSUE #1649: Retrieve last hidden state for RNNModel and BlockRNNModel
The last hidden state is an encoding of the information contained in the time series and is useful to retrieve in some projects.

Add a method that retrieve the last hidden state.

JanFidor:
Fixes #1649.

### Summary
I've added normalization parameter to the BlockRNNModel, I've brainstormed how to do it for RNNModel and I couldn't come up with a way that wouldn't require some type of dynamic aggregation of the hidden states, so I decided to make the PR for BlockRNN for now.
I added two torch modules to simplify the rnn sequence, not sure if it's the cleanest way to implement it, but it's at least very readable.
### Other Information
I also added layer norm, because it was a simple addition and it seems to be the recommended normalization for RNNs. I also considered adding group normalization, but it would either need constant num_groups parameter or additional constructor parameter for BlockRNNModel


JanFidor:
Some of the tests were failing, I'll check if continues after merging develop. One of them was test_fit_predict_determinism() which after debugging turned out to fail for ARIMA model, It wasn't in a scope of this PR so I'm unsure what might have happened. Might be a problem with my local build, I'll wait and see what the github actions say

JanFidor:
I've been thinking whether adding batch norm makes sense in this case, as repeated rescaling would cause gradient explosion, the very thing LSTM / GRU were supposed to combat. I'm inclined to only allow layer normalization (maybe also group norm), so that users don't accidentally fall into that trap. Let me know if you think that would fit with darts design philosophy !

darts/utils/torch.py
```diff
@@ -112,3 +112,34 @@ def decorator(self, *args, **kwargs) -> T:
...
     return decorator
+
+
+class TemporalBatchNorm1d(nn.Module):
+    def __init__(self, feature_size) -> None:
+        super().__init__()
+        self.norm = nn.BatchNorm1d(feature_size)
+
+    def forward(self, input):
+        input = self._reshape_input(input)  # Reshape N L C -> N C L
```
madtoinou, line 123: This line is more about swapping axes that reshaping, I would instead use the corresponding torch function:
```suggestion
        # Reshape N L C -> N C L
        input = input.swapaxes(1,2)  
```

darts/utils/torch.py
```diff
@@ -112,3 +112,34 @@ def decorator(self, *args, **kwargs) -> T:
...
+        input = self.norm(input)
+        input = self._reshape_input(input)
```
madtoinou, line 125:
```suggestion
        input = input.swapaxes(1,2)
```

madtoinou (commented):
Hi @JanFidor, thank you again for contributing to darts.

Found this [article](https://arxiv.org/abs/1510.01378), applying batch norm only on the output (not the hidden state). I would be curious to see a benchmark of the BlockRNNModel with and without normalization, check if we observe similar results.

JanFidor:
Hi @madtoinou , quick update! I've read the paper and have an idea how to implement it. It might need a little bit of magic to get the time_step_index into the model input, but I think it should be doable, I'll let you know when I'll get everything running or if I stumble into some problem

JanFidor:
Quick update @madtoinou. I've been browsing through the codebase and wanted to get your thoughts on my planned approach. I think that the simplest approach would be to manually add a past encoder with static position, but that would require expanding IntegerIndexEncoder which only supports 'relative' for now. That said, I'm not sure at which point the Encoders are applied to the TS and this approach depends on it happening before TS are sliced for training.

JanFidor:
Hi again @madtoinou! I wanted to get your thoughts on my new idea for the implementation. I went back to the paper and found the mention of using the batch norms specifically when training. Wouldn't it just suffice to store `input_chunk_length` batch norms instead, which would be significantly easier, might actually be more inline with what the paper proposes and wouldn't require TS of same length for training? I'll go ahead with this idea and give another update once I make some basic benchmarks.

JanFidor:
It took some playing around but I think I managed to fix most of the git history

darts/models/forecasting/block_rnn_model.py
```diff
@@ -183,6 +185,61 @@ def forward(self, x_in: Tuple):
...
+        return nn.Sequential(*modules)
+
+    def _fc_layer(
+        self,
+        input_size: int,
+        num_layers_out_fc: list[int],
+        target_size: int,
+        normalization: str = None,
+    ):
+        if not num_layers_out_fc:
```
VascoSch92, line 222: I don't get this point here.

`num_layers_out_fc` is a list of integers correct?
Suppose `num_layers_out_fc = []`, then not `num_layers_out_fc is True.`
So why `num_layers_out_fc = []` ?

darts/models/forecasting/block_rnn_model.py
```diff
@@ -183,6 +185,61 @@ def forward(self, x_in: Tuple):
...
+            num_layers_out_fc = []
+
+        last = input_size
+        feats = []
+        for feature in num_layers_out_fc + [
```
VascoSch92, line 227: i will rather use the extend method for lists

darts/models/forecasting/block_rnn_model.py
```diff
@@ -183,6 +185,61 @@ def forward(self, x_in: Tuple):
...
+            self.output_chunk_length * target_size * self.nr_params
+        ]:
+            if normalization:
+                feats.append(self._normalization_layer(normalization, last))
+            feats.append(nn.Linear(last, feature))
+            last = feature
+        return nn.Sequential(*feats)
+
+    def _normalization_layer(self, normalization: str, hidden_size: int):
```
VascoSch92, line 236: if `normalization` is different from `batch` and `layer` the method return `None`. is this intended?

PR #1716, opened 2023-04-18
Allows Path objects as paths

felixdivo:
**NOTE:** I am aware that this is not finished yet.

Fixes no opened issue.

### Summary

Loading and saving a model should allow for [path objects](https://docs.python.org/3/library/pathlib.html) to be passed.

PR #1754, opened 2023-05-10
Add support for PathLike objects to model save() and load()

### LINKED ISSUE #1751: [FEATURE] Add suport for PathLike objects
The `save` and `load` functions don't support `PathLike` objects from the [pathlib module](https://docs.python.org/3/library/pathlib.html), which is a standard library feature and used widely across the ecosystem.
Both functions already allow `str` as well as `BinaryIO` as input and handle these gracefully:

```Python
if isinstance(path, str):
    # save the whole object using pickle
    with open(path, "wb") as handle:
        pickle.dump(obj=self, file=handle, **pkl_kwargs)
else:
    # save the whole object using pickle
    pickle.dump(obj=self, file=path, **pkl_kwargs)
```

However, since the above code assumes the input to be `BinaryIO` in the case of the `else` path, we get the following error when invoking with a `PathLike` object:

```Bash
TypeError: file must have a 'write' attribute
```

I propose extending above code and other instances of this behavior accordingly as follows:

```Python
if isinstance(path, (str, os.PathLike)):
    # save the whole object using pickle
    with open(path, "wb") as handle:
        pickle.dump(obj=self, file=handle, **pkl_kwargs)
else:
    # save the whole object using pickle
    pickle.dump(obj=self, file=path, **pkl_kwargs)
```

Decide to not support `PathLike`.
In this case, I suggest making the above code more robust to input that is neither `str` nor `BinaryIO` and raising an error in that case.

**Additional context**
There already exists a [PR](https://github.com/unit8co/darts/pull/1716) on the topic, but it went stale and has no issue for discussion attached yet.

sudrich:
Fixes #1751.

### Summary

Added support for `PathLike` objects to the `save()` and `load()` functions of `ForecastingModel`.
This is achieved by explicitly including `os.PathLike` as a possible input for the path.
In addition, a check was added that raises a `ValueError` for invalid input arguments.


madtoinou:
As mentioned in the associated issue, this is a duplicate of #1716.

While scrolling through the changes, I noticed that the new lines in the changelog are not in the right section; you edited the release 0.24.0 instead of adding a new bullet point in the unreleased section. Can you please address this?

sudrich:
> As mentioned in the associated issue, this is a duplicate of #1716.
> 
> While scrolling through the changes, I noticed that the new lines in the changelog are not in the right section; you edited the release 0.24.0 instead of adding a new bullet point in the unreleased section. Can you please address this?

My bad. Moved the entry to the correct section.

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -154,8 +155,9 @@ def test_save_load_model(self):
 
         for model in [ARIMA(1, 1, 1), LinearRegressionModel(lags=12)]:
             model_path_str = type(model).__name__
-            model_path_file = model_path_str + "_file"
-            model_paths = [model_path_str, model_path_file]
+            model_path_pathlike = pathlib.Path(model_path_str + "_pathlike")
```
dennisbader, line 158: extending this test for invalid read/write path would be nice as well

dennisbader (changes requested):
Thanks @sudrich, looks great 🚀 

My only suggestion is to extend the test whether an error is raised for invalid paths (now that it should raise an error).

darts/tests/models/forecasting/test_local_forecasting_models.py
sudrich: Done. I put it in an extra test though to avoid inflating the original test too much.

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -195,6 +204,21 @@ def test_save_load_model(self):
...
+        model.fit(self.ts_gaussian)
+
+        # Use a byte string as path (, which is not supported)
+        model_path_invalid = b"invalid_path"
+
+        # test save
+        self.assertRaises(ValueError, model.save, model_path_invalid)
+        
+        # test load
+        self.assertRaises(ValueError, type(model).load, model_path_invalid)
```
dennisbader, line 219:
```suggestion
        # test save
        with pytest.raises(ValueError):
            model.save(model_path_invalid)
        
        # test load
        with pytest.raises(ValueError):
            ARIMA.load(model_path_invalid)
```

dennisbader (commented):
Looks good, just had one suggestion for using pytest over unit test.
Could you fix the current linting issues as well?

darts/tests/models/forecasting/test_local_forecasting_models.py
sudrich: Done.

sudrich:
> Looks good, just had one suggestion for using pytest over unit test. Could you fix the current linting issues as well?

Weird. Both my local `black` as well as the pre-commit hook don't trigger for the file.

dennisbader (approved):
LGTM, thanks 🚀

PR #1757, opened 2023-05-10
fix the support_covariates feature for NLinear and DLinear models

### LINKED ISSUE #1756: [BUG] Nlinear support_covariates
**Describe the bug**
For NLinear and DLinear model, "support_future/past_covariates" properties return True when Shared weights=True, although they should return False

 - Python version: [e.g. 3.9.15]
 - darts version [e.g. 0.24.0]

**Additional context**
PR https://github.com/unit8co/darts/pull/1583 should make covariates compatible with shared weights. If this fgoes through the bug is irrelevant

Loudegaste:
Fixes #1756

### Summary
For NLinear and DLinear model, "support_future/past_covariates" properties return True when Shared weights=True, although they should return False

### Other Information

PR https://github.com/unit8co/darts/pull/1583 should make covariates compatible with shared weights. If this fgoes through the bug is irrelevant


darts/tests/models/forecasting/test_dlinear_nlinear.py
```diff
@@ -253,6 +253,15 @@ def _eval_model(
                 )
                 self.assertLessEqual(e1, 0.40)
                 self.assertLessEqual(e2, 0.34)
+            for model in [DLinearModel, NLinearModel]:
```
dennisbader, line 256: could you add a check that fitting the model with past covariates and shared_weights=True raises an error?
```suggestion
            # can only fit models with past covariates when shared_weights=False
            for model in [DLinearModel, NLinearModel]:
```

dennisbader (commented):
Thanks @Loudegaste, looks good to me. Could you add a check that the model raises an error when fitting with covariates and shared_weights=True?

dennisbader (approved):
Looks good, thanks @Loudegaste 🚀

PR #1759, opened 2023-05-11
Fix/ loading metrics and loss in load_from_checkpoint

### LINKED ISSUE #1758: Unable to resume training from check point with custom loss function in NBEATS
**Describe the bug**

I am training an NBEATS model with a custom loss function defined by myself. I train the model (set save_checkpoints = True) for a certain amount of epochs and then I load it again (load_from_checkpoint) to continue training, but when I do so, the loss function switches to the default MSELoss instead of continuing with my custom loss.

 - Python version: [e.g. 3.7,13]
 - darts version [e.g. 0.22.0]

madtoinou:
Fixes #1758.

### Summary

Since `loss_fn` and `torch_metrics` are not saved in `PLForecastingModule` checkpoints, they must be re-created using the `model.model_params` values so that the training continue with the proper loss (and continue to report the desired torch metrics).

### Other Information

Added the corresponding unittests


darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -425,6 +414,26 @@ def epochs_trained(self):
 
         return current_epoch
 
+    @staticmethod
+    def configure_torch_metrics(
```
dennisbader, line 418: couldn't we instead add the loss and metrics to `PLForecastingModule.on_save_checkpoint/on_load_checkpoint` so that the saving / loading is automatically handled?

dennisbader (changes requested):
Thanks @madtoinou for this, Looks good :)

I would be interested to see if we can let PL handle the saving/loading of these parameters by adapting `PLForecastingModule.on_save_checkpoint` and `PLForecastingModule.on_load_checkpoint`.

darts/models/forecasting/pl_forecasting_module.py
madtoinou: Since they were excluded from `self.save_hyperparameters()` in `PLForecastingModule.__init__()`, I assumed that we wanted to avoid having them in the checkpoints but it should probably work. I will test and see if it works.

dennisbader (changes requested):
Awesome that it worked with the checkpointing :) 
Actually when you mentioned that we ignore loss_fn, and torch_metrics when saving the hyperparameters, I tested if we can achieve the same thing by removing the ignore, and it works :) I left a comment.

After this change we can merge 🚀

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -392,13 +381,34 @@ def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
...
+        checkpoint["torch_metrics_train"] = self.train_metrics
+        checkpoint["torch_metrics_val"] = self.val_metrics
 
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
         # by default our models are initialized as float32. For other dtypes, we need to cast to the correct precision
         # before parameters are loaded by PyTorch-Lightning
         dtype = checkpoint["model_dtype"]
         self.to_dtype(dtype)
 
+        # restoring attributes necessary to resume from training properly
```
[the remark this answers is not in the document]
madtoinou: I checked, it's already loaded when calling `load_weights_from_checkpoint()`. My guess is that since it's one of the constructor argument and that it does not require any processing, the de-serializing of the checkpoint by Pytorch Lightning does the job.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -471,6 +472,63 @@ def test_load_weights(self):
...
+                "RNN",
+                5,
+                1,
+                n_epochs=1,
+                work_dir=self.temp_work_dir,
+                model_name=model_name,
+                save_checkpoints=True,
+                force_reset=True,
+                torch_metrics=MeanAbsolutePercentageError(),
+                pl_trainer_kwargs={"logger": DummyLogger(), "log_every_n_steps": 1},
```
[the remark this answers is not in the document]
madtoinou: Just to remove some of the warnings when running the tests

darts/models/forecasting/pl_forecasting_module.py
madtoinou: Cool! I hope we will never discover why it was ignored in the first place...

I left the helper method `configure_torch_metrics()` but let me know if I should remove it since it's called only at one place in the code.

dennisbader (approved):
Very nice, looks great! Thanks a lot @madtoinou 💯 🚀

PR #1762, opened 2023-05-11
Add max version for numpy to allow for shap compatibility

### LINKED ISSUE #1761: [BUG] Incompatibility with latest numpy version
**Describe the bug**
Instantiation of `ShapExplainer` fails (see [here](https://github.com/unit8co/darts/actions/runs/4950175794/jobs/8853427228#step:9:848)) due to access to a previously deprecated attribute (see [here](https://github.com/unit8co/darts/actions/runs/4950175794/jobs/8853427228#step:9:901).

**To Reproduce**
Instantiate `ShapExplainer` with latest numpy version.

**Expected behavior**
Instantiation of `ShapExplainer` should not throw an error.

 - Python version: 3.7+
 - darts version: *master*

**Additional context**

sudrich:
Fixes #1761.

### Summary

Introduce a max version for `numpy` to ensure compatibility with the current version of `shap` (see [here ](https://github.com/slundberg/shap/issues/2911)for context on the issue).


dennisbader:
The issue here is that the shap repo doesn't seem to be very active, and that the fix is not coming anytime soon (see [here](https://github.com/slundberg/shap/pull/1890#issuecomment-1540767609)). We can fix the numpy version for now but have to think about an alternative before releasing any new Darts version.

sudrich:
> The issue here is that the shap repo doesn't seem to be very active, and that the fix is not coming anytime soon (see [here](https://github.com/slundberg/shap/pull/1890#issuecomment-1540767609)). We can fix the numpy version for now but have to think about an alternative before releasing any new Darts version.

Yes. There seems to be a fork that wants to take over development. We should probably migrate once they are ready to publish their first version.

sudrich:
> I opened a new PR that addresses other fixes for current unit tests (from new statsmodels version). Let's continue in #1766. Will close this one.

Thanks for fixing the issue. I saw it arise last week, but didn't have the time to look deeper into it over the weekend.

akshika47:
In the meantime I used Python=3.8 and Numpy=1.19.0 versions to get things running but I don't know what other issues that may arise :)

PR #1783, opened 2023-05-18
Use component names for plotting

sudrich:
### Summary

This PR changes `TimeSeries.plot()` to be in line with its documentation.

**Example:**

Imagine invoking `plot(label="prediction")` on a `TimeSeries` with the labels `[A, B, C]`.

Before the change, the labels are using an enumeration of the components: `[prediction_0, prediction_1, prediction_2]`.
With the change, the labels are using the component names: `[prediction_A, prediction_B, prediction_C]`.

### Other Information

The change also allows adding a prefix for a series with a single component. This was not possible for some reason.

**Example:**

Imagine invoking `plot(label="prediction")` on a `TimeSeries` with the labels `[A]`.

Before the change, the label would just be ignored: `[A]`.
With the change, the label is taken into account: `[prediction_A]`.

This behaviour is still possible to recreate by calling `plot()` or `plot(label="")`.

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -195,7 +195,6 @@ class TimeSeries:
             # We have to check manually if the index is complete for non-empty series. Another way could
             # be to rely on `inferred_freq` being present, but this fails for series of length < 3.
             if len(self._time_index) > 0:
-
                 is_index_complete = (
                     len(
                         pd.date_range(
@@ -2147,7 +2146,6 @@ class TimeSeries:
     def _split_at(
         self, split_point: Union[pd.Timestamp, float, int], after: bool = True
     ) -> Tuple["TimeSeries", "TimeSeries"]:
-
         # Get index with not after in order to avoid moving twice if split_point is not in self
         point_index = self.get_index_at_point(split_point, not after)
         return (
@@ -3046,7 +3044,6 @@ class TimeSeries:
         )
 
     def resample(self, freq: str, method: str = "pad", **kwargs) -> Self:
-
         """
         Build a reindexed ``TimeSeries`` with a given frequency.
         Provided method is used to fill holes in reindexed TimeSeries, by default 'pad'.
@@ -3499,7 +3496,6 @@ class TimeSeries:
 
         # run through all transformations in transforms
         for transformation in transforms:
-
             if "components" in transformation:
                 if isinstance(transformation["components"], str):
                     transformation["components"] = [transformation["components"]]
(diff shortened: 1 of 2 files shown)


darts/timeseries.py
```diff
@@ -3798,11 +3798,10 @@ def plot(
...
             else:
-                label_to_use = (
-                    (label + ("_" + str(i) if len(self.components) > 1 else ""))
-                    if label != ""
-                    else "" + str(comp_name)
-                )
+                if label == "":
+                    label_to_use = comp_name
+                else:
+                    label_to_use = f"{label}_{comp_name}"
```
dennisbader, line 3804: I agree with adding `label` as a prefix when there are multiple components.
For a single component, I think it's better to overwrite the name with `label`. WDYT?
```suggestion
                    label_to_use = label if len(self.components) == 1 else f"{label}_{comp_name}"
```

dennisbader (changes requested):
Thanks for this PR @sudrich, it looks good.
I had a minor suggestions for using only the `label` if the series only has a signel component. We could adapt the docstring for `label` accordingly.

darts/timeseries.py
sudrich: Fair enough. I've added it and changed the docstring accordingly.

darts/timeseries.py
```diff
@@ -3723,8 +3719,10 @@ def plot(
         default_formatting
             Whether or not to use the darts default scheme.
         label
-            A prefix that will appear in front of each component of the TimeSeries or a list of string of
-            length the number of components in the plotted TimeSeries (default "").
+            Can either be a string or list of strings. If it's a single string and the series has only has a single
+            component, it is used as the label of the series in the plot. If it's a single string and the series has
+            more components, it is used as a prefix that will appear in front of each component. If it's a list of
+            strings, it is used as the label of each component. (default: "")
```
dennisbader, line 3725:
```suggestion
            Can either be a string or list of strings. If a string and the series only has a single component, it is
            used as the label for that component. If a string and the series has multiple components, it is used as
            a prefix for each component name. If a list of strings with length equal to the number of components in 
            the series, the labels will be mapped to the components in order.
```

dennisbader (approved):
Looks good, thanks @sudrich 🚀 
Just had a minor suggestion for rephrasing the docstring :)

Could you resolve the merge conflicts from the recent updates to the CHANGELOG?

darts/timeseries.py
sudrich: Done.

sudrich:
> Looks good, thanks @sudrich 🚀 Just had a minor suggestion for rephrasing the docstring :)
> 
> Could you resolve the merge conflicts from the recent updates to the CHANGELOG?

Thanks for the review. I've addressed the outstanding issue and rebased to `master`.

darts/timeseries.py
dennisbader: 👍

dennisbader (approved):
Looks good, thanks a lot @sudrich 💯 🚀

PR #1787, opened 2023-05-19
Fix docstring ``predict()` -> ``predict()``

emptymalei:
Fixes #.

### Summary

In some of the docs, we see ([this page](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.lgbm.html))

image

``predict()`

should be

`predict()`

The reason that this happens is because some of the docstrings missed a `


dennisbader (approved):
Thanks @emptymalei 🚀

PR #1808, opened 2023-06-01
Update arima.py

tijmenvanetten:
ARIMA models are capable of making forecasts without retraining

Fixes #.


dennisbader:
This is already set in `TransferableFutureCovariatesLocalForecastingModel` that ARIMA inherits from:

https://github.com/unit8co/darts/blob/1f175807ab8ce220bc1c0dde6495d4b45795881d/darts/models/forecasting/forecasting_model.py#L2595

I'll close this for now. Let me know if I missed anything.

PR #1811, opened 2023-06-02
Feat/likelihood parameters prediction

### LINKED ISSUE #1445: Quantiles without Sampling
**Problem:**
Calculating the quantiles of a probabilistic forecast requires sampling.
When working with parameterised likelihood models such das the gaussian likelihood, one must use the num_samples parameter in the predict method to obtain samples. Then one can calculate the quantiles from the probabilistic time series.
In case of parameterised likelihood models one could directly calculate the the quantiles and avoid da calculation for sampling.
What is the advantage of sampling? 

**Proposed Solution:**
Extend the predict API to calculate quantiles using the internal estimated parameters of the distribution.

**Alternative Solution:**
Make raw outputs of the neural network accessible.

### LINKED ISSUE #1735: Accessing the parameters of marginal distribution
When working with probabilistic forecast, for example, the  `GaussianLikelihood` class. Sometimes I want to know the mean and variance of the predicted values, i.e. the parameters of the distribution. The `sample` function takes the input of mean and variance then sample from this distribution. I can estimate the mean and variance from the samples. Is there a way to extract the mean and variance without sampling?

madtoinou:
Fixes #1735, fixes #1445.

### Summary

Edit: Final version:
After fitting a regression or torch-based model with a likelihood, it's possible to pass the flag `predict_likelihood_parameters=True` to `predict()` and directly predict the distribution parameters instead of sampling (which remains the best way to simulate/visualize uncertainty).

Draft version:
The distribution parameters are returned by the `Likelihood.sample()` method (in addition of the sampled values), and processed in the same way as the target forecast values in `predict()`/`get_batch_prediction()` which remain necessary if the number of forecasted values is great (auto-regression). If the flag is set, the output of `predict()` contains only the distribution parameters (no target forecast)

### Other Information

Based on @hrzn comment in #1445, a warning should probably be raised if the predict TimeSeries contain many component to avoid dimension explosion. Another solution would eventually be to restrict this feature only to a few Likelihoods.

Verifying that the model converge to the good parameters is too time consuming for the unittests but I did check manually for some distributions (Gaussian, Poisson, Quantile) and it looked rather good.


darts/models/forecasting/regression_model.py
```diff
@@ -570,16 +571,34 @@ def predict(
...
             not self._is_probabilistic() and num_samples > 1,
             "`num_samples > 1` is only supported for probabilistic models.",
             logger,
         )
 
+        raise_if(
+            likelihood_parameters and not isinstance(self, _LikelihoodMixin),
+            f"{self.__class__} does not support `likelihood_parameters=True` as it does not inherit from "
+            f"`_LikelihoodMixin`. The models supporting it are `LinearRegressionModel`, `LGBMModel`, "
+            f"`CatboostModel` and `XGBoostModel`.",
```
dennisbader, line 592: we got to check also that the model is probablistic

```suggestion
            f"`likelihood_parameters=True` is only supported for probabilistic regression models.",
```

darts/utils/likelihood_models.py
```diff
@@ -160,10 +164,36 @@ def _params_from_output(self, model_output: torch.Tensor):
...
         """
         Samples a prediction from the probability distributions defined by the specific likelihood model
-        and the parameters given in `model_output`.
+        and the parameters given in `model_output`, return both the samples and the Likelihood parameters.
+        """
+        pass
+
+    def predict_likelihood_parameters(self, model_output: torch.Tensor) -> torch.Tensor:
+        """
+        Returns the distribution parametersas a single Tensor, extracted from the raw model outputs.
```
dennisbader, line 173:
```suggestion
        Returns the distribution parameters as a single Tensor, extracted from the raw model outputs.
```

darts/utils/likelihood_models.py
```diff
@@ -160,10 +164,36 @@ def _params_from_output(self, model_output: torch.Tensor):
     def sample(self, model_output: torch.Tensor) -> torch.Tensor:
         """
         Samples a prediction from the probability distributions defined by the specific likelihood model
-        and the parameters given in `model_output`.
+        and the parameters given in `model_output`, return both the samples and the Likelihood parameters.
```
dennisbader, line 167:
```suggestion
        and the parameters given in `model_output`.
```

darts/utils/likelihood_models.py
```diff
@@ -149,7 +151,9 @@ def _distr_from_params(self, params: Tuple) -> torch.distributions.Distribution:
         pass
 
     @abstractmethod
-    def _params_from_output(self, model_output: torch.Tensor):
+    def _params_from_output(
+        self, model_output: torch.Tensor
+    ) -> Union[Tuple[torch.Tensor, ...], torch.Tensor]:
```
dennisbader, line 156: in which case do we have a tuple of tensors?
I think even RNN drops the hidden_state before the likelihood

darts/utils/likelihood_models.py
```diff
@@ -264,6 +294,7 @@ def _distr_from_params(self, params):
 
     def sample(self, model_output: torch.Tensor) -> torch.Tensor:
         mu, sigma = self._params_from_output(model_output)
+        # concatenate the parameters on the component dimension
```
dennisbader, line 297: can this be removed?

darts/utils/likelihood_models.py
dennisbader, line 1: as discussed offline, the component names order is correct, but the tensors need to be reordered before creating the prediction time series.

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -236,24 +236,44 @@ def predict_step(
...
         batch_predictions = batch_predictions.cpu().detach().numpy()
 
+        if self.predict_likelihood_parameters:
+            custom_columns = [
+                self.likelihood.likelihood_components_names(input_series)
+                for input_series in batch_input_series
+            ]
+        else:
+            custom_columns = None
+
```
dennisbader, line 246: we can avoid this
```suggestion
```

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -236,24 +236,44 @@ def predict_step(
...
         ts_forecasts = Parallel(n_jobs=self.pred_n_jobs)(
             delayed(_build_forecast_series)(
                 [batch_prediction[batch_idx] for batch_prediction in batch_predictions],
                 input_series,
+                custom_columns=None
+                if custom_columns is None
+                else custom_columns[batch_idx],
```
dennisbader, line 253:
```suggestion
                custom_columns=self.likelihood.likelihood_components_names(input_series)
                if self.predict_likelihood_parameters else None,
```

darts/models/forecasting/rnn_model.py
```diff
@@ -86,7 +86,9 @@ def __init__(
         # The RNN module needs a linear layer V that transforms hidden states into outputs, individually
         self.V = nn.Linear(hidden_dim, target_size * nr_params)
 
-    def forward(self, x_in: Tuple, h=None):
+    def forward(
+        self, x_in: Tuple, h: Union[torch.Tensor, None] = None
```
dennisbader, line 90:
```suggestion
        self, x_in: Tuple, h: Optional[torch.Tensor] = None
```

darts/models/forecasting/rnn_model.py
```diff
@@ -120,11 +122,16 @@ def _produce_train_output(self, input_batch: Tuple):
         )
         return self(model_input)[0]
 
-    def _produce_predict_output(self, x: Tuple, last_hidden_state=None):
+    def _produce_predict_output(
+        self, x: Tuple, last_hidden_state: Union[torch.Tensor, None] = None
```
dennisbader, line 126:
```suggestion
        self, x: Tuple, last_hidden_state: Optional[torch.Tensor] = None
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1267,6 +1268,9 @@ def predict(
         mc_dropout
             Optionally, enable monte carlo dropout for predictions using neural network based models.
             This allows bayesian approximation by specifying an implicit prior over learned models.
+        predict_likelihood_parameters
+            If set to `True`, the model predict the parameters of its Likelihood parameters instead of the target. Only
+            supported for probablistic models, with `num_samples = 1` and `n<=output_chunk_length`. Default: ``False``
```
dennisbader, line 1273:
```suggestion
            supported for probabilistic models, with `num_samples = 1` and `n<=output_chunk_length`. Default: ``False``
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1389,6 +1402,9 @@ def predict_from_dataset(
         mc_dropout
             Optionally, enable monte carlo dropout for predictions using neural network based models.
             This allows bayesian approximation by specifying an implicit prior over learned models.
+        predict_likelihood_parameters
+            If set to `True`, the model predict the parameters of its Likelihood parameters instead of the target. Only
+            supported for probablistic models, with `num_samples = 1` and `n<=output_chunk_length`. Default: `False`.
```
dennisbader, line 1407:
```suggestion
            supported for probabilistic models, with `num_samples = 1` and `n<=output_chunk_length`. Default: `False`.
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1412,6 +1428,13 @@ def predict_from_dataset(
                 "`roll_size` must be an integer between 1 and `self.output_chunk_length`.",
             )
 
+        # prevent auto-regression when prediction the likelihood parameters
+        raise_if(
+            predict_likelihood_parameters and n > self.output_chunk_length,
+            "`n` must be inferior or equal to `output_chunk_length` when `predict_likelihood_parameters=True`.",
```
dennisbader, line 1434:
```suggestion
            "`n` must be smaller than or equal to `output_chunk_length` when `predict_likelihood_parameters=True`.",
```

darts/models/forecasting/regression_model.py
```diff
@@ -569,16 +569,30 @@ def predict(
...
+        if predict_likelihood_parameters:
+            raise_if_not(
+                isinstance(self, _LikelihoodMixin),
+                "`likelihood_parameters=True` is only supported for probabilistic regression models.",
+                logger,
+            )
+
+            raise_if(
+                n > self.output_chunk_length,
+                "`n` must be inferior or equal to `output_chunk_length` when `predict_likelihood_parameters=True`.",
```
dennisbader, line 591:
```suggestion
                "`n` must be smaller than or equal to `output_chunk_length` when `predict_likelihood_parameters=True`.",
```

darts/tests/models/forecasting/test_probabilistic_models.py
dennisbader, line 1: Tests look good, only tests for regression models are missing

dennisbader (changes requested):
Very nice @madtoinou , that looks very clean 🚀 
Had some refactoring suggestions here and there and the issue with the wrong component order for torch models.

darts/models/forecasting/forecasting_model.py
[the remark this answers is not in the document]
madtoinou: I prefer `predict_likelihood_parameters`

darts/utils/likelihood_models.py
madtoinou: When a distribution have several parameters, they are returned as separate Tensors (for example `GaussianLikelihood` with `mu` and `sigma`)

darts/models/forecasting/forecasting_model.py
```diff
@@ -244,7 +244,9 @@ def considers_static_covariates(self):
         return self._considers_static_covariates
 
     @abstractmethod
-    def predict(self, n: int, num_samples: int = 1) -> TimeSeries:
+    def predict(
+        self, n: int, num_samples: int = 1, predict_likelihood_parameters: bool = False
```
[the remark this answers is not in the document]
madtoinou: This property is already covered by `_is_probabilistic()`. Moreover, `RandomForest` inherits from `RegressionModel` which does not have the `likelihood` parameter in its constructor and check if self is an instance of `_LikelihoodMixin` in the `predict()` method.

darts/models/forecasting/regression_model.py
```diff
@@ -852,15 +897,28 @@ def _prepare_quantiles(quantiles):
 
         return quantiles, median_idx
 
+    def _likelihood_components_names(self, input_series: TimeSeries) -> List[str]:
```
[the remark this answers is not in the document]
madtoinou: I agree, it was very counter intuitive to find them here. They could probably go in the likelihood_models.py file with the others

darts/models/forecasting/regression_ensemble_model.py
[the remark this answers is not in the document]
madtoinou: Yes, I wanted to keep it for later because we need to perform additional checks on the likelihood used in the forecasting models & `output_chunk_length` value but it would probably be better to have everything available at once. I came up with something for the NaiveEnsemble, I need to slightly adapt it for the RegressionEnsemble

darts/utils/likelihood_models.py
dennisbader, line 1: I believe the parameter order is still wrong for the multivariate case, right?

darts/models/forecasting/ensemble_model.py
```diff
@@ -266,13 +270,21 @@ def predict(
             future_covariates=future_covariates,
             num_samples=num_samples,
             verbose=verbose,
+            predict_likelihood_parameters=predict_likelihood_parameters,
         )
 
-        # for multi-level models, forecasting models can generate arbitrary number of samples
+        if series is None:
```
dennisbader, line 276: were these changes required?

darts/models/forecasting/ensemble_model.py
```diff
@@ -362,6 +383,22 @@ def find_max_lag_or_none(lag_id, aggregator) -> Optional[int]:
     def _models_are_probabilistic(self) -> bool:
         return all([model._is_probabilistic() for model in self.models])
 
+    def _models_same_likelihood(self) -> bool:
```
dennisbader, line 386: for quantileregression we also need to check that the same quantiles were fitted, right?

darts/models/forecasting/baselines.py
```diff
@@ -213,16 +238,88 @@ def fit(
...
+                self._target_average(predictions, series)
+                if not predict_likelihood_parameters
+                else self._params_average(predictions, series)
+            )
+
+    def _target_average(self, prediction: TimeSeries, series: TimeSeries) -> TimeSeries:
+        """Average across the components, keep n_samples, rename components"""
+        n_forecasting_models = len(self.models)
+        n_components = series.n_components
+        prediction_values = prediction.all_values()
```
dennisbader, line 271:
```suggestion
        prediction_values = prediction.all_values(copy=False)
```

darts/models/forecasting/baselines.py
```diff
@@ -213,16 +238,88 @@ def fit(
...
+        likelihood: Union[str, Likelihood] = getattr(self.models[0], "likelihood")
+        if isinstance(likelihood, Likelihood):
+            likelihood_n_params = likelihood.num_parameters
         else:
-            return take_average(predictions)
+            likelihood_n_params = self.models[0].num_parameters
+        n_forecasting_models = len(self.models)
+        n_components = series.n_components
+        # aggregate across predictions [model1_param0, model1_param1, ..., modeln_param0, modeln_param1]
+        prediction_values = prediction.values()
```
dennisbader, line 304:
```suggestion
        prediction_values = prediction.values(copy=False)
```

dennisbader (changes requested):
Very nice with the support for EnsembleModels :) 

The last point missing is support for historical forecasts, and the corrections to the parameter order for torch likelihoods.

darts/models/forecasting/ensemble_model.py
madtoinou: Yes, because of the changes to support multivariate ts with the `NaiveEnsemble`, the argument `series` of `ensemble()` is not optional anymore. But I can move this to `NaiveEnsemble.ensemble()` if you prefer

darts/models/forecasting/ensemble_model.py
madtoinou: correct

darts/models/forecasting/forecasting_model.py
```diff
@@ -260,21 +260,18 @@ def predict(self, n: int, num_samples: int = 1) -> TimeSeries:
...
-        if not self._fit_called:
-            raise_log(
-                ValueError(
-                    "The model must be fit before calling predict(). "
-                    "For global models, if predict() is called without specifying a series, "
-                    "the model must have been fit on a single training series."
-                ),
-                logger,
-            )
+        raise_if_not(
```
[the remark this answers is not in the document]
madtoinou: Thanks for running the benchmarking

darts/models/forecasting/regression_model.py
```diff
@@ -569,16 +570,30 @@ def predict(
...
             univariate target series.
         """
-        raise_if(
-            not self._is_probabilistic() and num_samples > 1,
-            "`num_samples > 1` is only supported for probabilistic models.",
-            logger,
-        )
+
+        if predict_likelihood_parameters:
+            raise_if_not(
```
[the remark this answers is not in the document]
madtoinou: You're right, the validity check in `GlobalForecastingModel` should handle it

darts/models/forecasting/regression_model.py
```diff
@@ -1017,8 +1039,144 @@ def _quantile_sampling(self, model_output: np.ndarray) -> np.ndarray:
...
+
+        return samples_reshaped
+
+    def _quantiles_params(self, model_output: np.ndarray) -> np.ndarray:
+        # model_outputs has shape (n_series * n_samples, output_chunk_length, n_components, n_quantiles),
+        # with n_samples = 1
+
+        # output shape (n_series * n_samples, output_chunk_length, n_components*n_quantiles)
+        shape = model_output.shape
+        if len(shape) <= 4:
```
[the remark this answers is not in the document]
madtoinou: `model_output` have a standard shape after `predict_{likelihood}()`, this is a leftover from my previous approach

darts/models/forecasting/regression_model.py
```diff
@@ -1017,8 +1039,144 @@ def _quantile_sampling(self, model_output: np.ndarray) -> np.ndarray:
...
+            return model_output
+        else:
+            # last dim : [comp_1_q_1, ..., comp_1_q_n, ..., comp_n_q_1, ..., comp_n_q_n]
+            return model_output.reshape(shape[0], shape[1], shape[2] * shape[3])
+
+    def _poisson_params(self, model_output: np.ndarray) -> np.ndarray:
+        # model_output has shape (n_series * n_samples, output_chunk_length, n_components), with n_samples = 1
```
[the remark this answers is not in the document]
madtoinou: The output shape of predict and sample is somehow standardized, as `RegressionModel.predict()` expect something of the shape `(n_series * n_samples, output_chunk_length, n_components)` and then reshape it into `(n_series, output_chunk_length, n_components, n_samples)`.

I reworked the comments to make them less confusing, there were kind of notes to myself before that

It's the model_output shape for the gaussian likelihood (supported by CatBoost exclusively) that is slithly different from the two others, I can try to reshape it so that it's identical to the two others. Or maybe keep it for another PR, in which we could also move these LikelihoodMixin to the likelihood_models.py file. WDYT?

dennisbader (changes requested):

darts/models/forecasting/torch_forecasting_model.py
dennisbader, line 1: let's update the `self.output_chunk_length` of all TFMs to `self._output_chunk_length` as well, and then update the property accordingly

madtoinou: Updated, I tried to keep `self.output_chunk_length` when possible (available after the model/pl module is instantiated)

dennisbader (approved):
Congrats @madtoinou for this great PR 💯 
Ready to be merged 🚀

Issue #16, opened 2019-05-29
Theta


kstyrc:
Add Theta method to the library:
https://robjhyndman.com/papers/Theta.pdf

Issue #13, opened 2019-05-27
Non-Parametric Time Series (NPTS)


kstyrc:
AWS Forecast Recipe:
https://docs.aws.amazon.com/forecast/latest/dg/aws-forecast-recipe-npts.html

Issue #12, opened 2019-05-27
DeepAR+


kstyrc:
AWS Forecast Recipe:
https://docs.aws.amazon.com/forecast/latest/dg/aws-forecast-recipe-deeparplus.html

PR #1815, opened 2023-06-05
feat: EnsembleModel accepts pretrained global models

### LINKED ISSUE #1775: Add support for pre-trained models in EnsembleModel (first step)
- Do you guys have any documentation (in terms of papers) as to how this model works?
- If we save this model at a checkpoint, are the all of the individual components of this model also saved or do we have to save those separately?  IOW, how do we load up and call this model from another program?

### LINKED ISSUE #1785: [BUG] RegressionEnsembleModel.fit() does not have ability to use validation data to fit either base learners or ensemble learner
**Describe the bug**
The `RegressionEnsembleModel` does not allow a `val_series` or `val_past_covariates` to be passed in training the base models that are part of the ensemble.  As a result, right now, it seems impossible to use 'val_loss' or other metrics of the validation set to determine `EarlyStopping` criteria.

Even if I fit the individual base models separately, it appears that the `fit()` member function restarts the fitting for each of them.

 - Python version: [e.g. 3.9]
 - darts version [e.g. 0.14.0]

madtoinou:
Fixes #1775 and fixes #1785.

### Summary

- `EnsembleModel` accept pre-trained models if they are all global and all already fitted.
- `RegressionEnsembleModel.fit()` accepts an additional argument `retrain_forecasting_model` which is `False` by default to avoid catastrophic forgetting.  `retrain_forecasting_model=True` is accepted only if all the forecasting models are global and pre-trained.

### Other Information

- If a `NaiveEnsembleModel` was instantiated with pre-trained global models, the constructor sets `self._fit_called=True` so that `predict()` can be called directly without having to call `fit()`. I don't know if it's the most intuitive behavior but in case of single ts training/prediction, it makes the workflow more natural.
- Added some very basics tests


dennisbader (changes requested):
Thanks @madtoinou , this is a great start 🚀 

I have a couple of points:
- let's move `retrain_forecasting_models` to the constructor `__init__()` of the base `EnsembleModel`
- `retrain_forecasting_models` should be `True` by default to maintain the same behavior as we had previously. 
  - if `retrain_forecasting_models=True`, we use the `forecasting_models = [model.untrained_model() for model in forecasting_models]`. Like this we will not mess with any pre-trained model.
    - we can log an info if there were any pre-trained models in the original `forecasting_models`
- I think it would be a good idea at this point to rename `models` to `forecasting_models` in all ensemble models (base, naive) to have a common naming. This is a breaking change, but I think it is acceptable along with all the other changes of the next release.
- In theory `NaiveEnsembleModel` could be used without `fit()` if all the models were pretrained. Should we allow this?

darts/models/forecasting/regression_ensemble_model.py
dennisbader, line 1: How about we allow `-1` for `regression_train_n_points` when `retrain_forecasting_models=False`?
Like this whenever the user calls `fit()`, the regression model would be trained on the entire input?

darts/models/forecasting/ensemble_model.py
dennisbader, line 1: There are a still a couple of places where we refer to `models` in the docstrings/error messages, etc. 
Could you update these with the new `forecasting_models` ?

darts/models/forecasting/baselines.py
```diff
@@ -177,18 +178,26 @@ def __init__(
...
             List of forecasting models whose predictions to ensemble
+        retrain_forecasting_models
+            If set to `False`, the `forecasting_models` are not retrained on `series` (only supported if all the
+            `forecasting_models` are pretrained `GlobalForecastingModels`). Default: ``True``.
         show_warnings
             Whether to show warnings related to models covariates support.
         """
         super().__init__(
-            models=models,
+            forecasting_models=forecasting_models,
```
dennisbader, line 190: this super call has a lot of type hint warnings. E.g. traing_num_samples is supposed to be `int` but we give `None`. Can you fix these?

darts/models/forecasting/baselines.py
```diff
@@ -177,18 +178,26 @@ def __init__(
...
             train_num_samples=None,
             train_samples_reduction=None,
+            retrain_forecasting_models=retrain_forecasting_models,
             show_warnings=show_warnings,
         )
 
+        # ensemble model initialised with trained global models can call predict()
```
dennisbader, line 197:
```suggestion
        # ensemble model initialised with trained global models can directly call predict()
```

darts/models/forecasting/baselines.py
```diff
@@ -177,18 +178,26 @@ def __init__(
 
         Parameters
         ----------
-        models
+        forecasting_models
             List of forecasting models whose predictions to ensemble
+        retrain_forecasting_models
```
dennisbader, line 183: Can you add here that NaiveEnsemble doesn't have to be trained with retrain=False and predict can be call directly?

darts/models/forecasting/baselines.py
```diff
@@ -177,18 +178,26 @@ def __init__(
...
+            If set to `False`, the `forecasting_models` are not retrained on `series` (only supported if all the
```
dennisbader, line 184:
```suggestion
        retrain_forecasting_models
            If set to `False`, the `forecasting_models` are not retrained when calling `fit()` (only supported if all the
```

darts/models/forecasting/baselines.py
```diff
@@ -200,13 +209,14 @@ def fit(
...
-                kwargs["future_covariates"] = future_covariates
-            model.fit(**kwargs)
+        if self.retrain_forecasting_models:
+            for model in self.forecasting_models:
+                kwargs = dict(series=series)
+                if model.supports_past_covariates:
+                    kwargs["past_covariates"] = past_covariates
+                if model.supports_future_covariates:
+                    kwargs["future_covariates"] = future_covariates
+                model.fit(**kwargs)
```
dennisbader, line 219: here the tip again to return before in order to avoid indented sections

```suggestion
        if not self.retrain_forecasting_models:
            return self

        for model in self.forecasting_models:
            kwargs = dict(series=series)
            if model.supports_past_covariates:
                kwargs["past_covariates"] = past_covariates
            if model.supports_future_covariates:
                kwargs["future_covariates"] = future_covariates
            model.fit(**kwargs)
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -66,16 +70,20 @@ def __init__(
             If `forecasting models` are probabilistic and `regression_train_num_samples` > 1, method used to
             reduce the samples before passing them to the regression model. Possible values: "mean", "median"
             or float value corresponding to the desired quantile. Default: "median"
+        retrain_forecasting_models
+            If set to `False`, the `forecasting_models` are not retrained on `series` (only supported if all the
```
dennisbader, line 74:
```suggestion
            If set to `False`, the `forecasting_models` are not retrained when calling `fit()` (only supported if all the
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -115,6 +123,20 @@ def fit(
         past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
         future_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
     ):
+        """
+        Fits the forecasting models with the entire series except the last `regression_train_n_points` values.
```
dennisbader, line 127: this needs to be extended with the regression (Ensemble) model being training

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -144,17 +166,18 @@ def fit(
...
-            model._fit_wrapper(
-                series=forecast_training,
-                past_covariates=past_covariates
-                if model.supports_past_covariates
-                else None,
-                future_covariates=future_covariates
-                if model.supports_future_covariates
-                else None,
-            )
+        if self.retrain_forecasting_models:
```
dennisbader, line 169: It seems a bit strange to me that we ues `model.fit_wrapper()` here but [below](https://github.com/unit8co/darts/blob/8e93720e6277726089058cf4d039b97de0f57c34/darts/models/forecasting/regression_ensemble_model.py#L178) (before end of this method) we call `model.fit()` directly. Do you know if this was intended? If not intended then it's probably be better to use the same logic.

As far as I see it, _fit_wrapper() is handling the passing of covariates to the model based on their support.
We could probably refactor this method with the logic from [below](https://github.com/unit8co/darts/blob/8e93720e6277726089058cf4d039b97de0f57c34/darts/models/forecasting/regression_ensemble_model.py#L178). Then we can also avoid overwriting by the subclasses.

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -172,9 +195,11 @@ def fit(
         # prepare the forecasting models for further predicting by fitting them with the entire data
 
         # Some models (incl. Neural-Network based models) may need to be 'reset' to allow being retrained from scratch
-        self.models = [model.untrained_model() for model in self.models]
+        self.forecasting_models = [
```
dennisbader, line 198: we need to return here before when retrain=False
```
if not self.retrain_forecasting_models:
    return self
```

darts/tests/models/forecasting/test_ensemble_models.py
```diff
@@ -45,17 +45,55 @@ def test_untrained_models(self):
...
+        with self.assertRaises(ValueError):
+            NaiveEnsembleModel([local_model, global_model.untrained_model()])
+
+        # global trained, global untrained
+        with self.assertRaises(ValueError):
+            NaiveEnsembleModel([global_model, global_model.untrained_model()])
+
+        # both global trained, retrain = True
+        model_ens_retrain = NaiveEnsembleModel(
+            [global_model, global_model], retrain_forecasting_models=True
```
dennisbader, line 85: I wonder if it was better before (I know I suggested to implement what you have now :D ) when we raised an error instead of silently calling model.untrained_model().
Let me know what you think.

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -132,6 +132,35 @@ def test_accepts_one_model(self):
             model.fit(series=self.combined)
             model.predict(10)
 
+    def test_accept_pretrain_global_models(self):
```
dennisbader, line 135: 👍 cool test

darts/tests/models/forecasting/test_ensemble_models.py
```diff
@@ -45,17 +45,55 @@ def test_untrained_models(self):
...
         # an untrained ensemble model should also give untrained underlying models
         model_ens = NaiveEnsembleModel([NaiveDrift()])
         model_ens.fit(self.series1)
-        assert model_ens.models[0]._fit_called
+        assert model_ens.forecasting_models[0]._fit_called
         new_model = model_ens.untrained_model()
-        assert not new_model.models[0]._fit_called
+        assert not new_model.forecasting_models[0]._fit_called
+
+    def test_trained_models(self):
```
dennisbader, line 60: 👍 nice!

dennisbader (changes requested):
Really nice, thanks @madtoinou ! 🚀 

Had some last suggestions mainly regarding documentation and some typos.

One thing I thought could be nice to add:
- How about we allow `-1` for `regression_train_n_points` when `retrain_forecasting_models=False`? Like this whenever the user calls `fit()`, the regression model would be trained on the entire input?

darts/models/forecasting/regression_ensemble_model.py
madtoinou: I decided to update the `training_series` attribute of the models to make the behavior of `predict()` more intuitive (otherwise, there might be discrepancies between the forecasting models and the regressor, especially in the prediction time indexes)

darts/models/forecasting/regression_ensemble_model.py
madtoinou: Not sure of what you meant, let me know if the changes I applied correspond to your expectations.

darts/models/forecasting/regression_ensemble_model.py
madtoinou: Switched to fit_wrapper in all relevant places!

darts/tests/models/forecasting/test_ensemble_models.py
madtoinou: I prefer to raise error, to make it as explicit/transparent as possible

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -112,23 +131,77 @@ def _split_multi_ts_sequence(
...
+        past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries], None]] = None,
+        future_covariates: Optional[
+            Union[TimeSeries, Sequence[TimeSeries], None]
+        ] = None,
     ):
+        """
+        Fits the forecasting models with the entire series except the last `regression_train_n_points` values, which
+        are used to train the regressor model.
+
+        If `forecasting_models` contains fitted `GlobalForecastingModels` and `retrain_forecasting_model=False`, the
```
dennisbader, line 143:
```suggestion
        If `forecasting_models` contains fitted `GlobalForecastingModels` and `retrain_forecasting_model=False`, only the
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -112,23 +131,77 @@ def _split_multi_ts_sequence(
...
         )
 
         # spare train_n_points points to serve as regression target
         is_single_series = isinstance(series, TimeSeries)
-        if is_single_series:
-            train_n_points_too_big = len(self.training_series) <= self.train_n_points
-        else:
-            train_n_points_too_big = any(
-                [len(s) <= self.train_n_points for s in series]
+        if self.train_n_points == -1:
```
dennisbader, line 164: can we not use the all the points of all time series instead of using the largest possible length for all of them?
The global models are able to handle multiple time series with different lengths

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -170,17 +244,26 @@ def fit(
...
+                    if model.supports_past_covariates
+                    else None,
+                    future_covariates=future_covariates
+                    if model.supports_future_covariates
+                    else None,
+                )
+        # update training_series attribute to make predict() behave as expected
+        else:
+            for model in self.forecasting_models:
+                model.training_series = series if is_single_series else None
```
dennisbader, line 266: I would rather pass the series at prediction time (with `series=`) than mutating the models.
As far as I can tell, EnsembleModel.predict() passes the series to the forecasting `model.predict()`.
Since retrain_forecasting_models=False, we have only global models here which allow passing a series at prediction time.

darts/models/forecasting/baselines.py
```diff
@@ -200,13 +214,27 @@ def fit(
...
+                    series=series,
+                    past_covariates=past_covariates
+                    if model.supports_past_covariates
+                    else None,
+                    future_covariates=future_covariates
+                    if model.supports_future_covariates
+                    else None,
+                )
+        # update training_series attribute to make predict() behave as expected
+        else:
```
dennisbader, line 233: Here the same comment as I added for regression ensemble model.

darts/models/forecasting/regression_ensemble_model.py
dennisbader: 👍

dennisbader (changes requested):
Very nice, we're almost there 🚀 

Had a couple of suggestions, the main ones revolve around the regression_train_n_points=-1 where I think we can just use the series as is for the input to generate the regression model training data.

darts/models/forecasting/regression_ensemble_model.py
madtoinou: Because otherwise, there will be a shift in the time index of the "`future_covariates`" generated by the `forecasting models` and the target of the regressor: covariates will be shorter than the targets because of the lags/input_chunk_length

darts/models/forecasting/regression_ensemble_model.py
madtoinou: If the global models were (previously) fitted using single series (hence saved, with a time index potentially different than the one being fitted) and later, `predict()` is called with `series=None`, then the forecasting models could produce covariates with different time indexes and the ensemble won't work. I remember having this issue when running some tests

To prevent this, we could eventually retrieve the `ensemble.regression_model.training_series` when `series=None` instead of each `model.training_series` when `_make_multiple_predictions()` is called in `ensemble.predict()`.

dennisbader (changes requested):
Very cool, thanks @madtoinou 🚀  I also really like the notebook 😍 
Shows how powerful the ensemble models have become!

There are still one or two points we have to think about before merging.
Mainly how we handle the num_training_points and the train_using_historical_forecasts. I added some coments, maybe we can discuss a bit more next week/

examples/19-EnsembleModel-examples.ipynb
```diff
@@ -0,0 +1,868 @@
+{
```
dennisbader, line 1: <p>darts -&gt; Darts</p><p>either ensembling always with capital letter or never</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #12.        NaiveSeasonal,
</pre><p>not sure if import sorting will be checked in CI as soon as it's merged</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #16.        XGBModel,
</pre><p>let's drop XGBModel as it's not part of the standard <code>darts</code> installation </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #22.    from darts.utils.timeseries_generation import linear_timeseries
</pre><p>this line and the one above can be put together</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

dennisbader, line 1: <p>Ensembling combines the forecasts of several "weak" models to obtain a more robust and accurate model.</p><p> </p><p>All of Darts' ensembling models rely on the <strong>stacking technique</strong> (<a href="https://machinelearningmastery.com/essence-of-stacking-ensembles-for-machine-learning/" target="_blank">reference</a>). They provide the same functionalities as the other forecasting models. Depending on the ensembled models, they can:</p><ul><li>levarage covariates</li><li>be trained on multiple-series</li><li>predict multivariate targets</li><li>generate probabilistic forecasts</li><li>and more</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    plt.show()
</pre><p>we can drop plt.show() everywhere as we use %matplotlib inline at the biginning</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    naive_ensemble = NaiveEnsembleModel(forecasting_models=[NaiveSeasonal(K=12), NaiveDrift()])
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

dennisbader, line 1: <p><strong>Note</strong>: after looking at the each model's MAPE, one would notice that <code>NaiveSeasonal</code> is actually performing better on its own than ensembled with <code>NaiveDrift</code>. Checking the performance of the single models is in general a good practice before defining an ensemble. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

dennisbader, line 1: <p>we could use the same time range as in the previous plot so that the visual comparison becomes easier</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

dennisbader, line 1: <p>If not specified, a linear regression is used. Using a sklearn model is easy out-of-the-box, but using a <code>RegressionModel</code></p><p>...  -&gt; If not specified, Darts' <code>LinearRegressionModel</code> is used. Using a sklearn model is easy out-of-the-box, but using one of Darts' regression models ...</p><p> </p><p> </p><p> </p><p>Other comment: it sounds like it accepts only these three parameters. Maybe we can write:</p><p> </p><p>It accepts three parameters: -&gt; The main three parameters are:</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

dennisbader, line 1: <p>"which present an marked upward trend. " is there a typo?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #10.        regression_train_n_points=24,
</pre><p>we could use <code>regression_trian_n_points=-1</code> and <code>train_using_historical_forecasts=True</code> . Might yield better results?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #13.    # RegressionEnsemble must train the ensemble model, even if the forecasting models are already trained
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/1815/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

darts/models/forecasting/baselines.py
```diff
@@ -236,13 +250,27 @@ def fit(
...
-        for model in self.models:
-            kwargs = dict(series=series)
-            if model.supports_past_covariates:
-                kwargs["past_covariates"] = past_covariates
-            if model.supports_future_covariates:
-                kwargs["future_covariates"] = future_covariates
-            model.fit(**kwargs)
+        if self.retrain_forecasting_models:
+            # Some models may need to be 'reset' to allow being retrained from scratch, especially torch-based models
+            self.forecasting_models: List[ForecastingModel] = [
```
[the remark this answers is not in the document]
madtoinou: This line is actually unnecessary since the constructor raise an error if retrain_forecasting_models=True with some "already trained" model. I removed it

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -100,34 +115,155 @@ def __init__(
...
+    ) -> Union[TimeSeries, Sequence[TimeSeries]]:
+        """
+        For GlobalForecastingModel, when predicting n > output_chunk_length, `historical_forecasts()` generally
+        producebetter forecasts than `predict()`.
+
+        train_n_points are generated, starting from the end of the series
+        """
+        is_single_series = isinstance(series, TimeSeries) or series is None
+        predictions = [
+            model.historical_forecasts(
```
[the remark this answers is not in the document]
madtoinou: I decided to go with `forecast_horizon = stride = output_chunk_length` and `last_points_only=False`. When `output_chunk_length` is not a multiple of `regression_train_n_points`, the historical forecast is missing some values at the end (using a stride of 1 could solve the problem but then, there is the problem of the overlapping forecasts) so I added exception in the constructor to prevent this scenario. Let me know what you think.

examples/19-EnsembleModel-examples.ipynb
madtoinou: <p>Maybe "which contain an upward trend" would be clearer?</p>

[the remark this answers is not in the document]
madtoinou: <p>The performance is not that bad compared to the ensemble without the probabilistic forecasts, or even the one trained with historical forecast (MAPE: 5.12017).</p><p> </p><p><code>output_chunk_length=1</code> of the forecasting models does not help for sure... </p>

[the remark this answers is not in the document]
madtoinou: <p>I used TFTModel because I liked the logic around the covariates preparation, but I can swap it for TCNModel without problem.</p>

madtoinou: <p>I tried a few settings, I found something acceptable (without using <code>regression_trian_n_points=-1</code> and <code>train_using_historical_forecasts=True</code>) that should be enough for the sake of this example notebook</p>

darts/models/forecasting/baselines.py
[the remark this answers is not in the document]
dennisbader: 👍

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -39,14 +43,18 @@ def __init__(
...
         The regression model does not leverage the covariates passed to ``fit()`` and ``predict()``.
 
         Parameters
         ----------
         forecasting_models
             List of forecasting models whose predictions to ensemble
         regression_train_n_points
-            The number of points to use to train the regression model
+            The number of points to use to train the regression model. Can be set to `-1` to use the entire series
+            to train the regressor if `forecasting_models` are already fitted and `train_forecasting_models=False`.
```
dennisbader, line 57:
```suggestion
            The number of points per series to use to train the regression model. Can be set to `-1` to use the entire series
            to train the regressor if `forecasting_models` are already fitted and `train_forecasting_models=False`.
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -63,19 +71,29 @@ def __init__(
...
             reduce the samples before passing them to the regression model. Possible values: "mean", "median"
             or float value corresponding to the desired quantile. Default: "median"
+        train_forecasting_models
+            If set to `False`, the `forecasting_models` are not retrained when calling `fit()` (only supported
+            if all the `forecasting_models` are pretrained `GlobalForecastingModels`). Default: ``True``.
+        train_using_historical_forecasts
+            If set to `True`, use `historical_forecasts()` to generate the forecasting models' predictions used to
+            train the regression model in `fit()`. Available when `forecasting_models` contains only
+            `GlobalForecastingModels` with `output_chunk_length` multiple of `regression_train_n_points` and
+            recommended when `regression_train_n_points` is considerably greater than `output_chunk_length`.
```
dennisbader, line 84: I believe that ocl doesn't have to be multiple of train_n_points anymore, right?
```suggestion
            train the regression model in `fit()`. Available when `forecasting_models` contains only
            `GlobalForecastingModels`. Recommended when `regression_train_n_points` is greater than 
            `output_chunk_length` of the underlying `forecasting_models`.
```


```suggestion
            `GlobalForecastingModels`. Recommended when `regression_train_n_points` is greater than `output_chunk_length` of the underlying `forecasting_models`.
```

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -100,34 +118,168 @@ def __init__(
...
+                logger,
+            )
+        else:
+            # can perform this check without knowing the training series length
+            self._check_ocl_multiple_train_n_point(regression_train_n_points)
+
         self.train_n_points = regression_train_n_points
 
+        raise_if(
+            train_using_historical_forecasts and not self.is_global_ensemble,
```
dennisbader, line 136: they also must be pre-trained, right?

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -100,34 +118,168 @@ def __init__(
...
         # spare train_n_points points to serve as regression target
         is_single_series = isinstance(series, TimeSeries)
-        if is_single_series:
-            train_n_points_too_big = len(self.training_series) <= self.train_n_points
-        else:
-            train_n_points_too_big = any(
-                [len(s) <= self.train_n_points for s in series]
+        if self.train_n_points == -1:
+            # take the longest possible time index
+            self.train_n_points = (
```
dennisbader, line 254: let's make train_n_points relative to each series in case of multiple series

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -173,15 +224,55 @@ def test_torch_models_retrain(self):
...
-        forecast1 = model1_fitted.predict(10)
+        assert model1_fitted.training_series.time_index.equals(
+            model2.training_series.time_index
+        )
+        assert forecast1.time_index.equals(forecast2.time_index)
+        np.testing.assert_array_almost_equal(forecast1.values(), forecast2.values())
 
-        model2.fit(self.combined)
-        forecast2 = model2.predict(10)
+    @pytest.mark.skipif(not TORCH_AVAILABLE, reason="requires torch")
```
dennisbader, line 241: torch is not used in this one
```suggestion
```

examples/19-EnsembleModel-examples.ipynb
dennisbader, line 1: add the example to the doc examples.rst and merge worklow tests

dennisbader (commented):
Nice, really only minor comments 🚀

darts/models/forecasting/regression_ensemble_model.py
madtoinou: good catch!

darts/models/forecasting/forecasting_model.py
```diff
@@ -539,12 +539,27 @@ def _historical_forecasts_sanity_checks(self, *args: Any, **kwargs: Any) -> None
...
-            last_valid_pred_time = series.time_index[-1]
-        else:
-            last_valid_pred_time = series.time_index[-forecast_horizon]
-
+            return last_possible_time_index
+
+        # otherwise, we have to step `forecast_horizon` steps back.
+        last_valid_pred_time = series.time_index[-forecast_horizon]
+        # additionally, we check whether the `last_possible_time_index` was shifted back from the overall
+        # maximum possible prediction start point (which is by definition the first time step after the
```
madtoinou, line 556: let's call it the "latest prediction start point"

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -173,15 +224,154 @@ def test_torch_models_retrain(self):
...
-        model2.fit(self.combined)
-        forecast2 = model2.predict(10)
+    @pytest.mark.parametrize("config", [(1, 1), (5, 2), (4, 3)])
+    def test_train_with_historical_forecasts_no_covs(self, config):
+        """
+        Training regression model of ensemble with output from historical forecasts instead of predict should
+        yield better results when the forecasting models are global and regression_train_n_points >> ocl.
+
+        If the ocl of the forecasting models is not a multiple of the regression_train_n_points, the regression
+        model might be trained with less points.
```
madtoinou, line 248: This paragraph should be removed since the first points of the direct prediction as now used as "padding"

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -173,15 +224,154 @@ def test_torch_models_retrain(self):
...
+            (1, 1),
+            # (5, 2),
+            # (4, 3)
+        ],
+    )
+    def test_train_with_historical_forecasts_with_covs(self, config):
+        """
+        When the output_chunk_length of the forecasting models are not multiple of regression_train_n_points and
+        train_using_historical_forecasts=True, the operation used to align the historical forecasts might reduce
+        the number of points available to train the regression model of the ensemble.
```
madtoinou, line 310: same comment as above, this problem should not arise anymore

darts/tests/models/forecasting/test_regression_ensemble_model.py
```diff
@@ -173,15 +224,154 @@ def test_torch_models_retrain(self):
...
+                LinearRegressionModel(
+                    lags=1, lags_future_covariates=[1, 2], output_chunk_length=ocl2
+                ),
+            ],
+            regression_train_n_points=regression_train_n_points,
+            train_using_historical_forecasts=True,
+        )
+
+        # TODO: prediction of historical forecasts is missing the last two values, seems to be a bug
+        # covariates have the appropriate length
```
madtoinou, line 366: check if this problem is solved after the modifications to `get_last_prediction_point`

darts/timeseries.py
```diff
@@ -21,9 +21,9 @@
     - Have a monotonically increasing time index, without holes (without missing dates)
     - Contain numeric types only
     - Have distinct components/columns names
-    - Have a well defined frequency (
-    `date offset aliases <https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases>`_
-    for ``DateTimeIndex``, and step size for ``RangeIndex``)
+    - Have a well defined frequency (`date offset aliases
+    <https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases>`_
+    for ``DateTimeIndex``, or step size for ``RangeIndex``)
```
madtoinou, line 26: the formatting is still wrong in the compiled documention, need to find a solution, maybe moving this bullet point to the bottom

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -990,10 +995,15 @@ def test_torch_auto_start_multiple_no_cov(self, model_config):
         ), f"Model {model_cls} did not return a list of historical forecasts"
         theorical_forecast_length = (
             self.ts_val_length
-            - bounds[0]  # prediction input sample length
+            - (bounds[0] - 1)  # prediction input sample length
```
madtoinou, line 998: maybe short comment to explain that the `+1` is because the last valid prediction start point is not the last timestamp in the series but the point after (if the covariates are long enough)

dennisbader (approved):
Looks great now, thank for this PR @madtoinou 🚀

PR #1820, opened 2023-06-08
Feat/MIDAS transformer

### LINKED ISSUE #1570: Add transformer that converts higher frequency time series to lower frequency

madtoinou:
Fixes #1570.

### Summary

Based on @Beerstabr PR #1668, simplified some part of the code and made the transformer invertible.

The original high frequency is imputed based on the number of component introduced by the `transform()` method and the low frequency name (which is used to generate a `pd.Timedelta`). We could also allow the user to hint the target frequency by adding an argument to the function.

When the ratio between the low and the high frequency is not a fixed number (number of days in a quarter for example, after doing from months to quarter), the code relies on a rule-based logic. For anchored offset (`W-SUN` for example), the start time of the returned `TimeSeries` is not always identical to the original `TimeSeries` so I added an `offset` argument to allow the user to adjust for it

### Other Information

Not sure if the `strip` parameter should be set to `True` by default or even part of the transform as it can induce counter-intuitive results. @Beerstabr, did you include it for any specific reason?


darts/dataprocessing/transformers/midas.py
```diff
@@ -0,0 +1,328 @@
...
+
+        # remove the rows containing NaNs at the extremes, necessary to adjust the time index
+        first_finite_row = pd.core.missing.find_valid_index(series_values, how="first")
+        last_finite_row = pd.core.missing.find_valid_index(series_values, how="last")
+        series_values = series_values[first_finite_row : last_finite_row + 1]
+
+        start_time = series.start_time()
+        shift = 0
+        # adjust the start if was shifted due to the frequency change
+        if len(series.time_index) > 1:
```
dennisbader, line 252: It's a bit tricky to understand this part.
does it only work if either the `series` starts at the same time as original series, or it ends one time step after the end of the original series?

Ideally this should work for any point in time

darts/dataprocessing/transformers/midas.py
```diff
@@ -0,0 +1,328 @@
...
+            # shift is caused by the low frequency anchoring, fitted and inversed ts have the same start
+            if np.abs(transform_time_shift) <= low_freq_timedelta:
+                start_time = orig_ts_start_time
+            # shift is caused by the low frequency anchoring, inversed ts starts after the end of the fitted ts
+            elif series.start_time() - orig_ts_end_time <= low_freq_timedelta:
+                start_time = orig_ts_end_time
+                shift = 1
+
+        # create new time index and shift it if necessary
+        new_times = pd.date_range(
```
dennisbader, line 264: let's use `darts.utils.timeseries_generation.generate_index` instead

darts/dataprocessing/transformers/midas.py
dennisbader, line 1: I noticed that the name of the time index changes with the MIDAS

```
from darts.dataprocessing.transformers import MIDAS
ms = AirPassengersDataset().load()
midas = MIDAS(rule="As")
qs = midas.fit_transform(ms1)
ms_inv = midas.inverse_transform(qs)
print(ms.time_index.name)
print(qs.time_index.name)
print(ms_inv.time_index.name)
```

out: 
```
Month
Month
time
```

darts/tests/dataprocessing/transformers/test_midas.py
dennisbader, line 1: does MIDAS work for mutliple time series as well?

dennisbader (changes requested):
Thanks @madtoinou and @Beerstabr for this, looks very good 🚀 

Had a couple of suggestions here and there and some points that we need to check such as:
- multiple time series support
- inverse transforming a series with any starting point

darts/dataprocessing/transformers/midas.py
```diff
@@ -0,0 +1,328 @@
...
+        ----------
+        .. [1] https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#anchored-offsets
+        """
+        MIDAS._verify_series(series)
+        high_freq_name = params["fitted"]["high_freq_name"]
+        orig_ts_start_time = params["fitted"]["start"]
+        orig_ts_end_time = params["fitted"]["end"]
+
+        # retrieve the number of component introduced by midas
+        n_midas_components = int(series.components[-1].split("_")[-1]) + 1
```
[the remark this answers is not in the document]
madtoinou: We are relying on the time indexes (low and high frequencies) generated by pandas.resample to find the multiple, I will try to think of a more efficient way but it will probably not be straightforward.

darts/tests/dataprocessing/transformers/test_midas.py
madtoinou: Yes, I added the corresponding step (all the logic is handled by the `FittableDataTransformer` class)

darts/dataprocessing/transformers/midas.py
```diff
@@ -0,0 +1,328 @@
...
+)
+from darts.logging import get_logger, raise_if, raise_if_not
+
+logger = get_logger(__name__)
+
+
+class MIDAS(FittableDataTransformer, InvertibleDataTransformer):
+    def __init__(
+        self,
+        rule: str,
```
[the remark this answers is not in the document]
madtoinou: It will require a lot of changes since we heavily rely on the `pd.DataFrame.resample()` which require a `DatetimeIndex`. Possible but the whole logic would have to change. WDYT?

darts/dataprocessing/transformers/midas.py
```diff
@@ -0,0 +1,328 @@
...
+        Notes
+        -----
+        The high input frequency should always relate in the same rate to the low target frequency. For
+        example, there's always three months in quarter. However, the number of days in a month varies per month. So in
+        the latter case a MIDAS transformation does not work and the transformer will raise an error.
+
+        Parameters
+        ----------
+        rule
+            The offset string or object representing target conversion. Passed on to the rule parameter in
```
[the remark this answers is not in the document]
madtoinou: Changed the docstring to accepting string only, supporting object leads to a lot of issues as we rely on the `rule_code` attribute which raise a `NotImplementedError` for "non-canonical" `DateOffset`. We could add it in the future (along with the support for RangeIndex, I think that the logic will be quite similar)

darts/dataprocessing/transformers/midas.py
```diff
@@ -0,0 +1,328 @@
...
+        # make multiple low frequency columns out of the high frequency column(s)
+        midas_df = _create_midas_df(
+            series_df=series_df,
+            low_index_datetime=low_index_datetime,
+        )
+
+        # back to TimeSeries
+        midas_ts = TimeSeries.from_dataframe(
+            midas_df,
+            static_covariates=series.static_covariates,
```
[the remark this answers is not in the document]
madtoinou: Improved the logic, let me know what you think

darts/dataprocessing/transformers/midas.py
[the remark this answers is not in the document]
dennisbader: Yes, we can do it in a future PR :)

darts/tests/dataprocessing/transformers/test_midas.py
dennisbader: Cool :)

darts/tests/dataprocessing/transformers/test_midas.py
```diff
@@ -0,0 +1,460 @@
...
+        # verify prediction time index in both frequencies
+        self.assertTrue(pred_quarterly.time_index.equals(quarterly_test_ts.time_index))
+        self.assertTrue(pred_monthly.time_index.equals(monthly_test_ts.time_index))
+
+    def test_multiple_ts(self):
+        """
+        Verify that MIDAS works as expected with multiple series of different "high" frequencies.
+        """
+        quarterly_univariate_ts = TimeSeries.from_times_and_values(
+            times=pd.date_range(start="2000-01-01", periods=12, freq="Q"),
```
dennisbader, line 326: Could we use a monthly series and a quarterly series? then we can test the support for multiple ts with different high_freq to the same low freq.

madtoinou: This test is already fitting both monthly (`self.monthly_ts`) and quarterly to yearly, should it test an additional scenario?

dennisbader (approved):
I made some (hopefully last) updates:
- probabilistic supports
- leverage sliding windows for MIDAS.transform() instead of iterating through the dataframe
- reduce upsampling to a minumum

PR #1824, opened 2023-06-09
Fix/weights loading parameters

dennisbader:
### Summary
- now model weights can be loaded from a model with differen learning rate schedulers, optimizers, etc...


madtoinou (approved):
LGTM, thank you for fixing my flawed parameters check 🚀

PR #1829, opened 2023-06-14
Fix #1049 - add prior_scale and mode arguments to prophet model's add_seasonality

### LINKED ISSUE #1049: add_seasonality does not expose the condition_name parameter for the prophet model
The FB prophet add_seasonality method has a `condition_name` input variable that is not exposed by darts. This severely limits how custom seasonalities can be added.

The `add_seasonality` method in `Prophet` class should also take the `condition_name` as a parameter.

Additionally, the current method which adds the serasonality also ignores the `prior_scale` and `mode` parameters.

```

# prophet_model.py
for seasonality_name, attributes in self._add_seasonalities.items():
            self.model.add_seasonality(
                name=seasonality_name,
                period=attributes["seasonal_periods"] * interval_length,
                fourier_order=attributes["fourier_order"],
            )
# does not pass the mode, prior_scale and the condition_name parameters
```

id5h:
Fixes #1049 .

### Summary

When fitting a prophet model, the add_seasonality function will now properly add the prior_scale and mode to the custom seasonality, in accordance with the behaviour described in the docs.

### DIFF
--- a/darts/models/forecasting/prophet_model.py
+++ b/darts/models/forecasting/prophet_model.py
@@ -186,6 +186,8 @@ class Prophet(FutureCovariatesLocalForecastingModel):
                 name=seasonality_name,
                 period=attributes["seasonal_periods"] * interval_length,
                 fourier_order=attributes["fourier_order"],
+                prior_scale=attributes["prior_scale"],
+                mode=attributes["mode"],
             )
 
         # add covariates


madtoinou (commented):
Hi @id5h,

Thank you for contributing to darts. Your PR solves the issue of the "not passed" `prior_scale` and `mode` arguments, however it does not fully address the linked issue as `condition_name` remain unsupported. Would it be possible to also cover it in this PR?

id5h:
Hi @madtoinou ,

I pushed a simple update, where one can add conditional seasonalities as well.
I am not using the condition_name as an argument in the add_seasonality functionality. That is because in the FB prophet API that argument is used to identify the column that holds the mask - and that is not really necessary when using a univariate TimeSeries object.

id5h:
> Hi @id5h,
> 
> Sorry, last week was quite intense. I hope to be able to find some time during the week to review the code and run some tests.
> 
> I just re-launched the tests, let's hope Codedev does not fail (it has nothing to do with your code).

Alright. Thanks for the clarification and the prompt reply :)

darts/models/forecasting/prophet_model.py
```diff
@@ -182,10 +183,19 @@ def _fit(self, series: TimeSeries, future_covariates: Optional[TimeSeries] = Non
...
+            if attributes["condition_func"] is not None:
+                fit_df[seasonality_name] = fit_df["ds"].apply(
+                    attributes["condition_func"]
+                )
+                condition_name = seasonality_name
             self.model.add_seasonality(
                 name=seasonality_name,
                 period=attributes["seasonal_periods"] * interval_length,
                 fourier_order=attributes["fourier_order"],
+                prior_scale=attributes["prior_scale"],
```
dennisbader, line 196: thanks 👍

darts/models/forecasting/prophet_model.py
```diff
@@ -318,13 +334,20 @@ def add_seasonality(
         fourier_order: int,
         prior_scale: Optional[float] = None,
         mode: Optional[str] = None,
+        condition_func: Optional[types.FunctionType] = None,
```
dennisbader, line 337: I would suggest to use `condition_name` instead of a function and check at fit/predict time that the `future_covariates` series contains the boolean (binary) component `component_name`.

For this to work we also need to avoid adding the conditions as regressors in the following line: 
https://github.com/unit8co/darts/blob/b63543b8b6ec515c11fcb88cc54ad7728a6d654d/darts/models/forecasting/prophet_model.py#L200

I did a quick test and it seems to work fine.

dennisbader (changes requested):
Hi @id5h and thanks for this fix 🚀 
Looks good to me. I would only suggest to actually use `condition_name` similar to how Prophet does it.
At fit/predict time we can check that the user supplied the condition in the `future_covariates` series.
What do you think?

dennisbader:
Also, could you add an entry to the unreleased section in CHANGELOG.md? Thanks!

darts/models/forecasting/prophet_model.py
id5h: Moved in latest commit to use the `condition_name` and `future_covariates` instead of sending a function

id5h:
> Hi @dennisbader , thanks for the review. I implemented the changes you requested and added a test for the new functionality.

darts/tests/models/forecasting/test_prophet.py
```diff
@@ -234,3 +241,39 @@ def helper_test_prophet_model(self, period, freq, compare_all_models=False):
...
+        model.add_seasonality(
+            name="weekly_sun",
+            seasonal_periods=7,
+            fourier_order=2,
+            condition_name="is_sunday",
+        )
+
+        model.fit(ts, future_covariates=future_covariates)
+
+        forecast = model.predict(30, future_covariates=future_covariates)
```
dennisbader, line 274: also here, predicting 7 days should already be enough

darts/tests/models/forecasting/test_prophet.py
dennisbader, line 1: can you add tests that check that missing condition columns in future covariates and non-binary columns raise an error?

an example for this:
```
    model.fit(..., future_covariates=invalid_future_covariates)
...
```

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed a bug when loading the weights of a `TorchForecastingModel` trained with encoders or a Likelihood. [#1744](https://github.com/unit8co/darts/pull/1744) by [Antoine Madrona](https://github.com/madtoinou).
 - Fixed a bug when using selected `target_components` with `ShapExplainer. [#1803](https://github.com/unit8co/darts/pull/#1803) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed `TimeSeries.__getitem__()` for series with a RangeIndex with start != 0 and freq != 1. [#1868](https://github.com/unit8co/darts/pull/#1868) by [Dennis Bader](https://github.com/dennisbader).
+- Fixed an issue with `prophet_model.Prophet.add_seasonality()` to allow proper use of all passed parameters. [#1829](https://github.com/unit8co/darts/pull/#1829) by [Idan Shilon](https://github.com/id5h).
```
dennisbader, line 32: I would add this to the model improvements section.
"`Prophet` now supports conditional seasonalities, and properly handles all parameters passed to `Prophet.add_seasonality()` and model creation parameter `add_seasonalities`. ..."

dennisbader (changes requested):
Very nice, thanks @id5h .

Just some minor comments and then we can merge 🚀

darts/models/forecasting/prophet_model.py
```diff
@@ -220,6 +244,22 @@ def _predict(
...
 
+        for seasonality_name, attributes in self._add_seasonalities.items():
+            if attributes["condition_name"] is not None:
+                raise_if(
+                    future_covariates is None,
+                    f"Condition name '{attributes['condition_name']}' is required by "
+                    f"the custom seasonality '{seasonality_name}', but future_covariates is None. In addition, "
+                    f"the model should be re-trained with future_covariates.",
+                    logger,
+                )
```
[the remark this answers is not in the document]
id5h: This will be raised in case a user, for whatever reason, fits a model, calls add_seasonality() after fitting and then predict(). Hence the "model should be retrained with future_covariates" message.

darts/models/forecasting/prophet_model.py
```diff
@@ -181,23 +181,47 @@ def _fit(self, series: TimeSeries, future_covariates: Optional[TimeSeries] = Non
...
         # add user defined seasonalities (from model creation and/or pre-fit self.add_seasonalities())
         interval_length = self._freq_to_days(series.freq_str)
+        conditional_seasonality_covariates = []
+        if future_covariates is not None:
+            future_covariates_columns = future_covariates.columns
+        else:
+            future_covariates_columns = []
+
         for seasonality_name, attributes in self._add_seasonalities.items():
+            condition_name = attributes["condition_name"]
```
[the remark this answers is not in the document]
id5h: Makes sense 👍 
Done.

darts/tests/models/forecasting/test_prophet.py
id5h: Done

darts/tests/models/forecasting/test_prophet.py
id5h: Done

CHANGELOG.md
id5h: Done

darts/models/forecasting/prophet_model.py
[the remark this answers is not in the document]
id5h: The error is now raised according to the logic in `_check_seasonality_conditions()`

dennisbader (changes requested):
Really nice, thanks @id5h 🚀 
Last comment about slightly changing the error message and then we can merge 💯

id5h:
Great, thanks @dennisbader .
I will change to `raise_log` today most likely.

I have one suggestion before merging.
Because prophet accepts non-integer seasonalities, it would be nice to allow it also in darts.
I think the required change is minimal - the `seasonal_periods` type hint in `add_seasonality()` (_line 369_) and the `"dtype"` value in the `seasonality_properties` dictionary (_line 469_) need to be changed to `Union[int, float]`.

What do you think about adding this to the current PR?

id5h:
The `raise_log()` is imported now.
I also pushed the changes to accept float seasonality periods, including a small modification to the test - I hope that's fine with you.

dennisbader (changes requested):

darts/models/forecasting/prophet_model.py
```diff
@@ -377,10 +466,11 @@ def _store_add_seasonality_call(
 
         seasonality_properties = {
             "name": {"default": None, "dtype": str},
-            "seasonal_periods": {"default": None, "dtype": int},
+            "seasonal_periods": {"default": None, "dtype": Union[int, float]},
```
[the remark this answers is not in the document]
id5h: Thanks for the fix!
I pushed it together with the updated description.

dennisbader (approved):
Thanks for the contribution @id5h 🚀  I applied the last suggestion. We're good to merge now 💯

darts/models/forecasting/prophet_model.py
```diff
@@ -318,38 +396,53 @@ def predict_raw(
...
 
         Parameters
         ----------
         name
             name of the seasonality component
         seasonal_periods
-            number of timesteps after which the seasonal cycle repeats
+            number of timesteps after which the seasonal cycle repeats. This value will be multiplied by the inferred
+            number of days for the TimeSeries frequency (e.g. 365.25 for a yearly frequency) to be consistent with the
+            `add_seasonality()` method of Facebook Prophet.
```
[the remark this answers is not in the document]
id5h: Great - thanks :)

PR #1850, opened 2023-06-23
Fix NLinear for use with past covariates

felixdivo:
Previously, they caused a shape mismatch.

No issue exists for this yet.

### Summary

Only some part of `seq_last` may be re-added, in particular only the part that refers to the target. Apparently, this is the first part (see the beginning of the `self.shared_weights == True` case).

### Other Information

None.


felixdivo:
Oh, I've just seen that there is #1583!

madtoinou:
As mentioned above, this is kind of a duplicate of #1583, closing this to centralize the discussion in a single PR.

PR #1857, opened 2023-06-27
added acf_max_lags as argument to plot_residuals and have the param d…

### LINKED ISSUE #1659: [BUG] `plot_residuals_analysis` raises ValueError for len(TimeSeries) <= 24
**Describe the bug**
In the function `plot_residuals_analysis` from the `darts.utils.statistics` library, the `max_lag` argument in the ACF plot uses the default value of 24. This causes a `ValueError` if the `len(residuals)  <= 24`. 
**To Reproduce**
1. Start with a TimeSeries `train` of len <= 24
2. Fit a model on `train`
3. Call `plot_residuals_analysis(model.residuals(train))`
4. Throws `ValueError` from the `plot_acf` call

**Expected behavior**
Irrespective of the TimeSeries length, `plot_residuals_analysis(model.residuals(train))` should produce the residual analysis results.
 - Python version: 3.9.16
 - darts version: 0.23.1

**Additional context**
Having `max_lag = len(residuals)-1` as the argument to `plot_acf()` when called within `plot_residuals_analysis` will fix the issue.

Loudegaste:
Fixes #1659: plot_residuals_analysis uses plot_acf with default lag of 24 with no option for the user to change this value, making it fail on TS smaller than n=24.

### Summary

2 modifications:
1. add a acf_max_lag argument to plot_residuals_analysis
2. defaults acf_max_lag to len(ts)-1 if TS is smaller than 24

### DIFF
--- a/darts/tests/utils/test_statistics.py
+++ b/darts/tests/utils/test_statistics.py
@@ -205,5 +205,9 @@ class PlotTestCase(DartsBaseTestClass):
     def test_statistics_plot(self):
         plot_residuals_analysis(self.series)
         plt.close()
+        plot_residuals_analysis(self.series, acf_max_lag=10)
+        plt.close()
+        plot_residuals_analysis(self.series[:10])
+        plt.close()
         plot_pacf(self.series)
         plt.close()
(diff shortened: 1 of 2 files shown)


madtoinou (approved):

dennisbader (approved):
Thanks @Loudegaste 🚀

Issue #1807, opened 2023-05-31
Add TSMixer model


alexcolpitts96:
I recently found TSMixer (http://arxiv.org/abs/2303.06053).

It is very similar to TiDE (#1726) but with a few tweaks.

It should be pretty straight forward to implement based on the implementation of TiDE (#1727).

I will try to get started on it in the next few days.

alexcolpitts96:
Google Research implementation: https://github.com/google-research/google-research/blob/master/tsmixer/tsmixer_basic/models/tsmixer.py

Details in the paper aren't great; however, the source code clears things up.

joshua-xia:
@alexcolpitts96 did you get the paper implement tsmixer_extended? it seems support past/static/future covariable features

alexcolpitts96:
I have ran into a few things with the implementation and had some other PRs that I needed to cleanup.

I managed to implement reversible instance normalization, but there is a bug in the tests that only happens during the build process within Github. 

The rest of the model is pretty straightforward, I just need to find the time. I just started a new job so I am a little short on time as of late.

meteoDaniel:
Recently Google published a paper and an article on TSMixer:
https://blog.research.google/2023/09/tsmixer-all-mlp-architecture-for-time.html

alexcolpitts96:
I started working on it roughly two months ago. I have been busy wrapping up school and starting a new job. I should have some time to clean it up over the next few weeks.

I managed to get the skeleton written, but I still need to add covariates and probabilistic forecasting. 

https://github.com/alexcolpitts96/darts/blob/tsmixer/darts/models/forecasting/tsmixer_model.py

meteoDaniel:
From my point of view that looks good. 
Why do you think you need probabilistic forecasting? Does TSMixer provide it by nature?
Within tft , probabilistic forecast is a result of the quantile loss function. Maybe I am wrong but in case you want to add this feature to TSMixer, I think you just need to run it with QuantileLoss.

thijsjls:
@alexcolpitts96 Did you have any time to work on this further? Would be interested in using this model. Also open to contribute.

StatMixedML:
IBM has released its version of the [PatchTSMixer on HuggingFace](https://github.com/huggingface/transformers/blob/v4.36.1/src/transformers/models/patchtsmixer/configuration_patchtsmixer.py). Maybe this helps to have it available in darts soon

candalfigomoro:
[...]

PR #1865, opened 2023-06-29
Added Reversible Instance Normalization

### LINKED ISSUE #1121: Reversible Instance Normalization for Accurate Time-Series Forecasting against Distribution Shift.
I've seen pretty good results using [RevIN](https://github.com/ts-kim/RevIN) with larger datasets. I'm also waiting on a license.

### LINKED ISSUE #1861: Add Reversible Instance Normalization to Torch Models
Link to paper: https://openreview.net/forum?id=cGDAkQo1C0p

I have been implementing TSMixer (#1807) and noted that it was using RevIN (Reversible Instance Normalization).

I went to the original source paper to understand the purpose of it and it seems to obtain fantastic results as essentially a preprocessing step within the models themselves that helps address shifting distributions. The paper shows improved performance on Informer, N-BEATS, and SCINet (#1860).

image

Using my implementation and plugging it into N-HiTS, I was able to improve performance by a good marging using the Energy dataset.

with:
######################################
MSE:  0.2514766273597132
MAE:  0.334557913679719

without:
######################################
MSE:  0.30861193252719554
MAE:  0.41421373603271244

My main questions are around implementation. Can we add a wrapper around the forward method in the Torch-based models? Is there some other way to add this feature without having to edit a ton of the torch models?

alexcolpitts96:
Fixes #1121, fixes #1861.

Adds RIN as a normalization layer option requested in #1121. 

This is a prerequisite component for #1861 and #1807.


darts/models/components/layer_norm_variants.py
```diff
@@ -54,3 +54,82 @@ def __init__(self, input_size, **kwargs):
...
+
+            massage_shape = not (
+                (self.axis is not -1) and (self.axis is not x.ndim - 1)
+            )
+
+            # if axis isn't the last dimension, swap it to the last dimension
+            if massage_shape:
+                x = x.swapaxes(-2, self.axis)
+
+            x = x - self.affine_bias[target_slice]
```
dennisbader, line 126: could you add a comment why target_slice is needed?

darts/models/components/layer_norm_variants.py
```diff
@@ -54,3 +54,82 @@ def __init__(self, input_size, **kwargs):
...
+            x = x / self.affine_weight[target_slice]
```
dennisbader, line 127: official implementation uses:
```
x = x / (self.affine_weight[target_slice] + self.eps*self.eps)
```
did you intentionally drop the `eps` part?

darts/tests/models/components/test_layer_norm_variants.py
```diff
@@ -25,3 +28,37 @@ def test_lnv(self):
...
+                affine, axis, input_dim = run
+
+                if (x.shape[axis] is input_dim) and affine:
+
+                    with self.assertRaises(RuntimeError):
+                        rin = ReversibleInstanceNorm(
+                            axis=axis, input_dim=input_dim, affine=affine
+                        )
+                        x_norm = rin(x, "norm")
+                        x_denorm = rin(x_norm, "denorm")
```
dennisbader, line 55: it's better to only have the line which should raise the Error inside the with body.
The Norm creation should not raise an error. Will the error always be raised in the `x_norm` line?

Then we could rewrite as:
```suggestion
                    rin = ReversibleInstanceNorm(
                        axis=axis, input_dim=input_dim, affine=affine
                    )
                    with self.assertRaises(RuntimeError):
                        _ = rin(x, "norm")
                    with self.assertRaises(RuntimeError):
                        _ = rin(x, "denorm")
```

darts/models/components/layer_norm_variants.py
```diff
@@ -54,3 +54,82 @@ def __init__(self, input_size, **kwargs):
 class LayerNorm(nn.LayerNorm):
     def __init__(self, input_size, **kwargs) -> None:
         super().__init__(input_size, **kwargs)
+
+
+class ReversibleInstanceNorm(nn.Module):
```
dennisbader, line 59: I wonder if `RINorm` or `RevINorm` would be better names, as for longer ones such as RMSNorm we also use the abbreviation.
what do you think?

darts/models/components/layer_norm_variants.py
dennisbader, line 1: How should/can the `ReversibleInstanceNorm` be used in practice (in Darts)?

The paper shows very nice results also for NBEATS. Do you have something in mind how we could integrate it to these models?

Also, can it be used in all the places as the other layer norms here? For example as `norm_type` in `TFTModel` and `TransformerModel`? If so, then we should add it to the documentation of these models.

darts/tests/models/components/test_layer_norm_variants.py
```diff
@@ -25,3 +28,37 @@ def test_lnv(self):
                 ln = layer_norm(4)
                 inputs = torch.zeros(1, 4, 4)
                 ln(inputs)
+
+        def test_rin(self):
+
+            x = torch.randn(3, 4, 7)
```
dennisbader, line 34: this test is subjected to randomness and can eventually fail. Can we address this?

dennisbader (changes requested):
Very nice, thanks a lot for this very cool addition @alexcolpitts96 🚀 :)

I only had some minor suggestions and general questions about small deviations from the official implementation and the usage of the Norm in Darts

darts/models/components/layer_norm_variants.py
alexcolpitts96: I am not sure how to apply it for probabilistic models, but I was thinking there might be some way to wrap the forward function of a model. When I added it to NHiTS it was a simple as adding RIN to __init__ and calling transform before/after. Maybe add an argument to the torch model base class, "use_reversible_instance_norm"?

For covariates and probabilistic forecasts there will need to be some adjustments, but I expect it to be possible. Ignoring covariates on the denorm is easy, but denorming more samples is not something I can immediately think of a robust solution for.

darts/models/components/layer_norm_variants.py
alexcolpitts96: I had been writing RIN, but RINorm is cleaner and feels more in line with RMSNorm.

darts/models/components/layer_norm_variants.py
```diff
@@ -54,3 +54,82 @@ def __init__(self, input_size, **kwargs):
...
+            x = self._normalize(x)
+        elif mode == "denorm":
+            x = self._denormalize(x, target_slice)
+        else:
+            raise NotImplementedError
+        return x
+
+    def _get_statistics(self, x):
+        self.mean = torch.mean(x, dim=self.axis, keepdim=True)
+        self.std = torch.sqrt(torch.var(x, dim=self.axis, keepdim=True) + self.eps)
```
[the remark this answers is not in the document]
alexcolpitts96: Totally missed those. Added them back in.

darts/tests/models/components/test_layer_norm_variants.py
alexcolpitts96: I added RNG seeding for torch and numpy.

darts/models/components/layer_norm_variants.py
alexcolpitts96: I have no idea how I missed that. I added it.

darts/models/components/layer_norm_variants.py
```diff
@@ -54,3 +54,82 @@ def __init__(self, input_size, **kwargs):
...
+        x = x / self.std
+        if self.affine:
+
+            massage_shape = not (
+                (self.axis is not -1) and (self.axis is not x.ndim - 1)
+            )
+
+            # if axis isn't the last dimension, swap it to the last dimension
+            if massage_shape:
+                x = x.swapaxes(-2, self.axis)
```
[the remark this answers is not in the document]
alexcolpitts96: I seem to recall there was an issue with the matrix multiplication not playing nice if the normalization axis wasn't the last dimension. 

I added tests to cover this because it was a problem. The original source code I was basing my implementation off of assumed the time and feature dimensions were swapped. I will double check this, but it should be correct, but there may be a faster implementation.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
   - Model creation parameter `forecasting_models` now supports a mix of `LocalForecastingModel` and `GlobalForecastingModel` (single `TimeSeries` training/inference only, due to the local models). [#1745](https://github.com/unit8co/darts/pull/1745) by [Antoine Madrona](https://github.com/madtoinou).
   - Future and past covariates can now be used even if `forecasting_models` have different covariates support. The covariates passed to `fit()`/`predict()` are used only by models that support it. [#1745](https://github.com/unit8co/darts/pull/1745) by [Antoine Madrona](https://github.com/madtoinou).
   - `RegressionEnsembleModel` and `NaiveEnsembleModel` can generate probabilistic forecasts, probabilistics `forecasting_models` can be sampled to train the `regression_model`, updated the documentation (stacking technique). [#1692](https://github.com/unit8co/darts/pull/#1692) by [Antoine Madrona](https://github.com/madtoinou).
+  - Added `RINorm` as a new layer normalization option. [#1121](https://github.com/unit8co/darts/issues/1121) by [Alex Colpitts](https://github.com/alexcolpitts96)
```
dennisbader, line 22:
```suggestion
  - Added `RINorm` (Reversible Instance Norm) as a new layer normalization option. [#1121](https://github.com/unit8co/darts/issues/1121) by [Alex Colpitts](https://github.com/alexcolpitts96)
```

darts/tests/models/components/test_layer_norm_variants.py
```diff
@@ -25,3 +28,24 @@ def test_lnv(self):
...
+            torch.manual_seed(42)
+
+            x = torch.randn(3, 4, 7)
+            affine_options = [True, False]
+
+            # test with and without affine and correct input dim
+            for affine in affine_options:
+
+                rin = RINorm(input_dim=7, affine=affine)
+                x_norm = rin.forward(x)
```
dennisbader, line 44:
```suggestion
                x_norm = rin(x)
```

dennisbader (changes requested):
Nice, thanks @alexcolpitts96 🚀 
Only had some last minor suggestions.
After that we can merge :)

darts/models/components/layer_norm_variants.py
```diff
@@ -54,3 +54,55 @@ def __init__(self, input_size, **kwargs):
...
+        self.input_dim = input_dim
+        self.eps = eps
+        self.affine = affine
+
+        if self.affine:
+            self.affine_weight = nn.Parameter(torch.ones(self.input_dim))
+            self.affine_bias = nn.Parameter(torch.zeros(self.input_dim))
+
+    def forward(self, x):
+        calc_dims = tuple(range(1, x.ndim - 1))
```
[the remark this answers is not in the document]
alexcolpitts96: I added some comments. I want to keep it as taking a torch.Tensor for now and not a specific data set. I may change this when it comes to integrating it with models since it may make covariate handling very gross in the models. 

When I add it to TiDE (and maybe NHiTS for testing) I will see how it looks and update it accordingly with another PR.

dennisbader (approved):
Thanks a lot @alexcolpitts96, everything is looking great now! 💯 
Will merge as soon as all tests passed.

gdevos010:
@dennisbader @alexcolpitts96 I haven't had a chance to do a deep dive into the code changes or what TIDE/TSMixer model requires, but since this MR closed my issue, I thought I would chime in on my understanding of RevIN. 

RevIN is not a layer normalization.  You can see in the [Informer](https://github.com/ts-kim/RevIN/blob/34147dfabda0e0b9a4f3e78e30bd56ed1fccbf82/baselines/Informer2020/models/model.py#L77-L96) example and it implemented in [PatchTST](https://github.com/PatchTST/PatchTST/blob/bb0bc6058ddc421c02e8afe77e7e8db99f913957/PatchTST_supervised/layers/PatchTST_backbone.py#L60-L82) that it is applied at the start and end of the `forward()` method, not in-between the layers as layer norms are. To reduce the confusion, I would move it to a separate file since having it in `layer_norm_variants.py` does matter when doing we do the layer norm lookups in some of the models. 

The easiest way to implement REVIN is to add it `PLForecastingModule` as recommended [here ](https://github.com/unit8co/darts/issues/1861#issuecomment-1611137155) so that all models have access to RevIN. To me, #1121 is not completed until its added to `PLForecastingModule`. I see RevIN included in many of the new forecasting models, so I appreciate the effort to include it.

dennisbader:
You're right @gdevos010. We should add it to another file than the layer norm variants. So far RevIN is not used by any model, but @alexcolpitts96 mentioned integrating it while working on the other PRs.
I'll reopen your issue so we won't forget.

For non-probabilistic models (not using a likelihood) it will work fine at the beginning and end of the forward pass. We still have to check how to best do it for models using a likelihood.

PR #1866, opened 2023-06-29
Feat/historical_forecasts accept negative integer as start value

madtoinou:
### Summary

- Allow `historical_forecasts()` to receive a negative integer as `start` value by extending the `get_index_at_value()` method.
- Added a simple unit-test covering this scenario and updated the tests expecting exception in such cases

### Other Information

This feature is to be used in `EnsembleModel` so that `forecasting_models` can call `historical_forecasts()` instead of `predict()`, which should improve the results depending on the models being ensembled.


dennisbader (changes requested):
That's a nice attempt at supporting start < 0. 
However, changing the method get_timestamp_at_point() leads to ambiguous results as it's supposed to return the index position where the index value corresponds to the `point` value. For integer indexed TimeSeries that start at a negative index value, the method would result in the wrong index position.

Can we avoid this somehow? 

Also when a read the historical_forecasting documentation when `start` is an integer:

```
If an ``int``, the parameter will be treated as an integer index to the time index of
`series` that will be used as first prediction time.
```

As far as I can tell if we have a series with RangeIndex, then `start` must exist in the RangeIndex. So also there having an integer `start` can be ambiguous as it can either mean the position of the actual index value or just the index position.

darts/timeseries.py
```diff
@@ -2085,7 +2085,9 @@ def get_index_at_point(
             )
             point_index = int((len(self) - 1) * point)
         elif isinstance(point, (int, np.int64)):
-            if self.has_datetime_index or (self.start_time() == 0 and self.freq == 1):
+            if point < 0:
```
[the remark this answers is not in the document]
madtoinou: It's clear, I made the assumption that a RangeIndex would never start with a negative value. I don't think that the ambiguity can be solved with the type/value of the `start` without this assumption.

We could eventually accept a dict `start={"index":int, "from":"start/end"}` and treat it accordingly in `ts..get_timestamp_at_point()` and `get_index_at_point()`.

darts/models/forecasting/forecasting_model.py
```diff
@@ -628,6 +640,9 @@ def historical_forecasts(
             Note: Raises a ValueError if `start` yields a time outside the time index of `series`.
             Note: If `start` is outside the possible historical forecasting times, will ignore the parameter
             (default behavior with ``None``) and start at the first trainable/predictable point.
+        start_format
+            If set to 'index', `start` must be an integer and corresponds to the absolute position of the first point
```
dennisbader, line 644: can we rephrase this similar to how pandas describes it with labels (e.g. values of index) for "point" and positions (e.g. to positional index) for "index".
Maybe we can even rename the accepted literals?

madtoinou: Changed the name of the literal and tried to improve the phrasing, let me know what you think.

darts/models/forecasting/forecasting_model.py
```diff
@@ -610,7 +622,7 @@ def historical_forecasts(
             `min_train_series_length`.
         start
```
dennisbader, line 623: can we update the description for the integer start as well?
Maybe we change the start to something like this:

```
            Optionally, the first point in time at which a prediction is computed. This parameter supports:
            ``float``, ``int``, ``pandas.Timestamp``, and ``None``.
            If a ``float``, it is the proportion of the time series that should lie before the first prediction point.
            If an ``int``, it is either the index position of the first prediction point for `series` with a
            `pd.DatetimeIndex`, or the index value for `series` with a `pd.RangeIndex`. The latter can be changed to 
            the index position with `start_format="position"`.
            If a ``pandas.Timestamp``, it is the time stamp of the first prediction point.
            If ``None``, the first prediction point will automatically be set to:
```

dennisbader (approved):
Thanks @madtoinou this also looks great 🚀 
Just some minor suggestions mainly about the documentation. After those we can merge

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ### For users of the library:
 
+**Improvement**
+- `TimeSeries` with a `RangeIndex` starting in the negative start are now supported by `historical_forecasts`. [#1866](https://github.com/unit8co/darts/pull/1866) by [Antoine Madrona](https://github.com/madtoinou).
+- 🔴 Added a new argument `start_format` to `historical_forecasts()`, `backtest()` and `gridsearch` so that `start` can be provided as a 'positional index' (positive or negative) or a point of the time index ('value index', default behavior). [#1866](https://github.com/unit8co/darts/pull/1866) by [Antoine Madrona](https://github.com/madtoinou).
```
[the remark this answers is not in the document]
dennisbader: I think it's okay as `start` is a kwarg and at position 6, so small risk of users not using the keywords.

darts/models/forecasting/forecasting_model.py
```diff
@@ -560,6 +571,7 @@ def historical_forecasts(
         num_samples: int = 1,
         train_length: Optional[int] = None,
         start: Optional[Union[pd.Timestamp, float, int]] = None,
+        start_format: Literal["positional_index", "value_index"] = "value_index",
```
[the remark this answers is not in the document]
madtoinou: Completely agree!

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ### For users of the library:
 
+**Improvement**
```
dennisbader, line 13:
```suggestion
**Improved**
```

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
...
+- `TimeSeries` with a `RangeIndex` starting in the negative start are now supported by `historical_forecasts`. [#1866](https://github.com/unit8co/darts/pull/1866) by [Antoine Madrona](https://github.com/madtoinou).
+- Added a new argument `start_format` to `historical_forecasts()`, `backtest()` and `gridsearch`; an integer `start` can be used as a `RangeIndex` position (previously, had to be able of the index). [#1866](https://github.com/unit8co/darts/pull/1866) by [Antoine Madrona](https://github.com/madtoinou).
```
dennisbader, line 15:
```suggestion
- Added a new argument `start_format` to `historical_forecasts()`, `backtest()` and `gridsearch` that allows to use an integer `start` either as the index position or index value/label for `series` indexed with a `pd.RangeIndex`. [#1866](https://github.com/unit8co/darts/pull/1866) by [Antoine Madrona](https://github.com/madtoinou).
```

PR #1870, opened 2023-06-30
Fix: scikit-learn latest version handling

alexcolpitts96:
There have been a few issues relating to probablistic forecasting and the random number seeding.

I have had this test fail on my system and in Docker containers. The issue is that there are float differences between the values. Using np.isclose fixes the issue and, in my opinion, makes it more robust.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,23 @@
-FROM jupyter/base-notebook:python-3.9.5
+FROM ubuntu:latest
 
-RUN conda update --all -y --quiet \
- && conda install -c conda-forge ipywidgets -y --quiet \
- && conda clean --all -f -y
+# setup packages
+RUN apt-get update -y
+RUN apt-get install -y python3 python-is-python3 python3-pip default-jre
+RUN pip install --upgrade pip
 
-USER root
+# install python requirements before copying the rest of the files
+# this way we can cache the requirements and not have to reinstall them
+COPY requirements/ /app/requirements/
+RUN pip install -r /app/requirements/dev-all.txt
 
-# to build pystan
-RUN apt-get update \
- && apt-get -y install build-essential \
- && apt-get clean && rm -rf /var/lib/apt/lists/*
+# copy local files
+COPY . /app
 
-USER $NB_USER
+# set work directory
+WORKDIR /app
 
-ADD . /home/jovyan/work
+# install darts
+RUN pip install -e .
 
-WORKDIR /home/jovyan/work
-
-RUN pip install .
+# docker build . -t darts-test:latest
+# docker run -it -v $(pwd)/:/app/ darts-test:latest bash
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,20 @@
-FROM ubuntu:latest
+FROM jupyter/base-notebook:python-3.9.5
 
-# setup packages
-RUN apt-get update -y
-RUN apt-get install -y python3 python-is-python3 python3-pip default-jre
-RUN pip install --upgrade pip
+RUN conda update --all -y --quiet \
+ && conda install -c conda-forge ipywidgets -y --quiet \
+ && conda clean --all -f -y
 
-# install python requirements before copying the rest of the files
-# this way we can cache the requirements and not have to reinstall them
-COPY requirements/ /app/requirements/
-RUN pip install -r /app/requirements/dev-all.txt
+USER root
 
-# copy local files
-COPY . /app
+# to build pystan
+RUN apt-get update \
+ && apt-get -y install build-essential \
+ && apt-get clean && rm -rf /var/lib/apt/lists/*
 
-# set work directory
-WORKDIR /app
+USER $NB_USER
 
-# install darts
-RUN pip install -e .
+ADD . /home/jovyan/work
 
-# docker build . -t darts-test:latest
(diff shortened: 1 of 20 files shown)


madtoinou (approved):

alexcolpitts96:
I checked another failing PR (#1820) and it had the same issue:
`FAILED darts/tests/ad/test_scorers.py::ADAnomalyScorerTestCase::test_multivariate_componentwise_kmeans - AssertionError: 0.9900695134061569 != 0.9851 within 1e-05 delta (0.004969513406156922 difference)`

That being said, they have a different number than the failed test in this instance.
`AssertionError: 0.9900695134061569 != 0.9851 within 1e-05 delta (0.004969513406156922 difference)`

I will do some more digging to find the RNG seed problem and then a few PRs can get cleared up.

darts/tests/ad/test_scorers.py
```diff
@@ -1257,22 +1258,26 @@ def test_multivariate_componentwise_kmeans(self):
...
+        scorer_w10_cwtrue = KMeansScorer(
+            window=10, component_wise=True, n_init=10, random_state=rng_seed
+        )
         scorer_w10_cwtrue.fit(mts_train_kmeans)
         auc_roc_cwtrue = scorer_w10_cwtrue.eval_accuracy(
             anomalies_kmeans_per_width, mts_test_kmeans, metric="AUC_ROC"
         )
 
-        self.assertAlmostEqual(auc_roc_cwfalse, 0.9851, delta=1e-05)
+        self.assertAlmostEqual(auc_roc_cwfalse, 0.9841112214498511, delta=1e-05)
```
dennisbader, line 1278: you can round these similar to how it was before

darts/tests/models/forecasting/test_probabilistic_models.py
```diff
@@ -159,11 +159,11 @@ def test_fit_predict_determinism(self):
             model.fit(self.constant_noisy_ts)
             pred2 = model.predict(n=10, num_samples=2).values()
 
-            self.assertTrue((pred1 == pred2).all())
+            self.assertTrue(all(np.isclose(pred1, pred2)))
```
dennisbader, line 162: I am not 100% comfortable with changing the determinism tests to isclose.
We've had issues already with comparing float32 with float64 arrays that slipped through.

What machine are you running on, how do you launch the unit tests?

dennisbader (changes requested):
Hi @alexcolpitts96 and thanks for fixing these issues.

I'm not 100% confident in changing the tests for determinism to `isclose`, as we really want to check that they are exactly the same. What systems are you running on and how do you launch the tests?

From our side an CI-side, these tests are passing.

alexcolpitts96:
I will look to test that float32 being compared to float64 works.

I decided to take a step back and test this in a Docker container. I tested the main branch and it fails on the KMeans test.

```
FROM ubuntu:latest

# setup packages
RUN apt-get update -y
RUN apt-get install -y python3 python-is-python3 python3-pip default-jre
RUN pip install --upgrade pip

# install python requirements before copying the rest of the files
# this way we can cache the requirements and not have to reinstall them
COPY requirements/ /app/requirements/
RUN pip install -r /app/requirements/dev-all.txt

# copy local files
COPY . /app

# set work directory
WORKDIR /app

# install darts
RUN pip install -e .

# docker build . -t darts-test:latest
# docker run -it -v $(pwd)/:/app/ darts-test:latest bash
```

alexcolpitts96:
@dennisbader @madtoinou Root cause for `darts/tests/ad/test_scorers.py` is scikit-learn's latest [release](https://pypi.org/project/scikit-learn/#history). This is what is causing some builds after June 30th to break. 

From the change logs for their release:
> The sample_weight parameter now will be used in centroids initialization for [cluster.KMeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans), [cluster.BisectingKMeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.BisectingKMeans.html#sklearn.cluster.BisectingKMeans) and [cluster.MiniBatchKMeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.MiniBatchKMeans.html#sklearn.cluster.MiniBatchKMeans). This change will break backward compatibility, since numbers generated from same random seeds will be different. 

A few ideas to fix it:
1. Adjust the assert values based on sklearn version.
2. Look at the values and use greater or less than logic

I am inclined to go with 1 since it will ensure that there are only two outcomes and will allow for changes like this to be identified in the future; however, it will be more flexible if we go with 2.

I can adjust this PR once I know which option is preffered.

Dockerfile
dennisbader, line 1: If you want make changes to the Dockerfile, I'd suggest to do this in a dedicated PR :)

dennisbader (changes requested):
Thanks a lot @alexcolpitts96 for doing this thorough trace back of the issue and the fix! Great work 🚀 

Regarding the Dockerfile changes, if you want those to be part of Darts, then let's move it to a dedicated PR.

alexcolpitts96:
This should be good to go now. I will create a new PR for the Dockerfile changes.

dennisbader (approved):
Perfect, thanks a lot @alexcolpitts96 💯 🚀

Issue #1871, opened 2023-06-30
DTW alignment plot does not seem to work properly [BUG]


AhmetZamanis:
**Describe the bug**
I am using the `dtw` module to perform dynamic time warping on two univariate series. Everything seems to work fine except the `DTWAlignment.plot_alignment()` method, which does not produce the correct DTW alignment plot. It plots the two time series correctly, but the black alignment lines are just a vertical line around the start of the plot (see image below the reproducible code).

**To Reproduce**
Here's the code I used for everything. All other `DTWAlignment` methods seem to work fine except `.plot_alignment()`. The two series are univariate, and I've replicated the same issue with essentially the same code, using two other multivariate series.
```python
from darts.timeseries import TimeSeries
from darts.dataprocessing import dtw

# Retrieve the two univariate series (ToolWearMins across time steps, for Machine 1 and Machine 34)
ts_list = TimeSeries.from_group_dataframe(
  df = df.loc[df["MachineID"].isin([1, 34])],
  group_cols = "MachineID",
  time_col = "SurvivalTime",
  value_cols = "ToolWearMins"
)
ts_short = ts_list[0].with_columns_renamed("ToolWearMins", "Machine 1") # 78 time steps
ts_long = ts_list[1].with_columns_renamed("ToolWearMins", "Machine 34") # 101 time steps

# Perform DTW
alignment = dtw.dtw(ts_short, ts_long)

# Visualize the alignment (broken)
alignment.plot_alignment(series2_y_offset = 100)
_ = plt.gca().set_title("Alignment")
```
![plot_alignment_bug](https://github.com/unit8co/darts/assets/56347957/e3b50a50-b964-40ec-94b2-3c14f9a26fbe)

**Expected behavior**
I was able to manually create the correct plot using the alignment indices from `DTWAlignment.path()` (see image below). I'll share the code for this in the additional context, in case it's helpful.
![AlignmentPlotCorrect](https://github.com/unit8co/darts/assets/56347957/ed479bcf-240d-49de-b1b8-8dcd100dcc42)

**System:**
 - Python version: 3.10.8
 - darts version: 0.24.0

**Additional context**
Here's the code I used to create the second, correct alignment plot. I'm sure there's a cleaner way to do it.
```python
# Get dataframes where rows = "Machine1", "Machine34", cols = SurvivalTime, ToolWearMin (+100 for Machine34)
df_alignment = []
for pair in path:
  x1 = pair[0] + 1 
  x2 = pair[1] + 1
[...]

PR #1880, opened 2023-07-03
Fix/Make DTW alignment plot compatible with RangeIndex

AhmetZamanis:
Fixes issue #1871 by modifying the `DTWAlignment.plot_alignment()` method, so it works with both `DateTimeIndex` & `RangeIndex` time dimensions.

### Summary
Currently, `plot_alignment()` converts the x-axis coordinates for the alignment lines into DateTime (lines 154-155). The PR modifies this so the x-axis coordinates match the datatype of the series' time dimensions (DateTime or int). 

I also replaced `np.empty` with `np.zeros` (line 160), as doing `x_coords[2::3] = np.nan` raises an error (not sure why), but leaving `x_coords[2::3]` as all zeroes seems to work for RangeIndex time dimensions.

### Other Information
I tested this fix on the same series with both DateTime and RangeIndex time dimensions, and they produced the same plots.

### DIFF
--- a/darts/dataprocessing/dtw/_plot.py
+++ b/darts/dataprocessing/dtw/_plot.py
@@ -148,8 +148,8 @@ def plot_alignment(
     path = self.path()
     n = len(path)
 
-    time_dim1 = series1._time_dim
-    time_dim2 = series2._time_dim
+    time_dim1 = series1.time_dim
+    time_dim2 = series2.time_dim
 
     x_coords1 = np.array(xa1[time_dim1], dtype=xa1[time_dim1].dtype)[path[:, 0]]
     x_coords2 = np.array(xa2[time_dim2], dtype=xa2[time_dim2].dtype)[path[:, 1]]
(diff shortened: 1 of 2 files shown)


darts/dataprocessing/dtw/_plot.py
```diff
@@ -151,18 +151,19 @@ def plot_alignment(
     time_dim1 = series1._time_dim
```
dennisbader, line 151: can you change these to `series1.time_dim` and `series2.time_dim`?

darts/dataprocessing/dtw/_plot.py
```diff
@@ -151,18 +151,19 @@ def plot_alignment(
...
 
-    x_coords = np.empty(n * 3, dtype="datetime64[s]")
+    x_coords = np.zeros(n * 3, dtype=xa1[time_dim1].dtype)
     y_coords = np.empty(n * 3, dtype=np.float64)
 
     x_coords[0::3] = x_coords1
     x_coords[1::3] = x_coords2
-    x_coords[2::3] = np.datetime64("NaT")
+    if x_coords.dtype == "datetime64[s]":
+        x_coords[2::3] = np.datetime64("NaT")
```
dennisbader, line 166: Can you change lines 160-166 to the following?
The reason why `x_coords[2::3] = np.nan` failed for integer indexed series, is because only float arrays (not integer) allow NaNs 
```
    if series1.has_datetime_index:
        x_dtype = xa1[time_dim1].dtype
        x_nan = np.datetime64("NaT")
    else:
        x_dtype = np.float64
        x_nan = np.nan
    
    x_coords = np.empty(n * 3, dtype=x_dtype)
    y_coords = np.empty(n * 3, dtype=np.float64)

    x_coords[0::3] = x_coords1
    x_coords[1::3] = x_coords2
    x_coords[2::3] = x_nan
```

darts/dataprocessing/dtw/_plot.py
dennisbader, line 1: can you add a entry to the unreleased section in CHANGELOG.md with the fix?

dennisbader (changes requested):
Hi @AhmetZamanis and thanks a lot for this fix :) 🚀 

I only had two minor suggestions.

AhmetZamanis:
Thanks @dennisbader, I made the changes you required. Let me know if there is any issue.

dennisbader (approved):
Perfect, thanks a lot for your contribution @AhmetZamanis 🚀 💯

PR #1885, opened 2023-07-07
Optimisation of historical forecast for regression models

### LINKED ISSUE #1233: historical_forecast enhancement for `RegressionModel`
`RegressionModel` models type are specific models based on tabular data. There is a possibility to directly compute historical_forecast in the case **where retrain=False**, instead of using an iterative process for each time step. It can be particularly useful when some historical test set, after the train set chronologically, have to be computed to compute a test metric for example, and can save a lot of time.

We will try to use or reaadapt the same tabularization function `create_lagged_data` and will create a matrix with lags etc to direclty compute the historical forecasts.

madtoinou:
Fixes #1233

### Summary

Reduce `historical_forecasts()` runtime for all the `RegressionModels` using some tricks (enforce `retrain=False`, use boundaries of the "forecastable indexes" instead of a range, predictions are vectorized with stride).

### Other Information

`forecast_horizon > model.output_chunk_length` is not supported at the moment because it would require auto-regression ~~and `num_samples > 1` needs a bit of work before being fully supported~~.

**[EDIT]** : Some speed up statistics, also included some comparison between the refactored and "legacy" implementation.

```python
# parameters
multi_models = [True, False]
multivariate = [True, False]
forecast_horizon = [1,7]
stride = 1
start = [700, 800, 900, 990]
length_ts = 1000

ts_univariate = tg.linear_timeseries(start_value=1, end_value=length_ts, length=length_ts)
ts_multivariate = ts_univariate.stack(tg.sine_timeseries(length=length_ts))

# two models, to test forecast_horizon different or equal to output_chunk_length
model1 = LinearRegressionModel(lags=3, output_chunk_length=forecast_horizon
model2 = LinearRegressionModel(lags=3, output_chunk_length=7)

# loop iterating on all the parameters
t0 = time.time()
for i in range(replicates):
    hist_fct = model.historical_forecasts(
       series=ts,
        start=start,
        stride=stride,
        num_samples=num_samples,
        retrain=False,
        enable_optimisation=False
        )
t1 = time.time()
for i in range(replicates):
    opti_hist_fct =  model.historical_forecasts(
        series=ts,
        start=start,
        stride=stride,
        num_samples=num_samples,
        retrain=False,
        enable_optimisation=True
t2 = time.time()
```

Refactorization of the original method did not affect the performance, but it should be easier to read.

The ratio on the y axis is `(t1-t0)/(t2-t1)`

### Speed up when `last_points_only=True`
legacy_v_opti
When the conditions are met, the gain scales with the length of the forecasted period.

### Speed up when `last_points_only=False`
legacy_v_opti_allpoints
The bottleneck is the creation of the returned `TimeSeries`, the gain are less significant.


darts/utils/optimised_historical_forecasts.py
```diff
@@ -0,0 +1,427 @@
...
+            unit = freq if series_.has_datetime_index else 1
+
+            if not model.multi_models:
+                shift_start = model.output_chunk_length - 1
+
+                hist_fct_tgt_start -= shift_start * unit
+                hist_fct_pc_start -= shift_start * unit
+                hist_fct_fc_start -= shift_start * unit
+
+            if model.output_chunk_length == forecast_horizon:
```
dennisbader, line 322: can you explain why this is not required for ocl < forecast_horizon?

darts/utils/optimised_historical_forecasts.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # stride must be applied post-hoc to avoid missing values
+        X = X[0][:, :, 0]
+
+        # repeat rows for probabilistic forecast
+        forecast = model._predict_and_sample(
+            np.repeat(X, num_samples, axis=0), num_samples
+        )
+
+        require_auto_regression: bool = forecast_horizon > model.output_chunk_length
```
dennisbader, line 356: we can remove everything autoregressive related as we don't support it yet.
We can add this whenever we add support for it :)

darts/utils/optimised_historical_forecasts.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # reshape and stride the forecast into (forecastable_index, forecast_horizon, n_components, num_samples)
+        if model.multi_models:
```
dennisbader, line 359: this part is tricky to understand, could you add some more comments that describe how the output of `_predict_and_sample` is different for multi_model True/False, and share a bit more info for the slicing/reshaping.

We will be thankful in the future :)

darts/utils/optimised_historical_forecasts.py
dennisbader, line 1: This looks really great @madtoinou, congrats! This is an excellent boost historical forecasting 🚀 

Only suggestion is to give a bit more detail in some sections, as it can be tricky to wrap your head around all the different scenarios.

dennisbader (changes requested):
Great PR, thanks a lot and congratulations @madtoinou! This will give such a boost to historical forecasting and backtesting for RegressionModels! 🚀 

I left some comments and suggestions regarding what we discussed offline, to add some more information on specific parts, etc.

darts/models/forecasting/forecasting_model.py
```diff
@@ -1935,6 +2011,24 @@ def _get_default_model_params(cls):
...
+        forecast_horizon: int = 1,
+        stride: int = 1,
+        overlap_end: bool = False,
+        last_points_only: bool = True,
+        verbose: bool = False,
+        show_warnings: bool = True,
+    ) -> Union[
+        TimeSeries, List[TimeSeries], Sequence[TimeSeries], Sequence[List[TimeSeries]]
+    ]:
+        pass
```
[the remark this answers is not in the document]
madtoinou: To prevent user from calling this method directly? Wouldn't be more efficient to just directly raise an error?

Aside from raising the error, what else would you expect this method to do? Because even if the sanity check pass, it's the sub-class `_optimized_historical_forecasts()` which is responsible for calling the appropriate optimized method.

darts/utils/optimised_historical_forecasts.py
madtoinou: it's the other way around, forecast_horizon < ocl, and it's because these timestamps will be covered by the last prediction of length `output_chunk_length`. By leaving them, the prediction will go too far.

darts/utils/optimised_historical_forecasts.py
madtoinou: Updated the comments, let me know if it's enough

darts/utils/optimised_historical_forecasts.py
```diff
@@ -0,0 +1,427 @@
...
+
+    # target lags are <= 0
+    hist_fct_tgt_start, hist_fct_tgt_end = historical_forecasts_time_index
+    if min_target_lag is not None:
+        hist_fct_tgt_start += min_target_lag * freq
+    hist_fct_tgt_end -= 1 * freq
+    # past lags are <= 0
+    hist_fct_pc_start, hist_fct_pc_end = historical_forecasts_time_index
+    if min_past_cov_lag is not None:
+        hist_fct_pc_start += min_past_cov_lag * freq
```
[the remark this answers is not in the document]
madtoinou: Since `historical_forecasts_time_index` already corresponds to the "forecastable target" indexes, we could leave all the work to the tabularization module and pass the entire covariates series. I wanted to minimize the length of these series while making sure that tabularization had enough points to work with:
- push back the start of past covariates if some lags are defined
- push back the start of future covariates if some lags are negative
- push forward the end of future covariates if some lags are positive

I am going to push around the boundaries in all scenario, for extensiveness but I don't think that it'll make a big difference

darts/utils/optimised_historical_forecasts.py
```diff
@@ -0,0 +1,427 @@
...
+                    idx=series_idx,
+                    start=start,
+                    start_time_=start_time_,
+                    historical_forecasts_time_index=historical_forecasts_time_index,
+                )
+        else:
+            historical_forecasts_time_index = (
+                max(historical_forecasts_time_index[0], start_time_),
+                historical_forecasts_time_index[1],
+            )
```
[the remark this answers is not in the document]
madtoinou: I partially modularized this part, clearly separating these two aspects is quite difficult, let me know if it's enough!

darts/models/forecasting/regression_model.py
```diff
@@ -851,6 +856,151 @@ def supports_future_covariates(self) -> bool:
...
+            raise_log(
+                ValueError(
+                    "As `forecast_horizon > model.output_chunk_length`, historical forecast "
+                    "requires auto-regression which is not supported in this optimized routine."
+                ),
+                logger,
+            )
+
+        # retrieve stored covariates, usually handled by RegressionModel.predict()
+        if past_covariates is None and self.past_covariate_series is not None:
```
dennisbader, line 932: in the current implemenetation models using encoders are not yet optimizable.
Do you think we could add support for this?

dennisbader (changes requested):
Very, very cool stuff @madtoinou 🚀 

I really only had some minor suggestions :)
- add some tests
- add support for optimization with encoders (or at least make it non-optimizable if they use encoders)

We can

darts/models/forecasting/regression_model.py
```diff
@@ -851,6 +856,151 @@ def supports_future_covariates(self) -> bool:
     def supports_static_covariates(self) -> bool:
         return True
 
+    @property
+    def supports_optimized_historical_forecasts(self) -> bool:
```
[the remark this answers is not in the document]
madtoinou: I try to favour Bristish spelling in generall but don't mind using US spelling in this context.

darts/models/forecasting/regression_model.py
madtoinou: Good catch, I tend to forget about the encoders... Thank you for implementing this, I will adjust this PR as soon as the other one is merged.

darts/models/forecasting/forecasting_model.py
```diff
@@ -1997,6 +1973,28 @@ def _get_default_model_params(cls):
...
+        stride: int = 1,
+        overlap_end: bool = False,
+        last_points_only: bool = True,
+        verbose: bool = False,
+        show_warnings: bool = True,
+        predict_likelihood_parameters: bool = False,
+    ) -> Union[
+        TimeSeries, List[TimeSeries], Sequence[TimeSeries], Sequence[List[TimeSeries]]
+    ]:
+        logger.warning(
```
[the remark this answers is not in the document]
madtoinou: This method is never called for non-regression model, I wrote the warning just in case a user tries to call it directly.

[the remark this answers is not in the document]
dennisbader: True, then it's perfect :)

dennisbader (approved):
🚀 Super nice and great job @madtoinou 👏 
We're ready to merge now once unit tests have passed (i adapted slightly to reduce testing time).

Issue #672, opened 2021-12-06
[BUG] Implementation errors on `TransformerModel`.


Sangwon91:
I have looked into `[Transformer](https://github.com/unit8co/darts/blame/master/darts/models/forecasting/transformer_model.py)` and I have found some errors.

Frist,
 
In line 167, 170,
```python
src = self.encoder(src) * math.sqrt(self.input_size)
```
```python
tgt = self.encoder(tgt) * math.sqrt(self.input_size)
```

I don't think we have to multiply `math.sqrt(self.input_size)`  to inputs (`src` or `tgt`). 
Because `torch.nn.MultiheadAttention` take cares this normalization.

Second, 

In line 173 - 174,
```python
        x = self.transformer(src=src,
                             tgt=tgt)
````

There is no `tgt_mask` for this prediction. In order to use teacher forcing at training stage, user must feed `tgt_mask` to forward function (specifically `square_subsequent_mask` defined below). Otherwise decoder inputs before time t can see future decoder inputs (e.g, t+1, t+2, ...) which doesn't exist at inference stage.

```python
[docs]    @staticmethod
    def generate_square_subsequent_mask(sz: int) -> Tensor:
        r"""Generate a square mask for the sequence. The masked positions are filled with float('-inf').
            Unmasked positions are filled with float(0.0).
        """
        return torch.triu(torch.full((sz, sz), float('-inf')), diagonal=1)
```

I'm not sure these things are errors.
But, in my opinion, it seems this is not correct.

Thank you!

dennisbader:
Closing as merged in #1915.

PR #1915, opened 2023-07-23
Feature/transformer refactorisation

### LINKED ISSUE #601: Refactor Transformer model
Currently the Transformer is not really implemented as it should. We should revisit to implement it like the in original Transformer paper; including always training for predicting next sample (like language models), and calling the encoder+decoder in auto-regressive ways when producing forecasts. See: Attention Is All You Need 

 

Note from Francesco:
This current implementation is fully functional and can already produce some good predictions. However,        it is still limited in how it uses the Transformer architecture because the `tgt` input of        `torch.nn.Transformer` is not utlized to its full extent. Currently, we simply pass the last value of the        `src` input to `tgt`. To get closer to the way the Transformer is usually used in language models, we        should allow the model to consume its own output as part of the `tgt` argument, such that when predicting        sequences of values, the input to the `tgt` argument would grow as outputs of the transformer model would be        added to it. Of course, the training of the model would have to be adapted accordingly.

JanFidor:
Fixes  #601 #672

### Summary
- Added teacher forcing to the transformer model
- Added auto-regression for inference
- Changed the normalization factor to `math.sqrt(self.d_model)` (I tested deleting it altogether, but SunspotsDataset backtest MAPE was almost 2 times higher then)

### Other Information
One of the tricky parts was deciding on how to use encoders -> channel dimension for source consists of both (predicted_series and past covariates), while target only has the predicted_series. I decided to use 0 padding to substitute for the missing past covariates and use the same Linear layer for both. It seemed to give a little better results and is more in line with the original implementation.

When it comes to probabilistic forecasting on inference, I decided to just take an average over probabilistic dimension


JanFidor:
Hi @dennisbader , I've double checked the failing tests and everything seems to be running fine on my end locally. It also looks like the errors are cause by some weird pytorch interactions and I'm not 100% sure if the code is at fault or if CI/CD is flaky

dennisbader:
Hi @JanFidor and thanks for this PR. We'll have time to review next week.
For the meantime, the other unit test workflows run fine, so there is probably an issue with the changes in this PR.

JanFidor:
Quick update @dennisbader, I fixed the bug (transformer mask generation was creating tensor incorrectly) and tests are passing now, but it seems like codecov has an a connection problem. (Also, I had a tiny adventure with git history and some unpleasant rebases, but fortunately force-pushing and git reset saved the day :rocket: )

And for a neat conclusion, here's a change in the SunspotsDataset backtest performance (dataset from example notebook with a longer forecast horizon, which is where teacher-forcing should make a difference)

Old implementation performance:
Screenshot from 2023-08-07 17-02-04


New implementation performance:
Screenshot from 2023-08-07 16-52-52

darts/models/forecasting/transformer_model.py
```diff
@@ -276,47 +278,118 @@ def __init__(
             custom_decoder=custom_decoder,
         )
 
-        self.decoder = nn.Linear(
-            d_model, self.target_length * self.target_size * self.nr_params
-        )
+        self.decoder = nn.Linear(d_model, self.target_size * self.nr_params)
 
-    def _create_transformer_inputs(self, data):
+    def _permute_transformer_inputs(self, data):
```
madtoinou, line 283: if this method is reduced to a simple permutation of the dimensions, it would be better to get ride of it (and keep the comment)

darts/models/forecasting/transformer_model.py
```diff
@@ -276,47 +278,118 @@ def __init__(
...
+        pad_size = (0, self.input_size - self.target_size)
+
+        # start token consists only of target series, past covariates are substituted with 0 padding
+        start_token = self._permute_transformer_inputs(data[:, -1:, : self.target_size])
+        start_token_padded = F.pad(start_token, pad_size)
+
+        if len(x_in) == 3:
+            src, _, tgt = x_in
+            src = self._permute_transformer_inputs(src)
+            tgt_permuted = self._permute_transformer_inputs(tgt)
```
madtoinou, line 305: as mentioned above, this should be just `src.permute(1,0,2)` and `tgt.permute(1,0,2)`. The comment could be added to the docstring of this method.

darts/models/forecasting/transformer_model.py
```diff
@@ -276,47 +278,118 @@ def __init__(
...
-        src = self.positional_encoding(src)
+        src = self.encoder(src) * math.sqrt(self.d_model)
+        tgt = self.encoder(tgt) * math.sqrt(self.d_model)
 
-        tgt = self.encoder(tgt) * math.sqrt(self.input_size)
+        src = self.positional_encoding(src)
         tgt = self.positional_encoding(tgt)
 
-        x = self.transformer(src=src, tgt=tgt)
+        tgt_mask = torch.triu(
```
madtoinou, line 336: Is there a reason why you're not using the torch function [`torch.nn.Transformer.generate_square_subsequent_mask`](https://pytorch.org/docs/stable/generated/torch.nn.Transformer.html#torch.nn.Transformer.generate_square_subsequent_mask)?

darts/models/forecasting/transformer_model.py
```diff
@@ -276,47 +278,118 @@ def __init__(
...
-        predictions = out[0, :, :]
+        predictions = self._permute_transformer_inputs(out)
         predictions = predictions.view(
-            -1, self.target_length, self.target_size, self.nr_params
+            -1, target_length, self.target_size, self.nr_params
         )
 
         return predictions
 
+    # Allow teacher forcing
```
madtoinou, line 359: I would remove this comment, and add in the method docstring that this rely on teacher forcing (eventually with a link to some resources/documentation/section in the original article)

darts/models/forecasting/transformer_model.py
madtoinou, line 1: The disclaimer ([lines 594-600](https://github.com/unit8co/darts/pull/1915/files#diff-7539ea6efe65d5b74b5a17d6ccf4f07b3810f8c0d52beeb01e55d677aa9ff9bcR594-R600)) will have to be updated.

darts/models/forecasting/transformer_model.py
```diff
@@ -276,47 +278,118 @@ def __init__(
...
+        tgt = start_token_padded
+
+        predictions = []
+        for _ in range(self.target_length):
+            pred = self._prediction_step(src, tgt)[:, -1, :, :]
+            predictions.append(pred)
+            tgt = torch.cat(
+                [tgt, F.pad(pred.mean(dim=2).unsqueeze(dim=0), pad_size)],
+                dim=0,
+            )  # take average of quantiles
```
madtoinou, line 322: I think that you meant `samples` (in the darts way) instead of quantiles (since it depends on the likelihood used), I need to double check what the dimension of the target correspond to in this context.

madtoinou (changes requested):
Nice work @JanFidor 🚀 (and sorry for the delay)!

I wrote some high level comments before testing the model locally, let me know what you think.

I am not sure to understand your comment about `past_covariates` being unavailable for the start token, would you mind detailing this a bit?

JanFidor:
Thanks for the review @madtoinou ! About the covariates, as I understand it, they're expected to cover the same range of timestamps as the past_series (timestamps `1:L` where L is `input_chunk_length`). It's not a problem for the TransformerEncoder input, but the TransformerDecoder requires timestamps `L:L+H-1` where H is `output_chunk_length`. I couldn't find a way to access past covariates for `L+1:L+H-1`. I decided to drop the past_covariates values for the start token to generalize endocing input for TransformerDecoder-> only target channels have non-zero values. I was worried that otherwise encoding for TransformerDecoder could be unstable. For horizon of 2, first token would have diverse covariates, while second one's would be zero. I thought about extending the covariates of first token to the rest of target_series, but I was worried that the model could become to focused on them which would result in sub-optimal TransformerDecoder encodings.

darts/models/forecasting/transformer_model.py
```diff
@@ -8,6 +8,8 @@
 
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
+from torch.nn import Transformer
```
madtoinou, line 12: maybe import just `generate_square_subsequent_mask`

darts/models/forecasting/transformer_model.py
```diff
@@ -349,7 +410,7 @@ def __init__(
         The multi-head attention mechanism is highly parallelizable, which makes the transformer architecture
         very suitable to be trained with GPUs.
 
-        The transformer architecture implemented here is based on [1]_.
+        The transformer architecture implemented here is based on [1]_ and uses teacher forcing [2]_.
```
madtoinou, line 413: I think that I removed the reference entry by accident, can you please put it back? Instead of the toward data science article, could you please link the torch tutorial : https://github.com/pytorch/examples/tree/main/word_language_model

madtoinou (commented):
Thanks for adjusting the code.

Before approving, I would like to run some checks to make sure that there is no information leakage and tweak the default parameters (especially for the failing unit-test, it would be great to reach the same accuracy as before).

darts/models/forecasting/transformer_model.py
JanFidor: After reading up a little and checking out the implementation, it turns out that `generate_square_subsequent_mask` is a static method of `Transformer.` While it is possible to import just it (https://stackoverflow.com/questions/48178011/import-static-method-of-a-class-without-importing-the-whole-class) I don't think it's worth it. That said, I definitely agree that this import is a little intuitive and I think that a nice middle ground would be adding an implementation of `generate_square_subsequent_mask` to darts/utils as it's a very small function. What do you think?

JanFidor:
Thanks for another review and more great suggestions for improvement @madtoinou ! I'll try to play around with hyperparameters, but after reading the paper on DeepAR (https://arxiv.org/abs/1704.04110) I found a different approach to probabilistic forecasting which might help with the decrease in accuracy. DeepAR uses ancestral sampling (MonteCarlo simulation to generate each of the samples) (Section 3.1 last paragraph). It would require some changes to the implementation but I think the end result would be more expressive if the probability distribution of the errors would be known beforehand. Do you think it would be worth it considering the increase in the scope of the PR?

madtoinou:
Hi @JanFidor,

I would recommend keeping the implementation of the ancestral sampling in the `TransformerModel` for another PR so that we can more easily compare the change in performance and simplify the review process. Feel free to open an issue to track this improvement!

darts/models/forecasting/transformer_model.py
madtoinou: WDYT @dennisbader ?

darts/models/forecasting/transformer_model.py
```diff
@@ -410,7 +409,7 @@ def __init__(
         The multi-head attention mechanism is highly parallelizable, which makes the transformer architecture
         very suitable to be trained with GPUs.
 
-        The transformer architecture implemented here is based on [1]_ and uses teacher forcing [2]_.
+        The transformer architecture implemented here is based on [1]_ abd uses teacher forcing [4]_.
```
madtoinou, line 412: Small typo

JanFidor:
Fixed by [2261 ](https://github.com/unit8co/darts/pull/2261)

JanFidor:
Hi again, another PR I've revived @dennisbader @madtoinou . I've chenged the handling of starting token past_covariates, also, I've done some experiments and allowing teacher_forcing to be applied only on some batches (ex 90%). The idea being to train model in a similar regime it will work in during inference (autoregressively)  . I didn't make significant benchmarks, but tests on sunspots showed that it only reduced performance and extended training time, so teacher forcing is 'forced' during training for now. Though If you believe there might be value in providing that as an option I 'm open to adding it back. Have a good one and a sunny week :sunny:

JanFidor:
Old Implementation
image

Current Implementation
image

nice performance improvement on the tutorial notebook, sunspots dataset. Fairly slight training time increase comparatively ~20%

JanFidor:
Hi @dennisbader I totally get it if you're busy, but if possible would very much appreciate it if you let me know when you could get around to a review! Have an amazing week! ☀️ (on this pr too I'm afraid hahaha)

dennisbader (approved):
Thanks a lot for getting back to this and the updates @JanFidor 🚀 

I've applied some changes:
- integrate future_target into the general `forward()` pass `x_in` input. This unifies the behavior for all models (other models in the future might also benefit from having it), and removes the need for the custom train / val step logic.
- Correctly apply RIN to the future-target when teacher forcing
- Updated the notebooks
- Did some refactoring / cleanup of the TransformerModel logic

Logic-wise, everything is still as you implemented it. Only that I removed the `sqrt(d_model)` multiplier to be consistent. The linear layers should be able to handle this scaling out-of-the box.

From my side, everything is ready now for merging. Let me know if you want to have a look at it as well before I merge.

JanFidor:
Thanks for cleaning up everything @dennisbader ! I browsed through the changes and everything looks good. I will probably take a little more time later to learn how to better structure code from your changes (it really does look quite nice) but that's irrelevant to the merge. I believe the `sqrt(d_model)` line noticeably lowered my model evaluation MAPE in the notebook, but it looks like it's no longer the case when I tried to reproduce it. The only suggestion from me would be adding yourself as coauthor to the `CHANGELOG.md`, but that's of course up to you. Aside from that,  everything looks good and we can merge it! 
Have a great day! :sunny:

dennisbader:
Thanks @JanFidor for the update and for also for coming back to this and finishing the PR, always nice to see :) 🚀  

I added myself to the changelog entry and merging now 💯

PR #1917, opened 2023-07-25
Feature/multivariate wrapper

### LINKED ISSUE #1845: Add a multivariate wrapper around univariate models
It is currently a bit cumbersome to evaluate both multi- and univariate models in a single setting. Suppose we have one/many multivariate time series at hand and want to forecast with a multivariate model and also with (a collection of) univariate ones, per variate independently. We currently need to maintain a list of this ourselves and if we are looking at a univarite model, loop over the variates.

A wrapper around univariate models that applies `fit()`/`predict()`/... to the wrapped models, each per variate and else forwards most properties.

Have users implement it each time again.

**Additional context**
This is especially important if one wants to see how relvant the dynamics between multiple variates are. Related to #1844.

JanFidor:
Fixes #1845 .

### Summary
I've added a wrapper which allows a LocalForecastingModel to act like a GlobalForecastingModel

### Other Information
FutureCovariatesLocalForecastingModel support multivariate series, but I decided to treat them like the rest of LocalForecastingModels and train separate instances of the model on each of the components of the provided series


dennisbader:
Hi @JanFidor and thanks for this PR. #1845 is actually about a wrapper for multivariate series (a series with multiple target components / columns) rather than multiple series (multiple univariate or multivariate series in a list or sequence).

I don't think we will add a wrapper to make local models global, as this could potentially mean storing thousands of models. Also there would be a discrepancy between series used at training and prediction time.
The current global models train a single model on multiple series and accept any series at prediction time. For this wrapper, we don't know which trained model the series passed at prediction time should be mapped to.

JanFidor:
Hi @dennisbader ! Thanks for the explanation, I can definitely see why a GlobalForecastingModel wrapper would be problematic. I made some fixes and the wrapper now operates on a single TimeSeries (univariate or multivariate) and changed the name accordingly

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
+"""
+Regression ensemble model
+-------------------------
+
+An ensemble model which uses a regression model to compute the ensemble forecast.
+"""
```
madtoinou, line 6: This need to be updated

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+            model.predict(n=n, future_covariates=future_covariates)
+            if isinstance(model, FutureCovariatesLocalForecastingModel)
+            else model.predict(n=n)
+            for model in self._trained_models
+        ]
+
+        multivariate_series = concatenate(predictions, axis=1)
+        return multivariate_series
+
+    def __init__(self, model: LocalForecastingModel):
```
madtoinou, line 70: Can you move the constructor to the top of the class definition to make its docstring more visible?

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+    ]:
+        return None, None, False, self.supports_future_covariates, None, None
+
+    @property
+    def supports_multivariate(self) -> bool:
+        return True
+
+    @property
+    def supports_past_covariates(self) -> bool:
+        return False
```
madtoinou, line 125: this should rely on the base model property (even if no local model support past covariate at the moment)

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+    def supports_future_covariates(self) -> bool:
+        return self.base_model.supports_future_covariates
+
+    @property
+    def supports_static_covariates(self) -> bool:
+        return False
+
+    def _is_probabilistic(self) -> bool:
+        """
+        A GlobalForecastingModelWrappers is probabilistic if the base_model
```
madtoinou, line 137: Please update the docstring

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+        is probabilistic
+        """
+        return self.base_model._is_probabilistic()
+
+    def _supports_non_retrainable_historical_forecasts(self) -> bool:
+        return isinstance(
+            self.base_model, TransferableFutureCovariatesLocalForecastingModel
```
madtoinou, line 144: why not just using the `base_model` property?

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+            else model.predict(n=n)
+            for model in self._trained_models
+        ]
+
+        multivariate_series = concatenate(predictions, axis=1)
+        return multivariate_series
+
+    def __init__(self, model: LocalForecastingModel):
+        """
+        Wrapper around LocalForecastingModel allowing it to act like a GlobalForecastingModel
```
madtoinou, line 72: I would rephrase that, it's now about the components, not the multiple series support.

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+        The model doesn't use series supplied during predict() and instead predicts on the series it trained on.
+
+        Parameters
+        ----------
+        model
+            Model used to predict individual components
+        """
+        super().__init__()
+
+        self.base_model: LocalForecastingModel = model
```
madtoinou, line 84: I would check that the model passed is an instance of `LocalForecastingModel`. And call the attribute `model` to make it more intuitive.

madtoinou (changes requested):
Cool addition, good work @JanFidor!

I noticed some left-over from the "multiple series support" implementation and `EnsembleModel` "template" :)

In addition of exposing this class to the users, we could maybe also include it in `fit()` logic as it's done for the `RegressionModel` in `_fit_model()` (model is wrapped in `MultiOutputRegressor` when necessary). Maybe with a warning since each sub-model will not be able to leverage all the components to generate their forecasts. WDYT @dennisbader ?

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: I considered it, but I wasn't sure how rigid Darts code style is when it came to accessing private variables, so I decided to err on the side of caution and check it manually. That said, if it's not a problem I'd lean towards your suggestions which is significantly more readable

JanFidor:
Again, thanks for the review @madtoinou ! You caught me red-handed, I was certain I didn't leave anything from the  original  "template" ; ) . Btw. it seems like something happened to StatsForecastAutoETS, because it breaks CI/CD on a few unrelated PRs (this one and https://github.com/unit8co/darts/pull/1915 ) included.

madtoinou:
@JanFidor The error in the CI/CD was caused by a statsforacasting release where they fixed a bug that the unittest was actually "relying on", everything should run smoothly now.

I'll try to review your PR by the end of the weekend, thank you for addressing my points.

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,140 @@
...
+
+logger = get_logger(__name__)
+
+
+class MultivariateForecastingModelWrapper(FutureCovariatesLocalForecastingModel):
+    def __init__(self, model: LocalForecastingModel):
+        """
+        Wrapper around LocalForecastingModel allowing it to predict multivariate TimeSeries.
+
+        A copy of the provided model will be trained on each of the components of the provided series separately.
```
madtoinou, line 26: Wrapper for univariate LocalForecastingModel to enable multivariate series training and forecasting.

A copy of the provided model will be trained independently on each component of the target series, ignoring the potential interactions.

darts/timeseries.py
```diff
@@ -5253,3 +5253,10 @@ def concatenate(
         )
 
     return TimeSeries.from_xarray(da_concat, fill_missing_dates=False)
+
+
+def split_multivariate(time_series: TimeSeries):
```
madtoinou, line 5258: I don't think that this deserve a method

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,140 @@
...
+        components = split_multivariate(series)
+        if self.supports_future_covariates:
+            self._trained_models = [
+                self.model.untrained_model().fit(c, future_covariates)
+                for c in components
+            ]
+        else:
+            self._trained_models = [
+                self.model.untrained_model().fit(c) for c in components
+            ]
```
madtoinou, line 51: To make this more intuitive, I would rather loop on the components on the series directly and use an in-line if-else:

```suggestion
        for comp in series.components:
            self._trained_models.append(
                self.model.untrained_model().fit(
                    series=series[comp],
                    future_covariates=future_covariates if self.supports_future_covariates else None
                )
            )
```

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,140 @@
...
+        **kwargs,
+    ) -> TimeSeries:
+        predictions = [
+            model.predict(n=n, future_covariates=future_covariates)
+            if isinstance(model, FutureCovariatesLocalForecastingModel)
+            else model.predict(n=n)
+            for model in self._trained_models
+        ]
+
+        multivariate_series = concatenate(predictions, axis=1)
```
madtoinou, line 82: Let's get ride of this intermediary steps and directly return the concatenation

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,140 @@
...
+        future_covariates: Optional[TimeSeries] = None,
+        num_samples: int = 1,
+        verbose: bool = False,
+        **kwargs,
+    ) -> TimeSeries:
+        predictions = [
+            model.predict(n=n, future_covariates=future_covariates)
+            if isinstance(model, FutureCovariatesLocalForecastingModel)
+            else model.predict(n=n)
+            for model in self._trained_models
```
madtoinou, line 79: Instead of the `isinstance`, rather use the `supports_future_covariates`.

Have a single call the predict with an inline if-else to pass the future_covariates (check suggestion in the `fit()`).

Also, it would be great to at least check that there is the same number of components in the forecasted series as models in the `self_trained_models` attribute.

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def predict(
+        self,
+        n: int,
+        series: Optional[TimeSeries] = None,
+        future_covariates: Optional[TimeSeries] = None,
+        num_samples: int = 1,
+        **kwargs,
+    ) -> TimeSeries:
+        # we override `predict()` to pass a non-None `series`, so that historic_future_covariates
```
madtoinou, line 63: I don't see the `not-None` series being passed?

darts/tests/models/forecasting/test_multivariate_forecasting_model_wrapper.py
madtoinou, line 1: Your tests are not matching the style of the others tests, please look how they are defined in test_local_forecasting_models.py (especially the parametrize decorator and the helper functions)

darts/tests/models/forecasting/test_multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,107 @@
...
+            series_univariate,
+            series_multivariate,
+        ]
+
+        for combination in combinations:
+            preds = self.trained_model_predictions(
+                model, self.n_pred, combination, future_covariates
+            )
+            assert isinstance(preds, TimeSeries)
+            assert preds.n_components == combination.n_components
```
madtoinou, line 102: Would be nice to check that local models trained manually, on each component of a series, does return the exact same forecasts as one in the wrapper.

madtoinou (changes requested):
Hi @JanFidor,

A few additional comments about the docstring and simplification of the logic in `fit()` and `predict()`.

felixdivo:
Awesome that you takled this @JanFidor!

Do you plan to work on this further? I could, for example, help by adapting/extending the tests. I really want to soo this feature be completd since it makes my work quite a bit simpler. Do you want me to create a small PR to you branch `JanFidor:feature/multivariate-wrapper` for that?

JanFidor:
Hi @felixdivo , thanks for the message! Long story short, this university semester had much more work than I anticipated so I went into survival mode for a bit hahaha. As soon as the semester ends (~3 weeks) I'm planning on going back to my old PRs and get them merged. If you feel up for writing some tests I'd be very thankful, but I'd still need to survive until the exams end to get the PR merged (I'm really sorry, but this month is especially terrible when it comes to university workload so there's no way I'll be able to get back to it any faster)

felixdivo:
Hey @JanFidor, I feel you; this can definitely happen. I'll open a small PR patching a few things (like the tests, as advised by @madtoinou ), and you can just have a look once you're back from survival in creative mode. :) Feel free to reach out for help then. Good luck!

felixdivo:
I'd maybe rename the `model` in the `MultivariateForecastingModelWrapper` to something like `_template`, since it is never actually used to make forecasts itself, but only as a blueprint to generate copies form.

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Hi, I tried the approach with if else for passing the covariates, but it was throwing errors for models which didn't support future covariates

madtoinou: Because they were trained without the `future_covariates` (and hence theoretically support them but not in this context, at prediction time?) or because they actually did not support them?

darts/tests/models/forecasting/test_multivariate_forecasting_model_wrapper.py
felixdivo: This is now being tested.

darts/tests/models/forecasting/test_multivariate_forecasting_model_wrapper.py
felixdivo: Should be fixed now (8c1b573b9e1ef98ed84969db93c53b866d199c6a)

madtoinou: Instead of loops in the `test_fit_predict_local_models()` and `test_fit_predict_local_future_covariates_models()` what will make the test fail if one of the model fails, without indicating which one, it would be great to leverage `pytest.mark.parametrize` and replace the list of instantiated models with a list of kwargs (one for local, one for local that supports future covariates), that would then be used to create the model before fit/predict (as done [here](https://github.com/unit8co/darts/blob/master/darts/tests/models/forecasting/test_global_forecasting_models.py)).

Also, testing models that support future covariates without actually passing future covariates in `fit()` would be great.

darts/tests/models/forecasting/test_multivariate_forecasting_model_wrapper.py
madtoinou: Here, would be great to also leverage the `pytest.mark.parametrize` decorator and make the input series (univariate/multivariate) one of the parameter (in addition of the model, see comment above). So that if the test fail, it will be clear which series was involved.

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Because they actually did not support them, some of the LocalForecastingModels don't support future covariates, so an error is thrown about `future_covariates` not being a parameter for fit() and predict()

felixdivo:
@madtoinou are there unaddressed iossues in the PR? Maybe marking some comments above as resolved (if that is the case) helps maintain a better overview.

madtoinou:
Closed the conversations that were addressed, the comments about `pytest.mark.parametrize` are still valid.

felixdivo:
@JanFidor can you have a look at them?

darts/tests/models/forecasting/test_multivariate_forecasting_model_wrapper.py
JanFidor: Made the update, should be resolved!

JanFidor:
Hi @felixdivo @dennisbader @madtoinou  , due to some external circumstances I unfortunately took an extended break from OSS. Fortunately I'm back and this time I hope with a little more experience, in the one or 2 weeks I'll take a look at my other hanging PRs and decide whether to update them (and hopefully merge them) or close them for good. I do apologize the break took this long. For now here's a fixed and updated version of the MutivariateWrapper, best regards and have a sunny week :sunny:

JanFidor:
Also, there was a great suggestion from @felixdivo to at some point support parallel model training. Personally I like the idea a lot, but I'd rather create a separate issue for this functionality amd make sure the core model gets merged first. Let me know what you think

JanFidor:
Hi @dennisbader I totally get it if you're busy, but if possible would very much appreciate it if you let me know when you could get around to a review! Have a lovely week! ☀️

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+    TransferableFutureCovariatesLocalForecastingModel,
+)
+from darts.timeseries import TimeSeries, concatenate
+from darts.utils.ts_utils import seq2series
+
+logger = get_logger(__name__)
+
+
+class MultivariateForecastingModelWrapper(FutureCovariatesLocalForecastingModel):
+    def __init__(self, model: LocalForecastingModel):
```
dennisbader: Could we adapt this to be more in line with how we do it for example in `StatsForecastModel` (see [here](https://github.com/unit8co/darts/blob/006926315f8eafc63d8c69a90e1ad07a490eb896/darts/models/forecasting/sf_model.py#L45)) with `model` and `model_kwargs`?

This has some benefits:
- allow to setup the model via config files (e.g. yaml, ...) without any python objects
- in Darts, all model input parameters are stored under `model_params`. Storing only the model name and it's input parameters is more efficient.

```python
model = MultivariateForecastingModelWrapper("AutoETS", {"season_length": 12})
```

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+    LocalForecastingModel,
+    TransferableFutureCovariatesLocalForecastingModel,
+)
+from darts.timeseries import TimeSeries, concatenate
+from darts.utils.ts_utils import seq2series
+
+logger = get_logger(__name__)
+
+
+class MultivariateForecastingModelWrapper(FutureCovariatesLocalForecastingModel):
```
dennisbader: How about a shorter name like `MultivariateModelWrapper`? In Darts, we never include the "Forecasting" in the name. Same for the file name.


Also, we should better subclass from `TransferableFutureCovariatesLocalForecastingModel` to enable full support for all models. Then we also have to update the public facing method signatures

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+    def _fit(
+        self,
+        series: TimeSeries,
+        future_covariates: TimeSeries | None = None,
+        verbose: bool | None = None,
+    ):
+        super()._fit(series, future_covariates)
+        self._trained_models = []
+
+        series = seq2series(series)
```
dennisbader: Series is already a TimeSeries

```suggestion
```

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def predict(
+        self,
+        n: int,
+        series: TimeSeries | None = None,
+        future_covariates: TimeSeries | None = None,
+        num_samples: int = 1,
+        **kwargs,
+    ) -> TimeSeries:
+        return self._predict(n, future_covariates, num_samples, **kwargs)
```
dennisbader: make sure to pass all input parameters to `_predict`

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+        verbose: bool | None = None,
+    ):
+        super()._fit(series, future_covariates)
+        self._trained_models = []
+
+        series = seq2series(series)
+        for comp in series.components:
+            comp = series.univariate_component(comp)
+            component_model = (
+                self.model.untrained_model().fit(
```
dennisbader: we need to inspect what `fit()` parameters the underlying model supports and only pass those. For example passing the kwarg `future_covariates` is not supported by regular `LocalForecastingModel`

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+    def _predict(
+        self,
+        n: int,
+        future_covariates: TimeSeries | None = None,
+        num_samples: int = 1,
+        verbose: bool = False,
+        **kwargs,
+    ) -> TimeSeries:
+        predictions = [
+            model.predict(n=n, future_covariates=future_covariates)
```
dennisbader: same comment here with inspecting what the underlying model supports and only passing the supported params

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+            if self.supports_future_covariates
+            else model.predict(n=n)
+            for model in self._trained_models
+        ]
+
+        raise_if_not(
```
dennisbader: this is not needed here. You iterate over all trained models, so it is guaranteed that the predictions have the expected length

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+        self,
+    ) -> tuple[
+        int | None,
+        int | None,
+        bool,
+        bool,
+        list[int] | None,
+        list[int] | None,
+    ]:
+        return None, None, False, self.supports_future_covariates, None, None
```
dennisbader: why not use `self.model._model_encoder_settings`?

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+    @property
+    def supports_future_covariates(self) -> bool:
+        return self.model.supports_future_covariates
+
+    @property
+    def supports_static_covariates(self) -> bool:
+        return self.model.supports_static_covariates
+
+    @property
+    def _is_probabilistic(self) -> bool:
```
dennisbader: this is method is not available anymore in ForecastingModel.

Can you go through all the ForecastingModel properties, make sure we have all of them here as well and connect them to `self.model`?

dennisbader (changes requested):
Thanks a lot @JanFidor for getting back to this PR and the updates :) 

It's in a good state already. I've added a couple of suggestions.

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Sounds like a very good idea considering the benefits you outlined. From what I checked out it seems `darts/models/forecasting/ensemble_model.py` isn't in line with this approach, what do you think about making an issue for updating it to be in line with your suggestion? (i didn't find an already existing issue for it)

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Sounds good, code is updated

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Good catch, code is updated

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: added an if condition

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: added aggregation of params to kwargs based on model support and passing them with ** to the model

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Good catch, removed it

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: That, is actually a very good question, used it instead. (It's something i implemented back in the day, and didn't think to check if I can make it simpler when updating the PR)

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Checked all the properties and added missing ones, I think I got them all, but here's a quick list you can verify:
```
extreme_lags
_target_window_lengths
_model_encoder_settings
supports_multivariate
supports_past_covariates
supports_future_covariates
supports_static_covariates
supports_sample_weight
supports_probabilistic_prediction
supports_likelihood_parameter_prediction
supports_transferable_series_prediction
likelihood
_supports_range_index
_supports_non_retrainable_historical_forecasts
_supress_generate_predict_encoding
```

JanFidor:
Thank you @dennisbader  for the lovely message and great suggestions ^^! Another thing I thought a little about is making a notebook for this wrapper. I believe that it would be useful to showcase its usage when training & evaluating baselines (local models) in multivariate settings. Would be happy to hear your thoughts on the value of adding such a notebook! (In any case it would be a new PR, as this one is already quite the long running one hahaha). 
As always, have a lovely and a sunny (within reason as the temperatures are getting a little too high for my taste in my city) day :sunny:

darts/models/forecasting/multivariate_forecasting_model_wrapper.py
JanFidor: Modified code so that `predict() `only moves the input to the `_predict()` method. Specific logic is handled by the wrapped models

dennisbader (approved):
Thanks for the updates @JanFidor, looking great now 🚀 

I've updated some parts to be ready for merging:

- rename model to `MultivariateModel` to further simplify the naming
- Allow any forecasting model to be used as a base model
- Reference to the first model in the list of models as base model (since we want to know about the actual "trained" model, not only the model state before training)
- added the model to more unit tests
- added the model to the docs

JanFidor:
Thanks for adding the finishing touches!  As always, contributing was a pleasure 🔥

PR #1932, opened 2023-08-03
Solve some warnings triggered in unittests

madtoinou:
### Summary

Solving various warning messages appearing when running the unit-tests:
- removed deprecated `loffset` argument of `xarray.resample`, used by `TimeSeries.resample`.
- removed deprecated `verbose` argument of `granger_stationarity_test`
- `loss_fn` and `metrics` attributes of `TorchForecastingModels` checkpointing is handled in `on_save/on_load_checkpoints` and ignored in the constructor call to `save_hyperparameters`
- target series is passed to `create_lagged_prediction_data` only when the corresponding lags are defined (optimized historical forecast and shap explainer)
- closing matplotlib figures created in `TFTExplainer` unit-tests

### Other Information

Majority of remaining warnings are caused by statsforecast (ill-conditioned matrix, failure to converge, ...) or p-values outside of the tables that would require changing the dataset used.
Some warnings are caused by the PL callback `ModelCheckpoint` (created in the constructor) tracking the `val_loss`  when `save_checkpionts=True` even if no `val_series` is passed to the `fit()` method. Cannot think of a simple and elegant fix for this.

### DIFF
--- a/darts/explainability/shap_explainer.py
+++ b/darts/explainability/shap_explainer.py
@@ -751,9 +751,11 @@ class _RegressionShapExplainers:
         lags_future_covariates_list = self.model.lags.get("future")
 
         X, indexes = create_lagged_prediction_data(
-            target_series=target_series,
-            past_covariates=past_covariates,
-            future_covariates=future_covariates,
+            target_series=target_series if lags_list else None,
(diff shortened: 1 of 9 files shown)


dennisbader (approved):
Looks good, thanks a lot :) 🚀

PR #1933, opened 2023-08-04
Fix/unittests m1

dennisbader:
### Summary
Fixes unit tests to run locally on M1 devices
- run all torch models on CPU
  - by default most time series are float64 which is not supported by M1 GPU
  - not all torch functionalities that we use in our TorchForecastingModel are supported on M1 GPU
- fixes a bug in TiDE model where it was not possible to predict with `n < output_chunk_length`

### DIFF
--- a/darts/tests/models/forecasting/test_global_forecasting_models.py
+++ b/darts/tests/models/forecasting/test_global_forecasting_models.py
@@ -157,7 +157,7 @@ if TORCH_AVAILABLE:
 
     class GlobalForecastingModelsTestCase(
         DartsBaseTestClass
-    ):  # for running locally on M1 devices
+    ):
         # forecasting horizon used in runnability tests
         forecasting_horizon = 12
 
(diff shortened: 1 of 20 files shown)


darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -102,26 +125,39 @@
...
+                "pl_trainer_kwargs": tfm_kwargs["pl_trainer_kwargs"],
             },
             100,
         ),
     ]
 
-    class GlobalForecastingModelsTestCase(DartsBaseTestClass):
+    class GlobalForecastingModelsTestCase(
+        DartsBaseTestClass
+    ):  # for running locally on M1 devices
```
madtoinou, line 160:
```suggestion
    ):
```

madtoinou (approved):
LGTM, some minor comments about typo or coherence

PR #1952, opened 2023-08-10
Fix/Robuster parameters check when loading weights

### LINKED ISSUE #1946: [BUG] TFT model load failed from saved model.
**Describe the bug**

I get following error when I try to load TFT model from file.

ValueError: The values of the hyper parameter categorical_embedding_sizes should be identical between the instantiated model (None) and the loaded checkpoint ({}). Please adjust the model accordingly.

**To Reproduce**
I just save trained model and load it from file. 
And I get following error when I try to load model from file.

ValueError: The values of the hyper parameter categorical_embedding_sizes should be identical between the instantiated model (None) and the loaded checkpoint ({}). Please adjust the model accordingly.

This error is gone when I add following parameter to the model.

categorical_embedding_sizes={}

**Expected behavior**

 - Python version: 3.10
 - darts version: 0.25

**Additional context**

madtoinou:
Fixes #1946.

### Summary

Instead of comparing the `saved_model.model_params` with the hyper-parameters contained in the `.ckpt`, compare them with the `loaded_model.model_params`. Parameters with inconsistent names (TFM/PL Module) are now properly checked (e.g. `n_rnn_layers`/`num_layers` for `RNNModel`).

The ignored parameters remain the same to allow the user to change various parameters during the retraining.

### Other Information

Also added a new parameter `skip_checks` in case some user wants to load the weights from `.ckpt` file without depending on the `.pt` file. If set, the encoders cannot be loaded (potentially preventing direct inference) and weights loading related errors will be raised by `torch.load()` (missing keys, shape mismatch, ...).

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -2065,7 +2065,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
 
     def _check_ckpt_parameters(self, tfm_save):
         """
-        Check that the parameters used to instantiate the new model loading the weights match those of
+        Check that the positional parameters used to instantiate the new model loading the weights match those
         of the saved model, to return meaningful messages in case of discrepancies.
         """
         # parameters unrelated to the weights shape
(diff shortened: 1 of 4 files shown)


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -2058,6 +2063,42 @@ def _load_encoders(
...
+        ) + [
+            "loss_fn",
+            "torch_metrics",
+            "optimizer_cls",
+            "optimizer_kwargs",
+            "lr_scheduler_cls",
+            "lr_scheduler_kwargs",
+        ]
+        ckpt_model_params = tfm_save.model_params
+        for param_key, param_value in self.model_params.items():
```
dennisbader, line 2083: I'm not 100% but we might have to check the keys which are not in both dicts at the same time. If these keys are not in the `skipped_params`, we should raise an error as well (I believe `ModelMeta` only adds the positional arguments and the kwargs that were passed to `model_params`).

Also, can we capture all wrong parameters in a list, and then raise the error at the end for all wrong parameters?
Then the user can adjust all bad params at once.

dennisbader (commented):
Looks good thanks @madtoinou 🚀 
Had a small comment for first going through all bad params and then giving all of them in the error message.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -622,39 +587,158 @@ def create_DLinearModel(
...
+            Verify that the method comparing the parameters between the saved model and the loading model
+            behave as expected, used to return meaningful error message instead of the torch.load ones.
+            """
+            model_name = "params_check"
+            ckpt_name = f"{model_name}.pt"
+            # barebone model
+            model = DLinearModel(
+                input_chunk_length=4,
+                output_chunk_length=1,
+                work_dir=self.temp_work_dir,
```
dennisbader, line 677:
```suggestion
                work_dir=self.temp_work_dir,
                n_epochs=1,
```

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -622,39 +587,158 @@ def create_DLinearModel(
...
                     model_path_manual, map_location="cpu"
                 )
+            self.assertTrue(
+                str(error_msg.value).startswith(
+                    "The values of the hyper-parameters in the model and loaded checkpoint should be identical.\n"
+                    "incorrect"
+                )
+            )
+
+        def test_load_weights_from_checkpoint_params_check(self):
```
dennisbader, line 666:
```suggestion
        def test_load_weights_params_check(self):
```

dennisbader (approved):
Looks great 🚀 
Just had some last minor comments, after which we can merge

darts/utils/likelihood_models.py
```diff
@@ -232,6 +233,20 @@ def __eq__(self, other) -> bool:
...
 
+    def __repr__(self) -> str:
+        """Return the class and parameters of the instance in a nice format"""
+        cls_name = self.__class__.__name__
+        # only display the constructor parameters as user cannot change the other attributes
+        init_signature = inspect.signature(self.__class__.__init__)
+        params_string = ", ".join(
+            [
+                f"{str(v)}"
+                for _, v in init_signature.parameters.items()
```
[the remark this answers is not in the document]
madtoinou: It does display the name and the value (I don't exactly know why, the format of this dict is unexpected).

```python
>>> lkl = GaussianLikelihood()
>>> print(lkl)
GaussianLikelihood(prior_mu=None, prior_sigma=None, prior_strength=1.0, beta_nll=0.0)
```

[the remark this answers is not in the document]
dennisbader: 👍

PR #1957, opened 2023-08-16
Fix/no lambda function in add_encoders

### LINKED ISSUE #1843: [BUG] PicklingError: "Can't pickle : attribute lookup  on __main__ failed" happend when fit model
**Describe the bug**
when I test the model (maybe any forecast model, I test TFTModel, NlinearModel etc.)
I set the following parameters of model:

force_reset=True,
save_checkpoints=True,
optimizer_kwargs={"lr": results.suggestion()},
pl_trainer_kwargs={"callbacks": [early_stopper]},
add_encoders={
    'cyclic': {'future': ['month']},
    'datetime_attribute': {'future': ['hour', 'dayofweek']},
    'position': {'past': ['relative'], 'future': ['relative']},
    'custom': {'past': [lambda idx: (idx.year - 1950) / 50]},
    'transformer': Scaler(StandardScaler())
}

model.fit(
    series=[train_1, train_2, train_3],
    val_series=[val_1, val_2, val_3],
    past_covariates=[train_1_covariates, train_2_covariates, train_3_covariates],
    val_past_covariates=[val_1_covariates, val_2_covariates, val_3_covariates],
)

and train the model with validation data, it should trigger the model save the best model checkpoint.

but the following error happened:

    634 pickler.persistent_id = persistent_id
    635 print(type(obj))
--> 636 pickler.dump(obj)
    637 data_value = data_buf.getvalue()
    638 zip_file.write_record('data.pkl', data_value, len(data_value))

PicklingError: Can't pickle  at 0x2bc431280>: attribute lookup  on __main__ failed

when I remove add_encoders then the train work well.
I think the add_encoders lead to something can NOT dump correctly.

Anybody take a look this issue? it block me the model performance testing.

Thanks!

**To Reproduce**
see the description.

**Expected behavior**
model should be train and save the best checkpoint

 - Python version: 3.8
 - darts version  0.24.0

madtoinou:
Fixes #1843.

### Summary

- Raise an error when a lambda function is used in the `add_encoders` parameter so that model pickling works as intended when saving the model.
- Updated documentation and notebooks accordingly

### Other Information

- Since all the examples in the documentation (notebook included) used lambda function, this will probably cause issue for a lot of user.
- Instead of an exception, we could eventually just display a warning so that users can eventually use another pickling module that user lambda functions.


dennisbader (commented):
Looks good, thanks @madtoinou 🚀 
Only thing left is to update the `add_encoders` model docs that still use the lambda functions.

dennisbader (approved):
Nice, thanks @madtoinou 🚀 
Last thing to do is add it to the changelog, as a breaking change).

PR #1962, opened 2023-08-21
Feat/specify lags per component for RegressionModel

### LINKED ISSUE #1110: Different lags for each past_covariate in past_covariates
From what I understand, the lags_past_covariates argument takes a list or integer and applies the same lags to all past covariates. Is it possible to lag each past_covariate differently when using a regression model? For example, I have the following past_covariates feature1, feature2, feature3 and I want these lags for each:

`{'feature1': [-2, -4]
'feature2:' [-1, -7]
'feature3': [-2,  -3]}`

Any recommendations on how I could accomplish this?

Really enjoying the package by the way!

madtoinou:
Fixes #1110.

### Summary

- `RegressionModel`'s constructor accepts `Dict[str, Sequence[int]]` for the arguments `lags`, `lags_past_covariates` and `lags_future_covariates`. The keys in the dictionary must correspond to the name of the components of the series used for training (or the first series when using multiple series) as well as the components generated by the encoders (when applicable).
- the key "default_lags" can be used to define a set of lags for the components not indicated in the dictionnary
- `lagged_features_name` attribute was updated accordingly
- added tests, making sure the forecasts are the same when using equivalent definitions of the lags using int/list versus dict.
- added sanity checks for multiple series training, checking that they all contain the same number of components.

### Other Information

- The order of the components (as well as coefficient & features names) is different when using the dictionary definition of lags : instead of being grouped by lags, they are grouped by components.
- only `_create_lagged_data_by_moving_window` supports extracting the values for each component independently, it's the function used by default in `RegressionModel._create_lagged_data` and the user cannot change it from the API.
- The `ShapExplainer` is working as intented with component-specific lags.


darts/utils/data/tabularization.py
```diff
@@ -1243,6 +1279,10 @@ def _get_feature_times(
         [target_series, past_covariates, future_covariates],
         [lags, lags_past_covariates, lags_future_covariates],
     ):
+        # TODO: information is available in model.lags, not sure how to make the info get here
```
dennisbader, line 1282: can you explain a bit more what you describe here?

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -1519,6 +1594,90 @@ def test_integer_indexed_series(self, mode):
         # the time axis returned by the second model should be as expected
         assert all(preds[1].time_index == pd.RangeIndex(start=50, stop=70, step=2))
 
+    @pytest.mark.parametrize(
```
dennisbader, line 1597: 👍 Nice!
Can you add a multiple time series case?

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -516,6 +542,55 @@ def test_training_data_creation(self, mode):
         ]
         assert list(training_labels[0]) == [82, 182, 282]
 
+        # lags defined using dictionnaries
```
dennisbader, line 545:
```suggestion
        # lags defined using dictionaries
```

dennisbader (changes requested):
Thanks a lot @madtoinou, this looks great 🚀 

Only had some suggestions about simplifying the lag handling in the model constructor, and some docs rephrasing here and there.

darts/models/forecasting/forecasting_model.py
```diff
@@ -2078,6 +2078,20 @@ def fit(
             ):
                 self.static_covariates = series.static_covariates
         else:
+            # check that all the ts within one group have the same number of components
```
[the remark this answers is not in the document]
madtoinou: The error currently raised is the following:
`ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 3 and the array at index 1 has size 6`, not exactly informative...

We should maybe check an arbitrary number of series. Or try to catch this exception?

darts/models/forecasting/regression_model.py
```diff
@@ -174,18 +183,18 @@ def __init__(
...
             raise_if_not(
-                isinstance(_lags, (int, list)) or _lags is None,
-                f"`{lags_name}` must be of type int or list. Given: {type(_lags)}.",
+                isinstance(_lags, (int, list, dict)) or _lags is None,
+                f"`{lags_name}` must be of type int, list or dict. Given: {type(_lags)}.",
             )
             raise_if(
                 isinstance(_lags, bool),
-                f"`{lags_name}` must be of type int or list, not bool.",
+                f"`{lags_name}` must be of type int, list or dict, not bool.",
```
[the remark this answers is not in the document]
madtoinou: There is kind of a two step process where the types of the lags is first checked (lines 183 to 222) and then, based on the type, they are converted. I guess that it was to avoid processing all of them before realizing that the last lag has the wrong type.

I don't mind running the type check while converting the arguments.

darts/models/forecasting/regression_model.py
```diff
@@ -201,60 +210,168 @@ def __init__(
...
+            )
+        elif isinstance(lags_future_covariates, list):
+            self.lags["future"] = _check_list_future_lags(
+                lags_future_covariates, "lags_future_covariates"
+            )
+        elif isinstance(lags_future_covariates, dict):
+            conv_lags = _check_dict_lags(
+                lags_future_covariates, "lags_future_covariates"
+            )
+            if conv_lags is not None:
```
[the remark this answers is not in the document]
madtoinou: It was to make the type checking stop complaining :)

darts/utils/data/tabularization.py
madtoinou: Just that I need to recompute the min/max for each "variate", I was wondering if there was a nice way to pass this information to tabularization. But it's probably negligeable.

darts/models/forecasting/regression_model.py
```diff
@@ -170,94 +192,170 @@ def encode_year(idx):
...
+                            )
+                        tmp_components_lags[comp_name] = sorted(comp_lags)
+                    else:
+                        invalid_type = True
+                        supported_types = "strictly positive integer or a list"
+
+                if invalid_type:
+                    raise_log(
+                        ValueError(
+                            f"When passed in a dictionary, `{lags_name}` for component {comp_name} must be either a "
```
dennisbader, line 324: related to my previous comment
```suggestion
                            f"`{lags_name}` - `{comp_name}`: must either be a "
```

dennisbader (changes requested):
Nice, just a last few suggestions, then we can merge :)

darts/models/forecasting/regression_model.py
```diff
@@ -170,94 +192,170 @@ def encode_year(idx):
...
+        for lags_values, lags_name, lags_abbrev in zip(
+            [lags, lags_past_covariates, lags_future_covariates],
+            ["lags", "lags_past_covariates", "lags_future_covariates"],
+            ["target", "past", "future"],
+        ):
+            if lags_values is None:
+                continue
+
+            # check type of argument before converting to dictionary
+            if not isinstance(lags_values, dict):
```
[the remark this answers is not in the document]
madtoinou: I did it there to be able to distinguish between argument passed as dict and the other types (int/tuple/list).

I'll try to think of a way to convert everything to dict there while conserving the meaningful messages.

dennisbader (commented):
Looks great now, thanks a lot @madtoinou 🚀 
After we check the issue from our discussion today, we can merge this one

dennisbader (approved):
Very nice, thanks a lot @madtoinou, everything looks great now 💯 

🚀

PR #1967, opened 2023-08-24
Fix tft_explainer to import right module

Biebrya:
### Summary

In tft_explainer it imports shap instead of tft, nothing crazy


dennisbader (approved):
Thanks for the fix @Biebrya 🚀

PR #1974, opened 2023-08-30
Fix/stasforecast 1 6 release

madtoinou:
### Summary

- Stasforecast fixed a bug allowing `AutoETS` model to use the `"A*M"` and `"AM*"` combination when passing `"Z"` as parameter for the trend and/or seasonality. Darts' unittest was using `"AAZ"`, effectively relying on an invalid error value (which was increased from 4.2 to 7.3 in this PR)
- Updated the links to statsforecast documentation as they modified the repository organization.
- Updated the documentation of all the models to make them more homogeneous and included a little note about conformal prediction (introduced in their 1.6.0 release).

### Other Information
- The fitted parameters (values of the "Z" placeholder after training) can be found using `model.model.model_["components"]`

### DIFF
--- a/darts/models/forecasting/sf_auto_arima.py
+++ b/darts/models/forecasting/sf_auto_arima.py
@@ -32,15 +32,11 @@ class StatsForecastAutoARIMA(FutureCovariatesLocalForecastingModel):
         It is probabilistic, whereas :class:`AutoARIMA` is not.
 
         We refer to the `statsforecast AutoARIMA documentation
-        <https://nixtla.github.io/statsforecast/models.html#arima-methods>`_
-        for the documentation of the arguments.
+        <https://nixtla.github.io/statsforecast/src/core/models.html#autoarima>`_
+        for the exhaustive documentation of the arguments.
 
         Parameters
         ----------
-        autoarima_args
-            Positional arguments for ``statsforecasts.models.AutoARIMA``.
-        autoarima_kwargs
-            Keyword arguments for ``statsforecasts.models.AutoARIMA``.
         add_encoders
             A large number of future covariates can be automatically generated with `add_encoders`.
             This can be done by adding multiple pre-defined index encoders and/or custom user-made functions that
@@ -61,6 +57,10 @@ class StatsForecastAutoARIMA(FutureCovariatesLocalForecastingModel):
                     'transformer': Scaler()
                 }
             ..
+
+        .. note::
+            Positional and keyword arguments can be used to pass additional parameters to statsforecast's `AutoARIMA`
+            constructor, such as ``prediction_intervals`` to compute the conformal prediction intervals.
 
         Examples
         --------
--- a/darts/models/forecasting/sf_auto_arima.py
+++ b/darts/models/forecasting/sf_auto_arima.py
@@ -37,6 +37,8 @@ class StatsForecastAutoARIMA(FutureCovariatesLocalForecastingModel):
 
         Parameters
         ----------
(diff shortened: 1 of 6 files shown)


darts/models/forecasting/sf_auto_arima.py
```diff
@@ -32,15 +32,11 @@ def __init__(
...
-        for the documentation of the arguments.
+        <https://nixtla.github.io/statsforecast/src/core/models.html#autoarima>`_
+        for the exhaustive documentation of the arguments.
 
         Parameters
         ----------
-        autoarima_args
-            Positional arguments for ``statsforecasts.models.AutoARIMA``.
-        autoarima_kwargs
-            Keyword arguments for ``statsforecasts.models.AutoARIMA``.
```
dennisbader, line 43: I would revert the changes to this and the "note" section and do it similarly as we have it for TorchForecastingModels and the `**tfm_kwargs` arguments

darts/models/forecasting/sf_auto_ces.py
```diff
@@ -18,17 +18,27 @@ def __init__(self, *autoces_args, **autoces_kwargs):
...
+            * "S" for simple simple seasonality (lagged CES),
+            * "P" for partial seasonality (without complex part),
+            * "F" for full seasonality (lagged CES with both real and complex seasonality).
+
+            Furthermore, the character "Z" is a placeholder telling statsforecast
+            to search for the best parameter. Default: "Z".
+
+        .. note::
+            Positional and keyword arguments can be used to pass additional parameters to statsforecast's `AutoCES`
+            constructor, such as ``prediction_intervals`` to compute the conformal prediction intervals.
```
dennisbader, line 41: I think it's simpler if we leave it as before. Otherwise, we have to maintain this ourselves, and check whether it still complies with newer versions of statsforecast.

darts/tests/models/forecasting/test_local_forecasting_models.py
```diff
@@ -56,7 +56,7 @@
     (StatsForecastAutoARIMA(season_length=12), 4.6),
     (StatsForecastAutoTheta(season_length=12), 5.5),
     (StatsForecastAutoCES(season_length=12, model="Z"), 7.3),
-    (StatsForecastAutoETS(season_length=12, model="AAZ"), 4.1),
+    (StatsForecastAutoETS(season_length=12, model="AAZ"), 7.3),
```
dennisbader, line 59: just to check, was statsforecast previously wrongly using a different `model` for "AAZ", or was it just a bug in general?
If we the former applies, could we use that other model to achieve the 4.1 error?

dennisbader (changes requested):
Thanks a lot for checking this @madtoinou , glad that you found out where the issue is coming from 🚀 
I would suggest to revert the changes to the parameter descriptions as otherwise, we will have to maintain the descriptions in the future.

darts/tests/models/forecasting/test_local_forecasting_models.py
madtoinou: During the parameter optimization for the placeholder "Z" (seasonality state), the `"AAM"` combination was tested (which is apparently not allowed, I did not investigate why) and they fixed this bug. For the record, with 1.5.0, passing explicitly `"AAM"` would raise an error.

dennisbader: sounds good, thanks!

dennisbader (approved):

PR #1975, opened 2023-08-30
Added space so we have "contiguous in" and not "contiguousin".

twobitunicorn:
In a command line the missing space gives us the word "contiguousin" instead of "contiguous in".  Added said space.

You all rule.


dennisbader (approved):
Thanks for this @twobitunicorn 🚀

PR #1976, opened 2023-08-31
Fix/Updated tests with patch release 2.0.8 of PL

madtoinou:
### Summary

- Patch release of pytorch lightning 2.0.8 slightly changed the performance of the deep-learning based models, making the unittest failing for `TiDEModel`. Updated the expected value accordingly
- Separated the univariate and multivariate probabilistic forecasts tests using the `supports_multivariate` attribute, allowing for more granularity in the expected error (for facilitating the detection of other performance decrease in the future).

### DIFF
--- a/darts/tests/models/forecasting/test_probabilistic_models.py
+++ b/darts/tests/models/forecasting/test_probabilistic_models.py
@@ -209,26 +209,27 @@ class TestProbabilisticModels:
     constant_noisy_ts_short = constant_noisy_ts[:30]
 
     @pytest.mark.slow
-    def test_fit_predict_determinism(self):
-        for model_cls, model_kwargs, _, _ in models_cls_kwargs_errs:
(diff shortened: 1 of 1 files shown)


darts/tests/models/forecasting/test_probabilistic_models.py
```diff
@@ -201,7 +210,7 @@ class TestProbabilisticModels:
 
     @pytest.mark.slow
     def test_fit_predict_determinism(self):
-        for model_cls, model_kwargs, _ in models_cls_kwargs_errs:
+        for model_cls, model_kwargs, _, _ in models_cls_kwargs_errs:
```
dennisbader, line 213: can you parametrize this test as well?

dennisbader (approved):
Thanks a lot for this fix @madtoinou 🚀

PR #1988, opened 2023-09-09
feature/cumsum

### LINKED ISSUE #1981: Cumulative Sum on Timeseries class
I make frequent use of the TimeSeries.diff() function for forecasting, I was hoping to add the reverse of the function `cumulative sum aka cumsum` to the TimeSeries class in order to provide opposite functionality. I've found this useful in conjunction with baseline models such as NaiveMovingAverage in order to provide decent baselines.

optionally something that I would propose to add is a `starting value` which could be a value to start with when using cumsum. This is again useful with baseline predictions on differenced values.

example psudocode:

```python
ts = TimeSeries.from_values([1,2,3,4,5])
diffs = ts.diff() # [1,1,1,1]
reversed_diff = diffs.cumsum() # [1,2,3,4]

model = NaiveMovingAverage()
model.fit(diffs)
average_diffs = model.predict(5) # [1,1,1,1,1]
baseline_prediction = average_diffs.cumsum(starting_value=ts[-1]) # starting value of 5
# baseline prediction w/ NaiveMovingAverage would be [6,7,8,9,10]

```

I would implement the above pseudocode by mirroring the implementation of diff, except I would use the underlying xarray implementation of cumsum() (which relies on the numpy implementation of cumsum). 

Alternatively if this doesnt seem useful, its always possible to use the xarray implementation, but it requires a little bit more massaging. I figured it might be nice to have the reverse operation of `diff` as it was something that I found myself implementing at work.
 
**Additional context**
I'd be happy to contribute this issue if this is something the maintainers think would be a good addition! I already have a solution that I use, would just be modifying it & adding test cases to ensure functionality.

Eliotdoesprogramming:
Fixes #1981 

### Summary
adds the cumsum function to the Timeseries class. Implementation uses the xarray implementation of cumsum. 

Additionally adds unit tests for the cumsum function when n=1 and n=2

let me know if this needs any more refinement, thanks!

### DIFF
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -765,18 +765,24 @@ class TestTimeSeries:
         series.with_values(np.random.rand(5, 10, 2))
 
     def test_cumsum(self):
-        cumsum1 = TimeSeries.from_dataframe(self.series1.pd_dataframe().cumsum())
-        # cumsum on cumsum2 should be equal to n=2 cumsum on series 1
-        cumsum2 = TimeSeries.from_dataframe(cumsum1.pd_dataframe().cumsum())
-        assert self.series1.cumsum() == cumsum1
-        assert self.series1.cumsum(n=2) == cumsum2
-
-        with pytest.raises(ValueError):
-            self.series1.diff(n=0)
-        with pytest.raises(ValueError):
-            self.series1.diff(n=-5)
-        with pytest.raises(ValueError):
-            self.series1.diff(n=0.2)
+        cumsum_expected = TimeSeries.from_dataframe(
+            self.series1.pd_dataframe().cumsum()
+        )
+        # univariate deterministic
+        assert self.series1.cumsum() == TimeSeries.from_dataframe(
+            self.series1.pd_dataframe().cumsum()
+        )
+        # multivariate deterministic
+        assert self.series1.stack(self.series1).cumsum() == cumsum_expected.stack(
+            cumsum_expected
+        )
+        # multivariate stochastic
+        # shape = (time steps, components, samples)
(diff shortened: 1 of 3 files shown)


darts/tests/test_timeseries.py
```diff
@@ -764,6 +764,20 @@ def test_with_values(self):
...
+        cumsum2 = TimeSeries.from_dataframe(cumsum1.pd_dataframe().cumsum())
+        assert self.series1.cumsum() == cumsum1
+        assert self.series1.cumsum(n=2) == cumsum2
+
+        with pytest.raises(ValueError):
+            self.series1.diff(n=0)
+        with pytest.raises(ValueError):
+            self.series1.diff(n=-5)
+        with pytest.raises(ValueError):
+            self.series1.diff(n=0.2)
```
dennisbader, line 779: This tests diff rather then cumsum. From my comment below I would opt for dropping `n` support

dennisbader, line 779:
```suggestion
        cumsum = TimeSeries.from_dataframe(self.series1.pd_dataframe().cumsum())
        assert self.series1.cumsum() == cumsum
```

dennisbader (changes requested):
Thanks a lot for this PR @Eliotdoesprogramming 🚀 

I added some suggestions, mainly that I would opt for dropping support for `n` in cumsum.

After that we can merge :)

dennisbader (approved):
Thanks @Eliotdoesprogramming for this PR, looks great now 🚀 

I applied the requested changes as we want to release the new version soon.

Eliotdoesprogramming:
np thanks for doing it! I get a bit busy during the week. Happy to contribute

PR #1990, opened 2023-09-12
Fix/Updated tests with release 2.0.0 of XGBoost

madtoinou:
### Summary

Force `tree_method='exact'` in the unittests as the default value changed to `'hist'` (which was resulting in slightly worse performance in the tests).

### Other Information

[Link](https://github.com/dmlc/xgboost/releases) to the full changelog.


dennisbader (approved):
Perfect, thanks @madtoinou 🚀

PR #2006, opened 2023-09-21
make tcn loss only compute on output chunk

### LINKED ISSUE #1965: [Question, possible BUG] Output of _TCNModule is a function of input_chunk_length, not output_chunk_length
Hello!

I've been working with the _TCNModule of the library, and only noticed now that the size of the output is the following shape:
`x = x.view(batch_size, self.input_chunk_length, self.target_size, self.nr_params)`
This strikes me as a bit odd; why would we want to produce target predictions for the past? In that case the model ought to just output the targets it has access to directly.

Best regards,
Daniel

dennisbader:
Fixes #1965.
- TCNModel computed the training loss also on the input chunk (past/history). changed so that is only computed on the output chunk (future/horizon)
- removed now obsolete first_prediction_index and its occurences from torch models


madtoinou (commented):
Thanks for the fix @dennisbader,

The changes slightly increased the error in the tests, we should either update the hyper-parameters or the error threshold.

dennisbader:
Closing as #1965 was answered from @pennfranc.

PR #2011, opened 2023-09-28
Fix/plot alpha kwarg

### LINKED ISSUE #1536: BUG or FEATURE? `alpha` is not passed to `plot()`

in `timeseries.plot()`, the `alpha` parameter is not passed to the underlying `xarray.DataArray.plot()`.

Example:
```python
series.plot(color='blue', label='MyLabel', alpha=0.2)
```

Although this may make sense for _probabilistic_ series, it does not for _deterministic_

Somethin like this should work:
```python
# timeseries.py | Line: 3726
            # temporarily set alpha to 1 to plot the central value (this way alpha impacts only the confidence intvls)
            alpha = kwargs["alpha"] if "alpha" in kwargs else None
            # ***NEW** only set if not deterministic
            if not self.is_deterministic:
                kwargs["alpha"] = 1
```

State in docu

>args – some positional arguments for the plot() method
> kwargs – some keyword arguments for the plot() method*
> * be aware that `alpha` will be overridden

SimTheGreat:
Fixes #1536

### Summary
if is probabilistic changes alpha of main line.If deterministic keeps main line alpha = 1, but changes the alpha_confidence_intvls.


darts/timeseries.py
```diff
@@ -3748,6 +3748,9 @@ def plot(
         ax
             Optionally, an axis to plot on. If `None`, and `new_plot=False`, will use the current axis. If
             `new_plot=True`, will create a new axis.
+        alpha
+            Optional, if `None` and probalistic value is 1 if if `None` and probalistic value is 0.25.
+            Otherwise will set alpha to provided value. Value must be between 0 and 1.
```
dennisbader, line 3753:
```suggestion
            Optionally, set the line alpha for deterministic series, or the confidence interval alpha for 
            probabilistic series.
```

darts/timeseries.py
```diff
@@ -3821,9 +3824,11 @@ def plot(
             else:
                 central_series = comp.mean(dim=DIMS[2])
 
-            # temporarily set alpha to 1 to plot the central value (this way alpha impacts only the confidence intvls)
             alpha = kwargs["alpha"] if "alpha" in kwargs else None
-            kwargs["alpha"] = 1
+            if not self.is_deterministic:
+                kwargs["alpha"] = 1
+            elif "alpha" in kwargs:
+                kwargs["alpha"] = alpha
```
dennisbader, line 3831: this can be removed
```suggestion
```

dennisbader (commented):
Very nice @SimTheGreat and thanks for that 🚀 

Just two minor suggestions, after that, we can merge

dennisbader (approved):
Looks great, thanks @SimTheGreat 🚀

PR #2021, opened 2023-10-06
Feature/scalar with window

### LINKED ISSUE #1540: [FEATURE] Scaler with rolling/expanding window to eliminate look ahead bias
**Problem description**
Currently, [Scaler](darts/dataprocessing/transformers/scaler.py) transforms input series "globally", meaning that all values of the the input vector are considered:

```python
from darts import TimeSeries

scaler = Scaler()

s = [1, 2, 3, 4, 5, 6, 7, 8, 9]
s = TimeSeries.from_series(s)

s_norm = scaler.fit_transform(s)
s_norm.pd_series()

0    0.000
1    0.125
2    0.250
3    0.375
4    0.500
5    0.625
6    0.750
7    0.875
8    1.000
```

This is not a problem if data is manually split into train and test sets and scaler is fitted with training set.

However, if we go on to use this vector as an input to [historical forecasts](https://unit8co.github.io/darts/quickstart/00-quickstart.html?highlight=historical_forecast#Backtesting:-simulate-historical-forecasting), we risk introducing [look ahead bias](https://stats.stackexchange.com/questions/442739/look-ahead-bias-induced-by-standardization-of-a-time-series) into analysis (at least when performing normalization). To eliminate this bias, rolling window approach is at least sometimes used https://arxiv.org/abs/1907.09452.

One solution would be to add parameters to scaler like so:

```python
class Scaler(InvertibleDataTransformer, FittableDataTransformer):
    def __init__(
        self, scaler=None, type="Global", window=None, name="Scaler", n_jobs: int = 1, verbose: bool = False
    ):
   """
   Parameters
   ----------
   type
        The type to scale the data. Options:
        "Global" uses all data points 
        "Rolling" uses rolling window with window size specified by 'window' parameter
        "Expanding" uses expanding window with initial window size specified by 'window' parameter
    window
        Size of window if type is either "Rolling" or "Expanding"
    """
```

Another option is to integrate this functionality to historical_forecasts and backtest functions. This might be convenient because parameters above could be inferred by the desired backtesting setup.

**Additional context**
Thank you again for excellent software!

JanFidor:
Fixes #1540.

### Summary
I've added the ability to use Scalar inside historical_forecasts to avoid information leakage. 
As `RegressionModel._optimized_historical_forecasts` doesn't support model retraining,  decided to not add scalar with window there. Please let me know if the implementation makes sense or if the code could be changed for the better!
P.S. Sorry for the PR being quite late, but the past 2 months have been quite hectic. I'll have significantly less time during this school year, than I originally planned, but I'll do my best to find an hour or 2 a week to respond to reviews and slowly but surely bring this and the others PRs to a merge


darts/models/forecasting/forecasting_model.py
```diff
@@ -1032,6 +1047,9 @@ def retrain_func(
                     verbose=verbose,
                     predict_likelihood_parameters=predict_likelihood_parameters,
                 )
+                if is_scalar_used:
```
dennisbader, line 1050: Comment for line 1040 (just before model._predict_wrapper()): We could allow support for passing a transformer even if retrain=False. We would just assume that the transformer was fit before, or that it doesn't require fitting.

                # also apply for covariates, here we can assume that the scaler must have been fit before or not

```
                # require fitting at all
                if scaler is not None and not train_transfomed:
                    train_series = scaler.transform(train_series)
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -709,6 +711,8 @@ def historical_forecasts(
             Default: ``False``
         enable_optimization
             Whether to use the optimized version of historical_forecasts when supported and available.
+        scaler
+            Parameter Scaler applied to each historical forecast.
```
dennisbader, line 715: Can we add some more information here?
E.g. if model is retrained, data transformer re-fit on the training data at each historical forecast step.
The fitted transformer is used to transform the input either for model training or prediction.
Inverse transformation will be applied to the prediction.

ETTAN93:
Hi, since my issue #2134 has been linked to this, can I just double check if there is a timeline to finish implementing the usage of scalers in historical forecast?

JanFidor:
Hi @ETTAN93! I'll try to finish up the PR around end of January - early February. I'm terribly sorry for the delay, but unfortunately this university term was much harder than I hoped and I didn't have the free time to work on it.

darts/models/forecasting/forecasting_model.py
```diff
@@ -1977,6 +1995,7 @@ def _optimized_historical_forecasts(
         verbose: bool = False,
         show_warnings: bool = True,
         predict_likelihood_parameters: bool = False,
+        scaler=None,
```
VascoSch92, line 1998: missing type

darts/models/forecasting/regression_model.py
```diff
@@ -1117,6 +1117,7 @@ def _optimized_historical_forecasts(
         verbose: bool = False,
         show_warnings: bool = True,
         predict_likelihood_parameters: bool = False,
+        scaler=None,
```
VascoSch92, line 1120: missing type

darts/models/forecasting/regression_model.py
```diff
@@ -1168,6 +1169,7 @@ def _optimized_historical_forecasts(
                 overlap_end=overlap_end,
                 show_warnings=show_warnings,
                 predict_likelihood_parameters=predict_likelihood_parameters,
+                scaler=scaler,
```
VascoSch92, line 1172: missing type

darts/utils/historical_forecasts/optimized_historical_forecasts.py
```diff
@@ -137,22 +143,26 @@ def _optimized_historical_forecasts_regression_last_points_only(
...
-                    name=series_.time_index.name,
-                ),
-                values=forecast,
-                columns=forecast_components,
-                static_covariates=series_.static_covariates,
-                hierarchy=series_.hierarchy,
-            )
+        forecast_value = TimeSeries.from_times_and_values(
+            times=times[0]
+            if stride == 1 and model.output_chunk_length == 1
```
VascoSch92, line 148: Is it not better to do this in one more step? Because in this way it is really difficult to read and understand this method call...

darts/models/forecasting/forecasting_model.py
```diff
@@ -603,6 +604,7 @@ def historical_forecasts(
         show_warnings: bool = True,
         predict_likelihood_parameters: bool = False,
         enable_optimization: bool = True,
+        scaler: Scaler = None,
```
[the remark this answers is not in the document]
JanFidor: Great point with the `InvertibleDataTransformer` and covariates, it will definitely make the code cleaner and more robust!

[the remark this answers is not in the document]
JanFidor: Should I delete the InvertibleDataTransformer as an option for covariates as we don't need to invert them later on?

ETTAN93:
Hi, can I just check when this feature is planned to be merged?

JanFidor:
Hi everyone! The past term with the studies, work and the engineering Thesis I got a little burned out and OSS unfortunately fell on back burner for quite a bit. Fortunately with the term ending I got much less to keep on top of and with some vacation I'm once again back to doing some contributions. Sorry disappearing and thanks for all the reviews, I'll try to update / resolve all the comments!

darts/models/forecasting/forecasting_model.py
```diff
@@ -959,24 +964,34 @@ def retrain_func(
                 if train_length_ and len(train_series) > train_length_:
                     train_series = train_series[-train_length_:]
 
-                # testing `retrain` to exclude `False` and `0`
-                if (
+                forecast_time_index_correct = (
```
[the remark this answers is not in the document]
JanFidor: From what I'm seeing I think one of my older commits resolves it, please confirm

darts/models/forecasting/forecasting_model.py
```diff
@@ -959,24 +964,34 @@ def retrain_func(
...
-                        counter=_counter_train,
-                        pred_time=pred_time,
-                        train_series=train_series,
-                        past_covariates=past_covariates_,
-                        future_covariates=future_covariates_,
-                    ):
+                    if is_retrain_func:
                         # avoid fitting the same model multiple times
                         model = model.untrained_model()
+                        if is_scalar_used:
```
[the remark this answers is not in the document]
JanFidor: Good catch! Passing an already trained transformer makes sense as a feature, is resolved by current implementation

`if series_transformer is not None:
                            if isinstance(series_transformer, FittableDataTransformer):
                                series_transformer.fit(train_series)
                            # remember that series was already transformed for later forecast
                            train_transformed = True
                            train_series = series_transformer.transform(train_series)
`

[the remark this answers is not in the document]
JanFidor: Good catch! Pushed changes and moved all transforms into one place

darts/models/forecasting/forecasting_model.py
JanFidor: Makes sense to add allow that as a feature! Added it to the code, line 1196 onwards

darts/utils/historical_forecasts/optimized_historical_forecasts.py
JanFidor: You're right, now everything is done in one go without creating TimeSeries manually

madtoinou:
Hi @JanFidor,

Since this feature is pretty high on the roadmap and expected by many users, in order to speed things up a bit, we decided to branch out of your work and opened #2529.

PR #2031, opened 2023-10-16
Add root mean squared scaled error to metrics

Beerstabr:
Added root mean squared scaled error as used during the M5 competition to metrics.
[Also see this link for explanation.
](https://www.pmorgan.com.au/tutorials/mae%2C-mape%2C-mase-and-the-scaled-rmse/#:~:text=Root%20Mean%20Squared%20Scaled%20Error,root%20of%20the%20final%20product.&text=You%20could%20also%20think%20of,a%20one%2Dstep%20naive%20forecast.)

Is this something you are interested in? If so, I'll add some tests.


madtoinou:
It looks interesting but wouldn't be simpler to just add an argument to mse? Or at least reuse it in rmsse?

Beerstabr:
Thanks for having a look @madtoinou .

I think you are right about reusing rmse() in rmsse().  I changed it in the PR.

Adding an argument to mse() would make mse more complicated then it needs to be. 

I think the way rmsse() is now included is in line with how mase is included, as well as in line with how rmse() relates to mse().

What do you think?

darts/metrics/metrics.py
dennisbader, line 1: Thanks @Beerstabber for this contribution and sorry for the late review.
I think we can indeed add this to Darts (keeping in mind that RMSSE also has its cons, e.g. when the naive score is bad).

As @madtoinou mentioned, we should try to simplify this a bit.
It is using a similar logic to MASE, can we refactor it so that both functions make use of the same logic?

dennisbader (commented):
Hi @Beerstabr and thanks for this, looks like a great start 🚀 
You can go ahead with the PR!

Main suggestion was to refactor the MASE and RMSSE so that they both use the same functions.

darts/metrics/metrics.py
Beerstabr: @dennisbader I refactored the code such that mase and rmsse are wrapped similar to the other metrics. Might be possible to further refactor by for example combining multi_ts_support() and multi_ts_support_insample(). Not sure if it is worth the effort. Let me know what you think. 

Also just copied the mase tests for rmsse. Will refactor it if you like the other changes I've made so far.

Would like to and a Scaled Quantile Loss metric as well if you agree that this would be a worthwhile addition.

dennisbader:
Hi @Beerstabr, just wanted to check in whether you're still working on this PR? :)

PR #2034, opened 2023-10-25
Fix/operand error with encoders

### LINKED ISSUE #1875: [BUG] RegressionModel historical forecasts with specific encoder lags
**Describe the bug**
There is a bug when calling historical_forecasts on a RegressionModel that uses encoders with a non-divisible frequency RegressionModels and non-overlapping maximum lag with the target lags.

**To Reproduce**
```

model = LinearRegressionModel(
    lags=3,
    lags_past_covariates=[-3, -2],
    add_encoders={
        "cyclic": {"past": ["month"]},
    }
)

hc = model.historical_forecasts(series=series)
```

**Expected behavior**
Should not raise an error. The error 

 - Python version: [e.g. 3.10]
 - darts version [master]

**Additional context**
The error can be avoided by passing `series` also as `past_covariates` to historical forecasts. Reason is that encoders are generated on the fly with minimum time span requirements -> we have to calculate the expected start index of covariates because it is not available in the time index generated by the encoders.

### LINKED ISSUE #1991: [BUG] Error with XGBModel and Encoders
**Describe the bug**
I get a TypeError when trying to fit sparse data with encoders and XGBModel from the darts package.
If I remove the encoders and lags_past_covariates the bug goes away, but I'm not sure why.

**To Reproduce**
```python
import pandas as pd
import numpy as np
import darts

from datetime import datetime

from darts import TimeSeries
from darts.models import XGBModel

values = np.array([3.5, 0, 6.5, 0, 0, 0, 0, 0, 0])

dates = pd.date_range(start=datetime(2021, 5, 1),
              end=datetime(2022, 1, 1),
              freq='MS')

data = {'Date':dates,
        'Values':values}

encoders = {'cyclic': {'past':['month']}}
model_kwargs = {'lags':[-6, -3, -2],
                'lags_past_covariates':[-6],
                'add_encoders':encoders
               }

model = XGBModel(**model_kwargs)

df = pd.DataFrame(data)
ts = TimeSeries.from_dataframe(df, time_col='Date', value_cols=['Values'])

model.fit(ts)
```

**Expected behavior**
I expected the model to fit without error, but instead I get the following error:
TypeError: unsupported operand type(s) for //: 'Timedelta' and 'pandas._libs.tslibs.offsets.MonthBegin'

 - Python version: [3.9.4]
 - darts version [0.24.0]

madtoinou:
Fixes #1875, fixes #1991

### Summary

When encoders are used to generate covariates, they have the minimum time requirements. In tabularization, an arithmetic operation on `Timedelta` and `pandas.offset` must be performed to realign the covariate and target time indexes. However, some frequencies (`'M', 'Y' and 'y'`) conversion to `Timedelta` are ambiguous ([pandas doc](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_timedelta.html)), causing the unsupported operand error.

To solve the problem for these specific cases, a temporary `DatetimeIndex` is created and the information is extracted without relying on the conversion (slower than the arithmetic operation).

### DIFF
--- a/darts/tests/utils/tabularization/test_create_lagged_training_data.py
+++ b/darts/tests/utils/tabularization/test_create_lagged_training_data.py
@@ -1132,37 +1132,44 @@ class TestCreateLaggedTrainingData:
             assert np.allclose(expected_X, X[:, :, 0])
             assert np.allclose(expected_y, y[:, :, 0])
 
-    def test_lagged_training_data_extend_past_and_future_covariates_datetime_idx(self):
+    @pytest.mark.parametrize("freq", ["D", "MS", "Y"])
+    def test_lagged_training_data_extend_past_and_future_covariates_datetime_idx(
+        self, freq
+    ):
         """
         Tests that `create_lagged_training_data` correctly handles case where features
         and labels can be created for a time that is *not* contained in `past_covariates`
         and/or `future_covariates`. This particular test checks this behaviour by using
-        datetime index timeseries.
+        datetime index timeseries and three different frequencies: daily, month start and
+        year end.
 
         More specifically, we define the series and lags such that a training example can
         be generated for time `target.end_time()`, even though this time isn't contained in
         neither `past` nor `future`.
         """
-        # Can create feature for time `t = '1/11/2000'`, but this time isn't in `past` or `future`:
+        # Can create feature for time `t = '1/1/2000'+11*freq`, but this time isn't in `past` or `future`:
         target = linear_timeseries(
             start=pd.Timestamp("1/1/2000"),
-            end=pd.Timestamp("1/11/2000"),
             start_value=1,
             end_value=2,
+            length=11,
+            freq=freq,
         )
         lags = [-1]
         past = linear_timeseries(
             start=pd.Timestamp("1/1/2000"),
-            end=pd.Timestamp("1/9/2000"),
             start_value=2,
             end_value=3,
+            length=9,
+            freq=freq,
         )
         lags_past = [-2]
         future = linear_timeseries(
             start=pd.Timestamp("1/1/2000"),
-            end=pd.Timestamp("1/7/2000"),
             start_value=3,
             end_value=4,
+            length=7,
+            freq=freq,
         )
         lags_future = [-4]
         # Only want to check very last generated observation:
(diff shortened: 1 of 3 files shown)


darts/utils/data/tabularization.py
```diff
@@ -884,11 +884,25 @@ def _create_lagged_data_by_moving_window(
...
+                        len(time_index_i)
+                        - 1
+                        + (start_time - time_index_i[-1]) // series_i.freq
+                    )
+                else:
+                    # Create a temporary DatetimeIndex to extract the actual start index.
+                    start_time_idx = (
+                        len(
+                            pd.date_range(
+                                start=time_index_i[0],
```
dennisbader, line 899: It might be more efficient to generate the index from the end of the series instead of from the beginning and then just add the len(time_index_i) to it?

darts/utils/data/tabularization.py
dennisbader, line 1: Very nice and neat! Can you add a test for it?

dennisbader (commented):
Very nice, thanks a lot @madtoinou. 
Just had a minor suggestion and that we should add a test for it

darts/utils/data/tabularization.py
madtoinou: Absolutely, it would make it slightly faster.

The `else` condition will only occur with `DatetimeIndex`, I would rather rely on `pd.date_range` since darts `generate_index` is probably overkill.

madtoinou: Yup, I don't know if this is much faster but at least, it looks similar to the other case

darts/tests/utils/tabularization/test_create_lagged_training_data.py
```diff
@@ -1132,37 +1132,44 @@ def test_lagged_training_data_extend_past_and_future_covariates_range_idx(self):
             assert np.allclose(expected_X, X[:, :, 0])
             assert np.allclose(expected_y, y[:, :, 0])
 
-    def test_lagged_training_data_extend_past_and_future_covariates_datetime_idx(self):
+    @pytest.mark.parametrize("freq", ["D", "MS", "Y"])
```
dennisbader, line 1135: nice 👍

dennisbader (approved):
Nice, looks great thanks @madtoinou 🚀

PR #2039, opened 2023-10-27
Feat/Example notebook for the regression models

### LINKED ISSUE #588: Regression model example notebook
We need one or a few good examples for RegressionModels.

### LINKED ISSUE #1289: Prediction based only on covariates
I am trying to predict a target timeseries based on past and future covariates. However, I only have limited data for the target timeseries and in production it wont be available at all, but I have a lot of covariates, which are easily available and in principle should allow fairly good predictions. 

My issue is that there does not seem to be any way to train a model in such a way that the target timeseries is not also used as a predictor. It appears that all models and all fit methods will always also use past values from the target series as predictors, but this wont work for my problem, as these values are often not available. Furthermore, I find that if I train models with the target series as a predictor, the models become to reliant on extrapolating past values and don't really use the covariates (which is not what I want).

I am wondering if this could be addressed with a new type of Dataset, but it would seem to me that this would also have to be taken into consideration during model construction. Unfortunately I don't understand the inner workings of DARTS and Torch well enough to implement this myself...

Or am I missing something here? Is this already possible? It seems to me that this should be a fairly common type of prediction problem.

Also, thanks for publishing this fantastic package! and for any suggestions!

madtoinou:
Fixes #588, fixes #1289.

### Summary

Example notebook for regression models, covering both some basics and advanced features :
- basics about lags, forecast based on covariates only, component-specific lags
- impact of `multi_model` when `ouput_chunk_length > 1`, visualization of the data tabularization depending on the value of `multi_model`
- darts probabilistic forecasting, including direct inference of the distribution parameters
- sklearn-like models: built-in and custom implementations
- external librairies
- brief section about explainability

The example are mostly illustrative : they were not optimized for accuracy but aim at displaying how the various features can be used and their impact compared to the default parameters. The dataset was not studied in depth to extract correlation and no observation about the data are made.

### Other Information
- Added new dataset `ElectricityConsumptionZurichDataset`: The dataset contains the electricity consumption of households in Zurich, Switzerland from 2015-2022 on different grid levels. We also added weather measurements for Zurich which can be used as covariates for modelling.


docs/source/examples.rst
```diff
@@ -76,6 +76,16 @@ with Darts using the Optuna library for hyperparameter optimization.
 
    examples/17-hyperparameter-optimization.ipynb
 
+Regression Models
+=============================
```
dennisbader, line 80:
```suggestion
Regression Models
=================
```

CHANGELOG.md
```diff
@@ -15,7 +15,9 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
   - 🚀🚀 Optimized `historical_forecasts()` for pre-trained `TorchForecastingModel` running up to 20 times faster than before!. [#2013](https://github.com/unit8co/darts/pull/2013) by [Dennis Bader](https://github.com/dennisbader).
   - Added callback `darts.utils.callbacks.TFMProgressBar` to customize at which model stages to display the progress bar. [#2020](https://github.com/unit8co/darts/pull/2020) by [Dennis Bader](https://github.com/dennisbader).
 - Improvements to documentation:
-  - Adapted the example notebooks to properly apply data transformers and avoid look-ahead bias. [#2020](https://github.com/unit8co/darts/pull/2020) by [Samriddhi Singh](https://github.com/SimTheGreat). 
+  - Adapted the example notebooks to properly apply data transformers and avoid look-ahead bias. [#2020](https://github.com/unit8co/darts/pull/2020) by [Samriddhi Singh](https://github.com/SimTheGreat).
+  - New example notebook for the `RegressionModels` with information about the advanced features such as component-specific lags, `multi_model` or the `MultiOutputRegressor` wrapper. [#2039](https://github.com/unit8co/darts/pull/2039) by [Antoine Madrona](https://github.com/madtoinou).
```
dennisbader, line 19:
```suggestion
  - New example notebook for the `RegressionModels` explaining features such as (component-specific) lags, `output_chunk_length` in relation with `multi_models`, multivariate support, and more. [#2039](https://github.com/unit8co/darts/pull/2039) by [Antoine Madrona](https://github.com/madtoinou).
```

CHANGELOG.md
```diff
@@ -15,7 +15,9 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
...
+- New dataset `EnergyConsumptionZurich` available in `darts.datasets`, can be loaded with weather information as covariates. [#2039](https://github.com/unit8co/darts/pull/2039) by [Antoine Madrona](https://github.com/madtoinou).
```
dennisbader, line 20: update the "2015-??" from below with the actual time frame
```suggestion
- Added new dataset `EnergyConsumptionZurichDataset`: The dataset contains the electricity demand of households in Zurich, Switzerland from 2015-?? on different grid levels. We also added weather measurements for Zurich which can be used as covariates for modelling. [#2039](https://github.com/unit8co/darts/pull/2039) by [Antoine Madrona](https://github.com/madtoinou).
```

dennisbader (commented):
Thanks @madtoinou , a notebook for regression models was really needed 🚀 
So far, I just added some suggestions for the documentation, will start reviewing the notebook in a bit.

dennisbader (approved):
Very nice, thanks a lot @madtoinou 🚀

PR #2040, opened 2023-10-27
Fix/optimized historical forecast with component specific lags

madtoinou:
### Summary

Optimized historical forecasts with regression models would raise an exception when using component-specific lags with different number of lags because `create_lagged_prediction_data()` was calling `model.lags.get()` instead of the dedicated method `model._get_lags()` which properly accounts for the component-specific lags.

### Other Information

Code snippet to reproduce the bug:
```python

# future_cov components: darts_enc_fc_cyc_month_sin and darts_enc_fc_cyc_month_cos
model = LinearRegressionModel(
    lags=3,
    lags_future_covariates={'default_lags':[-4, -2],
                            'darts_enc_fc_cyc_month_sin':[-1]
                           },
    add_encoders={
        "cyclic": {"future": ["month"]},
    }
)

model.fit(series)
model.historical_forecasts(series=series, retrain=False)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,16 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 [Full Changelog](https://github.com/unit8co/darts/compare/0.26.0...master)
 
 ### For users of the library:
+
+**Improved**
 - Improvements to `TorchForecastingModel`:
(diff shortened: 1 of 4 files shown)


dennisbader (approved):
Looks great, thanks for fixing this so quickly @madtoinou 🚀 

I added a test, ready to be merged now 💯

PR #2044, opened 2023-10-31
Fix doc rendering in statistics.py

felixdivo:
It previously looked like this (where the "Returns" section is broken):

Screenshot 2023-10-31 at 12 07 11


dennisbader (approved):
Looks good as well, thanks a lot for improving the documentation @felixdivo 🚀

PR #2045, opened 2023-10-31
Fix doc formatting of LinearRegressionModel

felixdivo:
It previously looked like this:

Screenshot 2023-10-31 at 12 27 45

How can we check that it is now better?

### DIFF
--- a/darts/models/forecasting/fft.py
+++ b/darts/models/forecasting/fft.py
@@ -255,6 +255,7 @@ class FFT(LocalForecastingModel):
         >>> FFT(required_matches={'month'}, trend='exp')
 
         Simple usage example, using one of the dataset available in darts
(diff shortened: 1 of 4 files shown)


dennisbader:
Hi @felixdivo and thanks for this update.
To build the docs you can run the following from within the repository root:

```
# install darts locally so that documentation is updated
pip install -e .

# remove existing docs files
make --directory=./docs clean

# generate the new documentation (can be found under darts/docs/build/html, open in browser by double clicking on any of the generated files) 
make --directory=./docs build-all-docs
```

dennisbader (approved):
Looks great @felixdivo , thanks for that 🚀 

While checking the docs I found and fixed the same issue in some other models.

PR #2046, opened 2023-11-01
fix: torch_forecasting_model load_weights with float16 and float32

### LINKED ISSUE #1987: TorchForecastingModel model precision
I had a few questions around precision for TorchForecastingModels:

- It seems we can only choose between float64 & float32, I'm curious if we could use half-precision (float16 or bfloat16), it seems this is discouraged when setting through pl_trainer_kwargs.
- When loading the model, the precision seems to be reset to float64 (train_sample).

**Context:**

```python
import torch
import numpy as np

from darts import models

# create fake dataset
series_air = AirPassengersDataset().load().astype(np.float32) # can we use float16?

tft_accelerator = (
    {"accelerator": "gpu", "devices": -1}
    if torch.cuda.is_available()
    else {"accelerator": "cpu"}
)
def create_model(model_name: str):
    return models.TFTModel(
        input_chunk_length=4,
        output_chunk_length=4,
        categorical_embedding_sizes={}, # <-- error if not set + loading weights
        likelihood=QuantileRegression(), # <-- error if not set + loading weights
        n_epochs=10,
        model_name=model_name,
        work_dir="models",
        force_reset=True,
        save_checkpoints=True,
        pl_trainer_kwargs={**tft_accelerator},
    )

model = create_model("TFTModel")
model.fit(
    [series_air, series_air], 
    past_covariates=[series_air, series_air],
    future_covariates=[series_air, series_air],
)

model = create_model("TFTModel_finetune")
model.load_weights_from_checkpoint( # <-- train_sample now float64 regardless of series dtype
    model_name="TFTModel",
    work_dir="models",
    best=False,
)

model.predict(
    n=4,
    series=series_air[:-4], 
    past_covariates=series_air,
    future_covariates=series_air,
)
```

**System:**
Python version: `3.11.4`
darts version `0.25.0`

Hsinfu:
Fix bug for TorchForecastingModel calling load_weights with float16 and float32. Fixes #1987.

### Bug details

- Train the TorchForecastingModel with **float32** following the [instruction](https://unit8co.github.io/darts/userguide/torch_forecasting_models.html#build-your-timeseries-using-32-bits-data) and `.save` it to disk.
- Create a new model instance and call `.load_weights` to restore it.
- The `mock_train_sample` in `.load_weights_from_checkpoint` built from `np.zero` without specifying the `dtype`, so it sticks to the default **float64**, and causes the `self.train_sample` stays **float64**, too.

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -107,6 +107,12 @@ CHECKPOINTS_FOLDER = "checkpoints"
 RUNS_FOLDER = "runs"
 INIT_MODEL_NAME = "_model.pth.tar"
 
+TORCH_NP_DTYPES = {
+    torch.float16: np.float16,
+    torch.float32: np.float32,
+    torch.float64: np.float64,
+}
+
 # pickling a TorchForecastingModel will not save below attributes: the keys specify the
 # attributes to be ignored, and the values are the default values getting assigned upon loading
 TFM_ATTRS_NO_PICKLE = {"model": None, "trainer": None}
@@ -1872,8 +1878,9 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         )
 
         # pl_forecasting module saves the train_sample shape, must recreate one
+        np_dtype = TORCH_NP_DTYPES[ckpt["model_dtype"]]
         mock_train_sample = [
-            np.zeros(sample_shape) if sample_shape else None
+            np.zeros(sample_shape, dtype=np_dtype) if sample_shape else None
             for sample_shape in ckpt["train_sample_shape"]
         ]
         self.train_sample = tuple(mock_train_sample)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 **Fixed**
 - Fixed a bug when calling optimized `historical_forecasts()` for a `RegressionModel` trained with unequal component-specific lags. [#2040](https://github.com/unit8co/darts/pull/2040) by [Antoine Madrona](https://github.com/madtoinou).
 - Fixed a bug when using encoders with `RegressionModel` and series with a non-evenly spaced frequency (e.g. Month Begin). This raised an error during lagged data creation when trying to divide a pd.Timedelta by the ambiguous frequency. [#2034](https://github.com/unit8co/darts/pull/2034) by [Antoine Madrona](https://github.com/madtoinou).
+- Fixed a bug when loading a `TorchForecastingModel` that was trained with a precision other than `float64`. [#2046](https://github.com/unit8co/darts/pull/2046) by [Freddie Hsin-Fu Huang](https://github.com/Hsinfu).
 
 ### For developers of the library:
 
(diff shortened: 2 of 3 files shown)


darts/models/forecasting/torch_forecasting_model.py
dennisbader, line 1: Can you try adding a unit test for this?

The best place for this would be in darts/tests/models/forecasting/test_torch_forecasting_model.
Maybe something like `test_loading_weights_with_different_dtype`

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1872,8 +1872,9 @@ def load_weights_from_checkpoint(
         )
 
         # pl_forecasting module saves the train_sample shape, must recreate one
+        np_dtype = np.float32 if ckpt["model_dtype"] == torch.float32 else np.float64
```
dennisbader, line 1875: I'd suggest creating a mapping between the torch and numpy dtypes at the top of this file. With this we can avoid/catch some breaking changes (e.g. if we ever support other dtypes such as float16 and forget about this line).

```
TORCH_NP_DTYPES = {
    torch.float32: np.float32,
    torch.float64: np.float64,
}

...

mock_train_sample = [
    np.zeros(sample_shape, dtype=TORCH_NP_DTYPES[ckpt["model_dtype"]]) if sample_shape else None
    for sample_shape in ckpt["train_sample_shape"]
]

**Review by dennisbader — CHANGES_REQUESTED** [review]
Thanks @Hsinfu for contributing and this fix 🚀 

I had 2 minor suggestions regarding adding a unit test and making the dtype mapping more robust against future changes.

**Comment by madtoinou** [comment]
Just for the record, this PR fixed one of the point mentioned in #1987.

File: darts/models/forecasting/torch_forecasting_model.py
Comment (Hsinfu) [reply darts/models/forecasting/torch_forecasting_model.py:1]: `test_load_weights_with_float32_dtype`

File: darts/models/forecasting/torch_forecasting_model.py (hunk 1 above)
Comment (Hsinfu) [reply darts/models/forecasting/torch_forecasting_model.py]: also add float16

```python
TORCH_NP_DTYPES = {
    torch.float16: np.float16,
    torch.float32: np.float32,
    torch.float64: np.float64,
}
```

dennisbader:
Hi @Hsinfu , it seems like your branch is protected for us to adapt it.

Can you merge the current master head into your branch and add the following line to our `CHANGELOG.md` **Fixed** section?

```
- Fixed a bug when loading a `TorchForecastingModel` that was trained with a precision other than `float64`. [#2046](https://github.com/unit8co/darts/pull/2046) by [Freddie Hsin-Fu Huang](https://github.com/Hsinfu).
```

Hsinfu:
> - Fixed a bug when loading a `TorchForecastingModel` that was trained with a precision other than `float64`. [#2046](https://github.com/unit8co/darts/pull/2046) by [Freddie Hsin-Fu Huang](https://github.com/Hsinfu).

I just cloned the current master head, and rebase my branch on it. Also add the information in `CHANGELOG.md`. 
See if it works now.

dennisbader (approved):
Looks great, thanks a lot for this @Hsinfu 🚀 💯

PR #2048, opened 2023-11-02
Feat/summary_plot() returns the shap values

### LINKED ISSUE #1353: Getting shap_explain.summary_plot() in a python list rarther than an image
The output of ```shap_explain.summary_plot()``` is an image. How can I get the data in the image in a python-accesible format, say a list? ```ExplainabilityResult``` is a ```TimeSeries``` but I am more interested in a simple list of the average change in prediction for every feature (same numbers in in the ```shap_explain.summary_plot()```, just accessible in python).

madtoinou:
Fixes #1353.

### Summary
Simply appending the `shap.Explanations` used as input for the plot in a list then returned by the method. The shap values are stored in `.values` of each explanation.

### DIFF
--- a/darts/explainability/shap_explainer.py
+++ b/darts/explainability/shap_explainer.py
@@ -25,7 +25,7 @@ each of the (lagged) series.
 """
 
 from enum import Enum
-from typing import Dict, NewType, Optional, Sequence, Union
+from typing import Dict, List, NewType, Optional, Sequence, Union
 
 import matplotlib.pyplot as plt
 import pandas as pd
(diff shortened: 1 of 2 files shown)


darts/explainability/shap_explainer.py
```diff
@@ -395,6 +395,11 @@ def summary_plot(
             for the sake of performance.
         plot_type
             Optionally, specify which of the shap library plot type to use. Can be one of ``'dot', 'bar', 'violin'``.
+
+        Returns
```
dennisbader, line 399: Shall we just return the `shaps_` dict?

Might be a bit cumbersome to interpret which element belongs to which horizon/components

dennisbader (commented):
Thanks for this @madtoinou 🚀 
Had a minor suggestion

darts/explainability/shap_explainer.py
madtoinou: Good idea, changed it, updated the docstring and added a little test.

dennisbader (approved):
Looks great, thanks 🚀 💯

PR #2049, opened 2023-11-03
Feat/auto-regression and future values of past covariates documentation

### LINKED ISSUE #1822: [BUG] LightGBM past covariates look-ahead bias
```python
import pandas as pd
import numpy as np

from darts.datasets import EnergyDataset
from darts import TimeSeries
```

## Describe the bug
I am concerned that there is a specific corner case of look-ahead bias. This seems to be the case with the following configuration:

- `model == LightGBMModel`
- `output_chunk_length < horizon`
- `past_covariates != None`

## To Reproduce

After loading some arbitraty data, we will generate two versions of `past_covariates` series: 

1) `pc` that has not been manipulated and consisting of `pc_train` and `pc_test` series
2) `pc_zeros` in which `pc_test` has been floored to zero

```python
data = EnergyDataset().load().pd_dataframe().head(1000)

y = data['total load actual'].fillna(0)
pc = data['generation biomass'].fillna(0)

length = 400
y_train, y_test = y.iloc[:length], y.iloc[length:]
pc_train, pc_test = pc.iloc[:length], pc[length:]

zeros = np.zeros(600)
zeros = pd.Series(zeros, index=pc_test.index)
pc_zeros = pd.concat([pc_train, zeros])

pc.iloc[390:410].plot(label='pc');
pc_zeros.iloc[390:410].plot(label='pc_zeros');
plt.legend();
```

    
output_3_0

    

```python
# Convert to TimeSeries
pc = TimeSeries.from_series(pc)
y_train = TimeSeries.from_series(y_train)
pc_train = TimeSeries.from_series(pc_train)
pc_test = TimeSeries.from_series(pc_test)
pc_zeros = TimeSeries.from_series(pc_zeros)
```

Next we train a model where `output_chunk_length == horizon`. Then we predict 24 steps ahead using both `pc` and `pc_zeros` as an input. As expected, this produces identical results since manipulating `past_covariates` in prediction window should not affect the results.

```python
model = LightGBMModel(
    lags=10, 
    lags_past_covariates=10, 
    output_chunk_length=24,
    random_state=0)

model.fit(series=y_train, past_covariates=pc_train);

y_pred = model.predict(24, past_covariates=pc)
y_pred_noise = model.predict(24, past_covariates=pc_zeros)

y_pred.plot(label='pc')
y_pred_noise.plot(label='pc_zeros')
```

    
output_6_0

    

However, when `output_chunk_length < horizon`, the results are markedly different:

```python
model = LightGBMModel(
    lags=10, 
    lags_past_covariates=10, 
    output_chunk_length=1,
    random_state=0)

model.fit(series=y_train, past_covariates=pc_train);

y_pred = model.predict(24, past_covariates=pc)
y_pred_noise = model.predict(24, past_covariates=pc_zeros)

y_pred.plot(label='pc')
y_pred_noise.plot(label='pc_zeros')
```

    
output_8_0

    

## Conclusion

It seems that there is a look-ahead bias with specific LightGBM configuration. This is somewhat dangerous given that for `LightGBMModel` the `output_chunk_length` is an optional parameter defaulting to 1 and there is no warning in the current class [documentation](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.lgbm.html?highlight=lightgbmmodel#darts.models.forecasting.lgbm.LightGBMModel):

> output_chunk_length (int) – Number of time steps predicted at once by the internal regression model. Does not have to equal the forecast horizon n used in predict(). However, setting output_chunk_length equal to the forecast horizon may be useful if the covariates don’t extend far enough into the future.

## Proposed solution

As a quick fix, enforce `output_chunk_length==horizon`. 

## System

- Python version: 3.8.13
- Darts version: 0.24.0

### LINKED ISSUE #2005: Add warning for auto-regressive prediction with n>ocl and past covariates, update documentation
See #1822 

- raise a warning when predicting with `n>output_chunk_length` and the model uses past covariates
- update n, forecast_horizon, output_chunk_length, past covariates documentation to make users aware of potential look ahead bias when using future values of past covariates

madtoinou:
Fixes #1822, fixes #2005.

### Summary

- Added a warning in `GlobalForecastingModel.predict()` (class where output_chunk is defined) when `n > output_chunk_length` and the model uses past covariates.
- Added show_warnings parameter to hide the warning
- Updated the docstring of `output_chunk_length` for the `RegressionModels` to mention auto-regression and the usage of the past covariates

### Other Information

The covariates section of the User guide already contain the following passage :

> There are some extra nuances that might be good to know. For instance, deep learning models in Darts can (in general) forecast output_chunk_length points at a time. However it is still possible for models trained with past covariates to make forecasts for some horizon n > output_chunk_length if the past_covariates are known far enough into the future. In such cases, the forecasts are obtained by consuming future values of the past covariates, and using auto-regression on the target series. If you want to know more details, read on.

Should it be more emphasized?  Modified?


dennisbader (commented):
Thanks for this @madtoinou 🚀 
As discussed, it would be nice if the warning is raised once only for historical forecasting, but if it's too tedious, we can leave it like that.

darts/models/forecasting/forecasting_model.py
```diff
@@ -2210,6 +2220,18 @@ def predict(
...
+            show_warnings
+            and self.uses_past_covariates
+            and self.output_chunk_length is not None
+            and n > self.output_chunk_length
+        ):
+            logger.warning(
+                "Since `predict` was called with `n > output_chunk_length`, auto-regression is be used to forecast "
+                "the values after `output_chunk_length`. As this model uses past covariates, it will access future "
+                "values (compared to the first predicted timestep) of this covariates to produce each subsequent "
+                "`output_chunk_length` forecasts. To hide this warning, set `show_warnings=False`."
```
dennisbader, line 2233:
```suggestion
                "`predict()` was called with `n > output_chunk_length`: using auto-regression to forecast "
                "the values after `output_chunk_length` points. The model will access `(n - output_chunk_length)` "
                "future values of your `past_covariates` (relative to the first predicted time step). "
                "To hide this warning, set `show_warnings=False`."
```

darts/utils/historical_forecasts/optimized_historical_forecasts_torch.py
dennisbader, line 1: It looks like regression models never call the super().predict() and this warning would never be raised when going through the optimized routine. Can we add this somehow?

dennisbader (commented):
Thanks @madtoinou , looks good as well 🚀 
Had some minor suggestions and we should probably merge the fit/predict_kwargs PR first.

dennisbader (approved):
Looks great, thanks @madtoinou 🚀 

I updated the docs a bit for torch and regression models to have a more unified style for input/output chunk length and covariates support

PR #2050, opened 2023-11-03
Fix/add fit/predict_kwargs argument to historical_forecasts

### LINKED ISSUE #1767: [Minor BUG] num_workers when calling predict/historical_forecasts of specific models
Hello,

Not sure if it is a bug, or just little inconvenience (darts==0.23.1)

My scenario:
1. calling historic_forecast of TFT model without retrain, and only for few historic elements
2. getting a lot of ```PossibleUserWarning: The dataloader, predict_dataloader 0, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 4 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.```

After quick debugging:
1. forecasting_model.py(1802)_predict_wrapper() is calling TFTModel's `predict` function, but doesn't pass `num_loader_workers`
2. `predict_wrapper` is called from `historic_forecast` as almost last statement.

Proposition:
1. Historic_forecast should have extra parameter num_loader_workers - named, or as kwargs...
2. This parameter should be passed to `predict-wrapper`, and then to `predict`.
3. Not sure about interfering with other elements of Darts.

Btw. Darts is wonderfull and very usefull software. And big thanks to their developers :)\

madtoinou:
Fixes #1767.

### Summary

- Added `fit_kwargs` and `predict_kwargs` argument to `historical_forecasts`, `backtest` and `gridsearch`
- Added some logic in `fit_predict` and `predict_wrapper` to pass the arguments to `fit()/predict()` only when it's supported (based on method signature)

### Other Information

Rely on python to complain when unsupported or duplicated arguments are provided


dennisbader (commented):
Wouldn't this be a great opportunity to add fit and predict kwargs to historical forecasts? :)

darts/models/forecasting/forecasting_model.py
```diff
@@ -802,6 +824,51 @@ def retrain_func(
                 logger,
             )
 
+        if fit_kwargs is None:
+            fit_kwargs = dict()
+        if predict_kwargs is None:
+            predict_kwargs = dict()
+
+        # sanity checks of the arguments directly exposed by historical_forecasts
+        if "predict_likelihood_parameters" not in predict_kwargs:
```
dennisbader, line 833: I think here we can also just let python handle duplicated parameters.
It will throw an error:

```
SyntaxError: keyword argument repeated: predict_likelihood_parameters
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -802,6 +824,51 @@ def retrain_func(
...
+        if "num_samples" not in predict_kwargs:
+            predict_kwargs["num_samples"] = num_samples
+        elif predict_kwargs["num_samples"] != num_samples:
+            logger.warning(
+                "`num_samples` was provided with contradictory values, "
+                "retaining the value passed with `predict_kwargs`."
+            )
+
+        # fit/predict_kwargs cannot be used to pass arguments used by historical_forecast logic
+        forbiden_args = ["series", "past_covariates", "future_covariates"]
```
dennisbader, line 854: same comment as above

darts/models/forecasting/forecasting_model.py
```diff
@@ -802,6 +824,51 @@ def retrain_func(
...
+        fit_invalid_args = forbiden_args + [
```
dennisbader, line 855: Do we have to prohibit this?
In theory it would be valid to give a validation series that is further in the future than `series`.

I would leave the responsibility to the user until we implement a dedicated logic for validation series in historical forecasting

darts/models/forecasting/forecasting_model.py
```diff
@@ -31,10 +31,12 @@
 
 import numpy as np
 import pandas as pd
+from sklearn.multioutput import MultiOutputRegressor
```
dennisbader, line 34: not optimal to import this in the base classes. I think we can adapt the fit/predict wrappers as described in one of the comments

darts/models/forecasting/forecasting_model.py
```diff
@@ -1406,10 +1490,28 @@ def gridsearch(
...
-        # TODO: here too I'd say we can leave these checks to the models
-        # if covariates is not None:
-        #     raise_if_not(series.has_same_time_as(covariates), 'The provided series and covariates must have the '
-        #                                                       'same time axes.')
+        if fit_kwargs is None:
+            fit_kwargs = dict()
+        if predict_kwargs is None:
+            predict_kwargs = dict()
+
+        forbiden_args = ["series", "past_covariates", "future_covariates"]
```
dennisbader, line 1498: same comments as above

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -2044,7 +2043,7 @@ def _optimized_historical_forecasts(
         last_points_only: bool = True,
         verbose: bool = False,
         show_warnings: bool = True,
-        predict_likelihood_parameters: bool = False,
+        predict_kwargs: Optional[Dict[str, Any]] = None,
```
dennisbader, line 2046:
```suggestion
        **kwargs,
```

darts/models/forecasting/regression_model.py
```diff
@@ -1103,7 +1102,7 @@ def _optimized_historical_forecasts(
         last_points_only: bool = True,
         verbose: bool = False,
         show_warnings: bool = True,
-        predict_likelihood_parameters: bool = False,
+        predict_kwargs: Optional[Dict[str, Any]] = None,
```
dennisbader, line 1105:
```suggestion
        **kwargs
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -2065,16 +2064,15 @@ def _optimized_historical_forecasts(
...
             start=start,
             start_format=start_format,
             forecast_horizon=forecast_horizon,
             stride=stride,
             overlap_end=overlap_end,
             last_points_only=last_points_only,
             show_warnings=show_warnings,
-            predict_likelihood_parameters=predict_likelihood_parameters,
             verbose=verbose,
+            predict_kwargs=predict_kwargs,
```
dennisbader, line 2075:
```suggestion
            **kwargs
```

darts/models/forecasting/regression_model.py
```diff
@@ -1124,36 +1123,37 @@ def _optimized_historical_forecasts(
...
                 future_covariates=future_covariates,
-                num_samples=num_samples,
                 start=start,
                 start_format=start_format,
                 forecast_horizon=forecast_horizon,
                 stride=stride,
                 overlap_end=overlap_end,
                 show_warnings=show_warnings,
-                predict_likelihood_parameters=predict_likelihood_parameters,
+                **predict_kwargs,
```
dennisbader, line 1142: It looks like we don't pass the kwargs to regression model's predict() inside the optimized subroutines

dennisbader (commented):
Looks like a great start @madtoinou 🚀 

I added a couple of suggestions on how we could simplify the kwargs handling and the fit/predict wrappers.
We can let python handle some of edge cases, and potentially, we can adapt the fit/predict wrappers to generalize over all models (and reduce code).

darts/models/forecasting/forecasting_model.py
madtoinou: Sounds good, I was overly cautious

darts/models/forecasting/forecasting_model.py
```diff
@@ -328,10 +331,21 @@ def _predict_wrapper(
         num_samples: int,
         verbose: bool = False,
         predict_likelihood_parameters: bool = False,
+        num_loader_workers: int = 0,
```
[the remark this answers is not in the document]
madtoinou: Updated the wrappers, there is only one wrapper in `ForecastingModel` now and it relies on the signature of the `fit()/predict()` method to extract the supported arguments (first option you described).

I am going to run some additional tests to make sure it behave as expected.

darts/models/forecasting/forecasting_model.py
```diff
@@ -316,23 +316,25 @@ def _fit_wrapper(
...
-    ) -> TimeSeries:
-        kwargs = dict()
-        if self.supports_likelihood_parameter_prediction:
-            kwargs["predict_likelihood_parameters"] = predict_likelihood_parameters
-        return self.predict(n, num_samples=num_samples, verbose=verbose, **kwargs)
+        **kwargs,
+    ) -> Union[TimeSeries, Sequence[TimeSeries]]:
+        supported_params = inspect.signature(self.predict).parameters
+        kwargs_ = {k: v for k, v in kwargs.items() if k in supported_params}
+        return self.predict(n, num_samples=num_samples, **kwargs_)
```
dennisbader, line 337: Some thoghts here:
- Should we check here for past/future covariates similar to _fit_wrapper?

  E.g. add past/future_covariates as input parameters to _predict_wrapper and then:
  ```
  if self.uses_past_covariates:
      kwargs_["past_covariates"] = past_covariates
  ```

- as far as I can tell we could let `num_samples` be part of **kwargs.

- To avoid breaking changes, we can pop the supports_likelihood_parameter_prediciton from the `kwargs_` if it's not supported (as done before)

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -2044,7 +2043,7 @@ def _optimized_historical_forecasts(
         last_points_only: bool = True,
         verbose: bool = False,
         show_warnings: bool = True,
-        predict_likelihood_parameters: bool = False,
+        **kwargs,
```
dennisbader, line 2046: The signature here is now different from ForecastingModel._optimized_historical_forecasts and RegressionModel._opt....

Either leave or drop `num_samples` and `predict_likelihood_parameters` in all of them

darts/utils/historical_forecasts/optimized_historical_forecasts_torch.py
```diff
@@ -24,16 +26,15 @@ def _optimized_historical_forecasts(
...
     start: Optional[Union[pd.Timestamp, float, int]] = None,
     start_format: Literal["position", "value"] = "value",
     forecast_horizon: int = 1,
     stride: int = 1,
     overlap_end: bool = False,
     last_points_only: bool = True,
     show_warnings: bool = True,
-    predict_likelihood_parameters: bool = False,
     verbose: bool = False,
+    **kwargs,
```
dennisbader, line 37: same comment with function signature as above for `num_samples` and `predict_likelihood_parameters`

darts/utils/historical_forecasts/utils.py
```diff
@@ -208,6 +210,28 @@ def _historical_forecasts_general_checks(model, series, kwargs):
                 logger,
             )
 
+    # check that fit/predict_kwargs contain only supported parameters
+    if n.fit_kwargs is not None:
+        fit_args = set(inspect.signature(model.fit).parameters)
+        wrong_args = set(n.fit_kwargs) - fit_args
+        if len(wrong_args) > 0:
+            raise_log(
```
dennisbader, line 218: Should just raise a warning here and below to just tell the users which params will be ignored? Because the fit/predict wrappers will ignore the params which are not valid.
Could be nice to use the same hist_fc parameters for any model

darts/models/forecasting/forecasting_model.py
```diff
@@ -1457,16 +1498,23 @@ def _evaluate_combination(param_combination) -> float:
                     last_points_only=last_points_only,
                     verbose=verbose,
                     show_warnings=show_warnings,
+                    predict_kwargs=predict_kwargs,
```
dennisbader, line 1501:
```suggestion
                    fit_kwargs=fit_kwargs,
                    predict_kwargs=predict_kwargs,
```

darts/utils/historical_forecasts/utils.py
dennisbader, line 1: As discussed, to avoid the duplicate params between hist_fc input paramaters and fit/predict_kwargs, we should check this here once.

E.g. compare fit/predict_kwargs with historical_forecasts signature and raise an error if there are duplicates.
(`forecast_horizon` should be mapped to `n`)

dennisbader (commented):
Looking really good now, thanks @madtoinou  🚀
Just some minor suggestions

darts/models/forecasting/forecasting_model.py
madtoinou: You're right, all the `TorchForecastingModel` use the same `predict()` method but don't all support the same covariates. I agree for `num_samples`.

We don't need to do it for `predict_likelihood_parameters` as the `predict()` methods already indicate if the model support it or not (and we check if they are indeed probabilistic in the `predict()` call, which end up calling `GlobalForecastingModel.predict()`).

madtoinou:
Updated the PR, handling of the `fit/predict_kwargs` was improved and some tests were added.

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -1959,3 +1961,113 @@ def create_model(ocl, use_ll=True, model_type="regression"):
...
+            start=len(self.ts_pass_train) - n,
+            retrain=False,
+            enable_optimization=enable_optimization,
+            predict_kwargs=unsupported_predict_kwargs,
+        )
+
+        assert hist_fc.components.equals(self.ts_pass_train.components)
+        assert len(hist_fc) == n
+
+        # passing hist_fc parameters in predict_kwargs, interferring with the logic
```
dennisbader, line 2064:
```suggestion
        # passing hist_fc parameters in predict_kwargs, interfering with the logic
```

darts/models/forecasting/forecasting_model.py
dennisbader, line 1: we should add this to the changelog

dennisbader (approved):
Very nice, this looks awesome 🚀 
This time really only minimal suggestions, we can merge soon.

dennisbader (approved):
Everything looks good now, great job @madtoinou 🚀

pfk-beta:
Can't wait for release

PR #2059, opened 2023-11-07
Fix/exp smooth constructor args

### LINKED ISSUE #2024: Option to pass "smoothing_level" to ES model
Yes, we are using ES model to predict the value of the timeseries, now when the next point is being predicted, it is giving a very high weightage to the last point(s). So say if there was a drop in actual data for the last 2 points only, the predicted value goes down

Exponential smoothing model provides the option to pass "smoothing_level"  bounds which is not in Darts. If we get the option to pass this smoothing_level to ES via darts. It would probably resolve the issue. we can have a cap to restrict the weightage.

madtoinou:
Fixes  #2024.

### Summary

- Add `kwargs` argument to `ExponentialSmoothing` to expose all the parameters of the statsmodels model implementation
- Kept the `fit_kwargs` argument unchanged to preserve the API
- Slightly refactored the tests and added cases for the two kinds of kwargs


dennisbader (approved):
Looks great, thanks @madtoinou

PR #2072, opened 2023-11-17
Fix/nlinear normalization for multivariate series

### LINKED ISSUE #2035: [BUG] NLinear fails to predict >1 variates
**Describe the bug**
Predicting two or more variates fails.

**To Reproduce**

```python
from darts.datasets import ETTh1Dataset
from darts.models import NLinearModel

series = ETTh1Dataset().load().astype("float32")

target = series["HUFL"].stack(series["LULL"])
past_cov = series["MULL"]

model = NLinearModel(16, 4, normalize=True)
model.fit(series=target, past_covariates=past_cov)

---> RuntimeError: The size of tensor a (3) must match the size of tensor b (2) at non-singleton dimension 2
```

It works perfectly fine with `normalize=False`.

**Expected behavior**
Not a crash, but nice forecasts.

 - Python version: `3.10.13`
 - PyTorch version: `2.1.0`
 - darts version: `0.26.0`

**Additional context**
Possibly related to #1873.

madtoinou:
Fixes #2035.

### Summary

- Added slicing of the input tensor `x` when normalizing (only the target features based on the last value of the target features)

### Other Information

- The problem does not occur when `shared_weights=True` because covariates are not allowed (and normalization is applied to all the dimensions).


dennisbader (commented):
Looks great, thanks for fixing this @madtoinou

dennisbader (approved):

felixdivo:
Thanks @madtoinou!

PR #2077, opened 2023-11-20
Synthesize window-based normalization

LeonieFreisinger:
### Summary
This dart branch implements window-based normalization for the RNN. It considers using the model with or without window-based normalization depending on the preferences set in tot or hpdnorm.

### Features
- optionally enable window-based normalization 
- distinguish between instance and batch norm_type
- distinguish between affine_norm= True or False
- distinguish between mean and standard normalization


LeonieFreisinger:
Was not meant to be

PR #2087, opened 2023-11-22
Bump version of pytorch-lightning to 2.0.0

### LINKED ISSUE #2086: [BUG] torch_forecasting_model depends on pytorch-lightining>=2.0
**Describe the bug**
As of version 0.27 darts is not compatible with pytorch-lightning versions prior to 2.0 due to the dependency on the `pytorch_lightning.callbacks.ProgressBar` class. This class was introduced in pytorch-lightning 2.0. This isn't reflected in the requirements, which only specify pytorch-lightning>=1.5

**To Reproduce**
Attempt to import darts with pytorch-lightning<2.0. This will raise an error on line 43 of torch_forecasting_model.py, `from pytorch_lightning.callbacks import ProgressBar`.

eschibli:
Fixes #2086 

### Summary

Changed torch requirements to require pytorch-lightning>=2.0 due to dependency on ProgressBar


dennisbader (changes requested):
Hi @eschibli and thanks for this fix.

It might be a bit early for us to discard lightning<2.0.0.

What I would suggest is to adapt the ProgressBar import in `darts.models.forecasting.torch_forecasting_model.py`. 

Move the progressbar import in [this line](https://github.com/unit8co/darts/blob/dd08413318258f2f2d37aa07735d7ffc9b2e6de0/darts/models/forecasting/torch_forecasting_model.py#L43) into the version specific imports from [here](https://github.com/unit8co/darts/blob/dd08413318258f2f2d37aa07735d7ffc9b2e6de0/darts/models/forecasting/torch_forecasting_model.py#L99):

```
if pl_200_or_above:
    from pytorch_lightning.tuner import Tuner
    from pytorch_lightning.callbacks import ProgressBar
else:
    from pytorch_lightning.tuner.tuning import Tuner
    from pytorch_lightning.callbacks import ProgressBarBase
```

dennisbader:
Hi @eschibli. I applied the changes already, as we want to release a patch for this soon.
Thanks again for your contribution 🚀

dennisbader (approved):
LGTM, thanks @eschibli 🚀 💯

PR #2091, opened 2023-11-24
Fix/update doc

### LINKED ISSUE #1297: Documentation update regarding global models
Currently, [this page A](https://unit8co.github.io/darts/examples/01-multi-time-series-and-covariates.html) (section "Global Forecasting Models") says that there are 5 "global models" which support training on multiple series. However, [this page B](https://unit8co.github.io/darts/README.html) (section "Forecasting Models") has a tabular overview of all the forecasting models, where 12 models have a checkmark for "Multiple series training" (incl. RegressionModel, RandomForest, LinearRegressionModel etc.)

Could you please confirm that page A is out of date? My understanding is that the 12 models from page B are indeed global models, i.e. 1 model is trained that uses knowledge from multiple time series, _not_ multiple models are trained with one model specifically for each time series? 

It might make sense to update page A and refer to page B for a list of the global models. 

PS.: This is not a bug but a request regarding the documentation. There was no such option when creating a new issue.

### LINKED ISSUE #2001: add VARIMA, ... probabilistic support to readme

### LINKED ISSUE #2063: [BUG] timeseries split function doesn't  work align with percentage properly
Hi,
I was going through the [Quickstart](https://unit8co.github.io/darts/quickstart/00-quickstart.html) and ran into `series.split_before(0.75)`. As the `AirPassengersDataset`, total 144 lines(time steps) in the dataset. Given 75% percent we can get the index to split, which is 108. From my perspective, split_before index 108 means, the first half serie will have time steps from 0 to 107, which means there will be 108 time steps in the first half.

However, the result of `series.split_before(0.75)` splits the first half from index 0 to index 106 without the time step at 107, which is counter-intuitive. I've spotted the line of code which lead to this "issue". In [timeseries.py](https://github.com/unit8co/darts/blob/master/darts/timeseries.py#L2058), `point_index = int((len(self) - 1) * point)`. The split index is calculated by the `length - 1` times point(0.75 in the case above), which is `int((144 -1 ) * 0.75)=107`. Then the rest of the job is just split at index 107 without time step 107.

I'm not sure if its a determined bug, since the rest of Quickstart code split at index 108 other than 107. For example, just below the `series.split_before(0.75)`, there is `series1, series2 = series[:-36], series[-36:]`. I guess the intention of the author is to split the dataset with 75%(108 elements) and 25%(36 elements). But with `series.split_before(0.75)`, the dateset is split into 74%(107 elements) and 26%(37 elements).

Hope someone could address this, since the usage of percentage of `series.split_before(0.75)` is not completely align with the fraction.

**To Reproduce**
Simply run to the 3rd code block in the official [Quickstart](https://unit8co.github.io/darts/quickstart/00-quickstart.html).

 - Python version: 3.9.18
 - darts version: u8darts-all 0.26.0
 - OS: Linux

madtoinou:
Fixes #2001, fixes #1297, fixes #2063.

### Summary

- Updated the tables summarizing the models for probabilistic and multivariate support (naive models, VARIMA)
- Added a link tot he models table in the multi time series and multivariate notebook
- Replaced `split_before` with `split_after` in the quickstart notebook to make the example more coherent.


examples/01-multi-time-series-and-covariates.ipynb
```diff
@@ -167,13 +167,13 @@
    "metadata": {},
```
dennisbader, line 167: <p>"""</p><ul><li>The&nbsp;<strong>target time series</strong>&nbsp;is the time series we are interested to forecast (given its history)</li></ul><p>"""</p><p>maybe some words about static covariates and a link to the static covariates notebook could help here as well</p><p> </p><p> </p><p>----</p><p> </p><p>"""</p><p>Some models use only past covariates, others use only future covariates, and some models might use both. We will dive deeper in this topic in some other notebook, but for now it is enough to know this:</p><p>"""</p><p> </p><p>the list below can also be replaced by adding a link to the readme with covariates support</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/2091/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

dennisbader (approved):
Looks great, thanks @madtoinou 🚀 

Just had a minor suggestion about adapting things for covariates as well

examples/01-multi-time-series-and-covariates.ipynb
```diff
@@ -180,19 +180,18 @@
     "* The **target time series** is the time series we are interested to forecast (given its history)\n",
     "* A **covariate time series** is a time series which may help in the forecasting of the target series, but that we are not interested in forecasting. It's sometimes also called *external data*.\n",
     "\n",
+    "**Note**: Static covariates are invariant in time and correspond to additional information associated with the components of a `TimeSeries`. Detailed information about this kind of covariate can be found in the [static covariates notebook](https://unit8co.github.io/darts/examples/15-static-covariates.html).\n",
```
dennisbader, line 183: associated with the components of the **target time series**.

dennisbader (approved):
Thanks! One last adaption then we're good to go 🚀

examples/01-multi-time-series-and-covariates.ipynb
madtoinou: Can't they also be part of the covariates series?

dennisbader: They can be embedded into the covariates but will not be used by any model.
Static covariates are characteristics of the target series.

PR #2103, opened 2023-12-01
Fix/hist fc predict kwargs

### LINKED ISSUE #2097: [BUG] how to set mc_dropout for historical_forecasts
I am attempting to enable Monte Carlo Dropout for predictions in TFT. I have configured the `predict_kwargs` parameter for `historical_forecasts` as follows:

```
predict_kwargs={
    "mc_dropout": True,
}
predict_series = my_model.historical_forecasts(test_series[target_val],retrain=False,num_samples=2,predict_kwargs=predict_kwargs)
```

However, the results indicate that the samples are the same. And after executing the code multiple times there is still no any variation.

I have two specific questions:
1. How should I properly set up Monte Carlo Dropout (mc_dropout)?
2. When using historical_forecasts with `num_samples > 1` and `mc_dropout = True`, should I expect to obtain different samples in each run?

dennisbader:
Fixes #2097.

### Summary
- fixes a bug where some `fit_kwargs` and `predict_kwargs` were ignored in historical forecasts when the model's fit/predict methods had *args and **kwargs in the signature
- moved predict logic from mixed covariates models to MixedCovariatesTorchForecastingModels -> N/DLinear now use the proper logic
- fixed some text in docs


madtoinou (approved):
LGTM, thanks @dennisbader for the fix!

PR #2142, opened 2024-01-03
Improve description for ARIMA parameters (p, q, seasonal_orders and trend)

MarcBresson:
### Summary

It provides more context to some features that are available in statsmodels but were not mentionned in Darts documentation.

### Other Information

See https://github.com/statsmodels/statsmodels/pull/9065 that


MarcBresson:
Black has nothing to reformat but gradle still fails.

Screenshot 2024-01-16 at 15 09 50

dennisbader:
Hi @MarcBresson, simply run black like below (we don't use line length 120)

`black darts/models/forecasting.arima.py`

dennisbader:
@MarcBresson , now isort is complaining. I suggest installing the pre-commit hook for automatic lint/reformatting as described [here](https://github.com/unit8co/darts/blob/master/CONTRIBUTING.md#code-formatting-and-linting)

MarcBresson:
All the tests passed, I rebased the branch so its ready to be merged

darts/models/forecasting/arima.py
```diff
@@ -32,10 +32,17 @@ class ARIMA(TransferableFutureCovariatesLocalForecastingModel):
     def __init__(
         self,
         p: int = 12,
-        d: int = 1,
-        q: int = 0,
-        seasonal_order: Tuple[int, int, int, int] = (0, 0, 0, 0),
-        trend: Optional[str] = None,
+        d: Union[int, Iterable[int]] = 1,
```
VascoSch92, line 35: Why are you using Iterable instead of List? Can you give an example where you don't have a list of integer (or an integer)?

darts/models/forecasting/arima.py
```diff
@@ -32,10 +32,17 @@ class ARIMA(TransferableFutureCovariatesLocalForecastingModel):
...
+        q: Union[int, Iterable[int]] = 0,
+        seasonal_order: Tuple[
```
VascoSch92, line 37: You can also define this type at the top of the script and use it here. Because now it is not so readable.

darts/models/forecasting/arima.py
MarcBresson: Oops, it should be a sequence (it is a subtype of Iterable but without sets or dicts). A sequence is compatible with tuples, numpy arrays, torch tensors and lists. It is more general than lists.

darts/models/forecasting/arima.py
MarcBresson: I think this is the most readable way. Having a type alias at the top will force a super descriptive name that could be less clear to the end user.

MarcBresson: However, I do agree that seasonal_order is quite hard to grasps. I will check what I can do.

darts/models/forecasting/arima.py
VascoSch92: Ah ok. I was reading the docstring and I was sure we could have just a list or an int. Perhaps update the docstring also

MarcBresson: fixed. Btw, I just checked and torch.Tensor are *not* sequences since

```python
def head(x: Sequence[int]) -> int:
    return x[0]
```

but

```python
def head(x: torch.IntTensor) -> torch.IntTensor:
    return x[0]
```

(from https://discuss.pytorch.org/t/why-is-tensor-not-subclass-of-collections-abc-sequence/110126/2)

MarcBresson:
Should I update the test with

```python
def test_historical_forecasts_transferrable_future_cov_local_models(self):
    model = ARIMA(p=np.array([1, 2, 3, 4, 6]), q=(2, 3))
    ...
```
?

madtoinou:
Instead of modifying the current test, I would rather add new one so that the difference between the two set of parameters can be highlighted/verified.

darts/models/forecasting/arima.py
```diff
@@ -45,20 +53,29 @@ def __init__(
...
-            Parameter controlling the deterministic trend. 'n' indicates no trend,
-            'c' a constant term, 't' linear trend in time, and 'ct' includes both.
-            Default is 'c' for models without integration, and no trend for models with integration.
+        trend: Literal['n', 'c', 't', 'ct'] | Sequence[int], optional
+            Parameter controlling the deterministic trend. Can be specified as
+            a string where 'c' indicates a constant term, 't' indicates a
+            linear trend in time, and 'ct' includes both. Can also be specified
+            as an sequence defining a polynomial, as in `numpy.poly1d`, where
+            `[1,1,0,1]` would denote :math:`a + bt + ct^3`. Default is 'c' for
+            models without integration, and no trend for models with integration.
```
dennisbader, line 78: you can remove the left over line from above (trend: str)


```suggestion
        trend: Literal['n', 'c', 't', 'ct'] | Sequence[int], optional
            Parameter controlling the deterministic trend. Either a string or sequence of integers.
            If a string, can be 'n' for no trend, 'c' for a constant term, 't' for a linear trend in time,
            and 'ct' for a constant term and linear trend.
            If a sequence of integers, defines a polynomial according to `numpy.poly1d` [1]_. E.g., `[1,1,0,1]` would
            translate to :math:`a + bt + ct^3`.
            Default is 'c' for models without integration, and 'n' for models with integration.
```

For the reference (`[1]_`) you have to add the following below the parameter descriptions (above the example)

```
        References
        ----------
        .. [1] https://numpy.org/doc/stable/reference/generated/numpy.poly1d.html
```

darts/models/forecasting/arima.py
```diff
@@ -45,20 +53,29 @@ def __init__(
...
         add_encoders
```
dennisbader, line 79: the `random_state` parameter is missing, could you add it?

.pre-commit-config.yaml
```diff
@@ -6,7 +6,7 @@ repos:
         language_version: python3
 
   - repo: https://github.com/PyCQA/flake8
-    rev: 4.0.1
+    rev: 7.0.0
```
dennisbader, line 9: let's add this to the CHANGELOG.md under 
**UNRELEASED -> For developers of the library**

darts/models/forecasting/arima.py
dennisbader, line 1: Let's show your contribution in the CHANGELOG.md :) 

You can create a new entry for the documentation improvements under 

**UNRELEASED -> Improved**

dennisbader (commented):
Thanks a lot for this nice contribution @MarcBresson, I didn't even know that you could use a sequence of lags for more fine grained control over the parameters. That's great!

I added some suggestions that are just trying to rephrase some bits here and there to follow more our style from other models.

After that it's ready to be merged 🚀

darts/models/forecasting/arima.py
MarcBresson: Will the reference [1] in ARIMA's doc collide with the file docstring?

.pre-commit-config.yaml
MarcBresson: Should I update other pre-commit hooks ? That way we are up to date and developers will not have to `pre-commit install` anytime soon.

darts/models/forecasting/arima.py
MarcBresson: I don't see any comments in the release notes of statmodels 14.0 about this https://github.com/unit8co/darts/pull/1766. Furthermore, I searched for the diff between 13.5 and 14.0 and could not see anything related to non-reproducibility (https://github.com/statsmodels/statsmodels/compare/v0.13.5...v0.14.0). Is it parameter useful ?

.pre-commit-config.yaml
MarcBresson: I updated all pre-commit hooks and change the argument for pyupgrade. See changelog (or commit description) for more info

.pre-commit-config.yaml
dennisbader, line 1: actually, thinking a bit more about this, we should better apply these changes in a separate PR. This could introduce some issues, and I'd rather have it in another commit. Could you me this file out and remove the entry from the changelog?

.pre-commit-config.yaml
```diff
@@ -1,23 +1,23 @@
...
     hooks:
       -  id: isort
 
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.31.0
+    rev: v3.15.0
     hooks:
     -   id: pyupgrade
-        args: ['--py37-plus']
+        args: ['--py39-plus']
```
dennisbader, line 23: 3.8 is our minimum python version
```suggestion
        args: ['--py38-plus']
```

CHANGELOG.md
```diff
@@ -10,10 +10,13 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ### For users of the library:
 **Improved**
+- Improvements to `ARIMA` documentation: Specified possible `p`, `d`, `P`, `D`, `trend` advanced options that are available in statsmodels. More explanations on the behaviour of the parameters were added. [#2142](https://github.com/unit8co/darts/pull/2142) by [MarcBresson](https://github.com/MarcBresson)
 
 **Fixed**
 
 ### For developers of the library:
+- Updated pre-commit hooks to the latest version using `pre-commit autoupdate`.
+- Change `pyupgrade` pre-commit hook argument to `--py39-plus`. This allows for [type rewriting](https://github.com/asottile/pyupgrade?tab=readme-ov-file#pep-585-typing-rewrites).
```
dennisbader, line 19: See my comment above
```suggestion
```

darts/models/forecasting/arima.py
dennisbader: The random state is from darts to get reproducible result for probablistic forecasts.
see [here](https://www.statsmodels.org/stable/generated/statsmodels.tsa.arima.model.ARIMA.simulate.html).

I see we don't have this parameter described in the other local models either. We can do it in another PR then

dennisbader (changes requested):
Hi @MarcBresson, thanks for the changes. Everything loks fine now. 

As a last point, could you move the changes to the pre commit hooks to a separate PR? Like that we keep it clearly separated.

After that we can merge :) 🚀

.pre-commit-config.yaml
MarcBresson: Done ! https://github.com/unit8co/darts/pull/2228

CHANGELOG.md
MarcBresson: I droped the commit entirely

.pre-commit-config.yaml
dennisbader: No, all good. From your changes in this PR, it didn't update any other files.

CHANGELOG.md
dennisbader: Nice 🚀

PR #2152, opened 2024-01-10
Fix: Improve TimeSeries.__getitem__ frequency inference

### LINKED ISSUE #2089: [BUG] TimeSeries slicing silently changes the TimeSeries' frequency
**Describe the bug**
When using BusinessDay as frequency, selecting a slice e.g. ts[1:5] can result in the slice having a different frequency (Day).
This is problematic because subsequent concatenations can fail because the two TimeSeries won't be contiguous at the altered frequency (but would be contiguous at the original frequency)

**To Reproduce**
```python
from darts import TimeSeries, concatenate
import pandas as pd
import numpy as np

offset = pd.offsets.BusinessDay()   # Closed on Saturdays & Sundays
dates1 = pd.date_range("2023-11-01","2023-11-26", freq=offset)   # the 26th is a sunday, last date is the 24th
values1 = np.ones(len(dates1))
ts1  = TimeSeries.from_times_and_values(dates1,values1)
dates2 = pd.date_range("2023-11-26","2023-12-06", freq=offset)  # First date is the 27th
values2 = np.ones(len(dates2))
ts2 = TimeSeries.from_times_and_values(dates2, values2)

concatenate((ts1,ts2)) # works
print(ts1.freq) # <BusinessDay>
ts_subset = ts1[-4:]
print(ts_subset.freq) # <Day>
concatenate((ts_subset, ts2)) # Fails 
````

**Expected behavior**
TimeSeries slicing should not silently alter the frequency

 - Python version: 3.10
 - darts version: Tested on master and 0.25

DavidKleindienst:
Fixes #2089 .

### Summary
Currently when indexing a `TimeSeries`, the frequency of the new `TimeSeries` is inferred from dates of the new `TimeSeries`.
However, we often have information to do a more correct inference, in particular in the following cases:

- The `__getitem__` argument is a `pd.DatetimeIndex` with a set `.freq` attribute
- We index with a slice 

This PR contains the following improvements to frequency inference in the `TimeSeries.__getitem__` method:

- When a `pd.DatetimeIndex` with a set `.freq` attribute is used as `__getitem__` argument, we use this attribute as the new `TimeSeries` frequency
- When a `slice` without `step` parameter is used as `__getitem__` argument, the original `TimeSeries` frequency is used
- When a `slice` with `step` parameter is used as `__getitem__` argument, the new frequency is derived by multiplying `step` with the original `TimeSeries` frequency
- A unittest for frequency inference is added

This means that inferring the frequency from the dates of the new `TimeSeries` will only be used either when a `pd.DatetimeIndex` without `.freq` attribute is used as `__getitem__` argument or when indexing by a single `int` or `pd.Timestamp` value.

RangeIndexed `TimeSeries` are not affected by this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ### For users of the library:
 **Improved**
+- Improvements to `TimeSeries`:
+  - Improved the time series frequency inference when using slices or pandas DatetimeIndex as keys for `__getitem__`. [#2152](https://github.com/unit8co/darts/pull/2152) by [DavidKleindienst](https://github.com/DavidKleindienst).
 
 **Fixed**
 - Fixed a bug when using a `TorchForecastingModel` with `use_reversible_instance_norm=True` and predicting with `n > output_chunk_length`. The input normalized multiple times. [#2160](https://github.com/unit8co/darts/pull/2160) by [FourierMourier](https://github.com/FourierMourier).
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -17,7 +17,6 @@ from darts.utils.timeseries_generation import (
 
 
 class TestTimeSeries:
-
     times = pd.date_range("20130101", "20130110", freq="D")
     pd_series1 = pd.Series(range(10), index=times)
     pd_series2 = pd.Series(range(5, 15), index=times)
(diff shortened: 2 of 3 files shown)


dennisbader (approved):
Thanks for this @DavidKleindienst , looks great 🚀 

I made some minor adaptions so the frequency inference applies identically to integer indexed `TimeSeries`.

dennisbader:
@DavidKleindienst, you can ignore the failing unit tests for now, it is a problem with one of the dataset sources that they're fixing now.

PR #2170, opened 2024-01-19
fix deprecated method import from sklearn==1.4.0

dennisbader:
### Summary
- fixes failing unit tests due to new sklearn v1.4.0: method `_check_fit_params` was renamed to `_check_method_params`


AK6263:
Can you please provide an alternative fix I want to run the models

dennisbader:
@AK6263, you can just install a lower version of sklearn before installing darts.

```
pip install "scikit-learn<1.4.0"
pip install darts
```

PR #2176, opened 2024-01-20
Feat/shifted output

### LINKED ISSUE #2132: cannot figure out how to add 'offset' when training (gap between training data and forecasting period/testing)
Hello again 👋

I have tried to find to add a gap between testing and the forecasting period. Lets say i am training a model and i want it to predict days for, NOT next week, but the week after. So if the timestep is daily (one per day) , then i would need a gap of 7. 

I have read several of your tutorials, but as before, i could have misunderstood.  Of what i can discern, it says nothing about adding a 'gap' in either the TimeSeries.from_dataframe or model.fit(). The docs seem to be suggesting that i should look at fit_from_dataset() , but from what i can see, there is no explaining how to use that method for adding a gap. 

I am pretty sure i could add the gap with regular python for loop and then convert that to a timeseries object, though then (but i am not sure if it would work as i wanted). Tensorflow, in their timeSeries guide also shows how to add a gap/offset:
https://www.tensorflow.org/tutorials/structured_data/time_series.

Anyways, Merry Christmas and thanks for your awesome work on darts 🎄

### LINKED ISSUE #2139: question: training the model to predict a single future step

Hi, i have a question, I was wondering how we would specify the training data or pass the data into model.fit() if we want to provide a sequence to predict a single step thats further out into the future ? My understanding right now is that given a input chunk length, we are predicting the next `output_chunk_length` into the future. For example, given `input_chunk_length = 100`, and `output_chunk_length=1`, we are predicting the next step into the series. 

Is there a way for us to predict a step into the future without changing the `output_chunk_length` ? For example, using `input_chunk_length=100` and `output_chunk_length=1` but training the model to predict the 100th step into the future ?

Providing the option to specify the `stride_step` if needed, similar to the `forecast_horizon` method for models

There is also the function of `fit_with_dataset` for this, but i think providing the `stride_step` argument in `model.fit()` would be better.

**Additional context**

dennisbader:
Work in progress

Fixes #2139, fixes #2132

### Summary
- adds support for shifting the output chunk of global models with parameter `output_chunk_shift`
- add support to regression models
- fixes bug when using regression models with `lags=None` and `future_covariates` with positive only lags and `future_covariates` starting at or after the first predictable time step. 
- parametrize most of the tabularization unit tests
- introduces new helper function `darts.utils.utils.n_steps_between()` to efficiently compute the number of time steps (periods) between two points/timestamps with a given frequency -> improves efficiency for regression model tabularization by avoiding `pd.date_range()`.


noahvand:
Tried this version, and the shift output function works well, but the RNN models cannot work for some reason, where I got ```
"WARNING:darts.models.forecasting.rnn_model:ignoring user defined `output_chunk_length`. RNNModel uses a fixed `output_chunk_length=1`.
WARNING:darts.models.forecasting.rnn_model:ignoring user defined `output_chunk_shift`. RNNModel uses a fixed `output_chunk_shift=0`."
```

**Comment by madtoinou** [comment]
Hi @noahvand,

This is intentional because by definition `RNNModel` enforces `output_chunk_length=1` and the `hidden state` mechanism is not really compatible with "gaps" between input values and forecast: introducing such a shift would require the `RNNModel` to still forecast `output_chunk_shift` values before getting to the actual value of interest.

For the sake of uniform API, we could change this but `RNNModel` is kind of already in its own category anyway.

File: darts/tests/models/forecasting/test_dlinear_nlinear.py [hunk 1]
```diff
@@ -338,8 +339,92 @@ def test_optional_static_covariates(self):
...
+                    **tfm_kwargs,
                 )
                 model.fit(series.with_static_covariates(None))
                 preds = model.predict(n=2, series=series)
                 assert preds.static_covariates.equals(series.static_covariates)
+
+        @pytest.mark.parametrize(
+            "config", itertools.product([3, 7, 10], [NLinearModel, DLinearModel])
+        )
+        def test_output_shift(self, config):
```
madtoinou, line 351: Since this test is pretty much copy-pasted in every torch forecasting model unit-tests, would probably be better to move it to `test_torch_forecasting_model.py`?

Also, we should also add a short test to check that `historical_forecasts()` returns a ts with the expected time index (same length - assuming covariates are long enough-, but shifted)

darts/utils/data/inference_dataset.py
```diff
@@ -193,6 +197,23 @@ def __init__(
                 logger=logger,
             )
 
+        if output_chunk_shift and n > output_chunk_length:
+            raise_log(
+                ValueError(
+                    "Cannot perform auto-regression `(n > output_chunk_length)` with a model that uses a "
+                    "shifted output chunk `(output_chunk_shift > 0)`."
```
madtoinou, line 204: Nice!

madtoinou (commented):
Some very minor comments for the tests.

Nice that this could be implemented so elegantly!

darts/tests/models/forecasting/test_dlinear_nlinear.py
dennisbader: True!

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -1539,3 +1540,117 @@ def helper_create_DLinearModel(
...
+                    (BlockRNNModel, {}),
+                ],
+                [3, 7, 10],
+            ),
+        )
+        def test_output_shift(self, config):
+            """Tests shifted output for shift smaller than, equal to, and larger than output_chunk_length.
+            RNNModel does not support shift output chunk.
+            """
+            # model_cls = TFTModel
```
madtoinou, line 1565:
```suggestion
```

madtoinou (approved):
LGTM 🚀

darts/models/forecasting/forecasting_model.py
```diff
@@ -412,12 +412,13 @@ def extreme_lags(
         Optional[int],
         Optional[int],
         Optional[int],
+        Optional[int],
     ]:
         """
         A 6-tuple containing in order:
```
madtoinou, line 418:
```suggestion
        A 7-tuple containing in order:
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -412,12 +412,13 @@ def extreme_lags(
         Optional[int],
         Optional[int],
         Optional[int],
+        Optional[int],
```
madtoinou, line 415: Is it really `Optional`? I have the impression that it can be defined for all the models, it will just be 0 by default (and never `None`)?

darts/models/forecasting/forecasting_model.py
```diff
@@ -285,6 +285,13 @@ def output_chunk_length(self) -> Optional[int]:
         """
         return None
 
+    @property
+    def output_chunk_shift(self) -> Optional[int]:
+        """
+        Number of time steps that the output/prediction starts after the end of the input.
+        """
+        return 0
```
VascoSch92, line 293: the output should not be `None`? To be consistent with `output_chunk_length(self) -> Optional[int]` ?

or the `return` type should be `Optional`.

dennisbader: 👍

darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -1909,6 +1935,308 @@ def test_component_specific_lags(self, config):
...
+        # pred starts with a shift
+        for ocl_test in [ocl - 1, ocl]:
+            pred = model_target_only.predict(n=ocl_test)
+            assert pred.start_time() == series.end_time() + (shift + 1) * series.freq
+            assert len(pred) == ocl_test
+            assert pred.freq == series.freq
+
+        # check that shifted output chunk results with encoders are the
+        # same as using identical covariates
+        model = LinearRegressionModel(
```
madtoinou, line 2129: This model is not used in the rest of the test (aside from generating the covariates dictionary), `model_enc_shift` and `model_fc_shift` are. Wouldn't it be simpler to just check the keys in the lags dictionary?

madtoinou (approved):
Minor comments, amazing PR @dennisbader  🚀

PR #2205, opened 2024-02-03
Fix/RegressionEnsemble with single model regressor and coef access in LinearRegressionModel

### LINKED ISSUE #2200: Get coefficients of Regression and RegressionEnsemble model
Given N models, which are fit. In my case below I the models are already fit:

```
ensemble_probabilistic = RegressionEnsembleModel(forecasting_models=forecast_models, 
                                                 train_forecasting_models=False,
                                                 regression_train_n_points=-1,
                                                 train_using_historical_forecasts=False,
                                                 regression_model=LinearRegressionModel(
                                                     lags_future_covariates=[0],
                                                     lags=None, 
                                                     likelihood="quantile", quantiles=quantiles,
                                                    ),
                                                )

ensemble_probabilistic.fit(
        series=train_transformed,
        past_covariates=time_covariates_transformed,
        future_covariates=time_covariates_transformed,
)

pred_ens = ensemble_probabilistic.predict(len(val), num_samples=300)
```

It's not clear how to extract the N coefficients of the (probabilistic) Regression model.

I find that `len(ensemble_probabilistic.regression_model.model.estimators_)` == num_multivariate_ts ( or num of cols in the df)
but each of the `estimators_` is a `QuantileRegressor(quantile=0.9, solver='highs')` object. Even the first and last.

How can I extract the N coefficient weights? 
Separately, is there built in way to do pruning on the N models to select the K best? (K<=N) criteria like AIC, BIC, or an objective like MSE or MASE can work.

madtoinou:
Fixes #2200.

### Summary
- prevent creation of RegressionEnsemble with a `regression_model` with `multi_models=False` as it would require additional forecasts (and potentially auto-regression) from the `forecasting_models` when calling `predict()`.
- overwrite the `model` attribute of probabilistic `LinearRegressionModel` with the sub-models containers to make the access of the coefficients more intuitive

### DIFF
--- a/darts/models/forecasting/regression_ensemble_model.py
+++ b/darts/models/forecasting/regression_ensemble_model.py
@@ -123,6 +123,11 @@ class RegressionEnsembleModel(EnsembleModel):
                 lags=None, lags_future_covariates=[0], fit_intercept=False
             )
         elif isinstance(regression_model, RegressionModel):
+            raise_if_not(
+                regression_model.multi_models,
+                "`regression_model.multi_models = False` is not supported for `RegressionEnsembleModel.",
+                logger,
+            )
             regression_model = regression_model
         else:
             # scikit-learn like model
--- a/darts/models/forecasting/regression_ensemble_model.py
+++ b/darts/models/forecasting/regression_ensemble_model.py
@@ -125,7 +125,7 @@ class RegressionEnsembleModel(EnsembleModel):
         elif isinstance(regression_model, RegressionModel):
             raise_if_not(
                 regression_model.multi_models,
-                "`regression_model.multi_models = False` is not supported for `RegressionEnsembleModel.",
+                "Cannot use `regression_model` that was created with `multi_models = False`.",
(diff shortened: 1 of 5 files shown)


darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -123,6 +123,11 @@ def __init__(
                 lags=None, lags_future_covariates=[0], fit_intercept=False
             )
         elif isinstance(regression_model, RegressionModel):
+            raise_if_not(
+                regression_model.multi_models,
+                "`regression_model.multi_models = False` is not supported for `RegressionEnsembleModel.",
```
dennisbader, line 128: Maybe below could be more concise
```suggestion
                "Cannot use `regression_model` that was created with `multi_models = False`.",
```

dennisbader (approved):
Thanks for this fix @madtoinou, looks great! 🚀 
Just the changelog entry, then we're good to go 💯

PR #2207, opened 2024-02-05
fix: update hierarchy for single transform window_transform

### LINKED ISSUE #2204: [BUG] window_transform on TimeSeries with hierarchy
**Describe the bug**
window_transform on TimeSeries with hierarchy raises

```python
ValueError: The keys of the hierarchy must be time series components
```

**To Reproduce**
```python
from darts import TimeSeries, concatenate
from darts.datasets import AustralianTourismDataset
# Temp
tourism_series = AustralianTourismDataset().load()
hierarchy = dict()

reasons = ["Hol", "VFR", "Bus", "Oth"]
regions = ["NSW", "VIC", "QLD", "SA", "WA", "TAS", "NT"]
city_labels = ["city", "noncity"]

# Fill in grouping by reason
for reason in reasons:
    hierarchy[reason] = ["Total"]

# Fill in grouping by region
for region in regions:
    hierarchy[region] = ["Total"]

# Fill in grouping by (region, reason)
for region, reason in itertools.product(regions, reasons):
    hierarchy["{} - {}".format(region, reason.lower())] = [reason, region]

# Fill in grouping by (region, reason, <city>)
for region, reason, city in itertools.product(regions, reasons, city_labels):
    hierarchy["{} - {} - {}".format(region, reason.lower(), city)] = [
        "{} - {}".format(region, reason.lower())
    ]
tourism_series = tourism_series.with_hierarchy(hierarchy)
tourism_series.window_transform(transforms={
    "function": "mean",
    "mode": "rolling",
    "window": 28,
    "components": ["Total"]
})
```

**Expected behavior**
```python
tourism_series["Total"].window_transform(transforms={
    "function": "mean",
    "mode": "rolling",
    "window": 28,
})
```

Python version
3.11.5 | packaged by Anaconda, Inc. | (main, Sep 11 2023, 13:26:23) [MSC v.1916 64 bit (AMD64)]
Darts version
0.27.2

### LINKED ISSUE #2209: ‎TimeSeries.resample() Wrong method name in docs.
Wrong method name in docs.
https://github.com/unit8co/darts/blob/5b05d2b4a798be9a4206692a8c486d8836ec4c20/darts/timeseries.py#L3191
Current correct darts method name — `'bfill'`.  But i think `'backfill'` is more appropriate.
https://github.com/unit8co/darts/blob/5b05d2b4a798be9a4206692a8c486d8836ec4c20/darts/timeseries.py#L3234-L3238

madtoinou:
Fixes #2204, fixes #2209.

### Summary

- If `window_transform()` is called with only one transform on all the components of the series, update the hierarchy with the new components names, otherwise, it's dropped.

### DIFF
--- a/darts/tests/dataprocessing/transformers/test_window_transformations.py
+++ b/darts/tests/dataprocessing/transformers/test_window_transformations.py
@@ -128,6 +128,42 @@ class TestTimeSeriesWindowTransform:
             }  # forecating_safe=True vs center=True
             self.series_univ_det.window_transform(transforms=window_transformations)
 
+        # keep_old_names and overlapping transforms
+        with pytest.raises(ValueError):
+            window_transformations = [
+                {
+                    "function": "mean",
+                    "mode": "rolling",
+                    "window": 3,
+                    "components": self.series_multi_det.components[:1],
+                },
+                {
+                    "function": "median",
+                    "mode": "rolling",
+                    "window": 3,
+                    "components": self.series_multi_det.components,
+                },
+            ]
+            self.series_multi_det.window_transform(
+                transforms=window_transformations, keep_old_names=True
+            )
+
+        # keep_old_names and keep_non_transformed
+        with pytest.raises(ValueError):
+            window_transformations = [
+                {
+                    "function": "mean",
+                    "mode": "rolling",
+                    "window": 3,
+                    "components": self.series_multi_det.components[:1],
+                },
+            ]
+            self.series_multi_det.window_transform(
+                transforms=window_transformations,
+                keep_old_names=True,
+                keep_non_transformed=True,
+            )
+
     def test_ts_windowtransf_output_series(self):
         # univariate deterministic input
         transforms = {"function": "sum", "mode": "rolling", "window": 1}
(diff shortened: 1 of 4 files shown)


darts/timeseries.py
```diff
@@ -3607,6 +3607,20 @@ def _get_kwargs(transformation, forecasting_safe):
         if isinstance(transforms, dict):
             transforms = [transforms]
 
+        comp_names_map = None
+        if self.hierarchy:
+            several_transforms = len(transforms) > 1
+            partial_transform = (
+                "components" in transforms[0]
+                and transforms[0]["components"] != self.components
```
dennisbader, line 3615: It will raise an error, since we're comparing a list with a pandas Index.
Also this comparison is not order-agnostic. Can we improve it?

darts/timeseries.py
dennisbader, line 1: I noticed that when performing a single transform on all components, the hierarchy is also not preserved because the component names change.
Should we try to keep the hierarchy? Also for multiple transforms, we could keep the hierarchy across different transforms. What do you think?

darts/timeseries.py
```diff
@@ -3688,6 +3702,14 @@ def _get_kwargs(transformation, forecasting_safe):
                 [f"{name_prefix}_{comp_name}" for comp_name in comps_to_transform]
             )
 
+            if comp_names_map:
```
dennisbader, line 3705: at this point, `comp_names_map` will always be `None` or an empty dict

dennisbader (changes requested):
Thanks for giving this a go @madtoinou 🚀 
It seems like the issue has not yet been fully fixed with these changed (i tried running the code snippet from the issue).

I left some comments/suggestions

darts/timeseries.py
madtoinou: Yeah, this is the reason for the error mentioned in the issue.
For me, the hierarchy can be kept only if the transform is applied to all the components (otherwise, the relationship between the components might be lost).
If there are multiple transforms, the hierarchy can be kept (kind of duplicated) but it'll become quite difficult to read. Not sure how useful it will be but I will adapt the code so that it's supported.

darts/timeseries.py
```diff
@@ -3607,6 +3612,57 @@ def _get_kwargs(transformation, forecasting_safe):
...
+        convert_hierarchy = False
+        if self.hierarchy:
+            # the partial_transform covers for scenario keep_non_transformed = True
+            if several_transforms or partial_transforms:
+                logger.warning(
+                    "The hierarchy cannot be retained, either because there is more than one transform or "
+                    "because the transform is not applied to all the components of the series."
+                )
+            else:
+                comp_names_map = dict()
```
dennisbader, line 3663: Let's define this dict for all cases (outside the if/else scope). Then we avoid the "this variable might be undefined" warnings further down

darts/timeseries.py
```diff
@@ -3684,9 +3740,19 @@ def _get_kwargs(transformation, forecasting_safe):
...
+            if keep_old_names:
+                new_columns.extend(comps_to_transform)
+            else:
+                new_columns.extend(
+                    [f"{name_prefix}_{comp_name}" for comp_name in comps_to_transform]
+                )
+                if convert_hierarchy:
+                    comp_names_map.update(
+                        {
+                            comp_name: f"{name_prefix}_{comp_name}"
```
dennisbader, line 3752: Let's do zip() and use the new names from the list above

darts/tests/dataprocessing/transformers/test_window_transformations.py
```diff
@@ -128,6 +128,42 @@ def test_ts_windowtransf_input_dictionary(self):
...
+                {
+                    "function": "median",
+                    "mode": "rolling",
+                    "window": 3,
+                    "components": self.series_multi_det.components,
+                },
+            ]
+            self.series_multi_det.window_transform(
+                transforms=window_transformations, keep_old_names=True
+            )
```
dennisbader, line 149: Shall we check that it raises the correct error (message)?

darts/timeseries.py
dennisbader, line 1: Can we add the new arguments to the WindowTransformer as well?

dennisbader (commented):
Thanks @madtoinou :) some last few suggestions, then it's ready.

darts/timeseries.py
```diff
@@ -3607,6 +3612,57 @@ def _get_kwargs(transformation, forecasting_safe):
         if isinstance(transforms, dict):
             transforms = [transforms]
 
+        several_transforms = len(transforms) > 1
```
VascoSch92, line 3615: you are defining this variable here but you are using it after 40 lines. 
I will put the variable definition above the line where you use it, or I will not use this variable at all.

PR #2222, opened 2024-02-08
Fix: Using gridsearch with use_fitted_values=True raises unexpected error

madtoinou:
### Summary

When using `gridsearch()` with `use_fitted-values=True`, the model is not properly created to check if it indeed has the attribute `fitted_values` (due to incorrectly assuming that the class had default arguments).

```python
from darts.models import NHiTSModel
import darts.utils.timeseries_generation as tg

ts2 = tg.sine_timeseries(length=25, column_name="dummy_ts_2") + 100

# also fail for regression models
model = NHiTSModel.gridsearch(
    parameters={
        "input_chunk_length":[9],
        "output_chunk_length":[2,3],
    },
    series=ts2,
    use_fitted_values=True
)

>>> TypeError: init() missing 2 required positional arguments: 'input_chunk_length' and 'output_chunk_length
```

### Other Information

Added a check to ensure that each values in the `parameters` dictionary is indeed a list otherwise the `itertools.product` will fail.

### DIFF
--- a/darts/models/forecasting/forecasting_model.py
+++ b/darts/models/forecasting/forecasting_model.py
@@ -1483,10 +1483,21 @@ class ForecastingModel(ABC, metaclass=ModelMeta):
             logger,
         )
 
+        if not all(isinstance(params, list) for params in parameters.values()):
+            raise_log(
+                ValueError(
+                    "Every value in the `parameters` dictionary should be a list (possibly with only one element)."
+                ),
+                logger,
+            )
+
         if use_fitted_values:
             raise_if_not(
-                hasattr(model_class(), "fitted_values"),
-                "The model must have a fitted_values attribute to compare with the train TimeSeries",
+                hasattr(
+                    model_class(**{k: v[0] for k, v in parameters.items()}),
+                    "fitted_values",
+                ),
+                "The model must have a fitted_values attribute to compare with the train TimeSeries (local models)",
                 logger,
             )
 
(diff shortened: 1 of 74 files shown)


darts/models/forecasting/forecasting_model.py
```diff
@@ -1483,10 +1483,28 @@ def gridsearch(
...
         )
 
+        if not isinstance(parameters, dict):
+            raise_log(
+                ValueError(
+                    f"`parameters` should be a dictionary, received a: {type(parameters)}."
+                )
+            )
+
+        if not all(isinstance(params, list) for params in parameters.values()):
```
dennisbader, line 1493: Should we allow np arrays as well?
```suggestion
        if not all(isinstance(params, (list, np.ndarray)) for params in parameters.values()):
```

dennisbader (changes requested):
Thanks @madtoinou for this 🚀 

Should we allow for np arrays as well for the params values?

Also this introduces breaking changes that we need to list in the changelog:
- will raise an error now if params values are not list (or np ndarray)

darts/models/forecasting/forecasting_model.py
madtoinou: Sure, I restricted it to list because it's the type explicitly mentioned in the docstring.

dennisbader (approved):
Looks great, thanks a lot @madtoinou 🚀

PR #2237, opened 2024-02-20
Fix/ts prepend

### LINKED ISSUE #2230: [BUG] Component names are lost when using prepend_values
**Describe the bug**
When using `prepend_values` the component names are lost.

**To Reproduce**
```python
import numpy as np
from darts import concatenate

series1 = linear_timeseries(start_value=1, end_value=2, start=0, length=10, freq=2, column_name='A')
series2 = linear_timeseries(start_value=2, end_value=3, start=0, length=10, freq=2, column_name='B')

series = concatenate([series1, series2], axis=1)
series.columns # returns Index(['A', 'B'], dtype='object', name='component')

prepended = series.prepend_values(np.array([[1, 2], [1, 2]]))
prepended.columns # returns Index(['0', '1'], dtype='object', name='component')
```

**Expected behavior**
`prepended.columns` should be the same as `series.columns`.

**Additional context**
This issue does not happen when `append_values` is used.

```python
appended = series.append_values(np.array([[1, 2], [1, 2]]))
appended.columns # returns Index(['A', 'B'], dtype='object', name='component')
```

Probably it's enough to add `columns=self.columns` here: https://github.com/unit8co/darts/blob/master/darts/timeseries.py#L2875.

madtoinou:
Fixes #2230.

### Summary

- when calling `append/prepend` on two series, the hierarchy and components names of the series with the "earlier" index are retained
- when calling `append/prepend_values`, the hierarchy and components of the series are retained

### Other Information

Not sure if the hierarchy should be retained or not. Since the components don't change, it should be easy to assign the "old" hierarchy to resulting series but at the same time, adding values should not affect the relationship between the components so keeping by default sounds reasonnable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 **Fixed**
 - Fixed a bug in probabilistic `LinearRegressionModel.fit()`, where the `model` attribute was not pointing to all underlying estimators. [#2205](https://github.com/unit8co/darts/pull/2205) by [Antoine Madrona](https://github.com/madtoinou).
 - Raise an error in `RegressionEsembleModel` when the `regression_model` was created with `multi_models=False` (not supported). [#2205](https://github.com/unit8co/darts/pull/2205) by [Antoine Madrona](https://github.com/madtoinou).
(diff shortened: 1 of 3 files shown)


dennisbader (approved):
Looks great, thanks a lot @madtoinou 🚀

PR #2242, opened 2024-02-21
fix: datetime_attribute account for 0 or 1-indexing of the attributes

### LINKED ISSUE #2219: [BUG] datetime_attribute_timeseries does not create correct hot encoding
**Describe the bug**
Hello everyone, when using the darts utils datetime_attribute_timeseries to create hot encoding (one_hot=True) the generated encodings are not correct. Indeed, for some date time attributes. the last encoding is always 0 never 1.

**To Reproduce**
Here is a simple code to reproduce the problem:

```python
from darts.utils.timeseries_generation import datetime_attribute_timeseries as dt_attr

time_index = some_hourly_index_for_one_year
future_covariates = dt_attr(time_index, "hour", one_hot=True)
future_covariates = future_covariates.stack(dt_attr(time_index, "weekday", one_hot=True))
future_covariates = future_covariates.stack(dt_attr(time_index, "month", one_hot=True))
```
When I looked at the future_covariates matrix, the monthly encodings are OK but the hour_24 and weekday_7 are never equal to 1. It seems it comes from the fact that the date time attribute "month" begins at index 1 but hour and week begin at 0.

**Expected behavior**
The hot encoding function should take into account the fact that not all datetime attributes start at index 1 like the month. To make the code work in any case I added the starting and ending index of each datetime attribute I was using (not tested the others) and remove the "+1" in the code.

```python
def datetime_attribute_timeseries(
    time_index: Union[pd.DatetimeIndex, TimeSeries],
    attribute: str,
    one_hot: bool = False,
    cyclic: bool = False,
    until: Optional[Union[int, str, pd.Timestamp]] = None,
    add_length: int = 0,
    dtype=np.float64,
    with_columns: Optional[Union[List[str], str]] = None,
    tz: Optional[str] = None,
) -> TimeSeries:
    """
    Returns a new TimeSeries with index `time_index` and one or more dimensions containing
    (optionally one-hot encoded or cyclic encoded) pd.DatatimeIndex attribute information derived from the index.


    Parameters
    ----------
    time_index
        Either a `pd.DatetimeIndex` attribute which will serve as the basis of the new column(s), or
        a `TimeSeries` whose time axis will serve this purpose.
    attribute
        An attribute of `pd.DatetimeIndex`, or `week` / `weekofyear` / `week_of_year` - e.g. "month", "weekday", "day",
        "hour", "minute", "second". See all available attributes in
        https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DatetimeIndex.html#pandas.DatetimeIndex.
    one_hot
        (results in more columns).
    cyclic
        Boolean value indicating whether to add the specified attribute as a cyclic encoding.
        Alternative to one_hot encoding, enable only one of the two.
        (adds 2 columns, corresponding to sin and cos transformation)
    until
        Extend the time_index up until timestamp for datetime indexed series
        and int for range indexed series, should match or exceed forecasting window.
    add_length
        Extend the time_index by add_length, should match or exceed forecasting window.
        Set only one of until and add_length.
    dtype
        The desired NumPy dtype (np.float32 or np.float64) for the resulting series
    with_columns
        Optionally, specify the output component names.
        * If `one_hot` and `cyclic` are ``False``, must be a string
        * If `cyclic` is ``True``, must be a list of two strings. The first string for the sine, the second for the
            cosine component name.
        * If `one_hot` is ``True``, must be a list of strings of the same length as the generated one hot encoded
            features.
    tz
        Optionally, a time zone to convert the time index to before computing the attributes.

    Returns
    -------
    TimeSeries
        New datetime attribute TimeSeries instance.
    """

    time_index_ts, time_index = _process_time_index(
        time_index=time_index,
        tz=tz,
        until=until,
        add_length=add_length,
    )

    raise_if_not(
        hasattr(pd.DatetimeIndex, attribute)
        or (attribute in ["week", "weekofyear", "week_of_year"]),
        f"attribute `{attribute}` needs to be an attribute of pd.DatetimeIndex. "
        "See all available attributes in "
        "https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DatetimeIndex.html#pandas.DatetimeIndex",
        logger,
    )

    raise_if(one_hot and cyclic,
             "set only one of one_hot or cyclic to true", logger)

    num_values_dict = {
        "month": (1, 13),
        # "day": 31,
        "weekday": (0, 7),
        "dayofweek": (0, 7),
        "day_of_week": (0, 7),
        "hour": (0, 24),
        # "minute": 60,
        # "second": 60,
        # "microsecond": 1000000,
        # "nanosecond": 1000,
        # "quarter": 4,
        # "dayofyear": 365,
        # "day_of_year": 365,
        # "week": 52,
        # "weekofyear": 52,
        # "week_of_year": 52,
    }

    if attribute not in ["week", "weekofyear", "week_of_year"]:
        values = getattr(time_index, attribute)
    else:
        values = (
            time_index.isocalendar()
            .set_index("week")
            .index.astype("int64")
            .rename("time")
        )

    if one_hot or cyclic:
        raise_if_not(
            attribute in num_values_dict,
            f"Given datetime attribute `{attribute}` not supported with one-hot or cyclical encoding. "
            f"Supported datetime attribute: {list(num_values_dict.keys())}",
            logger,
        )

    if one_hot:
        values_df = pd.get_dummies(values)
        # fill missing columns (in case not all values appear in time_index)
        for i in range(num_values_dict[attribute][0], num_values_dict[attribute][1]):
            if not (i in values_df.columns):
                values_df[i] = 0
        values_df = values_df[range(
            num_values_dict[attribute][0], num_values_dict[attribute][1])]

        if with_columns is None:
            with_columns = [
                attribute + "_" + str(column_name) for column_name in values_df.columns
            ]

        raise_if_not(
            len(with_columns) == len(values_df.columns),
            "For the given case with `one_hot=True`,`with_columns` must be a list of strings of length "
            f"{values_df.columns}.",
            logger=logger,
        )

        values_df.columns = with_columns
    else:
        if cyclic:
            if attribute == "day":
                periods = time_index.days_in_month.values
                freq = 2 * np.pi * np.reciprocal(periods.astype(dtype))
            else:
                period = num_values_dict[attribute][1] - \
                    num_values_dict[attribute][0]
                freq = 2 * np.pi / period

            if with_columns is None:
                with_columns = [attribute + "_sin", attribute + "_cos"]

            raise_if(
                len(with_columns) != 2,
                "`with_columns` must be a list of two strings when `cyclic=True`. "
                "The first string for the sine component name, the second for the cosine component name.",
                logger=logger,
            )
            values_df = pd.DataFrame(
                {
                    with_columns[0]: np.sin(freq * values),
                    with_columns[1]: np.cos(freq * values),
                }
            )
        else:
            if with_columns is None:
                with_columns = attribute
            raise_if_not(
                isinstance(with_columns, str),
                "`with_columns` must be a string specifying the output component name.",
                logger=logger,
            )
            values_df = pd.DataFrame({with_columns: values})

    values_df.index = time_index_ts
    return TimeSeries.from_dataframe(values_df).astype(dtype)
```
 - Python version: 3.8.13
 - darts version 0.27.1

madtoinou:
Fixes #2219.

### Summary

- `datetime_attribute_timeseries()` was not taking into account the indexing of the attribute (1 for month, day or week for example), resulting in a shift of the "zeroed" columns
- 0-indexing is enforced for all the encodings (1-indexed attributes are shifted by one)

### Other Information

- added some tests


darts/tests/utils/test_timeseries_generation.py
```diff
@@ -368,69 +370,133 @@ def helper_routine(idx, attr, vals_exp, **kwargs):
...
-        vals = [1] + [0] * 11
-        vals = [vals for _ in range(48)]
-        helper_routine(idx, "month", vals_exp=vals, tz="CET", one_hot=True)
-
-        # ===> sine/cosine cyclic encoding
-        # hour (period = 24 hours in one day)
-        period = 24
         freq = 2 * np.pi / period
-        vals_dta = [i for i in range(24)] * 2
+        vals_dta = [i for i in range(shift, period + shift)] * 2
```
dennisbader, line 486: Actually, thinking about this: For cyclic encodings, shouldn't we make all freqs start at 0. So that also for example the first month (month=1) start at sin(0), cos(0), same as the first hour (hour=0) starts at sin(0), cos(0)

dennisbader (changes requested):
Very nice @madtoinou thanks a lot 🚀 

Had some minor suggestions. I think we should make the first "step" of each freq start at 0 for cyclic encodings (month=1 should result in sin(0), cos(0), similar to hour=0)

darts/tests/utils/test_timeseries_generation.py
madtoinou: I agree with you, would be better if all the attributes are "synchronized"

dennisbader: 👍 We should mention this is in the changelog, since it'll end up generating different predictions (probably)

darts/tests/utils/test_timeseries_generation.py
```diff
@@ -400,18 +402,52 @@ def test_datetime_attribute_timeseries(self):
...
+        [
+            ("M", "month", 12, "2000-01-01"),
+            ("H", "hour", 24, "2000-01-01 00:00:00"),
+            ("D", "weekday", 7, "2000-01-03"),
+            ("s", "second", 60, "2000-01-01 00:00:00"),
+            ("W", "weekofyear", 52, "2000-01-03"),
+            ("Q", "quarter", 4, "2000-01-01"),
+        ],
+    )
+    def test_datetime_attribute_timeseries_indexing_shift(self, config):
```
dennisbader, line 426: nice :)

dennisbader (approved):
Another great PR, thanks @madtoinou 🚀

darts/tests/utils/test_timeseries_generation.py
```diff
@@ -353,8 +353,10 @@ def helper_routine(idx, attr, vals_exp, **kwargs):
         vals_exp = np.array(vals_exp, dtype=ts.dtype)
         if len(vals_exp.shape) == 1:
             vals_act = ts.values()[:, 0]
+            # vals_act = ts.values()
         else:
             vals_act = ts.values()
+        print(vals_act, "----", vals_exp)
```
VascoSch92: use `logger` instead of `print`

Or it was just for debugging? :-)

PR #2243, opened 2024-02-21
Change default `gridsearch` kwarg value

### LINKED ISSUE #2236: [BUG] - Inconsistent default parameters between backtest and gridsearch
**Describe the bug**
The backtest and gridsearch methods of forecasting models have inconsistent default values to their start (None v.s. 0.5) keyword argument. This makes the scores returned by gridsearch inconsistent with the scores manually computed through backtest with the same model parameters 

**Expected behavior**
Set both to None or both to 0.5

**System
 - darts version: 0.27.0

I can make this PR

thomktz:
Fixes #2236

### Summary

Parameter default value change in `gridsearch`. I picked `None` as the default value since it seems to be the newer feature.


dennisbader (approved):
Looks great, thanks for this contribution @thomktz 🚀

PR #2267, opened 2024-03-05
Add optional inverse transform in historical forecast

### LINKED ISSUE #2260: Inverse scaling for the list of series (different lengths handling)
Currently the Scaler object trained on the list of the TimeSeries doesn't support inverse scaling for the list of lists of TimeSeries with the structure returned by the `historical_forecasts` method (in case of having the list of series as an input).

Let's assume that the Scaler is trained on the list of TimeSeries `[A, B, C]`. 
If the `historical_forecasts` is evaluated on the same list of TimeSeries, the example **output** could be:
`[[A_F1, A_F2, A_F3], [B_F1, B_F2], [C_F1, C_F2, C_F3]]` (here the TimeSeries `B` is an example of TimeSeries that is to short to produce 3 forecasts).
The Scaler trained on the mentioned above list of TimeSeries, can accept only triples of TimeSeries like `[A_FX, B_FX, C_FX]`. 

With this in mind, the following solution can be proposed: transformation of the **output** to the list of lists acceptable by the scaler, and iterating over them.

The potential alternative is that the user can write this function outside library, and transform the list of lists TimeSeries to the form acceptable by the Scaler.

alicjakrzeminska:
Edit:

Fixes #2260.

### Summary
- `InvertibleDataTransformer` now supports parallelized inverse transformation for `series` being a list of lists of `TimeSeries` (`Sequence[Sequence[TimeSeries]]`). This `series` type represents for example the output from `historical_forecasts()` when using multiple series.


darts/models/forecasting/forecasting_model.py
```diff
@@ -2106,6 +2115,57 @@ def _optimized_historical_forecasts(
...
+            The scaler trained on the same set of series that was used to generate forecasts.
+
+        Returns
+        -------
+        Union[List[TimeSeries], List[List[TimeSeries]]]
+            A list (of lists) of inversely transformed forecasts.
+
+        """
+
+        if not forecasts_list_transformed:
```
VascoSch92, line 2140: i don't get this point. If `forecasts_list_transformed` is empty you are doing nothing right?

Can you check that before calling the method? 

The method is private, so it should be used within this class.

Make it sense? :-)

darts/models/forecasting/forecasting_model.py
```diff
@@ -2106,6 +2115,57 @@ def _optimized_historical_forecasts(
...
+            return forecasts_list_transformed
+
+        if isinstance(forecasts_list_transformed[0], TimeSeries):
+            forecasts_list_inversely_transformed = scaler.inverse_transform(
```
VascoSch92, line 2144:
```suggestion
            return scaler.inverse_transform(
```
in this way you don't have to go down in the method to check what is returned in this case :-)

darts/models/forecasting/forecasting_model.py
```diff
@@ -2106,6 +2115,57 @@ def _optimized_historical_forecasts(
...
+                forecasts_list_transformed
+            )
+        else:
+            fill_value = TimeSeries.from_values(np.empty(1))
+
+            forecasts_list_inversely_transformed = list(
```
VascoSch92, line 2150: at this stage here the list is not inversely transformed. Perhaps another name for this variable.

darts/models/forecasting/forecasting_model.py
```diff
@@ -2106,6 +2115,57 @@ def _optimized_historical_forecasts(
         )
         return []
 
+    @staticmethod
+    def _inverse_transform_forecasts_list(
+        forecasts_list_transformed: Union[List[TimeSeries], List[List[TimeSeries]]],
+        scaler: Scaler,
+    ) -> Union[List[TimeSeries], List[List[TimeSeries]]]:
+        """Applies inverse transform on the list (of lists) of forecasts.
+        Uses the scaler trained previously on the same set of series.
```
VascoSch92, line 2124:
```suggestion
        """Applies inverse transform on the list (of lists) of forecasts,
        using the scaler trained previously on the same set of series.
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -737,6 +739,8 @@ def historical_forecasts(
             Whether to retain only the last point of each historical forecast.
             If set to True, the method returns a single ``TimeSeries`` containing the successive point forecasts.
             Otherwise, returns a list of historical ``TimeSeries`` forecasts.
+        scaler
+            Optionally, the scaler trained on the same `series`, performs inverse transform on the generated forecasts.
```
VascoSch92, line 743:
```suggestion
        scaler
            Optionally, the scaler trained on the same `series` is used to perform an inverse transform on the generated forecasts.
```

dennisbader (changes requested):
Thanks for this Draft PR @alicjakrzeminska. I would suggest to move the logic to `darts.dataprocessing.transformers.invertible_data_transformer.InvertibleDataTrasnformer`. In my opinion this can generally be applied to any `InvertibleDataTransfomer`. What do you think?

darts/dataprocessing/transformers/invertible_data_transformer.py
```diff
@@ -354,12 +354,18 @@ def inverse_transform_list(
...
-        Accepts output from the historical forecasting.
+        Assuming that the scaler was trained on series [A, B, C], this function accepts
+        input in the following form:
 
-        Allows for missing series in the end of the list,
-        e.g. if the scaler was trained on the list [A, B, C],
-        then this function can inversely transform the list [AX, BX],
-        where AX - series transformed by the scaler trained on series A.
+        * [A1, B1, C1] - list of series,
+        * [[A1, A2, A3], [B1, B2], [C1, C2, C3]] - list of lists of series,
```
VascoSch92, line 361:
```suggestion
        -  [A1, B1, C1] - list of series,
        -  [[A1, A2, A3], [B1, B2], [C1, C2, C3]] - list of lists of series,
```

darts/dataprocessing/transformers/invertible_data_transformer.py
```diff
@@ -354,12 +354,18 @@ def inverse_transform_list(
...
+
+        where AX - the timeseries transformed according to the parameters obtained with the training on series A.
+
+        Allows for different lengths of the lists.
+        Allows for missing series the list.
```
VascoSch92, line 366:
```suggestion
        Supports lists of varying lengths and missing elements within the list/s.
```

alicjakrzeminska:
@dennisbader, I moved the function to `InvertibleDataTransfomer `. Please, take a look, and if you find it ok, I'll add some tests.
@VascoSch92, thank you for your feedback, I applied your suggestions. :)

alicjakrzeminska:
@dennisbader, I integrated the functionality with `inverse_transform` method of `InvertibleDataTransfomer ` as you suggested. Please, have a look, and I'll write the docstrings and tests if it's ok for you.

darts/dataprocessing/transformers/invertible_data_transformer.py
```diff
@@ -306,15 +312,30 @@ def inverse_transform(
...
+            transformer_selector = range(len(series))
+            called_with_sequence_series = True
+        elif isinstance(series[0], list):
+            input_series = []
+            data = []
+            transformer_selector = []
+            for idx, series_list in enumerate(series):
+                for series_ in series_list:
+                    input_series.append(series_)
+                    data.append(series_)
```
VascoSch92, line 334:
```suggestion
                input_series.extend(series_list)
                data.extend(series_list
```

dennisbader (approved):
Really nice, thanks a lot for this great PR @alicjakrzeminska 🚀

Issue #2259, opened 2024-03-01
Add a `model.supports_sampled_predictions` attribute


felixdivo:
**Is your feature request related to a current problem? Please describe.**
It would be great to get the properties listed in [the nice table in the README](https://github.com/unit8co/darts?tab=readme-ov-file#forecasting-models) also from code. This helps with automated evaluation of multiple models. Additionally/Alternatively, having a `model.supports_probabilisitc_predictions` would be nice, where it would be equal to `model.supports_sampled_predictions or model.supports_likelihood_parameter_prediction`.

**Describe proposed solution**
I'd like to be able to do something along these lines
```python
my_model = ModelClass(some="properties")
print(f"can sample forecasts: {my_model.supports_sampled_predictions}")
print(f"can estimate distribution parameters: {my_model.supports_likelihood_parameter_prediction}")
print(f"can make probabilistic forecasts: {my_model.supports_probabilisitc_predictions}")
```

**Describe potential alternatives**
You make a list of the models that can do this by yourself. It works, but is annoying.

Also, one might discuss the distinction between `supports` (the implementation can do it, in principle) and `uses` (the parameters that are passed enabled it).

**Additional context**
Suggestions for other namings of these attributes are more than welcome. :)

dennisbader:
This is a good point @felixdivo . Currently we have this as a private model method `ForecastingModel._is_probabilistic()`.

I agree that we should make this public and similar to the other support properties.

Here are some thoughts/points to keep in mind:
- Currently, all models that support probabilistic predictions, support the `sampled` predictions. 
- Not all of these models support the `likelihood_parameter` predictions
- Global models are either probabilistic (when they use a `likelihood`) or deterministic (when they don't)
- On the other hand, the local models are deterministic but they can also support probabilistic predictions (ususally some kind of mc simulation).

Maybe converting the private method `ForecastingModel._is_probabilistic()` into a public property `ForecastingModel.supports_probabilistic_prediction` would do the trick?

felixdivo:
Sounds reasonable! Would you like me to propose a PR for it?

dennisbader:
Yes, go ahead @felixdivo, thanks!

madtoinou:
Fixed by #2269.

PR #2269, opened 2024-03-05
Add supports probabilistic prediction

felixdivo:
Implements #2259

### Summary

Renames the private property `ForecastingModel._is_probabilistic` to a public (still non-abstract) property `supports_probabilistic_prediction`.


dennisbader (commented):
Looks great @felixdivo, thanks a lot  🚀 Let's add an entry to the changelog and it's ready to be merged (under unreleased, since we released 0.28.0 today)

felixdivo:
@dennisbader Done

PR #2272, opened 2024-03-06
Fix/comp lags feat order

madtoinou:
### Summary

Order of the extracted features are not consistent when lags were defined component-wise compared to the shared lags approach.
- The lagged features are now reordered after extraction (sorted by lags, then by components) when defined component-wise to have a consistent order
- Lagged features names created was updated accordingly

### Other Information

- Should fix the problem with non-passing tests on some platform, due to the different ordering of the features
- Parametrized the test for `create_lagged_component_names()`
- Added missing tests to ensure lags passed as list and dict give the same result when the lags are identical across lags
- Added a test to make sure the extracted features are indeed in the desired order when the lags are passed as dict
- Added test to check the order of lagged feature names


darts/tests/utils/tabularization/test_create_lagged_training_data.py
```diff
@@ -17,6 +17,25 @@
...
+    n_components: int, components_names: Sequence[str] = None, **kwargs
+) -> TimeSeries:
+    """
+    Helper function that creates a `linear_timeseries` with a specified number of
+    components. To help distinguish each component from one another, `i` is added on
+    to each value of the `i`th component. Any additional keyword arguments are passed
+    to `linear_timeseries` (`start_value`, `end_value`, `start`, `end`, `length`, etc).
+    """
+    timeseries = []
+    if components_names is None or len(components_names) < n_components:
```
VascoSch92, line 30: what if `len(components_names) > n_components`?

it seems that just the first `n_components` are considered. Is this the behaviour expected? Or an error should be raised?

darts/tests/utils/tabularization/test_create_lagged_training_data.py
```diff
@@ -17,6 +17,25 @@
...
+def helper_create_multivariate_linear_timeseries(
+    n_components: int, components_names: Sequence[str] = None, **kwargs
+) -> TimeSeries:
+    """
+    Helper function that creates a `linear_timeseries` with a specified number of
+    components. To help distinguish each component from one another, `i` is added on
+    to each value of the `i`th component. Any additional keyword arguments are passed
+    to `linear_timeseries` (`start_value`, `end_value`, `start`, `end`, `length`, etc).
+    """
+    timeseries = []
```
VascoSch92, line 29: nit: deplace that on the top of the for-loop as it is used there.

VascoSch92 (commented):
Hey 👋 
sorry I saw just later that you were just moving the `create_multivariate_linear_timeseries` method in the other script.

I left some comment about this method.

Feel free to ignore if they don't have to do with the PR :-)

darts/tests/utils/tabularization/test_create_lagged_training_data.py
madtoinou: This is an helper function for the tests, if `components_names > n_components`, the last values will just be ignored.

darts/models/forecasting/regression_model.py
```diff
@@ -1149,6 +1135,44 @@ def _predict_and_sample(
...
+        t_pred: int,
+    ) -> np.ndarray:
+        """Extract, concatenate and reorder component-wise lags to obtain a features order
+        identical to tabularization.
+        """
+        # prepare index to reorder features by lags across components
+        comp_lags_reordered = np.concatenate(
+            [
+                np.array(comp_lags, dtype=int)
+                for comp_lags in self.component_lags[cov_type].values()
```
dennisbader, line 1153:
```suggestion
                comp_lags
                for comp_lags in self.component_lags[cov_type].values()
```

darts/models/forecasting/regression_model.py
```diff
@@ -1149,6 +1135,44 @@ def _predict_and_sample(
...
+        if cov_type == "target":
+            lags_shift = -shift - last_step_shift
+        else:
+            lags_shift = -self.lags[cov_type][0] + t_pred
+
+        # extract features
+        tmp_X = [
+            values_matrix[
+                :,
+                np.array(comp_lags) + lags_shift,
```
dennisbader, line 1167: Nitpicking: for such small arrays it will probably be more efficient with a list comprehension

darts/tests/utils/tabularization/test_create_lagged_training_data.py
```diff
@@ -550,15 +620,44 @@ def test_lagged_training_data_equal_freq(self, series_type: str):
...
-            assert y.shape[0] == len(times[0])
+            assert X.shape[0] == len(feats_times) == len(times[0])
+            assert y.shape[0] == len(feats_times) == len(times[0])
             # Check that outputs match:
             assert np.allclose(expected_X, X[:, :, 0])
             assert np.allclose(expected_y, y[:, :, 0])
             assert feats_times.equals(times[0])
 
+            # passing lags a dictionary
+            lags_as_dict = self.convert_lags_to_dict(
```
dennisbader, line 631: It looks like in most of the tests, this code is duplicated from the non-dict lags case. We could add the logic as local function inside each test (including lagged feature extraction and checks), and then just that call it once with non-dict lags and once with dict lags.

darts/tests/utils/tabularization/test_create_lagged_training_data.py
```diff
@@ -1304,6 +1546,167 @@ def test_lagged_training_data_positive_lags(self, config):
...
+    @pytest.mark.parametrize(
+        "config",
+        itertools.product(
+            [0, 1, 3],
+            [1, 2],
+            [True, False],
+            ["datetime", "integer"],
+        ),
+    )
+    def test_lagged_training_data_comp_wise_lags(self, config):
```
dennisbader, line 1583: nice 👍

dennisbader (changes requested):
Very nice @madtoinou, thanks for fixing this, and adding the missing tests 👍 

I had some suggestions, mainly:
- Add changelog entry
- moving the auto regressive logic into tabularization, and handle the entire X creation in there
- slight refactoring of the tests to avoid duplicate code

darts/utils/data/tabularization.py
```diff
@@ -775,10 +784,31 @@ def create_lagged_component_names(
 
         components = get_single_series(variate).components.tolist()
         if isinstance(variate_lags, dict):
+            if "default_lags" in variate_lags:
```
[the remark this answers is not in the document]
madtoinou: As the `RegressionModel` already handles it, I wanted to avoid code duplication (and circular imports) while preventing unexpected behavior if a user calls this method with parameters not directly extracted from a model.

darts/models/forecasting/regression_model.py
```diff
@@ -1149,6 +1135,44 @@ def _predict_and_sample(
...
+    def _extract_component_lags_autoregression(
+        self,
+        cov_type: str,
+        values_matrix: np.ndarray,
+        shift: int,
+        last_step_shift: int,
+        t_pred: int,
+    ) -> np.ndarray:
+        """Extract, concatenate and reorder component-wise lags to obtain a features order
+        identical to tabularization.
```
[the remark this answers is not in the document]
madtoinou: I put it there because the corresponding logic was previously contained in the `predict()` method but just moved it :)

darts/models/forecasting/regression_model.py
```diff
@@ -1032,24 +1032,16 @@ def predict(
                     )
                 # component-wise lags
                 if "target" in self.component_lags:
```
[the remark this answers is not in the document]
madtoinou: I agree. Would you prefer this logic to be in tabularization or regression_model?

darts/tests/utils/tabularization/test_create_lagged_training_data.py
madtoinou: Done, let me know what you think

dennisbader (approved):
Looks great now, thanks a lot for fixing this @madtoinou 🚀

PR #2293, opened 2024-03-21
Implement TSMixer Model

cristof-r:
Add TSMixer model #1807 with several unit tests.
Adopted PyTorch implementation from this repository: https://github.com/ditschuk/pytorch-tsmixer/
The paper can be found here: https://arxiv.org/pdf/2303.06053.pdf


cristof-r:
Any feedback is very welcome, for me it seems good so far.
I can add it to the different markdowns and make an example notebook if you like the implementation.

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+
+logger = get_logger(__name__)
+
+
+def time_to_feature(x: torch.Tensor) -> torch.Tensor:
+    """Converts a time series torch.Tensor to a feature torch.Tensor."""
+    return x.permute(0, 2, 1)
+
+
+feature_to_time = time_to_feature
```
VascoSch92, line 56: why are you doing that?

I have the following questions:

1. why are you not using the method directly?
2. it is confusing not? one is `time_to_feature` and the other `feature_to_time`. They are the opposite, but they are equal
3. Can you not use the name: `time_series_to_feature`, and use this method everywhere?

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+
+
+class TimeBatchNorm2d(nn.BatchNorm1d):
+    def __init__(self, normalized_shape: tuple):
```
VascoSch92, line 60:
```suggestion
    def __init__(self, normalized_shape: Tuple[int, int]):
```

and you have to `from typing import Tuple`

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+        This class extends nn.BatchNorm1d to apply batch normalization across time and
+        feature dimensions.
+
+        Parameters
+        ----------
+        normalized_shape: tuple
+            A tuple (num_time_steps, num_channels) representing the shape of the time and feature
+            dimensions to normalize.
+        """
+        num_time_steps, num_channels = normalized_shape
```
VascoSch92, line 74:
```suggestion
        self.num_time_steps, self.num_channels = normalized_shape
```

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+        super().__init__(reduce(operator.mul, normalized_shape))
+        self.num_time_steps = num_time_steps
+        self.num_channels = num_channels
```
VascoSch92, line 77:
```suggestion
```

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+            normalize_before=normalize_before,
+            norm_type=norm_type,
+        )
+
+    def forward(self, x: torch.Tensor, x_static: torch.Tensor | None) -> torch.Tensor:
+        if self.fr_static is None:
+            return self.fm(x)
+        v = self.fr_static(x_static)
+        repeat_factor = x.size(1) // v.size(1)
+        v = v.repeat(1, repeat_factor, 1)
```
VascoSch92, line 234:
```suggestion
        v = v.repeat(1,  x.size(1) // v.size(1), 1)
```
But this is just personal :-) in this way you see directly how v is transformed in the `forward`

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+            The type of normalization to use.
+        """
+        super().__init__()
+        self.norm = norm_type((sequence_length, input_dim))
+        self.activation = activation
+        self.dropout = MonteCarloDropout(dropout)
+        self.fc1 = nn.Linear(sequence_length, sequence_length)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x_temp = feature_to_time(x)
```
VascoSch92, line 275: why x_temp and not just x?

darts/models/forecasting/tsmixer_model.py
```diff
@@ -0,0 +1,943 @@
...
+        self.fc_out = nn.Linear(
+            hidden_output_size,
+            self.output_dim * self.nr_params,
+        )
+
+    def _build_mixer(
+        self, blocks: int, hidden_size: int, prediction_length: int, **kwargs
+    ):
+        """Build the mixer blocks for the model."""
+        # in case there a no future covariates, the feature_mixing_future layer
```
VascoSch92, line 535:
```suggestion
        # in case there are no future covariates, the feature_mixing_future layer
```

darts/tests/models/forecasting/test_tsmixer.py
```diff
@@ -0,0 +1,601 @@
...
+    TORCH_AVAILABLE = True
+
+except ImportError:
+    logger.warning("Torch not available. TSMixerModel tests will be skipped.")
+    TORCH_AVAILABLE = False
+
+
+if TORCH_AVAILABLE:
+
+    class TestTSMixerModel:
```
VascoSch92, line 28:
```suggestion
    @pytest.mark.skipif(TORCH_AVAIBLE is False, reason="Torch not available. TSMixerModel tests will be skipped.")
    class TestTSMixerModel:
```

You can use this decorator and the code becomes more readable :-) 


and you can delete the ` if TORCH_AVAILABLE:` above

darts/tests/models/forecasting/test_tsmixer.py
```diff
@@ -0,0 +1,601 @@
...
+                assert len(y_hat_i) == predict_n
+                assert y_hat_i.n_components == ts_i.n_components
+
+        def test_prediction_shape(self):
+            """checks whether prediction has same number of variable as input series and
+            whether prediction has correct length.
+            Test cases:
+                -   univariate
+                -   multivariate
+                -   multi-TS
```
VascoSch92, line 451: you can parametric these test cases using @pytest.mark.parametrize. In this way less code, same tests and it is more readable

VascoSch92 (commented):
Whoa... nice job

darts/models/forecasting/tsmixer_model.py
cristof-r: Yes, they both do the same thing.
The reason was to make it more readable in this case:

```
def forward(self, x: torch.Tensor) -> torch.Tensor:
    x_temp = feature_to_time(x) # permute the feature dim with the time dim
    x_temp = self.activation(self.fc1(x_temp))
    x_temp = self.dropout(x_temp)
    x_res = time_to_feature(x_temp) # permute back the time dim with the feature dim
    x_temp = self.norm(x + x_res)
    return x_temp
 ```
    
So a user who looks at the forward pass can better understand what is going on.
But it may be confusing to have the same function twice.

Maybe removing the second function and adding comments where necessary to the forward pass could be a good middle ground. 
I will do that.

File: darts/models/forecasting/tsmixer_model.py (hunk 2 above)
Comment (cristof-r) [reply darts/models/forecasting/tsmixer_model.py:60]: You are right, I probably should also exchange the "|" operator with Unions.
Forgot about that.

File: darts/models/forecasting/tsmixer_model.py (hunk 6 above)
Comment (cristof-r) [reply darts/models/forecasting/tsmixer_model.py:272]: Because it needs x later again and can not overwrite it.

```
    def forward(self, x: torch.Tensor) -> torch.Tensor:
        x_temp = feature_to_time(x)
        x_temp = self.activation(self.fc1(x_temp))
        x_temp = self.dropout(x_temp)
        x_res = time_to_feature(x_temp)
        x_temp = self.norm(x + x_res) # <--- here
        return x_temp
   ```

darts/models/forecasting/tsmixer_model.py
cristof-r: Updated

darts/tests/models/forecasting/test_tsmixer.py
cristof-r: Thanks, it's refactored.

dennisbader:
Hi @cristof-r, wow, this indeed looks amazing from a first glance!
Thanks a lot for this great PR, just give us some time to review it :) 🚀 🚀 🚀

leoniewgnr:
If there is anything I can help with, please let me know! Looking forward to this:)

dennisbader:
@cristof-r and @leoniewgnr, I'm currently reviewing the PR. There were a couple of things to change, so I started working on a new branch with a couple of adaptions to this PR. I'll soon open a PR to merge into this one.

Among other things it will improve the performance and reduce training time drastically.

While working on it I also made a little notebook for testing with the ETTh1Dataset. The model works pretty well, very close to TiDEModel.

I'll keep you updated.

cristof-r:
@dennisbader sounds great! Thank you very much

dennisbader (approved):
Thanks for this really nice PR @cristof-r. I took the freedom to push the changes already into this branch. I will comment on the most important changes:

**tests**
- removed some of the tests that were either already handled in other test files, or tests that were taking a "long" time to complete (e.g. model performance (accuracy) tests which took ~30 seconds to complete)


**model implementation**
- norms:
  - adapted `TimeBatchNorm2d` implementation to use actual 2d batch norm
  - removed support for RINorm since it should be used with model parameter `use_reversible_instance_norm`
- modules 
  - made all module classes private to hide them in the rendered documentation
  - removed `ConditionalFeatureMixing` and instead added the logic to `_ConditionalMixerLayer`
- model parameters
  - lowered the default parameters values (e.g. blocks, hidden_size, ...) to make a lighter default version
- main things that were fixed:
  -  before, `output_dim` for all modules was set to `hidden_output_size=hidden_size * output_dim * nr_params`, whereas it should just be `hdden_size`. This was why the model was getting really slow for multivariate target series or probabilistic models.
  - multi-component static covariates were not properly handled. We need to flatten the static covariates, and have `static_cov_dim=n components * n static features`
  - I believe the static mixing was not handled correctly before. It looks like it was done as described in the paper. However, I'm not sure if the paper described it correctly. 
    - before, static covariates were project to `hidden_size` with a linear layer, and then concatenated with `x`. In the first block `x` has only the actual number of input features.  So then for the concatenation, `x` has much lower dimensionalty than `x_static`.
    - Now at the beginning we apply feature mixing to historical, future, and static covariates separately, and then feed them together to the mixing layers.

**model example notebook**
- added an example notebook comparing a probabilistic TSMixer with TiDEmodel on a multivariate dataset, including future covariates (encoders) and static information

Let me know if you agree with the changes :) And again, thanks a lot for this great contribution, really appreciated!

cristof-r:
@dennisbader thank you very much for the improvements, I learned a lot. 
Also thank you very much for the darts library in general, it is really great. 
Do you already know when the next release will be?

dennisbader:
@cristof-r, we're aiming to release within the next two weeks.

PR #2295, opened 2024-03-27
add `DataLoader` related parameters to `fit()` and `predict()`

BohdanBilonoh:

### Summary

Add `torch.utils.data.DataLoader` related parameters to `fit()` and `predict()` of `TorchForecastingModel`

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -992,16 +992,18 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
                 f" provided input/output dimensions = {sample_shape(train_sample)}",
             )
 
-        if dataloader_kwargs is None:
-            dataloader_kwargs = {}
-
-        dataloader_kwargs["shuffle"] = True
-        dataloader_kwargs["batch_size"] = self.batch_size
-        dataloader_kwargs["drop_last"] = False
-        dataloader_kwargs["collate_fn"] = self._batch_collate_fn
-
         # Setting drop_last to False makes the model see each sample at least once, and guarantee the presence of at
         # least one batch no matter the chosen batch size
+        dataloader_kwargs = {
+            **{
+                "shuffle": True,
+                "batch_size": self.batch_size,
+                "drop_last": False,
+                "collate_fn": self._batch_collate_fn,
+            },
+            **(dataloader_kwargs or {}),
+        }
+
         train_loader = DataLoader(
             train_dataset,
             **dataloader_kwargs,
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -1218,7 +1218,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         n_jobs: int = 1,
         roll_size: Optional[int] = None,
         num_samples: int = 1,
-        data_loader_kwargs: Optional[Dict[str, Any]] = None,
+        dataloader_kwargs: Optional[Dict[str, Any]] = None,
         mc_dropout: bool = False,
         predict_likelihood_parameters: bool = False,
         show_warnings: bool = True,
@@ -1280,7 +1280,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         num_samples
             Number of times a prediction is sampled from a probabilistic model. Should be left set to 1
             for deterministic models.
-        data_loader_kwargs
+        dataloader_kwargs
             Optionally, a dictionary of keyword arguments to pass to the PyTorch DataLoader instances used to load the
             training and validation datasets. For more information on DataLoader, check out `this link
             <https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader>`_.
@@ -1367,7 +1367,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
             n_jobs=n_jobs,
             roll_size=roll_size,
             num_samples=num_samples,
-            data_loader_kwargs=data_loader_kwargs,
+            dataloader_kwargs=dataloader_kwargs,
             mc_dropout=mc_dropout,
             predict_likelihood_parameters=predict_likelihood_parameters,
         )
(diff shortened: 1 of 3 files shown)


madtoinou:
Hi @BohdanBilonoh,

It looks great, however to make it easier to maintain and more exhaustive, I think that it would be great to just add an argument called `dataloader_kwargs`, then check that the argument explicitly used by Darts are not redundant/overwritten and then pass this argument down to the `DataLoader` constructor.

It will allow users to specify more than just `prefetch_factor`, `persistent_workers` and `pin_memory`, while limiting copy-pasting from other library documentation (putting a link to the `torch.DataLoader` page does sound like a good idea for this argument however)

joshua-xia:
Hi @BohdanBilonoh, Would you please add multiprocessing_context parameter for Dataloader, it is useful when we use multi-workers for dataloader, Thanks!

joshua-xia:
> Hi @BohdanBilonoh, Would you please add multiprocessing_context parameter for Dataloader, it is useful when we use multi-workers for dataloader, Thanks!

@BohdanBilonoh refer to #2375

joshua-xia:
> > Hi @BohdanBilonoh, Would you please add multiprocessing_context parameter for Dataloader, it is useful when we use multi-workers for dataloader, Thanks!
> 
> @BohdanBilonoh refer to #2375

@BohdanBilonoh  My bad, it is good idea from @madtoinou to add dataloader_kwargs to let user input dataloader parameters as wish freely, not need to support multiprocessing_context parameter forcibly

BohdanBilonoh:
@madtoinou what do you think about hardcoded parameters like

```python
batch_size=self.batch_size,
shuffle=True,
drop_last=False,
collate_fn=self._batch_collate_fn,
```

should it be hard coded for new `dataloader_kwargs` like

```python
def _setup_for_train(
    self,
    train_dataset: TrainingDataset,
    val_dataset: Optional[TrainingDataset] = None,
    trainer: Optional[pl.Trainer] = None,
    verbose: Optional[bool] = None,
    epochs: int = 0,
    dataloader_kwargs: Optional[Dict[str, Any]] = None,
) -> Tuple[pl.Trainer, PLForecastingModule, DataLoader, Optional[DataLoader]]:

        ...

        if dataloader_kwargs is None:
            dataloader_kwargs = {}

        dataloader_kwargs["shuffle"] = True
        dataloader_kwargs["batch_size"] = self.batch_size
        dataloader_kwargs["drop_last"] = False
        dataloader_kwargs["collate_fn"] = self._batch_collate_fn

        # Setting drop_last to False makes the model see each sample at least once, and guarantee the presence of at
        # least one batch no matter the chosen batch size
        train_loader = DataLoader(
            train_dataset,
            **dataloader_kwargs,
        )

        dataloader_kwargs["shuffle"] = False

        # Prepare validation data
        val_loader = (
            None
            if val_dataset is None
            else DataLoader(
                val_dataset,
                **dataloader_kwargs,
            )
        )
```

or give a user full control on `dataloader_kwargs`?

tRosenflanz:
> @madtoinou what do you think about hardcoded parameters like
> 
> ```python
> batch_size=self.batch_size,
> shuffle=True,
> drop_last=False,
> collate_fn=self._batch_collate_fn,
> ```
> 
> should it be hard coded for new `dataloader_kwargs` like
> 
> ```python
> def _setup_for_train(
>     self,
>     train_dataset: TrainingDataset,
>     val_dataset: Optional[TrainingDataset] = None,
>     trainer: Optional[pl.Trainer] = None,
>     verbose: Optional[bool] = None,
>     epochs: int = 0,
>     dataloader_kwargs: Optional[Dict[str, Any]] = None,
> ) -> Tuple[pl.Trainer, PLForecastingModule, DataLoader, Optional[DataLoader]]:
> 
>         ...
> 
>         if dataloader_kwargs is None:
>             dataloader_kwargs = {}
> 
>         dataloader_kwargs["shuffle"] = True
>         dataloader_kwargs["batch_size"] = self.batch_size
>         dataloader_kwargs["drop_last"] = False
>         dataloader_kwargs["collate_fn"] = self._batch_collate_fn
> 
>         # Setting drop_last to False makes the model see each sample at least once, and guarantee the presence of at
>         # least one batch no matter the chosen batch size
>         train_loader = DataLoader(
>             train_dataset,
>             **dataloader_kwargs,
>         )
> 
>         dataloader_kwargs["shuffle"] = False
> 
>         # Prepare validation data
>         val_loader = (
>             None
>             if val_dataset is None
>             else DataLoader(
>                 val_dataset,
>                 **dataloader_kwargs,
>             )
>         )
> 
>         ...
> ```
> 
> or give a user full control on `dataloader_kwargs`?

you could extend your suggestion to allow overrides but with populated defaults 
```
defaults = dict(shuffle = True, batch_size = self.batch_size, drop_last = False, collate_fn = self._batch_collate_fn)
#combine with defaults but override them
dataloader_kwargs_train = {**defaults,**dataloader_kwargs}
#override shuffle
dataloader_kwargs_val = (**dataloader_kwargs_train, **dict(shuffle=False)} 
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1487,14 +1484,17 @@ def predict_from_dataset(
             mc_dropout=mc_dropout,
         )
 
+        if data_loader_kwargs is None:
```
dennisbader, line 1487: Shouldn't we allow then here as well the liberty to overwrite these defaults (expect shuffle)?

dennisbader (changes requested):
Thanks for the updates @BohdanBilonoh, I took the chance to fix some indentation issues in one of the docstrings and pushed the changes.

If we give the freedom to overwrite our default dataloader params when calling `fit()`, shouldn't we then also allow that during `predict()`?

Also, removing the `num_loader_workers` parameter is a breaking change. Can you document this in the CHANGELOG.md?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1281,11 +1280,10 @@ def predict(
         num_samples
             Number of times a prediction is sampled from a probabilistic model. Should be left set to 1
             for deterministic models.
-        num_loader_workers
-            Optionally, an integer specifying the ``num_workers`` to use in PyTorch ``DataLoader`` instances,
-            for the inference/prediction dataset loaders (if any).
-            A larger number of workers can sometimes increase performance, but can also incur extra overheads
-            and increase memory usage, as more batches are loaded in parallel.
+        data_loader_kwargs
+            Optionally, a dictionary of keyword arguments to pass to the PyTorch DataLoader instances used to load the
```
dennisbader, line 1284: it's referring to the train and val datasets, but it should be for the prediction dataset, same for the methods below

dennisbader (approved):
Thanks for the changes @BohdanBilonoh. I made some last updates including unit tests and some docstrings.

Looks great now, and ready to be merged once tests have passed!

Thanks a lot for the great work 🚀

madtoinou (approved):

PR #2312, opened 2024-04-10
Fix/mc dropout

### LINKED ISSUE #2300: [BUG] Monte Carlo dropout does not work with pytorch-lightning >=2.2.0
Monte Carlo dropout appears to silently fail with the current release of pytorch-lightning.

**To Reproduce**
See [this colab workbook](https://colab.research.google.com/drive/1C9BX6Ia6UmLExfFYgxS5o4KDJkp6oAit?usp=sharing). `TCNModel` instances with different `dropout` keywords produce identical CV results and inferences if given the same `random_state`, even with `dropout` set to 0.99, which should catastrophically underfit. I was able to reproduce this behaviour on my laptop using [this python environment](https://gist.github.com/eschibli/4c7afad3d9ba1ac28cbffbb29e1c5de2) obtained by running `pip install darts` in a fresh Python 3.11 venv. 

**Expected behavior**
The `dropout` keyword should impact model results. Extremely high dropout should result in severe underfitting.  I was able to obtain this expected behavior in [this environment](https://gist.github.com/eschibli/a75e2b612a9eabc17d2da5ce0b73b594) by downgrading pytorch-lightning to 2.1.2. This can also be observed in [colab](https://colab.research.google.com/drive/1DhMGYY6bZaZINeJ9PK1kgGVeqtR3fDbs?usp=sharing)

dennisbader:

Fixes #2300.

### Summary
- fixes Monte Carlo Dropout for all TorchForecastingModels. MC dropout training mode was not getting activated properly in the different model training/val/pred/.. stages.
- add `MonteCarloDropout` to all TorchForecastingModels that were not using them before.

## Additional Information

Below an example that it works properly now.

```
import numpy as np

from darts import concatenate
from darts.models import TCNModel, TSMixerModel

preds = []

dropouts = [0., 0.5, 0.99]
for dropout in dropouts:
    model = TCNModel(13, 12, dropout=dropout, random_state=42)
    model.fit(series, epochs=100)
    preds.append(
        concatenate(
            model.historical_forecasts(
                series=series,
                retrain=False,
                forecast_horizon=12,
                stride=12,
                last_points_only=False,
            ),
            axis=0
        )
    )

series.plot()
for pred, dropout in zip(preds, dropouts):
    pred.plot(label=f"d={dropout}")
plt.show()
```

image

### DIFF
--- a/darts/models/forecasting/pl_forecasting_module.py
+++ b/darts/models/forecasting/pl_forecasting_module.py
@@ -197,6 +197,7 @@ class PLForecastingModule(pl.LightningModule, ABC):
         self.pred_batch_size: Optional[int] = None
         self.pred_n_jobs: Optional[int] = None
         self.predict_likelihood_parameters: Optional[bool] = None
+        self.pred_mc_dropout: Optional[bool] = None
 
     @property
     def first_prediction_index(self) -> int:
@@ -240,6 +241,14 @@ class PLForecastingModule(pl.LightningModule, ABC):
         )
         self._calculate_metrics(output, target, self.val_metrics)
         return loss
+
+    def on_predict_start(self) -> None:
+        # optionally, activate monte carlo dropout for prediction
+        self.set_mc_dropout(active=self.pred_mc_dropout)
+
+    def on_predict_end(self) -> None:
+        # deactivate, monte carlo dropout for any downstream task
+        self.set_mc_dropout(active=False)
 
     def predict_step(
         self, batch: Tuple, batch_idx: int, dataloader_idx: Optional[int] = None
@@ -339,6 +348,7 @@ class PLForecastingModule(pl.LightningModule, ABC):
         batch_size: int,
         n_jobs: int,
         predict_likelihood_parameters: bool,
+        mc_dropout: bool,
     ) -> None:
         """to be set from TorchForecastingModel before calling trainer.predict() and reset at self.on_predict_end()"""
         self.pred_n = n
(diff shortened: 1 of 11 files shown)


darts/models/forecasting/tcn_model.py
```diff
@@ -29,7 +29,7 @@ def __init__(
         num_filters: int,
         kernel_size: int,
         dilation_base: int,
-        dropout_fn,
+        dropout: float,
```
madtoinou, line 32: This is a breaking change for users that were passing a function to this argument.

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -241,6 +242,14 @@ def validation_step(self, val_batch, batch_idx) -> torch.Tensor:
         self._calculate_metrics(output, target, self.val_metrics)
         return loss
 
+    def on_predict_start(self) -> None:
+        # optionally, activate monte carlo dropout for prediction
+        self.set_mc_dropout(active=self.pred_mc_dropout)
```
madtoinou, line 247: So this overwrite the dropout probability used during the training? So it is possible to have different values during training and inference, is it correct?

darts/models/forecasting/tcn_model.py
dennisbader: This is a hidden module so I wouldn't call it a breaking change.

darts/models/forecasting/pl_forecasting_module.py
dennisbader: No the probability is left unchanged. It just means that during prediction (when generating the samples) it applies dropout when `mc_dropout=True`.

madtoinou (approved):

PR #2323, opened 2024-04-12
lint: switch `flake8` to Ruff

### LINKED ISSUE #2195: switch to Ruff instead of `flake8`
It would address linting speed, as Ruff is much faster, and eventually extensibility, as Ruff can cover multiple linting packages with almost no performance drop...

replace linking with `flake8` preserving the same configuration

**Additional context**

Recently [Ruff](https://github.com/astral-sh/ruff) has become very popular among also mainstream projects for:
- much faster, they say 70x
- integrate several other tools under one roof, such as isort or bandit

Borda:

Fixes #2195

### Summary

Switch lint `flake8` to Ruff with the same basic checks.
As follow-up, we could also merge Black + `pyupgrade` + `isort` (after merging this one I'll prepare separate PRs for each of them)


darts/ad/anomaly_model/filtering_am.py
```diff
@@ -89,7 +89,7 @@ def fit(
         # TODO: add support for covariates (see eg. Kalman Filter)
 
         raise_if_not(
-            type(allow_model_training) is bool,
+            type(allow_model_training) is bool,  # noqa: E721
```
madtoinou, line 92: We are going to take advantage of this PR to actually comply with E721, would you mind replace the `type(...) is ...` with `instance(...)`?

requirements/dev.txt
```diff
@@ -1,7 +1,3 @@
-black[jupyter]==24.3.0
-flake8==7.0.0
-isort==5.13.2
 pre-commit
 pytest-cov
-pyupgrade==v3.15.0
```
madtoinou, line 6: Since this PR just replaces flake8, I think that you need to put back usort, black & pyupgrade in the requirements?

Also, you forgot to add ruff.

build.gradle
```diff
@@ -97,23 +97,9 @@ task pipInstall() {
...
-}
-
-task lint_isort(type: Exec) {
-    dependsOn pip_dev
-    commandLine "isort", "--check", "."
-}
-
-task lint {
-    dependsOn lint_black, lint_flake8, lint_isort
+    commandLine "pre-commit", "run", "--all-files"
```
madtoinou, line 102: Very nice!

madtoinou (changes requested):
Thanks a lot for this contribution, some minor comments (mostly about fixing E721).

darts/ad/anomaly_model/filtering_am.py
Borda: I had it that way, but the problem is - https://stackoverflow.com/a/37888668/4521646
```py
>>> instance(2, bool)
True
```
So many tests that check input fail.

requirements/dev.txt
Borda: I have run pre-commit instead of individual tools, and the packages in pre-commit have their own declaration and versions, so these installed packages are not used once you run pre-commit

darts/ad/anomaly_model/filtering_am.py
Borda: reported in https://github.com/astral-sh/ruff/issues/10948

Borda: @madtoinou let's land it with this noqa and see if ruff will ignore this case in the future?

madtoinou: Sounds good.

However, I think that one of the `type()` line that was changed to `isinstance()` doesn't pass the tests anymore, would you mind reverting the change?

Borda: sure, just curious why `int` has a problem :/

darts/ad/scorers/wasserstein_scorer.py
```diff
@@ -108,7 +108,7 @@ def __init__(
         #     only one sample
         #     - check if there is an equivalent Wasserstein distance for d-D distributions (currently only accepts 1D)
 
-        if type(window) is int:
+        if isinstance(window, int):
```
Borda, line 111:
```suggestion
        if type(window) is int:  # noqa: E721
```

darts/ad/scorers/scorers.py
```diff
@@ -33,7 +33,7 @@ class AnomalyScorer(ABC):
     def __init__(self, univariate_scorer: bool, window: int) -> None:
 
         raise_if_not(
-            type(window) is int,
+            isinstance(window, int),
```
Borda, line 36:
```suggestion
            type(window) is int,  # noqa: E721
```

madtoinou (approved):
LGTM, thanks a lot @Borda for initiating the transition to ruff!

I'll add an entry in the backlog, see if we can find a more elegant solution to these "type checks" to get ride of the `#noqa` in the long run.

PR #2329, opened 2024-04-16
Fix/historical forecasts torch models

dennisbader:

### Summary
- simplifies and improves historical forecast tests for torch models, by reducing input series length and checking for minimum length requirements. Speeds up tests from 2min to 1min 15 seconds
- Improvements to `RNNModel`:
  - 🔴 enforce `training_length>input_chunk_length`
  - the historical forecastable time index is now properly computed for RNNModels
- `ForecastingModel.exteme_lags()` now has an additional return element indicating the `max_target_lag` used during training (specific to `RNNModel`) 
- simplifies and improves historical forecast tests for torch models, by reducing input series length and checking for minimum length requirements. Speeds up tests from 2min to 1min 15 seconds
- fixes a bug when using historical forecasts with untrained torch models and using covariates, where the historical forecastable time index computation did not take the covariates into account

### DIFF
--- a/darts/tests/models/forecasting/test_historical_forecasts.py
+++ b/darts/tests/models/forecasting/test_historical_forecasts.py
@@ -1733,6 +1733,7 @@ class TestHistoricalforecast:
     @pytest.mark.skipif(not TORCH_AVAILABLE, reason="requires torch")
     @pytest.mark.parametrize("model_config", models_torch_cls_kwargs)
     def test_torch_auto_start_with_past_cov(self, model_config):
+        n_fcs = 3
         forecast_hrz = 10
         # Past covariates only
         model_cls, kwargs, bounds, cov_type = model_config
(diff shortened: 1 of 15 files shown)


darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -1752,231 +1700,410 @@ def test_torch_auto_start_with_past_cov(self, model_config):
...
+                pc_longer,
+                pc_longer_start,
+                pc_start_after,
+                pc_end_before,
             ],
             forecast_horizon=forecast_hrz,
             stride=1,
-            retrain=True,
+            retrain=retrain,
             overlap_end=False,
```
madtoinou, line 1794: discrepancy between the comment and the actual value of `overlap_end`

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -1752,231 +1700,410 @@ def test_torch_auto_start_with_past_cov(self, model_config):
...
+            future_covariates=[
+                fc_longer,
+                fc_longer_start,
+                fc_start_after,
+                fc_end_before,
+            ],
+            forecast_horizon=forecast_hrz,
+            stride=1,
+            retrain=retrain,
+            overlap_end=False,
```
madtoinou, line 1931: same as previous comment, discrepancy between the value and comment.

darts/tests/models/forecasting/test_historical_forecasts.py
```diff
@@ -1752,231 +1700,410 @@ def test_torch_auto_start_with_past_cov(self, model_config):
...
```
madtoinou, line 2097: Same comment as previously.

madtoinou (approved):
Good job spotting and fixing the bug, just noticed some discrepancies in the tests between the comment and the value `overlap_end`.

I hope that we will find a more elegant solution than adding a 8th output to `extreme_lags` but this should do for now.

PR #2340, opened 2024-04-17
lint: replace `pyupgrade` with Ruff's rule UP

Borda:
Checklist before merging this PR:
- [x] Mentioned all issues that this PR fixes or addresses.
- [ ] Summarized the updates of this PR under **Summary**.
- [ ] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md).

Follow-up to #2323

### Summary

`pyupgrade` functionality are fully covered with UP rule

PR #2341, opened 2024-04-18
lint: replace Black with `ruff-format`

Borda:

Follow-up to #2195 and https://github.com/unit8co/darts/pull/2340#pullrequestreview-2008101262

### Summary

Black functionality is fully covered with Ruff formatter


Borda:
@madtoinou @dennisbader, the major change is caused by line length...
Default Black is still something like 88, but you had it before `flake8` to 120 which I kept also for Ruff, but then when formatting updated with ruff, it also uses 120, so shall be change all (lint and formating) to 88 or upgrade to 120 which I feel is better :flamingo:

Borda:
seems there was some connection error fot the one falling check :)

dennisbader:
Hi @Borda, as far as I remember we had 120 for docstrings and 88 for code. 
Can we leave like this to keep the changes minimal? Also for code reviews I still find it useful sometimes to have it in 88. What do you think?

Borda:
> Hi @Borda, as far as I remember we had 120 for docstrings and 88 for code. Can we leave like this to keep the changes minimal?

yes, for sure

> Also for code reviews I still find it useful sometimes to have it in 88. What do you think?

I see here two issues, one technical and one logical:
- *technical* - it seems Ruff has one option for line length, which applies to the codebase, including docstrings. Moreover, there are many strings  (outside docs) in the codebase exceeding 88
- *logical* - docstrings are also part of the codebase and subject of eventual code review, so I do not really understand why they shall be treated differently :/ On the other hand, all docs or MD files could have any length, which is fine

As I said, I would suggest a) extending all to 120 or b) applying 88 to all docstrings.
If none of a or b (which is completely your choice) I will close this as it seems not to be feasible with Ruff formatted :flamingo:

Borda:
@dennisbader resolved and all seems to be passing for me locally

dennisbader:
@Borda, thanks for the changes 🙏 
We're soon to merge the anomaly detection refactor PR (within this week). Let's wait until then to avoid tedious conflicts.

Borda:
> We're soon to merge the anomaly detection refactor PR (within this week). Let's wait until then to avoid tedious conflicts.

@dennisbader sure, just ping me or lets link the PR so can track it when it is done :)

Borda:
@dennisbader It seems there are some collisions in meantime, which I will resolve just after getting some feedback if this is going the right direction :flamingo:

dennisbader:
@Borda. Yes, thanks. We're wrapping up the other PR. Not long anymore. I'll keep you updated

dennisbader:
Hi @Borda, we finally merged the other PR. Could you reapply the formatting based on the current master?
After that we're ready to merge 🚀

dennisbader (approved):
Looks great @Borda, thanks a lot again for another great PR and helping us migrate to Ruff 🚀 
The checks are failing due to Codecov uploads (sometimes it works, other times it doesn't, a bit frustrating) but I'll merge since the tests all passed. 💯

akshobhya95:
@dennisbader ,I was triggering some runs using DARTS forecasting models using ARIMA, Prophet for my test set case . today I was hitting some weird issue not sure if it is due to a side effect of this change, but the runs were not triggering as expected Kindly letme know  when I can trigger the runs using the DARTS model as normal. I wasn't aware of this check-in got to know bit later.

Issue #2384, opened 2024-05-17
[BUG] TSMixer TimeBatchNorm2d using BatchNorm instead of LayerNorm 


tRosenflanz:
**Describe the bug**
TSMixer implementation overall uses `LayerNorm` and doesn't give a `BatchNorm` option. In the paper it was shortly mentioned that the `LayerNorm` gave better results on smaller batches. However, the `TimeBatchNorm2d`  implementation bases `BatchNorm2d` which in my tests has been unstable agreeing with the authors. Inhereting from  `nn.InstanceNorm2d` in the class works out of the box and produces stable training and is a better representation of the paper. 

**To Reproduce**
Train a model with TimeBatchNorm2d normalization with lower batch size on non-stationary targets. Model doesn't train (in my case at least).
Replace the base class with `nn.InstanceNorm2d` - model trains just fine

**Expected behavior**
Model to train at least similarly to LayerNorm option

**System (please complete the following information):**
 - Python version: [e.g. 3.10]
 - darts version [e.g. 0.29.0]

**Additional context**
I think it is best to either add another option using `InstanceNorm2d` or replace the current implementation all together. Since BatchNorm isn't even an option to choose from currently, it is reasonable to choose the latter. 
Note `InstanceNorm` and `LayerNorm` are slightly different but very similar in practical terms compared to `BatchNorm` 
https://discuss.pytorch.org/t/is-there-a-layer-normalization-for-conv2d/7595/6

from the author:
https://github.com/ditschuk/pytorch-tsmixer/issues/2

tRosenflanz:
Attaching some quick comparisons on my datasets. 
Green is LayerNorm, Orange is BatchNorm2d , Grey is InstanceNorm2d. 
<img width="528" alt="Screenshot 2024-05-16 at 5 27 31 PM" src="https://github.com/unit8co/darts/assets/25041325/8d65be06-dcf0-49eb-a761-0881b3d5cec1">
val_loss
<img width="371" alt="Screenshot 2024-05-16 at 5 27 41 PM" src="https://github.com/unit8co/darts/assets/25041325/7f28b6f3-0f79-4635-9531-cff8482a5a1c">

dennisbader:
Hi @tRosenflanz, and thanks for raising this issue and the investigation 🚀 

Quick question:
- are the results you show from a use case where you applied the zeroing of past target and future covariates as mentioned in #2381?
[...]

PR #2386, opened 2024-05-17
TimeInstanceNorm2d normalization for TSMixer

tRosenflanz:
### Summary
Addresses https://github.com/unit8co/darts/issues/2384

Adds a new normalization type for TSMixer TimeInstanceNorm2d

### Other Information

To avoid duplicating code, used an abstract class to define the reshape logic and inherited from specific normalization layers in children classes. Let me know if you want it replaced with simpler code that just copypastes the logic.


tRosenflanz:
Unnecessary. LayerNormNoBias is identical

PR #2391, opened 2024-05-21
rework metrics logic to support states

### LINKED ISSUE #2389: [BUG] PLForecastingModule._calculate_metrics should call log_dict with batch_size
**Describe the bug**
When using additional metrics with PyTorch Lightning forecasting models, due to `PLForecastingModule._calculate_metrics` (called from `train_step()` or `validation_step()`) not calling `self.log_dict(...)` with the `batch_size=` parameter, the Trainer (`_ResultCollection.log()` is complaining about ambiguity (when `show_warnings=True` is enabled on the model):
> /opt/homebrew/Caskroom/miniconda/base/envs/tf/lib/python3.12/site-packages/pytorch_lightning/utilities/data.py:77: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 32. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.

**To Reproduce**
Add a custom metric via `torch_metrics=...` model parameter (e.g. BlockRNNModel), and run .fit().

**Expected behavior**
The warnings should be gone.

 - Python version: 3.12.3
 - darts version: 0.29.0

**Additional context**
suggested code modification (tested):
```python
def _calculate_metrics(self, output, target, metrics):
    if not len(metrics):
        return

    if self.likelihood:
        _metric = metrics(self.likelihood.sample(output), target)
    else:
        # If there's no likelihood, nr_params=1, and we need to squeeze out the
        # last dimension of model output, for properly computing the metric.
        _metric = metrics(output.squeeze(dim=-1), target)

    self.log_dict(
        _metric,
        on_epoch=True,
        on_step=False,
        logger=True,
        prog_bar=True,
        sync_dist=True,
        batch_size=target.shape[0],  # ADD THIS LINE
    )
```

### LINKED ISSUE #2390: [BUG] TorchMetrics loop implementation isn't compatible with stateful metrics
**Describe the bug**
Current metrics loop isn't valid for stateful metrics. 
Current metrics calculation boils down to:
1. Do metrics's forward on the batch 2. Mean the batch results at the end of epoch/evaluation. 

This breaks the  logic of the metrics' that shouldn't be computed per batch. E.g. My custom metric is Median MAE across the whole validation set. In Darts loop it is essentially replaced by mean median MAE of the batches which is a different stat - I have noticed a considerable divergence in results.

**To Reproduce**
Here is a simple example of such stateful metric
```
class Collector(Metric):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.add_state("preds", default=[], dist_reduce_fx="cat")


    def update(self, preds, target) -> None:
        if preds.shape != target.shape:
            raise ValueError("preds and target must have the same shape")
        self.preds.append(preds)

    def concat(self):
        return dim_zero_cat(self.preds),dim_zero_cat(self.targets)


#dummy tracker
class Non0Preds(Collector):
    def compute(self):
        preds,targets = self.concat()
        return sum(preds!=0)
```

Use in the prediction, observe that the metric is around the batch size rather than the size of the validation or train set as it should be. This is obviously a dummy example that  isreverse engineerable. Another example would be any ordering metric (medians, AUROC, etc.) where you can't reverse engineer the correct value. 

**Expected behavior**
A better pattern is:

1. Call metrics `.update` at every step 
2. Call `.compute` at the end of the epoch/evaluation 
3. Call .reset on the metric to reset its state (or make sure that users call it in compute) 

This allows the user to actually pass any metric and get the correct results AND is far more computationally effective since it avoids the internal state manipulations that Metric.forward does by default. 
Note that implementing `update` and `compute` is the recommended by TorchMetrics itself and you are not advised to mess with `forward` 

See the logic here for an example of implementation that should work for validation and can be replicated for training:
https://lightning.ai/docs/torchmetrics/stable/pages/overview.html#metriccollection

 - Python version: [ 3.10]
 - darts version [0.29.0]

tRosenflanz:
- [X] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md).

Fixes #2390, fixes #2389

### Summary

Instead of doing forward/log on metrics on each step it only updates the state at every step and does the computation/log/reset at the end of the epoch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed a bug when using a `RegressionModel` (that supports validation series) with a validation set, and encoders and/or component-specific lags, where the encodings and component specific lags were not added to the set. [#2383](https://github.com/unit8co/darts/pull/2383) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug where `n_steps_between` did not work properly with custom business frequencies. This affected metrics computation. [#2357](https://github.com/unit8co/darts/pull/2357) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug when calling `predict()` with a `MixedCovariatesTorchModel` (e.g. TiDE, N/DLinear, ...) `n<output_chunk_length` and a list of series with length `len(series) < n`, where the predictions did not return the correct number of series. [#2374](https://github.com/unit8co/darts/pull/2374) by [Dennis Bader](https://github.com/dennisbader).
-- Fixed a bug in torch metric evaluation loop that caused stateful metrics to be computed incorrectly as non-stateful. [#2391](https://github.com/unit8co/darts/pull/2391) by [Tim Rosenflanz](https://github.com/tRosenflanz)
+- Fixed a bug when using a `TorchForecastingModel` with stateful torch metrics, where the metrics were incorrectly computed as non-stateful. [#2391](https://github.com/unit8co/darts/pull/2391) by [Tim Rosenflanz](https://github.com/tRosenflanz)
 
 **Dependencies**
 - Improvements to linting via updated pre-commit configurations: [#2324](https://github.com/unit8co/darts/pull/2324) by [Jirka Borovec](https://github.com/borda).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,6 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed a bug when calling `predict()` with a `MixedCovariatesTorchModel` (e.g. TiDE, N/DLinear, ...) `n<output_chunk_length` and a list of series with length `len(series) < n`, where the predictions did not return the correct number of series. [#2374](https://github.com/unit8co/darts/pull/2374) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug in torch metric evaluation loop that caused stateful metrics to be computed incorrectly as non-stateful. [#2391](https://github.com/unit8co/darts/pull/2391) by [Tim Rosenflanz](https://github.com/tRosenflanz)
 
-
 **Dependencies**
 - Improvements to linting via updated pre-commit configurations: [#2324](https://github.com/unit8co/darts/pull/2324) by [Jirka Borovec](https://github.com/borda).
 - Improvements to unified linting by switch `isort` to Ruff's rule I. [#2339](https://github.com/unit8co/darts/pull/2339) by [Jirka Borovec](https://github.com/borda)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed a bug when using a `RegressionModel` (that supports validation series) with a validation set, and encoders and/or component-specific lags, where the encodings and component specific lags were not added to the set. [#2383](https://github.com/unit8co/darts/pull/2383) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug where `n_steps_between` did not work properly with custom business frequencies. This affected metrics computation. [#2357](https://github.com/unit8co/darts/pull/2357) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug when calling `predict()` with a `MixedCovariatesTorchModel` (e.g. TiDE, N/DLinear, ...) `n<output_chunk_length` and a list of series with length `len(series) < n`, where the predictions did not return the correct number of series. [#2374](https://github.com/unit8co/darts/pull/2374) by [Dennis Bader](https://github.com/dennisbader).
+- Fixed a bug in torch metric evaluation loop that caused stateful metrics to be computed incorrectly as non-stateful. [#2391](https://github.com/unit8co/darts/pull/2391) by [Tim Rosenflanz](https://github.com/tRosenflanz)
+
 
 **Dependencies**
 - Improvements to linting via updated pre-commit configurations: [#2324](https://github.com/unit8co/darts/pull/2324) by [Jirka Borovec](https://github.com/borda).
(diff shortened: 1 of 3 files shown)


darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -78,6 +79,20 @@
...
 
 
+class NumsCalled(Metric):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.add_state("preds", default=[], dist_reduce_fx="cat")
+
+    def update(self, preds, target) -> None:
+        if preds.shape != target.shape:
+            raise ValueError("preds and target must have the same shape")
```
VascoSch92, line 89:
```suggestion
            raise ValueError(f"preds and target must have the same shape, but got {preds.shape} for preds and {target.shape} for target.")
```

tRosenflanz: Thanks, updated. Note that this is just a test class, this error isn't raised in the training loop itself

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed a bug when using a `RegressionModel` (that supports validation series) with a validation set, and encoders and/or component-specific lags, where the encodings and component specific lags were not added to the set. [#2383](https://github.com/unit8co/darts/pull/2383) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug where `n_steps_between` did not work properly with custom business frequencies. This affected metrics computation. [#2357](https://github.com/unit8co/darts/pull/2357) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug when calling `predict()` with a `MixedCovariatesTorchModel` (e.g. TiDE, N/DLinear, ...) `n<output_chunk_length` and a list of series with length `len(series) < n`, where the predictions did not return the correct number of series. [#2374](https://github.com/unit8co/darts/pull/2374) by [Dennis Bader](https://github.com/dennisbader).
+- Fixed a bug in torch metric evaluation loop that caused stateful metrics to be computed incorrectly as non-stateful. [#2391](https://github.com/unit8co/darts/pull/2391) by [Tim Rosenflanz](https://github.com/tRosenflanz)
```
dennisbader, line 48:
```suggestion
- Fixed a bug when using a `TorchForecastingModel` with stateful torch metrics, where the metrics were incorrectly computed as non-stateful. [#2391](https://github.com/unit8co/darts/pull/2391) by [Tim Rosenflanz](https://github.com/tRosenflanz)
```

darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -239,9 +239,27 @@ def validation_step(self, val_batch, batch_idx) -> torch.Tensor:
             prog_bar=True,
             sync_dist=True,
         )
-        self._calculate_metrics(output, target, self.val_metrics)
+        self._update_metrics(output, target, self.val_metrics)
         return loss
 
+    def _compute_metrics(self, metrics):
+        res = metrics.compute()
```
dennisbader, line 246: Can you move this method to where method `_update_metric()` is defined?

```suggestion
        if not len(metrics):
            return

        res = metrics.compute()
```

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -78,6 +79,23 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.add_state("preds", default=[], dist_reduce_fx="cat")
+
+    def update(self, preds, target) -> None:
+        if preds.shape != target.shape:
+            raise ValueError(
+                "preds and target must have the same shape "
+                f"but got {preds.shape} for preds and {target.shape} for target."
+            )
```
dennisbader, line 92: Not required for this test

```suggestion
```

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -78,6 +79,23 @@
 ]
 
 
+class NumsCalled(Metric):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
dennisbader, line 84: Not required for this test
```suggestion
    def __init__(self):
        super().__init__()
```

dennisbader (changes requested):
Thanks for this great PR @tRosenflanz 🚀 

It looks really good, just had some minor suggestions. 
After those have been addressed, we can merge :)

darts/models/forecasting/pl_forecasting_module.py
tRosenflanz: Moved and "skipped" when needed. For some reason linter doesn't add empty line by default

darts/tests/models/forecasting/test_torch_forecasting_model.py
tRosenflanz: removed

darts/tests/models/forecasting/test_torch_forecasting_model.py
tRosenflanz: removed

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -1333,6 +1346,20 @@ def test_invalid_metrics(self):
...
+            12,
+            "RNN",
+            10,
+            10,
+            n_epochs=1,
+            torch_metrics=torch_metrics,
+            **tfm_kwargs,
+        )
+        model.fit(self.series)
+        assert model.model.trainer.logged_metrics["train_NumsCalled"] != 1
```
dennisbader, line 1361: Sorry, just saw this now. We should probably check that it is `> 1` since `0` would be incorrect.

```suggestion
        assert model.model.trainer.logged_metrics["train_NumsCalled"] > 1
```

tRosenflanz: Good call. This tests that update_metrics works correctly as well

dennisbader (approved):
Looks great, thanks a lot @tRosenflanz, read to merge 🚀 💯

PR #2362, opened 2024-04-30
Feat/ Adding sample weight

AntonRagot:
Checklist before merging this PR:
- [ ] Mentioned all issues that this PR fixes or addresses.
- [ ] Summarized the updates of this PR under **Summary**.
- [ ] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md).

Fixes #1175.

### Summary

- Adding the option to choose between three possibilities of predefined weights
- `fit` now has the option to support `sample_weight`, which can be a string or a TimeSeries

PR #2404, opened 2024-06-05
Feat/sample weights

### LINKED ISSUE #1175: Easy Sample Weights
Often in forecasting it makes sense to use sample weights that make your model focus more on the recent history. And with most Sklearn models you can introduce this through the fit method. It would be great if Darts could make it easy to implement sensible weighting schemes for forecasting such as an exponentially decaying weighting function. 

Many thanks for the library!

### LINKED ISSUE #2107: Applying sample weights to the training process

Hi, I was wondering if its possible to provide sample weights of certain observations when making predictions

Providing sample weights to the training function 

**Additional context**

dennisbader:

Fixes #1175, fixes #2107.

(this is a continuation of #2362)

Summary
- adds support for training regression models with sample weights.
- `fit()` now has two new parameters  `sample_weight` and `val_sample_weight` which allow to pass sample weights for training and validation (`val_sample_weights` only for models that support validation sets)

```
sample_weight
            Optionally, some sample weights to apply to the target `series` labels.
            They are applied per observation, per label (each step in `output_chunk_length`), and per component.
            If a string, then the weights are generated using built-in weighting functions. The available options are
            `"linear_decay"` or `"exponential_decay"`. The weights are only computed the longest series in `series`,
            and then applied globally to all `series` to have a common time weighting.
            If a `TimeSeries` or `Sequence[TimeSeries]`, then those weights are used. The number of series must
            match the number of target `series` and each series must contain at least all time steps from the
            corresponding target `series`. If the weight series only have a single component / column, then the weights
            are applied globally to all components in `series`. Otherwise, for component-specific weights, the number
            of components must match those of `series`.
val_sample_weight
```


darts/models/forecasting/regression_model.py
```diff
@@ -809,6 +827,17 @@ def fit(
...
 
+        # Checking for correct values the provided sample weights
+        if isinstance(sample_weight, str):
+            raise_if(
+                sample_weight not in ["equal", "linear_decay", "exponential_decay"],
+                f"Invalid value for `sample_weight`: {sample_weight}."
+                f"Possible values are: equal, linear_decay, exponential_decay.",
+            )
+        elif isinstance(sample_weight, TimeSeries):
+            # The error is caught later, should we still verify it here?
```
VascoSch92, line 838: why this is an error if in the docstring you are saying that `If a TimeSeries is passed, then those weights are used.`?

madtoinou: This is a placeholder, we are trying to think if it's worth it to check that the time index of the passed series matches the index of the training series for example.

darts/models/forecasting/regression_model.py
```diff
@@ -714,6 +751,17 @@ def fit(
         n_jobs_multioutput_wrapper
             Number of jobs of the MultiOutputRegressor wrapper to run in parallel. Only used if the model doesn't
             support multi-output regression natively.
+        sample_weight
+            Optionally, some sample weights to apply to the target `series` labels.
+            They are applied per observation, per label (each step in `output_chunk_length`), and per component.
+            If a string, then the weights are generated using built-in weighting functions. The available options are
+            `"linear_decay"` or `"exponential_decay"`. The weights are only computed the longest series in `series`,
```
madtoinou, line 758:
```suggestion
            `"linear_decay"` or `"exponential_decay"`. The weights are only computed for the longest series in `series`,
```

darts/utils/data/tabularization.py
```diff
@@ -261,18 +272,55 @@ def create_lagged_data(
...
+                ),
+                logger=logger,
+            )
+        if target_series is None:
+            raise_log(
+                ValueError("Must supply `target_series` when using `sample_weight`."),
+                logger=logger,
+            )
+        # create global time weights based on the longest target series
+        max_len = max(len(target_i) for target_i in target_series)
```
madtoinou, line 302: Since the weights are shared across all the series in case of multiple series training, should the generated weights have a time index index spanning between the most extremes dates across all series instead of just the longest series (since there is no guarantee of index overlap?)

madtoinou (approved):
Great work!

One minor comment about the index for the generated weights; maybe we should consider the two most extremes index values rather than the length of the longest series? WDYT?

darts/utils/data/tabularization.py
dennisbader: As discussed, the (global) weights are computed based on the length of the longest series to be time agnostic (e.g. we can ignore also ignore different frequencies across multiple series). The weights for each series are extracted from the end of those global weights. That means that each series, irrelevant of when in time it was, will weigh the time steps equally based on the index position relative to the end.

PR #2410, opened 2024-06-13
Feat/sample weight torch

### LINKED ISSUE #1175: Easy Sample Weights

### LINKED ISSUE #2107: Applying sample weights to the training process

dennisbader:

Fixes #1175, fixes #2107.

(this is a continuation of #2404)

### Summary
- adds support for training torch forecasting models with sample weights.
- `fit()` now has two new parameters  `sample_weight` and `val_sample_weight` which allow to pass sample weights for training and validation

```
sample_weight
            Optionally, some sample weights to apply to the target `series` labels.
            They are applied per observation, per label (each step in `output_chunk_length`), and per component.
            If a string, then the weights are generated using built-in weighting functions. The available options are
            `"linear_decay"` or `"exponential_decay"`. The weights are only computed the longest series in `series`,
            and then applied globally to all `series` to have a common time weighting.
            If a `TimeSeries` or `Sequence[TimeSeries]`, then those weights are used. The number of series must
            match the number of target `series` and each series must contain at least all time steps from the
            corresponding target `series`. If the weight series only have a single component / column, then the weights
            are applied globally to all components in `series`. Otherwise, for component-specific weights, the number
            of components must match those of `series`.
val_sample_weight
```

### DIFF
--- a/darts/utils/data/training_dataset.py
+++ b/darts/utils/data/training_dataset.py
@@ -141,10 +141,7 @@ class TrainingDataset(ABC, Dataset):
             )
 
             if covariate_type is not CovariateType.NONE:
-                target_time_index = target_series._time_index
-                covariate_time_index = covariate_series._time_index
-
-                # not CovariateType.Future -> both CovariateType.PAST and CovariateType.HISTORIC_FUTURE
+                # not CovariateType.FUTURE -> both CovariateType.PAST and CovariateType.HISTORIC_FUTURE
                 start = (
                     future_start
                     if covariate_type is CovariateType.FUTURE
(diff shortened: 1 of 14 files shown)


darts/utils/data/training_dataset.py
```diff
@@ -110,9 +113,184 @@ def _memory_indexer(
...
+            past_end,
+            future_start,
+            future_end,
+            covariate_start,
+            covariate_end,
+            sample_weight_start,
+            sample_weight_end,
+        )
+
+    def _memory_indexer2(
```
madtoinou, line 251: What is the difference between this method and `_memory_indexer` above?

darts/tests/models/forecasting/.lr_find_7603e48a-4678-44b9-98ec-fd749d4af7df.ckpt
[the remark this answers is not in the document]
dennisbader: 👍

darts/utils/data/training_dataset.py
dennisbader: forgot to remove this method, thanks

PR #2441, opened 2024-07-02
Added the IQR Detector

### LINKED ISSUE #2319: Add IQR Anomaly Detector
I'd like to detect outliers using the Interquartile range (IQR) method, i.e. an outlier is a point that lies more than 1.5 * IQR above(below) the upper(lower) quartile

Similar to the QuantileDetector but use calculation IQR = (Q3-Q1) where Q1 and Q3 are the 25th and 75th quartiles. Then set the upper fence at Q3 + 1.5*IQR and the lower fence to Q1 - 1.5*IQR

I'm not aware of any alternative.

**Additional context**

u8-igor:

Closes #2319.

### Summary
Added IQRDetector, allowing for detecting anomalies using the interquartile range algorithm.
It is based on the QuantileDetector and has a `scale` parameter for adjusting the acceptable range.

### DIFF
--- a/darts/ad/detectors/iqr_detector.py
+++ b/darts/ad/detectors/iqr_detector.py
@@ -1,0 +1,102 @@
+"""
+Interquartile Range (IQR) Detector
+-----------------
+
+Flags anomalies that are beyond the IQR (between the third and the first quartile)
+of historical data by some factor of it's difference (typically 1.5).
+This is similar to a threshold-based detector, but the thresholds are
+computed as distances from the IQR of historical data when the detector is fitted.
+"""
+
+from typing import Sequence, Union
+
+import numpy as np
+
+from darts.ad.detectors.quantile_detector import QuantileDetector
+from darts.ad.detectors.threshold_detector import ThresholdDetector
+from darts.logging import get_logger, raise_log
+from darts.timeseries import TimeSeries
+
+logger = get_logger(__name__)
+
+
+class IQRDetector(QuantileDetector):
+    def __init__(self, scale: Union[Sequence[Union[float, None]], float] = 1.5) -> None:
+        """IQR Detector
+
+        Flags values that lie outside of the interquartile range (IQR)
+        by more than a certain factor of IQR's value as anomalies.
+        The factor is passed in the `scale` parameter.
+
+        If a single value is provided for `scale`,
+        this same value will be used across all components of the series.
+
+        If a sequences of values is given for the `scale` parameter,
+        it's length must match the dimensionality of the series passed.
+        Unless it's length is 1, then it is treated as a single value.
+        Infinity may be used to disable flagging on that dimension.
+
+        Parameters
+        ----------
+        scale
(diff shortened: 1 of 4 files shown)


darts/ad/detectors/iqr_detector.py
```diff
@@ -0,0 +1,102 @@
...
+        scale
+            (Sequence of) scale(s) used to indicate what distance from the IQR constitutes an anomaly.
+            Defaults to `1.5`. Must be non-negative. If a sequence, must match the dimensionality of the series
+            this detector is applied to.
+        """
+
+        # Parent QuantileDetector will compute Q1 and Q3 thresholds
+        super().__init__(low_quantile=0.25, high_quantile=0.75)
+
+        self.scale = np.array(scale)
```
dennisbader, line 50: let's make sure we convert all `scale` values into a 1d array 
```suggestion
        scale = [scale] if isinstance(scale, float) else scale
        self.scale = np.array(scale)
```

then you can raise an error if `self.scale.ndim != 1`. That simplifies things for the downstream operations as we can assume a fixed dimensionality.

darts/ad/detectors/iqr_detector.py
```diff
@@ -0,0 +1,102 @@
...
+        if np.ndim(self.scale) > 0 and any(x < 0.0 for x in self.scale):
+            raise_log(
+                ValueError("All values in `scale` must be non-negative numbers."),
+                logger=logger,
+            )
+        if np.ndim(self.scale) == 0 and self.scale < 0.0:
+            raise_log(
+                ValueError("Parameter `scale` must be non-negative"),
+                logger=logger,
+            )
```
dennisbader, line 71: with my comment from above we can convert this to:

```suggestion
        if (self.scale < 0.0).any():
            raise_log(
                ValueError("All values in `scale` must be non-negative numbers."),
                logger=logger,
            )
```

darts/ad/detectors/iqr_detector.py
```diff
@@ -0,0 +1,102 @@
...
+    def _fit_core(self, series: Sequence[TimeSeries]) -> None:
+        super()._fit_core(series)
+
+        # if `scale` is not a single value, then check it matches input width
+        # Note: Due to array broadcasting mechanics the len check is also necessary
+        if (
+            np.ndim(self.scale) > 0
+            and len(self.scale) > 0
+            and len(self.scale) != series[0].width
+        ):
```
dennisbader, line 82: with comment from above:

```suggestion
        if len(self.scale) > 1 and len(self.scale) != series[0].width:
```

dennisbader (changes requested):
Thanks for this great PR @u8-igor, it already looks really good 🚀 
Your code is well-written.

I just added a couple of minor suggestions where we could try to simplify a bit here and there.

dennisbader (approved):
Looks perfect now, thanks a lot @u8-igor 🚀  Ready to merge once all tests have passed 💯

PR #2445, opened 2024-07-03
fix the sample_weights bug

### LINKED ISSUE #2439: [BUG] Models that fit method do not support the parameter sample_weights cannot be fitted
**Describe the bug**
In Darts version 0.30.0, the parameter sample_weights was added. However, there are some regressors in scikit-learn where the fit method does not support the parameter sample_weights. I have used the following code to list the models in scikit-learn where the fit method does or does not support the parameter sample_weights:
```python
from sklearn.utils import all_estimators
import inspect

for name, clf in all_estimators(type_filter='regressor'):
    try:
        # Try to instantiate the model and check the fit method
        model = clf()
        if 'sample_weight' in inspect.signature(model.fit).parameters:
            print(name, 'support sample_weight')
        else:
            print(name, 'not support sample_weight')
    except TypeError as e:
        # If a TypeError occurs during instantiation, it means that the model requires additional parameters, skip this model
       continue
```
If using `RegressionModel `with scikit-learn `GaussianProcessRegressor`、`GaussianProcessRegressor `or other models whose fit method does not support the parameter sample_weights, and run .fit(), TypeError would occur:
```
TypeError: fit() got an unexpected keyword argument 'sample_weight'
```

**To Reproduce**
Use `RegressionModel `with scikit-learn `GaussianProcessRegressor`、`GaussianProcessRegressor `or other models whose fit method does not support the parameter sample_weights, and run .fit().

**Expected behavior**
The errors should be gone.

 - Python version: 3.9.17
 - darts version 0.30.0

**Additional context**
I've made initial changes in **regression_model.py** to address this issue and have been able to get models whose fit methods don't support the sample_weights parameter to run successfully. As a dedicated user of the amazing darts library, and a newcomer to the open source community, I would be incredibly proud and excited to be allowed to contribute to this issue. I am looking forward to hearing more from you, thank you very much!
```python
import inspect
...
        if sample_weights is not None:
            fit_parameters = inspect.signature(self.model.fit).parameters
            if 'sample_weight' in fit_parameters:
                self.model.fit(
                    training_samples, training_labels, sample_weight=sample_weights, **kwargs
                )
            else:
                self.model.fit(
                    training_samples, training_labels, **kwargs
                )
```

cnhwl:

Fixes #2439.

### Summary

fixes the sample_weights bug by making conditional judgement on whether sample_weights is none and whether the parameter sample_weights exists in the model fit method.

### DIFF
--- a/darts/models/forecasting/regression_model.py
+++ b/darts/models/forecasting/regression_model.py
@@ -703,21 +703,19 @@ class RegressionModel(GlobalForecastingModel):
                 max_samples_per_ts=max_samples_per_ts,
             )
 
-        # check whether model's fit method supports parameter sample_weights
+        # check whether model's fit method supports parameter `sample_weight`
+        sample_weight_kwargs = dict()
         if sample_weights is not None:
-            fit_parameters = inspect.signature(self.model.fit).parameters
-            if "sample_weight" in fit_parameters:
-                self.model.fit(
-                    training_samples,
-                    training_labels,
-                    sample_weight=sample_weights,
-                    **kwargs,
-                )
+            if "sample_weight" in inspect.signature(self.model.fit).parameters:
+                sample_weight_kwargs = {"sample_weight": sample_weights}
             else:
-                logger.warning("Provided `sample_weights` wasn't used.")
-                self.model.fit(training_samples, training_labels, **kwargs)
-        else:
-            self.model.fit(training_samples, training_labels, **kwargs)
+                logger.warning(
+                    "`sample_weight` was ignored since underlying regression model's "
+                    "`fit()` method does not support it."
+                )
+        self.model.fit(
+            training_samples, training_labels, **sample_weight_kwargs, **kwargs
+        )
 
         # generate and store the lagged components names (for feature importance analysis)
         self._lagged_feature_names, self._lagged_label_names = (
--- a/darts/models/forecasting/regression_model.py
+++ b/darts/models/forecasting/regression_model.py
@@ -27,6 +27,7 @@ When static covariates are present, they are appended to the lagged features. Wh
 if their static covariates do not have the same size, the shorter ones are padded with 0 valued features.
 """
 
+import inspect
 from collections import OrderedDict
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


darts/models/forecasting/regression_model.py
```diff
@@ -701,9 +702,22 @@ def _fit_model(
...
+                    training_samples,
+                    training_labels,
+                    sample_weight=sample_weights,
+                    **kwargs,
+                )
+            else:
+                logger.warning("Provided `sample_weights` wasn't used.")
+                self.model.fit(training_samples, training_labels, **kwargs)
+        else:
+            self.model.fit(training_samples, training_labels, **kwargs)
```
dennisbader, line 720: Just a slight modifictation to write the `fit()` call only once :)

```suggestion
        # check whether model's fit method supports parameter `sample_weight`
        sample_weight_kwargs = dict()
        if sample_weights is not None:
            if "sample_weight" in inspect.signature(self.model.fit).parameters:
                sample_weight_kwargs = {"sample_weight": sample_weights}
            else:
                logger.warning(
                    "`sample_weight` was ignored since underlying regression model's "
                    "`fit()` method does not support it."
                )
        self.model.fit(
            training_samples, training_labels, **sample_weight_kwargs, **kwargs
        )
```

dennisbader (changes requested):
Thanks for the fix and the contribution @cnhwl  🚀  It's highly appreciated :) 

I just had a minor suggestion for code. And there are some things that we should address before merging:
- add a unit test to `darts/tests/models/forecasting/test_regression_models.py` that checks that it works as expected with models that don't support sample weights
- in `darts/utils/multioutput.py` - `MultiOutputRegressor.fit()` we currently raise an error if the user provided `sample_weight` but the model doesn't support it. The MultiOutputRegressor is applied if we need to predict more than one target component and / or output_chunk_length > 1. Ideally, we should have the same behavior there as well. So if you can replace raising an error with a warning, that should be fine. (these lines [here](https://github.com/unit8co/darts/blob/b95930d19ea6bd8d97e9e99dd5704ded1c97f8d4/darts/utils/multioutput.py#L105))
- can you add an etry to our CHANGELOG.md under the Unreleased / Fixed section?

darts/models/forecasting/regression_model.py
cnhwl: Thank you very much for your advice! @dennisbader Of course, I will complete the code as soon as possible and add some describition of this fix in the changelog.

cnhwl: Hi! @dennisbader I've completed the code changes and changelog update, please check it out.🤝

dennisbader (approved):
Thanks for the updates @cnhwl 🚀 

I took the liberty to apply the last open points directly (e.g., adaptions to MultiOutputRegressor, and adding the KNeighborsRegressor only to the sample weights tests).

Looks great now! We can merge after all tests have passed 💯

cnhwl:
Thank you so much! @dennisbader ❤️
I have benefited a lot from your code. It's been a pleasure working with you on this merge. Looking forward to contributing more to the darts library! 🚀

PR #2455, opened 2024-07-09
Fix Issue 2454

### LINKED ISSUE #2454: Documentation Inconsistency: `num_loader_workers` Parameter
The documentation for deep learning models in Darts is outdated regarding the `num_loader_workers` parameter.

### Current Documentation
[This page](https://github.com/unit8co/darts/blob/master/docs/userguide/torch_forecasting_models.md#tune-num_loader_workers), states:
>All deep learning models in Darts have a parameter `num_loader_workers` in their `fit()` and `predict()` functions, which
configures the `num_workers` parameter in the PyTorch `DataLoaders`. By default it is set to 0, which means that the main process will also take care of loading the data. 

### Actual Implementation (as of v0.30.0)
According to the changelog, this is no longer accurate:
> Removed parameter num_loader_workers from fit*(), predict*(), find_lr(). You can now set the parameter through the dataloader_kwargs dict.

### Proposed Update
I suggest updating the documentation to reflect this change:
> All deep learning models in Darts have a parameter `dataloader_kwargs` in their `fit()` and `predict()` functions, which configures the PyTorch DataLoaders. The `num_workers` parameter for PyTorch DataLoaders can be set using the `num_workers` key in the `dataloader_kwargs` dictionary.

Note: The default value for `num_workers` is (I think) the PyTorch default (0), and so the second sentence can be omitted, but this should be confirmed.

lomira:
- [x] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md) << **I don't think this is relevant here.**

### Fixes 
Fixes #2454
### Summary
No code change, only documentation
As of v0.30.0,  num_load was moved inside dataloader_kwargs. Reflect this change in documentation.

### DIFF
--- a/docs/userguide/torch_forecasting_models.md
+++ b/docs/userguide/torch_forecasting_models.md
@@ -484,9 +484,8 @@ A larger batch size tends to speed up the training because it reduces the number
 of backward passes per epoch and has the potential to better parallelize computation. However it also changes the training dynamics (e.g. you might need more epochs, and the convergence dynamics is affected). Furthermore larger batch sizes increase memory consumption. So here too some testing is required.
 
 ### Tune `num_loader_workers`
-All deep learning models in Darts have a parameter `num_loader_workers` in their `fit()` and `predict()` functions, which
-configures the `num_workers` parameter in the PyTorch `DataLoaders`. By default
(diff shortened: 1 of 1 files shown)


dennisbader (approved):
Looks great, thanks for this @lomira 🚀

PR #2474, opened 2024-07-29
Minor fix: Use pd_series3 for series3 in tests

Joelius300:
Checklist before merging this PR:
- [x] Mentioned all issues that this PR fixes or addresses.
- [x] Summarized the updates of this PR under **Summary**.
- [x] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md).

I didn't open an issue for this tiny change because it's so small and internal to the tests.
Hope that's okay; will of course open issues for anything substantial.

### Summary

Minor fix to actually use `pd_series3` for `series3` in `test_timeseries.py` (instead of `pd_series2`).

Luckily only one test needed to be adjusted for everything to pass.

### Other Information

No entry in changelog because it's a negligible and internal change.

I read the contribution guidelines but it's my first contribution to Darts nonetheless, so I am more than happy to get notes or feedback :)

PR #2476, opened 2024-07-29
Allow broadcasting single-sample TimeSeries onto multi-sample TimeSeries

### LINKED ISSUE #2475: Allow broadcasting for element-wise operations on TimeSeries

Currently, you can only do a select few broadcasting operations on TimeSeries, namely with scalar values (e.g. `series + 3`).
This means you cannot add a TimeSeries with 10 elements, 1 component and 1 sample to a TimeSeries with 10 elements, 1 component and 5 samples. When working with probabilistic forecasts, this was something that I was missing because I regularly shifted a multi-sample forecast by some non-constant series.

```python
times = pd.date_range("20130101", "20130110", freq="D")
# time-series with multiple samples
# time: 10, components: 1, samples: 3
seriesA = TimeSeries.from_times_and_values(
    times,
    np.expand_dims([
        range(10),
        range(5, 15),
        range(15, 25),
    ], axis=1).T
)

# time-series with one sample
# time: 10, components: 1, samples: 1
seriesB = TimeSeries.from_series(
    pd.Series([2 for _ in range(10)], index=times)
)

broadcast_constant = seriesA + 2
broadcast_series = seriesA + seriesB
assert broadcast_constant == broadcast_series, "Broadcasting scalar to TimeSeries does not give same result as Broadcasting single-sample TimeSeries to multi-sample TimeSeries"
```

**Expected result:** Runs without error or assertion failure

**Actual result:** `ValueError: Attempted to perform operation on two TimeSeries of unequal shapes.`

I propose allowing certain types of broadcasting, especially on sample level because I think that's the most useful case. \
This might be a good opportunity to discuss what other types of broadcasting could be useful.

At the moment, you need to extend the series explicitly when you want to add it to another.

E.g. this is what I did to work around this limitation and make my function compatible with both single- and multi-sample series.

```python
if num_samples > 1:
    series = darts.concatenate([series] * num_samples, axis=2)
```

**Additional context**

I'm currently testing some things in this regard and will open a draft pull request later on. Edit: See #2476

Joelius300:

Fixes #2475

### Summary

Allows broadcasting TimeSeries with a single sample onto TimeSeries with multiple samples for element-wise operations like addition, subtraction, etc. Note that the underlying implementation already supports broadcasting, this PR only relaxes the safeguard regarding the shape-compatibility when combining two TimeSeries. Depending on the other uses of the `_combine_arrays` method, there could also be an argument that allows broadcasting like this but given that I don't know the code base well enough to judge that, I implemented the simpler version (and made sure that all the tests still pass).

### Other Information

This PR is a draft because it's still up for discussion what other types of broadcasting might be interesting to allow.

I read the contribution guidelines but it's my first non-negligible contribution to Darts nonetheless, so I am more than happy to get notes or feedback :)

Tested to work before and after merging #2474, but working on this PR here is also where I found that minor bug in the tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ## [Unreleased](https://github.com/unit8co/darts/tree/master)
 
-- Added support for broadcasting single-sample TimeSeries onto multi-sample TimeSeries. [#2476](https://https://github.com/unit8co/darts/pull/2476)
+- Added support for broadcasting to TimeSeries on component and sample level. [#2476](https://https://github.com/unit8co/darts/pull/2476)
   by [Joel L.](https://github.com/Joelius300).
 - Fixed bug preventing TimeSeries to be divided by xarray or ndarray. [#2476](https://https://github.com/unit8co/darts/pull/2476)
   by [Joel L.](https://github.com/Joelius300).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,6 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ## [Unreleased](https://github.com/unit8co/darts/tree/master)
 
-- Added support for broadcasting to TimeSeries on component and sample level. [#2476](https://https://github.com/unit8co/darts/pull/2476)
-  by [Joel L.](https://github.com/Joelius300).
-- Fixed bug preventing TimeSeries to be divided by xarray or ndarray. [#2476](https://https://github.com/unit8co/darts/pull/2476)
-  by [Joel L.](https://github.com/Joelius300).
-
 [Full Changelog](https://github.com/unit8co/darts/compare/0.30.0...master)
 
 ### For users of the library:
@@ -19,6 +14,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Added hyperparameters controlling the hidden layer sizes for the feature encoders in `TiDEModel`. [#2408](https://github.com/unit8co/darts/issues/2408) by [eschibli](https://github.com/eschibli).
 - Made README's forecasting model support table more colorblind-friendly. [#2433](https://github.com/unit8co/darts/pull/2433)
 - Updated the Ray Tune Hyperparameter Optimization example in the [user guide](https://unit8co.github.io/darts/userguide/hyperparameter_optimization.html) to work with the latest `ray` versions (`>=2.31.0`). [#2459](https://github.com/unit8co/darts/pull/2459) by [He Weilin](https://github.com/cnhwl).
+- Added support for broadcasting to TimeSeries on component and sample level. [#2476](https://https://github.com/unit8co/darts/pull/2476) by [Joel L.](https://github.com/Joelius300).
 
 **Fixed**
 
@@ -26,6 +22,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed a bug with `xgboost>=2.1.0`, where multi output regression was not properly handled. [#2426](https://github.com/unit8co/darts/pull/2426) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug when using `ShapExplainer.explain()` with some selected `target_components` and a regression model that natively supports multi output regression: The target components were not properly mapped. [#2428](https://github.com/unit8co/darts/pull/2428) by [Dennis Bader](https://github.com/dennisbader).
 - Fixed a bug when using `fit()` with a `RegressionModel` that uses an underlying `model` which does not support `sample_weight`. [#2445](https://github.com/unit8co/darts/pull/2445) by [He Weilin](https://github.com/cnhwl).
+- Fixed a bug preventing TimeSeries to be divided by xarray or ndarray. [#2476](https://https://github.com/unit8co/darts/pull/2476) by [Joel L.](https://github.com/Joelius300).
 
 **Dependencies**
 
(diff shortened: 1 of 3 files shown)


darts/timeseries.py
dennisbader, line 1: I'm thinking a bit about what types of broadcasting could be useful:
Let's say `self` shape is (n time steps (T), n components (C), n samples (S)), then we could technically allow `other` to have shapes:
- (T, C, N): same shape
- (T, C, 1): broadcasting on sample level
- (T, 1, 1): broadcasting on component and sample level
- (T, 1, S): broadcasting on component level

What do you think? Are there any others? I believe broadcasting on time level shouldn't be supported.
I'm open for discussion as well.

darts/tests/test_timeseries.py
```diff
@@ -975,6 +975,47 @@ def test_ops(self):
             # Cannot divide by 0.
             self.series1 / 0
 
+    def test_ops_samples_broadcasting(self):
```
dennisbader, line 978: Cool test 👍 

I would maybe check the operations using a non-constant seriesB, so we can make sure that the broadcasting is performed properly along all axis/dimensions. What do you think?

dennisbader (changes requested):
Hi @Joelius300 and thanks for this great first Darts PR 🚀 This will be a neat addition to Darts!

I've only had minor suggestions, and some thoughts about what other broadcasting types we could allow. Let me know what you think.

darts/timeseries.py
Joelius300: Looks great, I can implement those if you want. I'm assuming we want to enforce that `other` is always three-dimensional? 

For two-dimensional, there might be ambiguity how to broadcast if one dimension has a size greater than 1, e.g. how would you broadcast (M, N) onto (M, N, N). \
When one dimension is 1, it might still be a bit tricky because (T, C, S) + (T, C) would be assumed to be equivalent to (T, C, S) + (T, C, 1), so you need to ensure that C != S, otherwise you cannot differentiate (T, C, S) + (T, C) from (T, C, S) + (T, S). For that reason, I would enforce a three-dimensional `other` except for the following.

In the case (T, C, S) + (T, 1, 1), I think allowing `other` to be (T,) or (T, 1) would also be fine, what do you think?

darts/tests/test_timeseries.py
Joelius300: Can do. I started with that but it's much more work (and my implementation would just always fail because #2474 😅 ) so I thought I'd do a simpler one first.

darts/timeseries.py
dennisbader: If you want to implement those it would be great of course :) 

Yes, let's enforce 3D for now to keep it safe. Also regarding (T, C, S) + (T, 1, 1), we should also enforce 3D. 
- if `other` is a `TimeSeries`, it is guaranteed to be 3D
- if `other` is an array-like (let's say a numpy array), the user would just have to reshape it with `arr.reshape(-1, 1, 1))`. That's not too cumbersome in my opinion.

What do you think?

darts/tests/test_timeseries.py
dennisbader: I think you can also just create a new `TimeSeries` within the test, if the pre-defined examples are not suitable for this test.

- I would generate the expected results using numpy arrays and operations since we can assume that numpy does the broadcasting as expected. 
- Then create a TimeSeries from the expected numpy arrays (with the time index and columns from the source/left series) and check that the TimeSeries broadcasting operations give the same result as these expected TimeSeries

darts/timeseries.py
Joelius300: Done, (t, 1, 1), (t, c, 1) and (t, 1, s) are all allowed now.

darts/tests/test_timeseries.py
Joelius300: Added a more extensive test. Also found another bug, you could not divide TimeSeries by xarray or ndarray. I added another test for that and fixed it.

Joelius300:
Thank you for the review! I hope I could resolve all of it. I decided not to open another PR for the small division bug I found, but am happy to do so if you'd like (the commits cd25d9d and 50e15f6 would be moved out of this PR).

dennisbader (changes requested):
Thanks for the changes @Joelius300! Looks great, I only had two very small suggestions. After that it's ready to be merge 🚀

darts/timeseries.py
```diff
@@ -4984,12 +4992,18 @@ def __truediv__(self, other):
...
         elif isinstance(other, (TimeSeries, xr.DataArray, np.ndarray)):
-            if not (other.all_values(copy=False) != 0).all():
+            if isinstance(other, TimeSeries):
+                other_vals = other.data_array(copy=False).values
+            elif isinstance(other, xr.DataArray):
+                other_vals = other.values
+            else:
+                other_vals = other
+            if not (other_vals != 0).all():
                 raise_log(
```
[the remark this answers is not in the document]
Joelius300: Okay, so I'll leave this as is?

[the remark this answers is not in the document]
dennisbader: yes

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ## [Unreleased](https://github.com/unit8co/darts/tree/master)
 
+- Added support for broadcasting to TimeSeries on component and sample level. [#2476](https://https://github.com/unit8co/darts/pull/2476)
```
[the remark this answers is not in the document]
Joelius300: Ah I'm sorry, I misunderstood the format. Should be fixed now.

Joelius300:
Thanks for the quick feedback, should be all fixed now. Do you want me to squash some commits together or is it fine like this?

dennisbader:
> Thanks for the quick feedback, should be all fixed now. Do you want me to squash some commits together or is it fine like this?

This is done automatically when merging :)

dennisbader (approved):
Looks perfect now, thanks a lot @Joelius300 🚀 💯

dennisbader:
The dataset download fails sometimes, that's okay. I'll merge

hrzn:
Nice one @Joelius300 !

PR #2504, opened 2024-08-17
Feature add activation to BlockRNN (#2492)

### LINKED ISSUE #2492: [BUG] No activations in BlockRNNModel output MLP
**Describe the bug**
I've been trying to understand the differences between `BlockRNNModel` and `RNNModel`; in particular, why one only supports future covariates, one only supports past covariates, and neither support static covariates. This is still unclear to me, however it's not the reason for this issue.

While I was looking at the source, I noticed that in `BlockRNNModel`, an MLP is being created which is just a stack of linear layers with no activations in between, which seems like a mistake.

https://github.com/unit8co/darts/blob/a646adf10fa73d8facb16a611f8a3682dc8a1191/darts/models/forecasting/block_rnn_model.py#L158-L163

**Expected behavior**
There should be an nonlinear activation function between the linear layers, otherwise it is the same as just having a single linear layer. An argument to `BlockRNNModel` should be provided that allows the user to choose this activation function.

SzymonCogiel:

Fixes #2492 .

### Summary

* Added support for specifying PyTorch activation functions (`ReLU`, `Sigmoid`, `Tanh`, or `None`) in the `BlockRNNModel`.
* Ensured that activation functions are applied between fully connected layers, but not as the final layer.
* Implemented a check to raise an error if an activation function is set but the model only contains one linear layer.
* Updated documentation to reflect the new activation parameter and usage examples.
* Added test cases to verify the correct application of activation functions and to handle edge cases.

### Other Information

Consider using a logging warning instead of raising an error if an activation function is set but the model only contains one linear layer.


eschibli:
Should a warning be raised if `activation` is None and `hidden_fc_sizes` isn't? That would always be suboptimal and users who don't read the changelog might not be aware the default activation is None.

darts/models/forecasting/block_rnn_model.py
```diff
@@ -243,6 +259,9 @@ def __init__(
             Sizes of hidden layers connecting the last hidden layer of the RNN module to the output, if any.
         dropout
             Fraction of neurons afected by Dropout.
+        activation
+            The name of the activation function to be applied between the layers of the fully connected network.
+            Options include "ReLU", "Sigmoid", "Tanh", or None for no activation. Default: None.
```
madtoinou, line 264: I would not give example
```suggestion
            The name of a torch.nn activation function to be applied between the layers of the fully connected network. Default: None.
```

darts/models/forecasting/block_rnn_model.py
```diff
@@ -430,11 +449,31 @@ def encode_year(idx):
...
+                    ValueError(
+                        f"Invalid activation function: {activation}. "
+                        "Please use a valid torch.nn activation function name."
+                    ),
+                    logger=logger,
+                )
+            raise_if(
+                activation and (hidden_fc_sizes is None or len(hidden_fc_sizes) == 0),
+                "The activation function has been set, but the model contains only one linear layer. "
+                "Activation function will not be applied.",
```
madtoinou, line 467: Since this error interrupt the code, I would change the last sentence to make it actionable
```suggestion
                "The activation function has been set, but the model does not contain hidden fully connected layers; "
                "either set `activation=None` or increase `hidden_fc_sizes`.",
```

darts/models/forecasting/block_rnn_model.py
```diff
@@ -430,11 +449,31 @@ def encode_year(idx):
...
+                logger=logger,
+            )
+
```
madtoinou, line 470: As suggested by @eschibli, I would add a check that raise an error when activation is None and `hidden_fc_sizes` is greater than 0, to tell the user to select an activation function (otherwise, `hidden_fc_sizes > 1` is useless as you correctly reported).

darts/tests/models/forecasting/test_block_RNN.py
```diff
@@ -180,3 +222,18 @@ def helper_test_pred_length(self, pytorch_model, series):
 
     def test_pred_length(self):
         self.helper_test_pred_length(BlockRNNModel, self.series)
+
+    def test_varied_chunk_lengths(self):
```
madtoinou, line 226: What is the point of this test?

madtoinou (changes requested):
Thank you for the PR, some minor comments about the sanity checks.

darts/tests/models/forecasting/test_block_RNN.py
SzymonCogiel: This test checks if the BlockRNNModel correctly handles different input and output chunk lengths and ensures the predictions have the expected shape and length. It verifies that the model functions properly in various scenarios that might occur in practice.

darts/models/forecasting/block_rnn_model.py
SzymonCogiel: Done

darts/models/forecasting/block_rnn_model.py
```diff
@@ -464,6 +512,7 @@ def _create_model(self, train_sample: Tuple[torch.Tensor]) -> torch.nn.Module:
             num_layers=self.n_rnn_layers,
             num_layers_out_fc=hidden_fc_sizes,
             dropout=self.dropout,
+            activation=self.activation,
```
dennisbader, line 515: Adding `activation` will break loading the weights from earlier python versions.

You would have to add the method `_check_ckpt_parameters` to handle this case.
You can look at the TiDEModel as reference. Here would be the method for the BlockRNNModel:

```
    def _check_ckpt_parameters(self, tfm_save):
        # new parameters were added that will break loading weights
        new_params = ["activation"]
        for param in new_params:
            if param not in tfm_save.model_params:
                tfm_save.model_params[param] = "ReLU"
        super()._check_ckpt_parameters(tfm_save)
```

Having done this, only the case when a user trained a model with multiple fc layers will fail (which is okay, since no activation there is bad practice in any case).

darts/models/forecasting/block_rnn_model.py
```diff
@@ -430,11 +449,40 @@ def encode_year(idx):
                     logger=logger,
                 )
 
+        raise_if(
```
dennisbader, line 452: With the comment from above, we could remove this check

darts/models/forecasting/block_rnn_model.py
```diff
@@ -430,11 +449,40 @@ def encode_year(idx):
...
+                getattr(nn, activation)
+            except AttributeError:
+                raise_log(
+                    ValueError(
+                        f"Invalid activation function: {activation}. "
+                        "Please use a valid torch.nn activation function name."
+                    ),
+                    logger=logger,
+                )
+            raise_if(
```
dennisbader, line 473: and we could remove this as well

darts/models/forecasting/block_rnn_model.py
```diff
@@ -195,6 +210,7 @@ def __init__(
         n_rnn_layers: int = 1,
         hidden_fc_sizes: Optional[List] = None,
         dropout: float = 0.0,
+        activation: Optional[str] = None,
```
dennisbader, line 213: I'm thinking whether it'd be better to just use a default activation:

- If there are no hidden layers, then no activiation will be added in any way.
- In case of hidden layers, then we use the default.

Otherwise every time you want to use `hidden_fc_sizes` you'd have to also specify the activiation function which might be annoying.

@madtoinou what do you think?

```suggestion
        activation: str = "ReLU",
```

dennisbader (changes requested):
Hi @SzymonCogiel and thanks for this great PR 🚀  We're close, I just had a couple more suggestions. Mainly that I would set a default activation, to make it easier for the user in case he uses multiple fc layers.

darts/models/forecasting/block_rnn_model.py
```diff
@@ -135,7 +144,8 @@ def __init__(
         Outputs
         -------
         y of shape `(batch_size, output_chunk_length, target_size, nr_params)`
-            Tensor containing the prediction at the last time step of the sequence.
+            Tensor containing the prediction at the last time step of the sequence, where optional activation
+            functions may have been applied between the layers of the fully connected network.
```
dennisbader, line 148: you can use the original description here in my opinion.

darts/models/forecasting/block_rnn_model.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         nr_params: int,
         num_layers_out_fc: Optional[List] = None,
         dropout: float = 0.0,
+        activation: Optional[str] = None,
```
dennisbader, line 33: This is not optional anymore. Can you update it everywhere, and also adjust the parameter description by removing the "Optionally, "?
```suggestion
        activation: str = "ReLU",
```

dennisbader (changes requested):
Thanks for the updates @SzymonCogiel. Last little adaptions and then we're ready to merge :)

madtoinou (approved):

dennisbader (approved):
Looks good now, thanks @SzymonCogiel 🚀

PR #2512, opened 2024-08-28
Fix/bug time series.from group dataframe 2491

### LINKED ISSUE #2491: [BUG] TimeSeries.from_group_dataframe incompatible with integer timestamps
**Describe the bug**
`TimeSeries.from_group_dataframe` should support both timestamp and integer time columns, but it internally converts to `DatetimeIndex` regardless of what is passed.

https://github.com/unit8co/darts/blob/a646adf10fa73d8facb16a611f8a3682dc8a1191/darts/timeseries.py#L869-L871

**To Reproduce**
```python
import pandas as pd
import darts
df = pd.DataFrame({
    "group": ["a", "a", "a", "b", "b", "b"],
    "t": [0, 1, 2, 0, 1, 2],
    "x": [1.0, 2.0, 3.0, 2.0, 3.0, 4.0],
})
ts = darts.TimeSeries.from_group_dataframe(df, group_cols="group", time_col="t")

ts[0].time_index
# DatetimeIndex([          '1970-01-01 00:00:00',
#                '1970-01-01 00:00:00.000000001',
#                '1970-01-01 00:00:00.000000002'],
#               dtype='datetime64[ns]', name='t', freq='ns')
```

**Expected behavior**
`time_index` should be a `RangeIndex`.

 - Python version: 3.10
 - darts version 0.30.0

AlessiopSymplectic:

Fixes #2491 

### Summary

In the function "from_grou_dataframe()" there was an instance that casted the index as a Datetime index before sorting the index. The line was removed.

A test called "test_grom_group_dataframe()" was added to test_timeseries to verify the intended functionality.

### DIFF
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -165,7 +165,7 @@ class TestTimeSeries:
         # getting index for idx should return i s.t., series[i].time == idx
         assert series.get_index_at_point(16) == 3
 
-    def test_grom_group_dataframe(self):
+    def test_from_group_dataframe(self):
         # checks that the time_index is of RangeIndex type when the time_col is a list/Rangeindex
         group = ["a", "a", "a", "b", "b", "b"]
         values = np.random.randn(len(group))
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -165,6 +165,31 @@ class TestTimeSeries:
         # getting index for idx should return i s.t., series[i].time == idx
         assert series.get_index_at_point(16) == 3
 
+    def test_grom_group_dataframe(self):
+        # checks that the time_index is of RangeIndex type when the time_col is a list/Rangeindex
+        group = ["a", "a", "a", "b", "b", "b"]
+        values = np.random.randn(len(group))
+
+        # for time as a list
+        time = [i for i in range(len(group))]
+        df = pd.DataFrame({
+            "group": group,
+            "time": time,
+            "x": values,
+        })
+        ts = TimeSeries.from_group_dataframe(df, group_cols="group", time_col="time")
+        assert isinstance(ts[0].time_index, pd.RangeIndex)
+
+        # for time as RangeIndex
+        time = pd.RangeIndex(len(group))
+        df = pd.DataFrame({
+            "group": group,
+            "time": time,
+            "x": values,
+        })
+        ts = TimeSeries.from_group_dataframe(df, group_cols="group", time_col="time")
+        assert isinstance(ts[0].time_index, pd.RangeIndex)
+
     def test_integer_indexing(self):
         n = 10
         int_idx = pd.Index([i for i in range(n)])
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -166,12 +166,12 @@ class TestTimeSeries:
         assert series.get_index_at_point(16) == 3
 
     def test_from_group_dataframe(self):
-        # checks that the time_index is of RangeIndex type when the time_col is a list/Rangeindex
+        # checks that the time_index is of RangeIndex type when the time_col is a(n) (unsorted) list and/or a Rangeindex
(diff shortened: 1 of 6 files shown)


darts/tests/test_timeseries.py
```diff
@@ -165,6 +165,31 @@ def test_integer_range_indexing(self):
         # getting index for idx should return i s.t., series[i].time == idx
         assert series.get_index_at_point(16) == 3
 
+    def test_grom_group_dataframe(self):
```
madtoinou, line 168: Small typo in the name of the function
```suggestion
    def test_from_group_dataframe(self):
```

darts/tests/test_timeseries.py
```diff
@@ -165,6 +165,31 @@ def test_integer_range_indexing(self):
...
+        # for time as RangeIndex
+        time = pd.RangeIndex(len(group))
+        df = pd.DataFrame({
+            "group": group,
+            "time": time,
+            "x": values,
+        })
+        ts = TimeSeries.from_group_dataframe(df, group_cols="group", time_col="time")
+        assert isinstance(ts[0].time_index, pd.RangeIndex)
+
```
madtoinou, line 192: Can you add a test for `time` made out of `pd.Timestamps`, just to make sure they are converted to the expected type (`DatetimeIndex`)?

You could also try to leverage the `pytest.parametrize` functionality to make the test a bit shorter.

Also, would be nice to add a case where the index (either list of int or list of Timestamps) are not ordered, to verify that the sorting of the index behave as expected.

madtoinou (commented):
Looks great, minors comments about the tests to make them a bit more exhaustive & ensure that everything behave as expected.

darts/tests/test_timeseries.py
AlessiopSymplectic: Thanks for the comments and the corrections, much appreciated. There are some more tests on the group_from function inside "test_timeseries_static_covariates" already, but I agree, it's better to test and unsorted index (and datetimeindex with timestamps).

darts/tests/test_timeseries_static_covariates.py
```diff
@@ -104,6 +104,48 @@ def test_ts_from_x(self, tmpdir_module):
...
+        time = [2, 1, 0, 0, 1, 2]
+        df = pd.DataFrame({
+            "group": group,
+            "time": time,
+            "x": values,
+        })
+        ts = TimeSeries.from_group_dataframe(df, group_cols="group", time_col="time")
+        assert isinstance(ts[0].time_index, pd.RangeIndex)
+        assert (ts[0].time_index == pd.RangeIndex(3)).all()
+        assert (ts[1].time_index == pd.RangeIndex(3)).all()
```
dennisbader, line 122: You can write it like this everywhere (and remove the assert in the line above)
```suggestion
        assert ts[0].time_index.equals(pd.RangeIndex(3))
```

dennisbader, line 122: Also adding a check that the values in the TimeSeries are in the expected order:
You can extract the values with `ts[0].values()`

darts/timeseries.py
```diff
@@ -865,11 +865,19 @@ def from_group_dataframe(
...
-        # sort on entire `df` to avoid having to sort individually later on
         if time_col:
-            df.index = pd.DatetimeIndex(df[time_col])
-            df = df.drop(columns=time_col)
-        df = df.sort_index()
+            df = df.set_index(df[time_col])
+
+        if df.index.is_monotonic_increasing:
+            logger.warning(
+                "UserWarning: The `df` (time) index set from `time_col` is monotonically increasing. This "
```
dennisbader, line 873:
```suggestion
                "UserWarning: The (time) index from `df` is monotonically increasing. This "
```

dennisbader (changes requested):
Looks great, thanks @AlessiopSymplectic 🚀 
Just some last minor comments, then we're ready to merge.

dennisbader (approved):
Looks great now, thanks a lot again @AlessiopSymplectic 🚀

PR #2520, opened 2024-09-02
Fix/bug 2491 revisited

AlessiopSymplectic:

Fixes #2512 .


darts/timeseries.py
```diff
@@ -866,7 +866,40 @@ def from_group_dataframe(
...
+                    )
+            elif np.issubdtype(df[time_col].dtype, np.datetime64):
+                df.index = pd.DatetimeIndex(df[time_col])
+                df.drop(columns=time_col)
+            else:
+                raise_log(
+                    AttributeError(
+                        "Invalid type of `time_col`: it needs to be of either 'str', 'datetime' or 'int' dtype."
+                    )
+                )
```
dennisbader, line 902: That's a nice start to handle the different types. We can actually simplify this here a bit. What we can/should assume is that if the column was a str/object or np.datetime64 then we should create a DatetimeIndex. Otherwise, it must be an integer index.

So, we can let pandas handle the exceptions in case it does not work. Then the code becomes something as below:

```suggestion
            if np.issubdtype(df[time_col].dtype, object) or np.issubdtype(
                df[time_col].dtype, np.datetime64
            ):
                df.index = pd.DatetimeIndex(df[time_col])
                df = df.drop(columns=time_col)
            else:
                df = df.set_index(time_col)
```

Also as a general tip: **try** to avoid try/except as much as possible. Especially when you don't raise the error in the `except:`. This can very quickly lead to unintended behavior.

darts/timeseries.py
```diff
@@ -876,7 +909,7 @@ def from_group_dataframe(
             )
 
         # sort on entire `df` to avoid having to sort individually later on
-        if not df.index.is_monotonic_increasing:
+        else:
```
dennisbader, line 912: 👍

darts/tests/test_timeseries_static_covariates.py
```diff
@@ -154,6 +154,42 @@ def test_from_group_dataframe(self):
         assert (ts[0].values().flatten() == [values[2], values[1], values[0]]).all()
         assert (ts[1].values().flatten() == [values[3], values[4], values[5]]).all()
 
+        # for time as string dates
```
dennisbader, line 157: I would take this test as a way to get familiar with test parametrization (e.g. writing a generic test that can runs/test on collection of inputs).

You would need to adapt the test so that it expects an input time index type, and from that it will generate the corresponding index and do the entire logic only once.

You can then use the `@pytest.mark.parametrize(...)` decorator to run different test configurations.

Spoiler: Below is an example of how this test could be simplified:

```
    @pytest.mark.parametrize(
        "index_type",
        ["int", "dt", "str"]
    )
    def test_from_group_dataframe(self, index_type):
        """Tests correct extract of TimeSeries groups from a long DataFrame with unsorted (time/integer) index"""
        group = ["a", "a", "a", "b", "b", "b"]
        values = np.arange(len(group))

        # time as an unsorted list
        if index_type == "int":
            index_expected = pd.RangeIndex(3)
            time = [2, 1, 0, 0, 1, 2]
        else:  # datetime index
            index_expected = pd.date_range("2024-01-01", periods=3)
            time = index_expected[::-1].append(index_expected)
            if index_type == "str":
                time = time.astype(str)

        # for time as a unsorted list
        df = pd.DataFrame({
            "group": group,
            "time": time,
            "x": values,
        })
        ts = TimeSeries.from_group_dataframe(df, group_cols="group", time_col="time")

        # check the time index
        assert ts[0].time_index.equals(index_expected)
        assert ts[1].time_index.equals(index_expected)

        # check the values
        assert (ts[0].values().flatten() == [values[2], values[1], values[0]]).all()
        assert (ts[1].values().flatten() == [values[3], values[4], values[5]]).all()
```

dennisbader (changes requested):
Thanks for this @AlessiopSymplectic, looks already like a good start 🚀 
I had some suggestions / tips on how we could try to simplify a bit.

darts/timeseries.py
AlessiopSymplectic: Completely agree, thanks 👍

darts/tests/test_timeseries_static_covariates.py
AlessiopSymplectic: thanks (done so in the latest commit)

dennisbader (approved):
Looks great, thanks a lot for the fix @AlessiopSymplectic 🚀

PR #2522, opened 2024-09-03
Fix/bug time series.append and prepend

### LINKED ISSUE #2516: [BUG] TimeSeries.append and prepend with empty array produces an unhelpful error message
**Describe the bug**

`TimeSeries.prepend_values` produces a less-than-clear error message when called from a TimeSeries with a datetime axis and passed an empty array

**To Reproduce**
Steps to reproduce the behavior, preferably code snippet.
```
ts = darts.TimeSeries.from_times_and_values(pd.date_range(start='2021-01-01', periods=9, freq='W-SUN'),
   np.array([1, 2, 3, 4, 5, 6, 7, 8, 9]))

ts.append_values(np.array([]))  # Long stacktrace, including IndexError: index 0 is out of bounds for axis 0 with size 0
ts.prepend_values(np.array([]))  # Long stacktrace, including IndexError: index -1 is out of bounds for axis 0 with size 0

# Works with an integer axis
ts = darts.TimeSeries.from_values(np.array([1, 2, 3, 4, 5, 6, 7, 8, 9])
ts.prepend_values(np.array([]))  # Returns unchanged `ts`

```
**Expected behavior**
Either a clear error message that confirms the unsuitability of the array for pre/appending, or for the unchanged timeseries to be returned. 

 - Python 3.11
 - darts 0.30.0

**Additional context**
The error raised if the array has an unsuitable shape is clearer, but could also be improved if the array dims were checked for suitability, rather than letting `append` raise an error for mismatched numbers of components.
```
ts = darts.TimeSeries.from_values(np.array([1, 2, 3, 4, 5, 6, 7, 8, 9]).reshape(3, 3))
ts.append_values(np.array([10, 11, 12]))  # "ValueError: Both series must have the same number of components."
```

### LINKED ISSUE #2523: [BUG] Time index name is lost after using prepend_values
**Describe the bug**
The name of the time_index changes to 'time' after using `prepend_values`.

**To Reproduce**
```python
import numpy as np
import pandas as pd
from darts import TimeSeries
from darts.utils.timeseries_generation import generate_index

start = pd.Timestamp("2000-01-01")
end = pd.Timestamp("2000-12-31")
freq = pd.Timedelta(weeks=1)

index = generate_index(start=start, end=end, freq=freq, name='date')
values = np.random.normal(0, 1, size=len(index))

ts = TimeSeries.from_times_and_values(index, values, freq=freq, columns=pd.Index(['value']))
ts2 = ts.prepend_values(np.repeat(0, 30))

ts.time_index.name, ts2.time_index.name # ('date', 'time')
```

**Expected behavior**
The name of the time_index should stay the same when values are prepended.

AlessiopSymplectic:

Fixes #2516 and fixes #2523.


darts/timeseries.py
```diff
@@ -2849,7 +2854,10 @@ def append(self, other: Self) -> Self:
             "Both series must have the same number of components.",
             logger,
         )
-        if self._has_datetime_index:
+        # In the Rangeindex we never call this (especially when empty), thus not resulting in the
+        # bug when ultimately calling "other.start_time()"
+        # Checking that other.time_index is NOT EMPTY and only then entering should fix the issue
+        if self._has_datetime_index and (not other.time_index.empty):
```
dennisbader, line 2860: Thinking about, it might be better to also check this for the non-datetime index (range index). This will introduce a breaking change (which we need to highlight in the CHANGELOG.md) but at least we're consistent with our notion of time across all index types.

Also this check requires `self` not to be empty either

```suggestion
        if len(self) > 0 and len(other) > 0:
```

darts/timeseries.py
```diff
@@ -2849,7 +2854,10 @@ def append(self, other: Self) -> Self:
...
             raise_if_not(
                 other.start_time() == self.end_time() + self.freq,
                 "Appended TimeSeries must start one time step after current one.",
```
dennisbader, line 2863:
```suggestion
                "Appended TimeSeries must start one (time) step after current one.",
```

darts/timeseries.py
```diff
@@ -2886,6 +2894,29 @@ def append_values(self, values: np.ndarray) -> Self:
         TimeSeries
             A new TimeSeries with the new values appended
         """
+
+        values = np.array(values) if not isinstance(values, np.ndarray) else values
+
+        # If the component and sample dimension don't match, we try to recast the values into the
+        # appropriate shape:
+        if values.shape[1:] != self._xa.values.shape[1:]:
```
dennisbader, line 2902: As discussed, let's use the existing `expand_arr()`, and then raise an error if the two last dimensions don't match (component, and sample dimension)

darts/timeseries.py
```diff
@@ -2946,6 +2978,28 @@ def prepend_values(self, values: np.ndarray) -> Self:
             A new TimeSeries with the new values prepended.
         """
 
+        values = np.array(values) if not isinstance(values, np.ndarray) else values
```
dennisbader, line 2981: same here

darts/tests/test_timeseries.py
```diff
@@ -810,6 +810,50 @@ def test_append_values(self):
         assert appended.time_index.equals(expected_idx)
         assert appended.components.equals(series.components)
 
+    def test_pre_and_append_values_empty(self):
```
dennisbader, line 813: we should check it for datetime index as well (we can use pytest parametrization)

darts/tests/test_timeseries.py
```diff
@@ -810,6 +810,50 @@ def test_append_values(self):
...
+        series = linear_timeseries(start=1, length=9, freq=2)
+        appended = series.append_values([])
+        assert np.allclose(appended.values(), series.values())
```
dennisbader, line 816: checking for time index equality could also be good

darts/tests/test_timeseries.py
```diff
@@ -810,6 +810,50 @@ def test_append_values(self):
...
+        appended = series.prepend_values([])
+        assert np.allclose(appended.values(), series.values())
+
+    def test_pre_and_append_values_shape(self):
```
dennisbader, line 820: after our discussion, we probably have to adapt here a bit. 
We only naively expand an input array-like to 3-D with the `expand_arr()` function, assuming that the input dimension follow the shape (time, optional(component), optional(samples)).

darts/utils/utils.py
```diff
@@ -523,3 +523,18 @@ def expand_arr(arr: np.ndarray, ndim: int):
     if len(shape) != ndim:
         arr = arr.reshape(shape + tuple(1 for _ in range(ndim - len(shape))))
     return arr
+
+
+def expand_arr_comp_sample(arr: np.ndarray, shape_0: tuple):
```
dennisbader, line 528: we can remove it :)

dennisbader (changes requested):
Thanks for this @AlessiopSymplectic, looks nice :) 
Had some suggestions, mainly from our discussion the other day.

dennisbader (approved):
Looks great now, thanks @AlessiopSymplectic 🚀

PR #2529, opened 2024-09-12
Feat/scalar with window

### LINKED ISSUE #1540: [FEATURE] Scaler with rolling/expanding window to eliminate look ahead bias

madtoinou:

Fixes #1540.

### Summary

- Possibility to pass `Scaler/Pipeline` to `historical_forecasts()`, `backtest()` and `gridsearch()` to avoid data-leakage
  - When `enable_optimization=True`, the `Pipeline` must already be fitted (when applicable), all the series are transformed in one pass
  - `Pipeline` are systematically fitted and applied to the series between each forecast horizon (regardless of the `retrain` parameter value)

### Other Information

The input series to these methods must always be "un-processed" when providing `data_transformers` in order to avoid "double scaling" of the series.

This PR is based on #2021.


darts/models/forecasting/forecasting_model.py
```diff
@@ -1053,6 +1080,19 @@ def retrain_func(
...
+                if len(data_transformers) > 0:
+                    # data transformers are retrained between iterations to avoid data-leakage
+                    train_series, past_covariates_, future_covariates_ = (
+                        _apply_data_transformers(
+                            series=train_series,
+                            past_covariates=past_covariates_,
+                            future_covariates=future_covariates_,
+                            data_transformers=data_transformers,
+                            max_future_cov_lag=model.extreme_lags[5],
+                            fit_transformers=True,
```
dennisbader, line 1092: this should be defined according to `retrain`, right?

darts/models/forecasting/forecasting_model.py
```diff
@@ -1193,6 +1241,9 @@ def backtest(
         reduction: Union[Callable[..., float], None] = np.mean,
         verbose: bool = False,
         show_warnings: bool = True,
+        data_transformers: Optional[
```
dennisbader, line 1244: add description

darts/models/forecasting/forecasting_model.py
```diff
@@ -1533,6 +1585,9 @@ def gridsearch(
         verbose=False,
         n_jobs: int = 1,
         n_random_samples: Optional[Union[int, float]] = None,
+        data_transformers: Optional[
```
dennisbader, line 1588: add description

darts/models/forecasting/forecasting_model.py
```diff
@@ -1773,27 +1838,49 @@ def _evaluate_combination(param_combination) -> float:
                     last_points_only=last_points_only,
                     verbose=verbose,
                     show_warnings=show_warnings,
+                    data_transformers=data_transformers,
                     fit_kwargs=fit_kwargs,
                     predict_kwargs=predict_kwargs,
                     sample_weight=sample_weight,
                 )
             else:  # split mode
+                if len(data_transformers) > 0:
```
dennisbader, line 1847: should we also apply this to the `use_fitted_values=True` case?

darts/models/forecasting/forecasting_model.py
```diff
@@ -1716,6 +1771,16 @@ def gridsearch(
                 logger,
             )
 
+        if data_transformers is None:
```
dennisbader, line 1774: could be nice to have a utils function for that somewhere

darts/models/forecasting/forecasting_model.py
```diff
@@ -1773,27 +1838,49 @@ def _evaluate_combination(param_combination) -> float:
...
-                    past_covariates=past_covariates,
-                    future_covariates=future_covariates,
+                    series=series_,
+                    past_covariates=past_covariates_,
+                    future_covariates=future_covariates_,
                     num_samples=1,
                     verbose=verbose,
                     **predict_kwargs,
                 )
+                if (
```
dennisbader, line 1879: also for this, a utils function would be nice, just to have the reoccuring logic covered in one function

darts/utils/historical_forecasts/optimized_historical_forecasts_regression.py
```diff
@@ -173,6 +191,9 @@ def _optimized_historical_forecasts_last_points_only(
                 hierarchy=series_.hierarchy,
             )
         )
+    # single data transformer accross all series and forecasts, using optimized inverse_transform
```
dennisbader, line 194:
```suggestion
    # single data transformer across all series and forecasts, using optimized inverse_transform
```

darts/utils/historical_forecasts/optimized_historical_forecasts_regression.py
```diff
@@ -33,20 +37,34 @@ def _optimized_historical_forecasts_last_points_only(
...
+        data_transformers = dict()
     forecasts_list = []
     iterator = _build_tqdm_iterator(series, verbose)
     for idx, series_ in enumerate(iterator):
         past_covariates_ = past_covariates[idx] if past_covariates is not None else None
         future_covariates_ = (
             future_covariates[idx] if future_covariates is not None else None
         )
+        # Pipeline/DataTransformer must already be fitted, transform everything in one go
+        if len(data_transformers) > 0:
```
dennisbader, line 58: I think the transformation should be applied on all series together, and not individually.

Also, in my opinion, we can move the transform / inverse transform out of this function and into `ForecastingModel.historical_forecasts()`. Apply transform before calling `model_optimized_historical_forecasts()`, and inverse transform before returning.

That would simplify things quite a bit

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -815,3 +821,122 @@ def test_residuals(self):
...
+        "params",
+        product(
+            [
+                (
+                    {"series": sine_3_ts + 3},
+                    {"target": Scaler()},
+                ),
+                (
+                    {"series": sine_3_ts + 3, "past_covariates": sine_1_ts + 3},
+                    {"past": Scaler(scaler=MinMaxScaler())},
```
dennisbader, line 835: That is the default already

```suggestion
                    {"past": Scaler()},
```

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -815,3 +821,122 @@ def test_residuals(self):
...
+                        "past_covariates": sine_2_ts + 3,
+                        "future_covariates": sine_1_ts + 3,
+                    },
+                    {"target": Scaler(), "past": Scaler()},
+                ),
+            ],
+            [True, False],
+        ),
+    )
+    def test_historical_forecasts_with_scaler(self, params):
```
dennisbader, line 853: nice one 👍  :) 
Could you check for 2 historical forecast iterations? Also checking `last_points_only=False`, and `enable_optimization=True` would be important.

dennisbader (changes requested):
Thanks a lot @madtoinou this looks really great 🚀  I like the approach with the Pipeline instead of Transformers :) 

Had some remarks and suggestions, we can discuss later.

darts/dataprocessing/pipeline.py
```diff
@@ -217,6 +221,32 @@ def invertible(self) -> bool:
         """
         return self._invertible
 
+    def fittable(self) -> bool:
```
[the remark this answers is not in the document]
madtoinou: I agree, I did not like the fact that API was different for the pipeline (compared to models for example)

darts/utils/historical_forecasts/utils.py
```diff
@@ -211,6 +212,23 @@ def _historical_forecasts_general_checks(model, series, kwargs):
                 logger,
             )
 
+    # ensure DataTransformer are already be fitted when using the optimized historical forecasts
+    if isinstance(n.data_transformers, dict) and n.enable_optimization:
```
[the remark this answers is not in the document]
madtoinou: I reused the naming convention of the lags, but probably better to go with the name of the series directly.

darts/models/forecasting/forecasting_model.py
madtoinou: `fitted_values` is actually supported only by very models, I am not sure that it's relevant to add data transformer here... But going to do it, just in case

darts/utils/historical_forecasts/optimized_historical_forecasts_regression.py
madtoinou: I thought about it, decided to make it redundant in case some people want to call the optimized routine directly but I don't mind moving most of the logic to `historical_forecasts` and have only the inverse transform occur in the optimized methods.

dennisbader: I would say we can leave the optimized routines naive to transformations. Also, applying it on all series/hist fc output is more efficient due to parallelization :)

madtoinou:
<p>I added it to hide the warnings, but I don't mind removing it.</p>

---

 <a href='https://app.reviewnb.com/unit8co/darts/pull/2529/discussion/#comment-2470029056'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

madtoinou:
<p>Updated the notebook; two examples (one with, one without the auto-scaling) while minimizing the changes</p>

---

 <a href='https://app.reviewnb.com/unit8co/darts/pull/2529/discussion/#comment-2470029094'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -2630,6 +2648,342 @@ def test_probabilistic_optimized_hist_fc_regression(self, config):
...
+                    },
+                    {"series": Scaler(), "past_covariates": Scaler()},
+                ),
+            ],
+            [True, False],  # retrain
+            [True, False],  # last point only
+            models,
+        ),
+    )
+    def test_historical_forecasts_with_scaler(self, params):
```
dennisbader, line 2769: nice test 👍

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -112,6 +123,8 @@
 
     NB_EPOCH = 1
 
+    models += [NLinearModel, RNNModel]
```
dennisbader, line 126: one torch model should be enough

```suggestion
    models += [NLinearModel]
```

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -2630,6 +2648,342 @@ def test_probabilistic_optimized_hist_fc_regression(self, config):
...
+                columns=manual_hf_0.components,
+            )
+            self.helper_compare_hf(tmp_ts, hf_auto, last_points_only)
+        else:
+            self.helper_compare_hf(
+                hf_auto, [manual_hf_0, manual_hf_1], last_points_only
+            )
+
+    def test_historical_forecasts_with_scaler_errors(self):
+        """"""
```
dennisbader, line 2860: missing docstring

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -2630,6 +2648,342 @@ def test_probabilistic_optimized_hist_fc_regression(self, config):
...
+                data_transformers={"series": Scaler()},
+                retrain=False,
+            )
+        assert str(err.value).startswith(
+            "All the fittable entries in `data_transformers` must already be fitted when `retrain=False`."
+        )
+
+        # retrain=False and data transformers fitted on multplie series with fit_global=False
+        with pytest.raises(ValueError) as err:
+            tmp_scaler = Scaler(global_fit=False)
```
dennisbader, line 2886: we need to add support for this (see my previous comment in `darts/utils/historical_forecasts/utils.py`)

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -2630,6 +2648,342 @@ def test_probabilistic_optimized_hist_fc_regression(self, config):
...
+            )
+            assert len(opti_hf_auto) == len(hf_auto) == 2
+            # check that all the nested forecasts are identicals
+            for hf_list, opti_hf_list in zip(hf_auto, opti_hf_auto):
+                for fc_auto, fc_opti in zip(hf_list, opti_hf_list):
+                    assert fc_auto.time_index.equals(fc_opti.time_index)
+                    np.testing.assert_almost_equal(
+                        fc_auto.values(),
+                        fc_opti.values(),
+                    )
```
dennisbader, line 2985: here also

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -2630,6 +2648,342 @@ def test_probabilistic_optimized_hist_fc_regression(self, config):
...
+        manual_hf_1_1 = self.helper_manual_scaling_prediction(
+            model, {"series": series[1]}, hf_scaler, retrain, -ocl, ocl
+        )
+
+        # check that all the nested forecasts are identicals
+        for auto_hf_list, manual_hf_list in zip(
+            hf_auto, [[manual_hf_0_0, manual_hf_0_1], [manual_hf_1_0, manual_hf_1_1]]
+        ):
+            for fc_auto, fc_manual in zip(auto_hf_list, manual_hf_list):
+                assert fc_auto.time_index.equals(fc_manual.time_index)
```
dennisbader, line 2961: you could use your hfc compare helper from above

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 **Improved**
 
 - Improvements to `ForecastingModel`: Improved `start` handling for historical forecasts, backtest, residuals, and gridsearch. If `start` is not within the trainable / forecastable points, uses the closest valid start point that is a round multiple of `stride` ahead of start. Raises a ValueError, if no valid start point exists. This guarantees that all historical forecasts are `n * stride` points away from start, and will simplify many downstream tasks. [#2560](https://github.com/unit8co/darts/issues/2560) by [Dennis Bader](https://github.com/dennisbader).
+- Added `data_transformers` argument to `historical_forecasts`, `backtest` and `gridsearch` that allows scaling of the series without data-leakage. [#2529](https://github.com/unit8co/darts/pull/2529) by [Antoine Madrona](https://github.com/madtoinou) and [Jan Fidor](https://github.com/JanFidor)
```
dennisbader, line 15: `residuals()` support is missing

dennisbader (commented):
This looks really good @madtoinou 🚀 

I added a couple of suggestions, and two things which we should add:
- support data transformers in residuals
- support pre-trained data transformers regardless of `global_fit`

darts/utils/historical_forecasts/utils.py
```diff
@@ -193,6 +198,97 @@ def _historical_forecasts_general_checks(model, series, kwargs):
...
+                            f"When multiple series are provided, their number should match the number of "
+                            f"`TimeSeries` used to fit the data transformers `n={max(fitted_params_pipelines)}` "
+                            f"(only relevant for fittable transformers that use `global_fit=False`)."
+                        ),
+                        logger,
+                    )
+            else:
+                # at least one pipeline was fitted on several series with `global_fit=False` but only
+                # one series was passed
+                if max(fitted_params_pipelines) > 1:
```
dennisbader, line 286: ignore this when `ignore_warnings=True`

dennisbader (changes requested):
Beautiful 😍 I pushed some minor changes:
- adapt the docs a bit here and there
- renamed `idx_series` to `series_idx`
- converted `transform()` of fitted_values (in gridsearch) to `inverse_transform()` (should be inverse transform I think ;) )

There is one last change that we throw a warning instead of an error when `global_fit=True` and `len(series)>1`

dennisbader (approved):
Great work @madtoinou and @JanFidor 🚀 Thanks for all the work!

PR #2530, opened 2024-09-14
Feat/metrics quantiles

### LINKED ISSUE #2527: Compute multiple quantiles for Metrics
Add (multi) quantile support to all metrics. This comes in handy for conformal prediction and evaluation.

dennisbader:

Fixes #2527.

### Summary
- adds support for computing metrics on one or multiple quantiles `q`, either from probabilistic predictions or predicted quantiles.
- adds support for computing backtest and residuals on these quantile metrics
- added property `TimeSeries.shape` to get the shape of the time series
- adds first interval metrics: `miw` (Mean Interval Width, time aggregated) and `iw` (Interval Width, per time step / non-aggregated) which compute the width of quantile intervals `q_intervals` (expected to be a tuple or sequence of tuples with (lower quantile, upper quantile).


darts/metrics/metrics.py
```diff
@@ -241,35 +284,68 @@ def wrapper_multivariate_support(*args, **kwargs) -> METRIC_OUTPUT_TYPE:
...
+    stochastic_quantile
+        Optionally, a tuple with (quantile values, quantile values if `pred_series` is stochastic
+        and the quantile component names otherwise).
     """
+    if stochastic_quantile is not None:
+        q, q_names = stochastic_quantile
+    else:
+        q, q_names = None, None
+
     if vals.shape[2] == 1:  # deterministic
```
madtoinou, line 325: maybe leverage the `SMPL_AX` variable to make it clearer?

darts/metrics/metrics.py
```diff
@@ -241,35 +284,68 @@ def wrapper_multivariate_support(*args, **kwargs) -> METRIC_OUTPUT_TYPE:
...
+            # `q_names` are the component names of the predicted quantile parameters
+            # we extract the relevant quantile components with shape (times, components * quantiles)
+            vals = vals[:, vals_components.get_indexer(q_names)]
+            # rearrange into (times, components, quantiles)
+            vals = vals.reshape((len(vals), len(actual_components), -1))
+        return vals
+
+    # stochastic
+    if stochastic_quantile is None:
+        return vals
```
madtoinou, line 336: This condition is not clear, since it's technically an `else` of the `vals.shape[2] == 1` and actually directly return the `vals` input argument, I would rather move it at the very top? Before even unpacking the `stochastic_quantile` argument.

darts/metrics/metrics.py
```diff
@@ -241,35 +284,68 @@ def wrapper_multivariate_support(*args, **kwargs) -> METRIC_OUTPUT_TYPE:
...
+
+    # compute multiple quantiles for all times and components; with shape: (quantiles, times, components)
+    out = np.quantile(vals, q, axis=2)
```
madtoinou, line 339: Again, I would maybe use the `SMPL_AX` constant.

darts/metrics/metrics.py
```diff
@@ -2816,22 +3000,14 @@ def ql(
...
-        )
-
     y_true, y_pred = _get_values_or_raise(
         actual_series,
         pred_series,
         intersect,
         stochastic_quantile=q,
         remove_nan_union=True,
     )
+    q, _ = q
```
madtoinou, line 3010: This line works only when `q` is passed as a tuple, a float or list of float.

darts/metrics/metrics.py
```diff
@@ -2873,11 +3049,11 @@ def mql(
         The (sequence of) actual series.
     pred_series
         The (sequence of) predicted series.
-    q
-        The quantile (float [0, 1]) of interest for the loss.
     intersect
         For time series that are overlapping in time without having the same time index, setting `True`
         will consider the values only over their common time interval (intersection in time).
+    q
+        The quantile (float [0, 1]) or list of quantiles of interest to compute the metric on.
```
madtoinou, line 3056: does not mention the option of passing a tuple.

darts/models/forecasting/forecasting_model.py
```diff
@@ -1312,6 +1318,12 @@ def backtest(
             Whether to print progress.
         show_warnings
             Whether to show warnings related to parameters `start`, and `train_length`.
+        predict_likelihood_parameters
+            If set to `True`, the model predict the parameters of its Likelihood parameters instead of the target. Only
+            supported for probabilistic models with `likelihood="quantile"`, `num_samples = 1` and
+            `n<=output_chunk_length`. Default: ``False``.
+        enable_optimization
+            Whether to use the optimized version of historical_forecasts when supported and available.
```
madtoinou, line 1326: Indicate that the default is `True`?

darts/models/forecasting/forecasting_model.py
```diff
@@ -1491,22 +1505,24 @@ def __getitem__(self, index) -> TimeSeries:
...
 
             if len(metric) == 1:
                 # shape `(n metrics, *)` -> `(*,)`
                 errors_series = errors_series[0]
-            elif not last_points_only and reduction is None:
+            else:
                 # shape `(n metrics, *)` -> `(*, n metrics)`
-                errors_series = errors_series.T
+                errors_series = errors_series.transpose(
+                    tuple(i for i in range(1, errors_series.ndim)) + (0,)
```
madtoinou, line 1524: Why would you concatenate a tuple containing just `0` to the error series?

darts/models/forecasting/forecasting_model.py
```diff
@@ -2051,13 +2083,37 @@ def residuals(
...
                 if len(res.shape) != 3:
-                    res = np.reshape(res, (len(fc), fc.n_components, 1))
-                res_list_out.append(res if values_only else fc.with_values(res))
+                    res = np.reshape(res, (len(fc), n_comp_out, 1))
+                if (
+                    values_only
+                    or q is None
+                    or (len(q) == 1 and res.shape[1] == fc.shape[1])
+                ):
+                    res = res if values_only else fc.with_values(res)
```
madtoinou, line 2105: for clarity, let's treat the `value_only` case separately from the other two.

darts/tests/metrics/test_metrics.py
```diff
@@ -1539,3 +1568,265 @@ def helper_test_non_aggregate(self, metric, is_aggregate, val_exp=None):
...
+        """Test output types and shapes for time aggregated metrics:
+        for single and multiple univariate or multivariate series, in combination
+        with different component and series reduction functions."""
+        np.random.seed(42)
+        metric, is_univar, is_single = config
+        params = inspect.signature(metric).parameters
+
+        n_comp = 1 if is_univar else 2
+
+        qs_all = [0.1, 0.5, 0.8]
```
madtoinou, line 1620: Is is intentional to have "asymmetric" quantiles (aka having 0.1 and 0.8 instead of 0.1 with 0.9 or 0.2 with 0.8)?

darts/tests/models/forecasting/test_backtesting.py
```diff
@@ -1252,6 +1253,154 @@ def time_reduced_metric(*args, **kwargs):
...
+            historical_forecasts=hfc,
+            metric=metric,
+            last_points_only=lpo,
+            reduction=None,
+            metric_kwargs=metric_kwargs,
+        )
+        assert isinstance(bts, list) and len(bts) == 2
+        if lpo:
+            bts = [[bt] for bt in bts]
+        # `ae` with time and component reduction is equal to `mae` with component reduction
```
madtoinou, line 1306: Nice!

darts/tests/models/forecasting/test_residuals.py
```diff
@@ -612,3 +619,304 @@ def test_sample_weight(self, config):
...
+    @pytest.mark.parametrize(
+        "config",
+        itertools.product(
+            [True, False],  # last_points_only
+            [False, True],  # from stochastic predictions (or predicted quantiles)
+        ),
+    )
+    def test_quantiles_metrics(self, config):
+        """Tests backtest with different metric_kwargs based on historical forecasts generated on a sequence
+        `series` with last_points_only=False"""
```
madtoinou, line 632: The comment is contradictory with the parameters in the product, where lpo can take either the `True` or `False` value.

darts/tests/models/forecasting/test_residuals.py
```diff
@@ -612,3 +619,304 @@ def test_sample_weight(self, config):
...
+    @pytest.mark.parametrize(
+        "config",
+        itertools.product(
+            [True, False],  # last_points_only
+            [False, True],  # from stochastic predictions (or predicted quantiles)
+        ),
+    )
+    def test_metric_quantiles_lpo_false(self, config):
+        """Tests backtest with different metric_kwargs based on historical forecasts generated on a sequence
+        `series` with last_points_only=False"""
```
madtoinou, line 726: Same comment as above, contradiction between the values taken by `lpo` and the docstring/name of the test.

darts/tests/models/forecasting/test_residuals.py
```diff
@@ -612,3 +619,304 @@ def test_sample_weight(self, config):
...
+                assert bt.shape[:2] == shape_expected
+                np.testing.assert_array_almost_equal(bt[:, :, 0], bt_expected)
+
+    @pytest.mark.parametrize(
+        "lpo",
+        [True, False],  # last_points_only
+    )
+    def test_quantiles_from_model(self, lpo):
+        """Tests backtest with different metric_kwargs based on historical forecasts generated on a sequence
+        `series` with last_points_only=False"""
```
madtoinou, line 817: same as above.

darts/utils/utils.py
```diff
@@ -67,6 +67,34 @@ class ModelMode(Enum):
 }
 
 
+def likelihood_component_names(
```
madtoinou, line 70: Let's add a docstring

darts/metrics/metrics.py
```diff
@@ -60,12 +68,7 @@ def wrapper_multi_ts_support(*args, **kwargs):
 
         params = signature(func).parameters
         n_jobs = kwargs.pop("n_jobs", params["n_jobs"].default)
-        if not isinstance(n_jobs, int):
-            raise_log(ValueError("n_jobs must be an integer"), logger=logger)
-
         verbose = kwargs.pop("verbose", params["verbose"].default)
-        if not isinstance(verbose, bool):
-            raise_log(ValueError("verbose must be a bool"), logger=logger)
```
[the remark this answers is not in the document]
dennisbader: Looked at codecov, we don't test for this, and through type hinting it's given

darts/metrics/metrics.py
dennisbader: I didn't want to mention this in the API because it's what the wrappers gives, and maybe too complicated for the users.
I could add it in the description, but I feel like they should only use float, list[float]

darts/models/forecasting/forecasting_model.py
dennisbader: this translates to `errors_series.transpose(1, 2, 0)` for a 3D array. That should be correct, no?

darts/tests/metrics/test_metrics.py
dennisbader: Yes, in theory you can compute it on any quantile.

darts/metrics/metrics.py
```diff
@@ -212,15 +253,16 @@ def wrapper_multivariate_support(*args, **kwargs) -> METRIC_OUTPUT_TYPE:
             num_series_in_args += 1
 
         vals = func(*input_series, *args[num_series_in_args:], **kwargs)
-        if not 1 <= len(vals.shape) <= 2:
+        # bring vals to shape (n_time, n_comp, n_quantile)
+        if not 2 <= len(vals.shape) <= 3:
```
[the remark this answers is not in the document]
dennisbader: Good catch 👍 updated the message

darts/metrics/metrics.py
```diff
@@ -212,15 +253,16 @@ def wrapper_multivariate_support(*args, **kwargs) -> METRIC_OUTPUT_TYPE:
...
             raise_log(
                 ValueError(
                     "Metric output must have 1 dimension for aggregated metrics (e.g. `mae()`, ...), "
                     "or 2 dimension for time dependent metrics (e.g. `ae()`, ...)"
                 ),
                 logger=logger,
             )
-        elif len(vals.shape) == 1:
+        elif len(vals.shape) == 2:
```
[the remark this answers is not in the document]
dennisbader: We could assume that the if it's 1dim then it must be the components axis. But I wonder it's better to raise more errors here to make sure we actually get the expected shapes 🤔

darts/metrics/metrics.py
dennisbader: Good one, makes it more simple 👍

madtoinou (approved):

PR #2544, opened 2024-09-27
Feat/show_anomalies for multivariate

### LINKED ISSUE #2114: show_anomalies for multivariate
Hi, I've experimented with the anomaly detection example, and it has proven to be exceptionally helpful in saving me a significant amount of time while establishing the anomaly detection test. I do have a question: 

I am currently engaged in **multivariate forecasting** and wish to conduct **anomaly detection on individual components**. To achieve this, I have employed `component_wise=True` for the scorer, and it has been effective. However, when attempting to utilize `show_anomalies` (which gives visually appealing result figures), all the results are put into a single figure, making it challenging to discern individual results. I have been unable to identify a method to plot each variable separately. Could you provide guidance on this matter?Thanks！

cnhwl:

Fixes #2114.

### Summary
I determine whether this feature is enabled by adding parameter `multivariate_plot: bool = False` to `show_anomalies()`, which is implemented in the `show_anomalies_from_scores` function. 

My general idea is to iterate through the components in the series and separately plot each component (including series, pred_series, pred_scores and anomalies). The following is a simple example, with the output shown below:

```
import numpy as np
import pandas as pd

from darts import TimeSeries
from darts.ad.utils import (
    eval_metric_from_scores,
    show_anomalies_from_scores,
)
from darts.ad import (
    ForecastingAnomalyModel,
    NormScorer,
    WassersteinScorer,
)

def generate_data_ex1(random_state: int):
    np.random.seed(random_state)

    # create the train set using standard normal distribution
    comp1 = np.expand_dims(np.random.normal(loc=0, scale=1, size=200), axis=1)
    comp2 = np.expand_dims(np.random.normal(loc=0, scale=1, size=200), axis=1)
    
    # Calculate means and standard deviations
    mean1, std1 = np.mean(comp1), np.std(comp1)
    mean2, std2 = np.mean(comp2), np.std(comp2)
    
    # Identify anomalies
    anomalies1 = (comp1 > mean1 + 2 * std1).astype(int)
    anomalies2 = (comp2 > mean2 + 2 * std2).astype(int)
    
    # Concatenate anomalies
    anomalies = np.concatenate([anomalies1, anomalies2], axis=1)
    
    # Concatenate the original values
    vals = np.concatenate([comp1, comp2], axis=1)
    
    return vals, anomalies

# Example usage
data, anomalies = generate_data_ex1(random_state=42)

series = TimeSeries.from_values(data, columns=["comp1", "comp2"])
series_train = series[:120]
series_test = series[120:]
anomalies_series = TimeSeries.from_values(anomalies, columns=["comp1_anomalies", "comp2_anomalies"])
anomalies_series_test = anomalies_series[120:]

anomaly_model = ForecastingAnomalyModel(
    model=RegressionModel(lags=10),
    scorer=[
        NormScorer(component_wise=True),
        WassersteinScorer(component_wise=True)
    ],
)

anomaly_model.fit(series_train, allow_model_training=True, verbose=True)

anomaly_model.show_anomalies(
    series=series_test,
    anomalies=anomalies_series_test,
    multivariate_plot=True
)
```

output

I would appreciate more input on how to further improve this feature, thank you very much!


darts/ad/anomaly_model/anomaly_model.py
```diff
@@ -247,6 +247,7 @@ def show_anomalies(
         names_of_scorers: Union[str, Sequence[str]] = None,
         title: str = None,
         metric: Optional[Literal["AUC_ROC", "AUC_PR"]] = None,
+        multivariate_plot: bool = False,
```
madtoinou, line 250: Please update the docstring to include a description of what this argument is doing

darts/ad/anomaly_model/forecasting_am.py
```diff
@@ -440,6 +440,7 @@ def show_anomalies(
         names_of_scorers: Union[str, Sequence[str]] = None,
         title: str = None,
         metric: Optional[Literal["AUC_ROC", "AUC_PR"]] = None,
+        multivariate_plot: bool = False,
```
madtoinou, line 443: Here as well, an entry in the docstring would be great (should be identical to the one in anomaly model)

darts/ad/scorers/scorers.py
```diff
@@ -176,6 +176,7 @@ def show_anomalies_from_prediction(
         anomalies: TimeSeries = None,
         title: str = None,
         metric: Optional[Literal["AUC_ROC", "AUC_PR"]] = None,
+        multivariate_plot: bool = False,
```
madtoinou, line 179: Update the docstring

darts/ad/utils.py
```diff
@@ -422,105 +425,259 @@ def show_anomalies_from_scores(
...
         nbr_plots = nbr_plots + len(set(window))
 
-    fig, axs = plt.subplots(
-        nbr_plots,
-        figsize=(8, 4 + 2 * (nbr_plots - 1)),
-        sharex=True,
-        gridspec_kw={"height_ratios": [2] + [1] * (nbr_plots - 1)},
-        squeeze=False,
-    )
+    if multivariate_plot:
```
madtoinou, line 428: I think that it would be great to refactor the logic to make it more concise and avoid code duplication; you could define an helper function with the "old" logic, potentially adding some features (for the axis/title naming for example) that `show_anomalies_from_scores()` would call in slightly different ways depending on the value of `multivariate_plot`.

madtoinou (changes requested):
Thank you for the contribution @cnhwl, sorry for making you wait so much for a review.

The documentation of the new argument is missing, also, some refactoring of the code to reduce code duplication would be great. Other than that, it looks great!

cnhwl:
Thank you very much for your code review and guidance! @madtoinou
I've resolved the code changes and am ready to merge anytime now! 🚀

darts/ad/utils.py
```diff
@@ -311,6 +311,7 @@ def show_anomalies_from_scores(
     names_of_scorers: Union[str, Sequence[str]] = None,
     title: str = None,
     metric: Optional[Literal["AUC_ROC", "AUC_PR"]] = None,
+    multivariate_plot: bool = False,
```
madtoinou, line 314: Docstring for `multivariate_plot` is missing

darts/ad/utils.py
```diff
@@ -422,105 +425,92 @@ def show_anomalies_from_scores(
...
-    _plot_series(series=series, ax_id=axs[index_ax][0], linewidth=0.5, label_name="")
-
-    if pred_series is not None:
-        _plot_series(
-            series=pred_series,
-            ax_id=axs[index_ax][0],
-            linewidth=0.5,
-            label_name="model output",
+    if multivariate_plot:
+        series = _check_input(
```
madtoinou, line 429: This is redundant with the check line 358, let's call `_check_input` only once

darts/ad/utils.py
```diff
@@ -422,105 +425,92 @@ def show_anomalies_from_scores(
...
+        if pred_scores is not None:
+            for pred_score in pred_scores:
+                pred_score = _check_input(
+                    pred_score,
+                    name="pred_score",
+                    width_expected=series.width,
+                    check_multivariate=True,
+                )[0]
+
+        series_width = series.n_components
```
madtoinou, line 461: This line is a duplicate of line 434

darts/ad/utils.py
```diff
@@ -422,105 +425,92 @@ def show_anomalies_from_scores(
...
+        if pred_series is not None:
+            pred_series = _check_input(
+                pred_series,
+                name="pred_series",
+                width_expected=series.width,
+                check_multivariate=True,
+            )[0]
+
+        if anomalies is not None:
+            anomalies = _check_input(
```
madtoinou, line 445: If this check is performed, we should make sure this series contains binary values

darts/ad/utils.py
```diff
@@ -422,105 +425,92 @@ def show_anomalies_from_scores(
...
-            label_name="anomalies",
-            color="red",
+        fig.suptitle(title)
+    else:
+        fig, axs = plt.subplots(
+            nbr_plots,
+            figsize=(8, 4 + 2 * (nbr_plots - 1)),
+            sharex=True,
+            gridspec_kw={"height_ratios": [2] + [1] * (nbr_plots - 1)},
+            squeeze=False,
```
madtoinou, line 496: Let's avoid having this snippet twice and call it before treating the two different cases by including `multivariate_plot` in the logic;

```python
plots_per_ts = nbr_plots * series_width if multivariate_series else nbr_plots
fig, axs = plt.subplots(
            plots_per_ts,
            figsize=(8, 4 + 2 * (plots_per_ts - 1)),
            sharex=True,
            gridspec_kw={"height_ratios": [2] + [1] * (plots_per_ts - 1)},
            squeeze=False,
)
```

darts/ad/utils.py
```diff
@@ -422,105 +425,92 @@ def show_anomalies_from_scores(
...
+                else None,
+                pred_series=pred_series[pred_series.components[i]]
+                if pred_series is not None
+                else None,
+                pred_scores=pred_scores,
+                window=window,
+                names_of_scorers=names_of_scorers,
+                metric=metric,
+                axs=axs,
+                index_ax=index_ax,
```
madtoinou, line 485:
```suggestion
                index_ax=i * nbr_plots,
```

madtoinou (changes requested):
Thanks you a lot for the changes @cnhwl , I still have some improvement to suggest

cnhwl:
Thanks again for your help! @madtoinou 🤝

darts/ad/utils.py
```diff
@@ -421,106 +434,77 @@ def show_anomalies_from_scores(
...
+            )[0]
+
+        if pred_scores is not None:
+            for pred_score in pred_scores:
+                pred_score = _check_input(
+                    pred_score,
+                    name="pred_score",
+                    width_expected=series.width,
+                    check_multivariate=True,
+                )[0]
```
madtoinou, line 472: these checks could actually also be conducted when `multivariate_plot=False`, and also check `num_series_expected` while we are at it.

darts/ad/utils.py
```diff
@@ -421,106 +434,77 @@ def show_anomalies_from_scores(
             )
 
         nbr_plots = nbr_plots + len(set(window))
```
madtoinou, line 436:
```suggestion
        nbr_plots += len(set(window))
```

madtoinou (changes requested):
Last changes request, after this, this PR should be ready to be reviewed by @dennisbader 🚀

cnhwl:
Happy New Year! @madtoinou @dennisbader 
Please recheck the code and feel free to request any changes. 🤝

madtoinou (commented):
Thanks for addressing the comments! LGTM, let's wait for dennis' opinion :D

dennisbader (commented):
Thanks @cnhwl for the PR which looks nice, and @madtoinou for the review.

I saw one or two issues where I took the liberty to change some things:

- fix issue where fig, axes were not defined in case `pred_scores=None`
- fix issue which raised an error when input `series` was multivariate, but anomalies and scorers are `component_wise=False`

I'm not sure whether `plot_multivariate` is the best name. To me it sounds like it should plot all components on one plot if `True` (but it's the opposite). Should we rather use `component_wise: bool = False` (dafault False), as we define it also for the scorers?

My last concern is about plotting all the components separately in one figure. This can get quite big for larger numbers of components. But maybe it's a bit an overkill to improve this.

At least I find that we should improve the spacing a bit between suptitle and axes. 
The box for the suptitle should be fixed regardless of the number of components, whereas now it is scaled with the number of plots (see my comment).

Let me know what you think, happy to discuss further :)

darts/ad/utils.py
```diff
@@ -420,106 +424,81 @@ def show_anomalies_from_scores(
...
+                num_series_expected=1,
+                check_multivariate=multivariate_plot,
+            )[0]
+
+    plots_per_ts = nbr_plots * series_width if multivariate_plot else nbr_plots
     fig, axs = plt.subplots(
-        nbr_plots,
-        figsize=(8, 4 + 2 * (nbr_plots - 1)),
+        plots_per_ts,
+        figsize=(8, 4 + 2 * (plots_per_ts - 1)),
```
dennisbader, line 462: total figsize is scaled here by the number of axes / components. This results in the suptitle having more and more space with larger number of components (if `multivariate_plot=True`).

Can we improve the spacing a bit? See for example the figure in the PR description, where the suptitle covers a lot of space.

cnhwl: I use `plt.subplots(constrained_layout=True)` to improve the spacing between suptitle and axes. Please let me know if there is a better way to do this!

dennisbader (approved):
@cnhwl, thanks for the updates. I made some last changes to address the figure size issue and some changes to the documentation.

Thanks again for the contribution and this nice PR 🚀 Always a pleasure to work together!

madtoinou (approved):

Issue #2551, opened 2024-10-02
[Question] deprecated mean_squared_error


quant12345:
Hello everyone!

I would like to correct the following warning: FutureWarning: `'squared' is deprecated in version 1.4 and will be removed in 1.6. To calculate the root mean squared error, use the function'root_mean_squared_error'.`

 In the tests build (3.8, all) the version is scikit-learn-1.3.2. 'root_mean_squared_error' appeared in scikit-learn-1.4.

Question: Do you plan to upgrade scikit-learn to 1.4(in builds 3.8)?
Code processing depends on this.

madtoinou:
Hi @quant12345,

Since Python 3.8 is reaching its end of support at the end of the month, we are planning on doing one last release for it. Once this is done, we will be able to focus on things such as fixing some warning/deprecation messages, remove the version capping on numpy and simplify the typing imports/synthax.

If you want to contribute to darts, I would recommend checking the issues labeled with the `good first issue` or `pr welcomed` :)

quant12345:
Considering that after scikit-learn-1.3.2 the next version is 1.4.0, it seems to me that it would be more appropriate to try to switch to scikit-learn-1.4.0 (if possible).

Of course, I can do scikit-learn version processing and depending on the version choose `mean_squared_error, {"squared": False}` or `root_mean_squared_error`. But then if you switch to 1.4.0, there will be unnecessary lines of code with version checking that will need to be removed (that is, changing the code again).

madtoinou:
Hi @quant12345,

Since there is no cap on scikit-learn's version as you can see [here](https://github.com/unit8co/darts/blob/master/requirements/core.txt), you are actually free to use another version. I just checked the environment that I am using to work with darts and it's actually relying on scikit-learn 1.4.1.

You can safely ignore the warning, we will make sure to solve it before the release of scikit-learn 1.6.

quant12345:
I looked in scikit-learn 1.4.0 supports Python versions 3.9 to 3.12. I locally replaced `"squared": False` from `mean_squared_error `to `root_mean_squared_error` and the tests pass without warning:
```
pytest darts/tests/metrics/test_metrics.py::TestMetrics::test_metrics_deterministic[config10]
pytest darts/tests/metrics/test_metrics.py::TestMetrics::test_metrics_deterministic[config17]
```
Once you switch to 3.9, I'm ready to do a PR.

PR #2554, opened 2024-10-04
replacing deprecated "squared": False from mean_squared_error

quant12345:

Removes warning:`FutureWarning: 'squared' is deprecated in version 1.4 and will be removed in 1.6. To calculate the root mean squared error, use the function'root_mean_squared_error'.`

Discussion #2551

I didn't do versioning to choose [mean_squared_error](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html) ` {"squared": False} `or `root_mean_squared_error`, depending on which version of `scikit-learn`.

As far as I understand, you are stopping support for Python 3.8.
The next version of `scikit-learn `after `1.3.2 ` will `1.4.0` (where [root_mean_squared_error](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.root_mean_squared_error.html#sklearn.metrics.root_mean_squared_error) appeared) and version `1.4.0` supports Python versions starting from `3.9`.

**That is, tests will fail on Python 3.8, since for scikit-learn >= 1.4.0, Python >= 3.9 is needed.**
But you can just take your time with this PR)

### Summary

Replacing deprecated `"squared": False` from `mean_squared_error`, instead it is used `root_mean_squared_error.`


quant12345:
In Actions `01-multi-time-series-and-covariates.ipynb` it gives: DatasetLoadingException: `Could not download the dataset. Reason:BadZipFile('File is not a zip file').`

Probably some failure in loading?

Issue #2510, opened 2024-08-25
[FEAT] Add tsmixer-basic


eschibli:
tsmixer was original reported as two different models, tsmixer-basic (which allows for past covariates and is called simply tsmixer in the paper) and tsmixer-ext, which allows for past, future, and static covariates. All results in the paper except for the m5 dataset used tsmixer-basic. The darts implementation is based on tsmixer-ext. 

However, tsmixer-ext isn't identical to tsmixer-basic when there are no static or future covariates. The key difference is that while tsmixer-basic projects to `output_chunk_length` in the final layer, effectively encoding the historical data while maintaining it's time dimension, tsmixer-ext projects the historical and static data to `output_chunk_length` in the first layer. I don't think this is optimal as this will limit the usefulness of the residual connections. My testing with the original google-research source code shows a performance regression of about two points of MAE and MSE with the `weather` dataset when moving the temporal project step to the top of the model. 

If the maintainers think this would be valuable I can impliment this. I think the most sensible way to do so would be to add a `project_first=True` keyword.

madtoinou:
Hi @eschibli,

If you think that you can elegantly make the `tsmixer-basic` architecture easily available through the `TSMixerModel` API/constructor, it would be for sure valuable to have a variation of this model that performs better when no future covariates are available, which can occur in many situations. I would maybe just call the argument `first_layer_projection` instead of just `project_first`, but we can discuss it in your PR.

You will also need to add checks in the `fit()` method so that an error is raised if `first_layer_projection=False` and future/static covariates are provided.

eschibli:
Hi @madtoinou - in my work I have seen significantly better performance in problems with both past and future covariates using an encoder-decoder model, projecting in the middle of the backbone so that both the past and future values are mixed in their original time space. This was not reported in the original paper but seems like a logical extension, which reduces to both original models with `project_after=0` and `project_after=num_blocks`. 

ie, 

```diff
class TSMixerModel(MixedCovariatesTorchModel):
    def __init__(
        self,
        ...,
+       project_after=0
        )

class TSMixeModule(...)
...

-self.conditional_mixer = ...

+self.encoder_mixer = self._build_mixer(
+    sequence_length=self.input_chunk_length,
+    num_blocks=project_after,
+    future_cov_dim=0
+    hidden_size=hidden_size,
+    static_cov_dim=static_cov_dim,
+    **mixer_params
+    )


+self.decoder_mixer = self._build_mixer(
[...]

PR #2555, opened 2024-10-08
Imp/tsmixer basic

eschibli:

Implements [#2510](https://github.com/unit8co/darts/issues/2510)

### Summary

Adds the option to project to the output temporal space at the end of TS-Mixer, rather than the beginning. This was how most of the results in the original google-research paper were achieved (ie, the architecture in [Fig #1 of the paper](https://arxiv.org/pdf/2303.06053)). This may allow higher performance in cases where past covariates are important by allowing a more direct series of residual connections along the input time dimension.

I allowed support for future covariates by instead projecting them into the lookback temporal space, but this probably won't perform well in cases where they are more important than the historical targets and past covariates. 

### Other Information

The original paper and source code do not clarify whether the final temporal projection should go before or after the final feature projection as they hardcoded `hidden_size` to `output_dim` and therefore did not have need a final feature projection. I erred on the side of putting the temporal projection  first, as otherwise the common  `output_dim==1` could lead to unexpected, catastrophic compression before the temporal projection step.


eschibli:
> Hi @eschibli,
> ...

Thanks @madtoinou. I was not able to get Gradle running on my machine and didn't realize ruff was that easy to set up so sorry for spamming your test pipeline. 

I don't believe the failing mac build is a result of my changes so it should be good for review now.

dennisbader:
Hi @eschibli, thanks for the PR. Yes, the failing mac tests are unrelated to your PR, we're working on it :). 
Also, give us some time to review, our capacity is currently a bit limited 🙏

eschibli:
Understood Dennis

darts/tests/models/forecasting/test_tsmixer.py
```diff
@@ -362,3 +362,25 @@ def test_time_batch_norm_2d_gradients(self):
...
+            # Cover case of projecting future covs back to input dims
+            add_encoders={"cyclic": {"future": "hour"}},
+            **tfm_kwargs,
+        )
+        model.fit(ts)
+
+        if project_first_layer:
+            assert model.model.sequence_length == output_len
+        else:
+            assert model.model.sequence_length == input_len
```
madtoinou, line 386: Can the test also include a call to `predict()` to make sure it works as well (even if the forward pass is already occurring in the call to `fit()`)?

madtoinou (commented):
It looks great, thank you for this nice PR @eschibli! Some minor comment about the order of the operations/projections to make the flow more intuitive.

Could you also extend the [TSMixer notebook](https://github.com/unit8co/darts/blob/master/examples/21-TSMixer-examples.ipynb) to include a section where the difference in performance with "project_first_layer=True/False"  and future covariates can be visualized?

madtoinou:
@eschibli, do you think that you would have the time to apply the modifications mentioned above so that we can get closer to merge this in `main`?

eschibli:
Hi @madtoinou:  Sorry I was very busy in December and January. I should have some time this week or next week to finish this. 

Also, in my own work I have had more success with an intermediate variant, which I mentioned in the original feature request thread. If we would consider adding that additional functionality I will add it to this pull request.

madtoinou:
Sure! We will then try to compare each one of them on small use-cases, to make sure that they bring something but based on your experience, it seems to already be the case :)

darts/models/forecasting/tsmixer_model.py
```diff
@@ -491,31 +513,62 @@ def forward(
...
+            x = _time_to_feature(x)
+            # linear transformations to SL (T in this case)
+            # (B, H, L) -> (B, H, SL)
+            x = self.fc_hist(x)
+            # Transpose back
+            # (B, H, T) -> (B, T, H)
+            x = _time_to_feature(x)
+
+        # Otherwise, encoder-style model with residual blocks in input time dimension
+        # In the original paper this was not implimented for future covariates,
```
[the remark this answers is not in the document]
eschibli: Spelling resolved everywhere.

darts/models/forecasting/tsmixer_model.py
```diff
@@ -491,31 +513,62 @@ def forward(
...
+        # but rather than ignoring them or raising an error we remap them to the input time dimension.
+        # Suboptimal but may be useful in some cases.
+        elif self.future_cov_dim:
```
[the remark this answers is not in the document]
eschibli: This implementation is out-of-date with the current version of the PR, and I think the current version makes more sense.

darts/models/forecasting/tsmixer_model.py
```diff
@@ -491,31 +513,62 @@ def forward(
...
+            # repeat to match time dim: (B, 1, C * S) -> (B, SL, C * S)
+            x_static = x_static.repeat(1, self.sequence_length, 1)
 
         for mixing_layer in self.conditional_mixer:
-            # conditional mixer layers with static covariates (B, T, 2 * H_S), (B, T, C * S) -> (B, T, H_S)
+            # conditional mixer layers with static covariates (B, SL, 2 * H_S), (B, SL, C * S) -> (B, SL, H_S)
             x = mixing_layer(x, x_static=x_static)
 
+        # If we are in the input time dimension, we need to project to the output time dimension.
+        # The original paper did not a fc_out layer (as hidden_size == output_dim)
```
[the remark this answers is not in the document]
eschibli: Fixed I believe

darts/models/forecasting/tsmixer_model.py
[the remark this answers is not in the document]
eschibli: I set it up this way originally because the TSMixer performance test was failing if I didn't generate the layers in exactly the same order. In this PR I have relaxed the tolerance as it wasn't stable.

darts/tests/models/forecasting/test_tsmixer.py
eschibli: Done.

eschibli:
@madtoinou  are you happy with these changes?

eschibli:
@madtoinou  @dennisbader  sorry to nag but could you please clarify if you are unhappy with this addition or if more changes are needed to approve this?

dennisbader:
@madtoinou could you give an update on this one? :)

madtoinou:
Hi @eschibli,

Sorry for the delay, I started reviewing but it took me longer than expected.

These changes are welcomed, I have the impression that it will bring considerable performance gain but I want to make sure that the current behavior is still accessible and reproducible.

eschibli:
Sorry to nag but is there anything I could do to get this merged? It should reproduce current behavior when `project_after` isn't passed a positive integer -- though it's not trivial to prove this to be exactly as the changes @madtoinou suggested result in initializing weights in a different order -- and it allows the other architecture in the paper to be used when `project_after` is passed `num_blocks`. In my work I've found better results with an intermediate value when past and future covs are important but I can't share those unfortunately.

dennisbader:
Hi @eschibli and really sorry that this got a bit lost. We'll review it until next Friday (most likely on next Friday).

darts/models/forecasting/tsmixer_model.py
```diff
@@ -397,13 +402,17 @@ def __init__(
             "normalize_before": normalize_before,
         }
 
+        # Projects from the input time dimension to the output time dimension
```
dennisbader, line 405: `self.fc_hist` is not used anymore (I guess replaced by the new `encoder_to_decoder`?). We should remove one of the two

darts/models/forecasting/tsmixer_model.py
```diff
@@ -570,8 +616,10 @@ def __init__(
             The hidden state size / size of the second feed-forward layer in the feature mixing MLP.
         ff_size
             The size of the first feed-forward layer in the feature mixing MLP.
-        num_blocks
-            The number of mixer blocks in the model. The number includes the first block and all subsequent blocks.
+        num_encoder_blocks
+            The number of mixer blocks in the encoder.
+        num_decoder_blocks
+            The number of mixer blocks in the decoder.
```
dennisbader, line 622: These are not available, and description of `project_after_n_blocks` is missing

```suggestion
        num_blocks
            The number of mixer blocks in the model. The number includes the first block and all subsequent blocks.
```

darts/models/forecasting/tsmixer_model.py
```diff
@@ -793,6 +842,22 @@ def _create_model(self, train_sample: TorchTrainingSample) -> nn.Module:
...
+            num_encoder_blocks < 0
+            or num_decoder_blocks < 0
+            or (num_encoder_blocks + num_decoder_blocks != self.num_blocks)
+        ):
+            raise_log(
+                ValueError(
+                    f"Invalid number of encoder and decoder blocks. "
+                    f"project_after_n_blocks must be between 0 and {self.num_blocks} inclusive."
+                ),
+            )
```
dennisbader, line 859: We should move this sanity check to `__init__`, simply check that `0 <= project_after_n_blocks <= num_blocks`, and rephrase the message to 

```
f"`project_after_n_blocks` must be between 0 and {num_blocks} inclusive."`
```

darts/models/forecasting/tsmixer_model.py
```diff
@@ -478,39 +509,53 @@ def forward(self, x_in: PLModuleInput) -> torch.Tensor:
...
+        if self.static_cov_dim:
+            # (B, C, S) -> (B, 1, C * S)
+            x_static_hist = x_static.reshape(x_static.shape[0], 1, -1)
+            # repeat to match lookback time dim: (B, 1, C * S) -> (B, L, C * S)
+            x_static_hist = x_static_hist.repeat(1, self.input_chunk_length, 1)
+
+            # (B, C, S) -> (B, 1, C * S)
+            x_static_future = x_static.reshape(x_static.shape[0], 1, -1)
+            # repeat to match horizon time dim: (B, 1, C * S) -> (B, T, C * S)
+            x_static_future = x_static_future.repeat(1, self.output_chunk_length, 1)
```
dennisbader, line 526: This can be simplified. Also, it would be nice to only create x_static_hist / x_static_future if they are really required (e.g. future only if the encoder is not used).

Any type of operation that can be avoided has positive impact on the model throughput :)

```suggestion
            # (B, C, S) -> (B, 1, C * S)
            x_static = x_static.reshape(x_static.shape[0], 1, -1)
            # repeat to match lookback time dim: (B, 1, C * S) -> (B, L, C * S)
            x_static_hist = x_static.repeat(1, self.input_chunk_length, 1)
            # repeat to match horizon time dim: (B, 1, C * S) -> (B, T, C * S)
            x_static_future = x_static.repeat(1, self.output_chunk_length, 1)
```

darts/models/forecasting/tsmixer_model.py
```diff
@@ -478,39 +509,53 @@ def forward(self, x_in: PLModuleInput) -> torch.Tensor:
...
-            x_static = x_static.reshape(x_static.shape[0], 1, -1)
-            # repeat to match horizon (B, 1, C * S) -> (B, T, C * S)
-            x_static = x_static.repeat(1, self.output_chunk_length, 1)
+        # Apply encoder mixer layers
+        for layer in self.encoder_mixer:
+            x = layer(x, x_static_hist)
+
+        # Project time dimension (B, L, H_S) -> (B, T, H_S)
+        x = x.transpose(1, 2)
+        x = self.encoder_to_decoder(x)  # Linear map
```
dennisbader, line 544: Hmm.. The `encoder_to_decoder` (previously called `fc_hist`) is now applied after `feature_mixing_hist`.

This means `feature_mixing_hist` is only applied on the input chunk (L), instead of on the output chunk (T). From Figure 4 in the paper, the temporal projection for historic features is applied also before the `feature_mixing_hist`.

Meaning, users will probably not be able to reproduce results from earlier Darts versions.

darts/models/forecasting/tsmixer_model.py
```diff
@@ -478,39 +509,53 @@ def forward(self, x_in: PLModuleInput) -> torch.Tensor:
...
+
+        # If future covariates are provided, mix and concatenate them with the encoder output
+        if self.future_cov_dim:
+            # (B, T, H_S) -> (B, T, 2 * H_S)
+            x = torch.cat([x, x_future], dim=-1)
 
-        for mixing_layer in self.conditional_mixer:
-            # conditional mixer layers with static covariates (B, T, 2 * H_S), (B, T, C * S) -> (B, T, H_S)
-            x = mixing_layer(x, x_static=x_static)
+        # Apply decoder mixer layers
```
dennisbader, line 552: let's add the shape transformation

darts/models/forecasting/tsmixer_model.py
```diff
@@ -478,39 +509,53 @@ def forward(self, x_in: PLModuleInput) -> torch.Tensor:
...
             x_future = self.feature_mixing_future(x_future)
-            # (B, T, H_S) + (B, T, H_S) -> (B, T, 2*H_S)
-            x = torch.cat([x, x_future], dim=-1)
 
-        if self.static_cov_dim:
-            # (B, C, S) -> (B, 1, C * S)
-            x_static = x_static.reshape(x_static.shape[0], 1, -1)
-            # repeat to match horizon (B, 1, C * S) -> (B, T, C * S)
-            x_static = x_static.repeat(1, self.output_chunk_length, 1)
+        # Apply encoder mixer layers
```
dennisbader, line 538: can you add the shape transformation (e.g. `(B, ...) -> (B, ...)`)

dennisbader (changes requested):
Thanks a lot for the PR and the updates at @eschibli, it looks really good 🚀 

We're almost there :) 

I added a couple of minor suggestions and one where we should discuss a bit more (the one where `encoder_to_decoder` is not applied at a different step compared to the original implementation).

Sorry again for the delay on this one 🙏

PR #2588, opened 2024-11-08
Fix/hfc opti reg prob

### LINKED ISSUE #2461: How are samples generated with RegressionModel / MLPRegressor?
Hi there,

I tried sklearn's `MLPRegressor` with the `RegressionModel` wrapper, and to my surprise, I was able to generate samples with `historical_forecasts` (e.g. with `num_samples = 1000`). How is this possible? Neither `RegressionModel` nor `MLPRegressor` accepts any kind of likelihood or loss function as an argument or am I missing something? 

`model.supports_probabilistic_prediction` is actually false in my case but I can still generate samples.

I would be happy to use this but it is not officially supported, right?

madtoinou:

Fixes #2461.

### Summary

Thanks to #2534, the reshaping of the features logic was fixed. This PR just makes sure an exception is raised when trying to use `num_samples > 1` with a deterministic regression model and the optimized historical forecasts (which does not call the `ForecastingModel.predict()` method, where some sanity checks are implemented).


darts/tests/models/forecasting/test_regression_models.py
```diff
@@ -1289,6 +1289,49 @@ def test_historical_forecast(self, mode):
         )
         assert len(result) == 21
 
+    @pytest.mark.parametrize("mode", [True, False])
```
dennisbader, line 1292: The added logic is the same for both modes, so I think we can simply check only one of the two

dennisbader (approved):
Thanks @madtoinou, looks good to me. Just a minor suggestion

PR #2592, opened 2024-11-12
Feat/slice intersect multi series

### LINKED ISSUE #2042: Union function to find the intersection of time series
Hello!
I think it would be incredibly valuable to have a function that finds the intersection of a list of time series. This would be something akin to xarray's "align" function, which finds the intersection across a list of variables.

I commonly find I have to manually convert my TimeSerries back to an xarray data array, run an align, then convert back to a TimeSeries. For instance:

```
import xarray as xr

ts1 = TimeSeries.from_csv('file1.csv', time_col='time')
ts2 = TimeSeries.from_csv('file2.csv', time_col='time')
ts3 = TimeSeries.from_csv('file3.csv', time_col='time')


ts1_, ts2_, ts3_ = xr.align( ts1.data_array(), ts2.data_array(), ts3.data_array(), exclude = 'component')

ts1 =  TimeSeries(ts1_)
ts2 =  TimeSeries(ts2_)
ts3 =  TimeSeries(ts3_)
```

So something like:
```
ts1 = TimeSeries.from_csv('file1.csv', time_col='time')
ts2 = TimeSeries.from_csv('file2.csv', time_col='time')
ts3 = TimeSeries.from_csv('file3.csv', time_col='time')

ts1, ts2, ts3 = Timeseries.intersection(ts1, ts2, ts3)
```

if the intersection doesn't exist, it could raise a warning like: `Warning: there is no overlapping times between the time series`.

Another option would also be an option for "stack" to take an argument to do this intersection:
```
full_ts = ts1.stack(ts2, intersection = True).stack(ts3, intersection = True)
```

And finally, if this could be added in the pipeline API, that would be excellent!

Apologies if these features already exists! I am new to the package, but I couldn't find an example or something like this in the API. I am loving it so far though!

ymatzkevich:

Fixes #2042.

### Summary

The function `TimeSeries.slice_intersect()` ([see documentation](https://unit8co.github.io/darts/generated_api/darts.timeseries.html?highlight=timeseries#darts.timeseries.TimeSeries.slice_intersect)) allows to intersect a `TimeSeries` with another one so that they end up with the same time indices. However, if one wants to intersect multiple series, that function would need to be called several times or the intersection would need to be done by hand using e.g. `xarray`. The new function `slice_intersect()` introduced with this PR solves this issue for an arbitrary number of `TimeSeries`. 

Essentially, given a list of `TimeSeries` having the same time index type, `slice_intersect()` will output the aligned list meaning that all `TimeSeries` in it will have the same start and end time (if the intersection exists).

### Other Information

If the given `TimeSeries` do not have all the same time index type (e.g. some have a `RangeIndex` and some `DateTimeIndex`), the function will raise an error.


darts/timeseries.py
```diff
@@ -5659,6 +5659,39 @@ def concatenate(
     return TimeSeries.from_xarray(da_concat, fill_missing_dates=False)
 
 
+def intersect(series: Sequence[TimeSeries]):
+    """Returns the intersection with respect to the time index of multiple ``TimeSeries``.
+
```
dennisbader, line 5664:
```suggestion
    """Returns a list of ``TimeSeries``, where all `series` have been intersected along the time index.
```

darts/timeseries.py
```diff
@@ -5659,6 +5659,39 @@ def concatenate(
...
+    Parameters
+    ----------
+    series : Sequence[TimeSeries]
+        sequence of ``TimeSeries`` to intersect
+
+    Returns
+    -------
+    Sequence[TimeSeries]
+        Intersected series
```
dennisbader, line 5673:
```suggestion
        Intersected series.
```

darts/timeseries.py
```diff
@@ -5659,6 +5659,39 @@ def concatenate(
     return TimeSeries.from_xarray(da_concat, fill_missing_dates=False)
 
 
+def intersect(series: Sequence[TimeSeries]):
```
dennisbader, line 5662: Adding the return type in the function signature is always preferred

```suggestion
def slice_intersect(series: Sequence[TimeSeries]) -> Sequence[TimeSeries]:
```

darts/timeseries.py
```diff
@@ -5659,6 +5659,39 @@ def concatenate(
...
+    Sequence[TimeSeries]
+        Intersected series
+    """
+    if not series:
+        return []
+
+    data_arrays = []
+    has_datetime_index = series[0].has_datetime_index
+    for ts in series:
+        if ts.has_datetime_index != has_datetime_index:
```
dennisbader, line 5681: You should add a test for this. You can see which lines are not covered by your tests when looking at the missing lines in the `codecov` report in your PR :)

darts/timeseries.py
```diff
@@ -5659,6 +5659,39 @@ def concatenate(
...
+            raise_log(
+                IndexError(
```
dennisbader, line 5683: We almost always use `ValueError` in Darts, I think we should keep it that way for now
 
```suggestion
                ValueError(
```

darts/timeseries.py
```diff
@@ -5659,6 +5659,39 @@ def concatenate(
...
+                    "The time index type must be the same for all TimeSeries in the Sequence."
```
dennisbader, line 5684:
```suggestion
                    "Found mismatch between time index types. All `series` must either be "
                    "indexed with a `pd.DatetimeIndex` or a `pd.RangeIndex`."
```

darts/tests/test_timeseries.py
```diff
@@ -603,40 +604,64 @@ def check_intersect(other, start_, end_, freq_):
...
+            ]
+
+            if start_ is None:  # empty slice
+                assert len(s_int) == 0
+                assert len(o_int) == 0
+                return
+
+            assert s_int.start_time() == o_int.start_time() == start_
+            assert s_int.end_time() == o_int.end_time() == end_
+            assert s_int.freq == o_int.freq == freq_
```
dennisbader, line 624: This here is redundant, since we know already that `intersect([...])` `series.slice_intersect(...)` are identical (from above). Since you call this check always after the check for `TimeSeries.slice_intersect()`, we can assume that everything is okay

darts/tests/test_timeseries.py
dennisbader, line 1: Would be good to add a test that checks that it's also working for 1 series (<2), and 3 series (> 2)

dennisbader (changes requested):
Thanks for this PR @ymatzkevich, it looks really good already 🚀 
Just had some minor suggestions here and there. After that we can merge

ymatzkevich:
In order to test the efficiency of the new logic, I performed a test (removed from the unit testing because of numerical costs) of intersecting a large number of TimeSeries. For example, something like:
`sequence = [seriesA, seriesB]*500
  int_sequence = slice_intersect(sequence)`
This allowed to compare between the different options available to implement `slice_intersect`. While using `xarray.align` to perform this task proved to be slightly faster, it did not preserve mixed frequencies properly. The new logic is still scaling efficiently and passes all the unit tests.

ymatzkevich:
In order to compare between the different options available to implement `slice_intersect`, I have written the following benchmark:

```python
import time
import itertools
import numpy as np
import pandas as pd

from darts import TimeSeries
from darts.timeseries import slice_intersect
from darts.utils.utils import generate_index

def helper_test_intersect(freq, is_mixed_freq: bool, N):
    start = pd.Timestamp("20130101") if isinstance(freq, str) else 0
    freq = pd.tseries.frequencies.to_offset(freq) if isinstance(freq, str) else freq

    # handle identical and mixed frequency setup
    if not is_mixed_freq:
        freq_other = freq
        n_steps = 11
    elif "2" not in str(freq):  # 1 or "1D"
        freq_other = freq * 2
        n_steps = 21
    else:  # 2 or "2D"
        freq_other = freq / 2
        n_steps = 11
    freq_other = int(freq_other) if isinstance(freq_other, float) else freq_other

    idx = generate_index(start=start, freq=freq, length=n_steps)
    end = idx[-1]

    # we construct 2 different series that will be used for the intersection
    startA = start
    endA = end
    idxA = generate_index(startA, endA, freq=freq_other)
    seriesA = TimeSeries.from_series(pd.Series(range(len(idxA)), index=idxA))

    startB = start + freq
    endB = startB + 6 * freq_other
    idxB = generate_index(startB, endB, freq=freq_other)
    seriesB = TimeSeries.from_series(pd.Series(range(len(idxB)), index=idxB))

    iterations = 100 # to have a statistical sample from which we compute mean time

    start_time = time.time()

    for _ in range(iterations):
        sequence = [seriesA, seriesB]*N
        int_sequence = slice_intersect(sequence) # we do not need to use the intersected sequence, just to compute it for benchmarking

    end_time = time.time()
    time_taken = end_time - start_time
    mean_time = time_taken/iterations 

    return mean_time

freq_list = ["D", "2D", 1, 2] # different types of frequencies 
is_mixed_freq_list = [False, True] # mixed frequencies
N = 5 # determines size of sequence that we are intersecting

combinations = list(itertools.product(freq_list, is_mixed_freq_list)) # we test all combinations
length = len(combinations) # number of combinations

total_time = 0
for i, (freq, is_mixed_freq) in enumerate(combinations):
    print(f"combination {i+1}/{length}: (freq,is_mixed_freq)=({freq},{is_mixed_freq})")
    total_time += helper_test_intersect(freq, is_mixed_freq, N)

mean_time = total_time / length

print(f"N={N}, mean_time={mean_time}")


```

The results of the benchmark are shown here, in seconds:

|                                    | `N`=5    | `N`=50   | `N`=500  |
|------------------------------------|--------|--------|--------|
| using `TimeSeries.slice_intersect()` | 0.0021 s | 0.0229 s | 0.2374 s |
| using `pandas intersection` function | 0.0031 s | 0.0312 s | 0.3275 s |
| using `xarray.align`                 | 0.0022 s | 0.0225 s | 0.2345 s |

While using `xarray.align` to perform this task compares to the present logic in terms of speed, it does not preserve mixed frequencies properly and hence does not pass all the unit tests. While first intersecting on the time indexes using the `intersection` function from `pandas` could seem faster, it is still slower than the Darts implementation which takes advantage of the `TimeSeries` structure.

dennisbader (approved):
Thank a lot @ymatzkevich for this very nice PR and performance report 💯 

Everything looked fine, I took the opportunity to make some minor adaptions. 
Now it's ready to be merged 🚀

PR #2634, opened 2024-12-26
FIX-2633: Use correct time indices when running historical forecasts …

### LINKED ISSUE #2633: [BUG] optimized historical_forecasts has incorrect time indices under certain conditions
**Describe the bug**
When generating a historical forecast using regression model and last point forecasting on a model with `output_chunk_shift > 0`, time indices are not shifted according to the `output_chunk_shift`. This is only the case when doing an 'optimized' historical forecast (when you get into `darts.utils.historical_forecasts.optimized_historical_forecasts_regression._optimized_historical_forecasts_last_points_only`.

**To Reproduce**
```

model = RegressionModel(lags=20, output_chunk_shift=5)
model.fit(series)

historical = model.historical_forecasts(series=series,
                                        retrain=False,
                                        forecast_horizon=1)


series.plot(label="actual")
historical.plot(label="historical")
```

This will plot this image:
image

**Expected behavior**
The returned historical series don´t have their time index shifted respecting the output chunk shift.
The following result would be correct:
image

 - Python version: 3.12
 - darts version 0.31.0

**Additional context**
I will create a PR for this.
I've also inspected `darts.utils.historical_forecasts.optimized_historical_forecasts_regression._optimized_historical_forecasts_all_points`, which does seem to always take into account the `output_chunk_shift` unconditionally. I also looked at `darts.utils.historical_forecasts.optimized_historical_forecasts_tocrch._optimized_historical_forecasts`, which also doesn´t seem affected.

MattiasDC:
…on regression models with 'output_chunk_shift > 0' and 'output_chunk_length == 1'. Extended unit tests to cover this

Fixes #2633.

### Summary

When creating historical forecasts using the optimized version, use the correct time indices when `output_chunk_shift > 0` and `output_chunk_length == 1`. Extended the unit tests to cover this case


darts/utils/historical_forecasts/optimized_historical_forecasts_regression.py
```diff
@@ -162,19 +162,24 @@ def _optimized_historical_forecasts_last_points_only(
         else:
             forecast = forecast[:, 0]
 
+        if (
+            stride == 1
+            and model.output_chunk_length == 1
+            and model.output_chunk_shift == 0
```
MattiasDC, line 168: This if-check is extended with an additional condition, which fixes the issue.

dennisbader (approved):
Thanks @MattiasDC for finding and fixing this bug! It's greatly appreciated :) 
PR also looks great. I made some changes to the tests (the bounds were not correct, but we didn't properly check that before) and updated the changelog.

I'll merge once all tests have passed. Thanks again 🚀

MattiasDC:
I'm on my phone now, but you have a typo in test comments 'sampels' iso 'samples'

PR #2639, opened 2025-01-08
title as parameter for TimeSeries.plot()

### LINKED ISSUE #2622: `TimeSeries.plot` Title as parameter
The `TimeSeries.plot` method sets the plot title to the name of the underlying xarray's name.

However, it is unclear to me (and to my knowledge not documented) how to set this name without touching the private `._xa`.

In my experience, the name attribute is not set, if it is not constructed directly from a named xarray, which is rarely the case.

I propose a new parameter for the plot function, that can overwrite this behaviour.

- title (str): Set a title for the figure. Default is None, which plots the name of the underlying xarray (if any)

https://github.com/unit8co/darts/blob/71a1902b1c28b579328984cd3a37349cd9a9aba9/darts/timeseries.py#L4257
would change then to:
`ax.set_title(title if title else self._xa.name)`

- Extend the TimeSeries API so a name can be set. E.g. `TimeSeries.set_name(self, name: str)`
- Provide a name parameter for all TimeSeries construction methodes, such as e.g. `TimeSeries.from_array(..., name: str)`

I would prefer proposed solution which is more flexible to the user

**PR**

I can to a PR for that

jonathankoch99:

Fixes #2622 .

### Summary

Adds the `title` attribute to `TimeSeries.plot()` as requested in #2622 . If no title is provided as attribute the method reverts to the original behavior.


madtoinou (changes requested):
Thank you for this contribution @jonathankoch99! Looks good to me, can you please just add an entry in the Unreleased section of the changelog and it should be ready to be merged!

darts/timeseries.py
```diff
@@ -4254,7 +4257,7 @@ def plot(
                     )
 
         ax.legend()
-        ax.set_title(self._xa.name)
+        ax.set_title(title if title else self._xa.name)
```
dennisbader, line 4260:
```suggestion
        ax.set_title(title if title is not None else self._xa.name)
```

darts/timeseries.py
```diff
@@ -4106,6 +4107,8 @@ def plot(
             interval is shown if `high_quantile` is None (default 0.95).
         default_formatting
             Whether to use the darts default scheme.
+        title
+            The title of the plot.
```
dennisbader, line 4111:
```suggestion
            Optionally, a custom plot title. If `None`, will use the name of the underlying `xarray.DataArray`.
```

dennisbader (changes requested):
Thanks a lot @jonathankoch99 for this PR, and welcome to the Darts community :D 🚀 

Just some minor comments, then we can merge.

dennisbader (approved):
Looks great now, thanks again 💯 🚀

madtoinou (approved):

PR #2643, opened 2025-01-17
Extend TimeSeries.resample() to support more resampling methods

### LINKED ISSUE #699: Extend `TimeSeries.resample()` to be more like pandas `resample()`
**Describe the bug**
Hi, I'm trying to downsample a high-frequency Darts time series object to daily/monthly/yearly mean/sum/min/max etc. But the behavior is different than Pandas, so my only current option is to downsample before converting to a time series object. 

**To Reproduce**
```
import pandas as pd
import numpy as np
from darts import TimeSeries
from darts.datasets import MonthlyMilkDataset, MonthlyMilkIncompleteDataset

series = MonthlyMilkDataset().load()

series.resample('Y').sum()

```
Results in 

`component
Pounds per cow    10094.0
dtype: float64`

**Expected behavior**
Expected a time series with the sum of each year, rather than the sum of all years. That is the default behaviour of the pandas resample method and I'm having a hard time understanding why this wouldn't work in Darts.

 - Python version: 3.9
 - darts version: 0.14.0

jonasblanc:

Fixes #699 

### Summary

- Expose all aggregation functions of [xarray.DataArrayResample](https://docs.xarray.dev/en/stable/generated/xarray.core.resample.DataArrayResample.html) through the "method" argument of `TimeSeries.resample()`.
- Let users specify method specific arguments through `method_args`.
- Extend tests with at least one test per additional method.

### Other Information

- Not all methods make sense for either down/up-sampling. I chose not to restrict the use of such methods in situation where they are not expected to be used (ie. xarray is responsible for handling such case). However while testing I encountered this bug: [issue xarray asfreq](https://github.com/orgs/unit8co/projects/12/views/1?pane=issue&itemId=80891935&issue=unit8co%7Cdarts%7C699) which happens when `asfreq` is used for down-sampling.
- I chose to give access to `any()` and `all()` aggregation methods, to be consistent with `TimeSeries`, I cast `bool` back to `int` after the aggregation.
- I chose to let `pad` as the default method so as not to break the default behavior, even though it's not the most intuitive method for down-sampling.

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -2496,7 +2496,7 @@ class TimeSeries:
         """
         if other.has_same_time_as(self):
             return self.__class__(self._xa)
-        if other.freq == self.freq:
+        elif other.freq == self.freq and len(self) and len(other):
             start, end = self._slice_intersect_bounds(other)
             return self[start:end]
         else:
@@ -2815,9 +2815,9 @@ class TimeSeries:
         """
         if len(other) != len(self):
             return False
-        if other.freq != self.freq:
+        elif other.freq != self.freq:
             return False
-        if other.start_time() != self.start_time():
+        elif other.start_time() != self.start_time():
             return False
         else:
             return True
@@ -5660,6 +5660,35 @@ def concatenate(
         )
 
     return TimeSeries.from_xarray(da_concat, fill_missing_dates=False)
+
+
+def slice_intersect(series: Sequence[TimeSeries]) -> list[TimeSeries]:
+    """Returns a list of ``TimeSeries``, where all `series` have been intersected along the time index.
+
+    Parameters
+    ----------
+    series : Sequence[TimeSeries]
+        sequence of ``TimeSeries`` to intersect
+
+    Returns
+    -------
+    Sequence[TimeSeries]
+        Intersected series.
+    """
+    if not series:
+        return []
+
+    # find global intersection on first series
+    intersection = series[0]
+    for series_ in series[1:]:
+        intersection = intersection.slice_intersect(series_)
+
+    # intersect all other series
+    series_intersected = [intersection]
+    for series_ in series[1:]:
+        series_intersected.append(series_.slice_intersect(intersection))
+
+    return series_intersected
 
 
 def _finite_rows_boundaries(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - New model: `StatsForecastAutoTBATS`. This model offers the [AutoTBATS](https://nixtlaverse.nixtla.io/statsforecast/src/core/models.html#autotbats) model from Nixtla's `statsforecasts` library. [#2611](https://github.com/unit8co/darts/pull/2611) by [He Weilin](https://github.com/cnhwl).
 - Added the `title` attribute to `TimeSeries.plot()`. This allows to set a title for the plot. [#2639](https://github.com/unit8co/darts/pull/2639) by [Jonathan Koch](https://github.com/jonathankoch99).
 - Added parameter `component_wise` to `show_anomalies()` to separately plot each component in multivariate series. [#2544](https://github.com/unit8co/darts/pull/2544) by [He Weilin](https://github.com/cnhwl).
+- Added general function `darts.slice_intersect()` to intersect a sequence of `TimeSeries` along the time index. [#2592](https://github.com/unit8co/darts/pull/2592) by [Yoav Matzkevich](https://github.com/ymatzkevich).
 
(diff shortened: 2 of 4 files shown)


darts/timeseries.py
```diff
@@ -3367,20 +3367,22 @@ def resample(
...
         freq
             The new time difference between two adjacent entries in the returned TimeSeries.
             Expects a `pandas.DateOffset` or `DateOffset` alias.
         method:
-            Method to fill holes in reindexed TimeSeries (note this does not fill NaNs that already were present):
-
-            'pad': propagate last valid observation forward to next valid
-
-            'backfill': use NEXT valid observation to fill.
+            Method to either aggregate grouped values (for down-sampling) or fill holes (for up-sampling)
```
madtoinou, line 3378: let's chaneg the type from `str` to `Literal`, it will simplify the sanity checks. Since there are a lot of possible values, we could use a constant/alias to store them.

darts/timeseries.py
```diff
@@ -3367,20 +3367,22 @@ def resample(
...
+            in reindexed TimeSeries. For more information, see the `xarray DataArrayResample documentation
+            <https://docs.xarray.dev/en/stable/generated/xarray.core.resample.DataArrayResample.html>`_.
+            Supported methods: ["all", "any", "asfreq", "backfill", "bfill", "count", "ffill",
+                                "first", "interpolate", "last", "max", "mean", "median", "min",
+                                "nearest", "pad", "prod", "quantile", "reduce", "std", "sum", "var"].
         kwargs
+            use method_args:dict[str:Any] to specify method specific arguments, otherwise
```
madtoinou, line 3385: This is not clear, did you mean `method_args=dict[str, Any]`?

darts/timeseries.py
```diff
@@ -3410,25 +3412,67 @@ def resample(
...
+        >>> downsampled_mean_ts = ts.resample(freq="30min", method="mean")
+        >>> print(downsampled_mean_ts.values())
+        [[0.5]
+        [2.5]
+        [4.5]]
+        >>> downsampled_reduce_ts = ts.resample(freq="30min", method="reduce", method_args={"func":np.mean})
+        >>> print(downsampled_reduce_ts.values())
+        [[0.5]
+        [2.5]
+        [4.5]]
```
madtoinou, line 3425: Nice!

madtoinou (changes requested):
Nice contribution @jonasblanc, some minor comments to improve the readibility

darts/timeseries.py
jonasblanc: As discussed waiting for @dennisbader opinion on this. Typing as Literal is not sufficient for sanity check of `method`. So it would look something like this.

```
ResamplingMethod = Literal["all", ... "var",]

def resample(
        self, freq: Union[str, pd.DateOffset], method: ResamplingMethod = "pad", **kwargs
    ) -> Self:
    ...
    if method in get_args(ResamplingMethod):
    ...
```

darts/timeseries.py
```diff
@@ -3363,24 +3367,30 @@ def add_holidays(
...
             The new time difference between two adjacent entries in the returned TimeSeries.
             Expects a `pandas.DateOffset` or `DateOffset` alias.
         method:
-            Method to fill holes in reindexed TimeSeries (note this does not fill NaNs that already were present):
-
-            'pad': propagate last valid observation forward to next valid
-
-            'backfill': use NEXT valid observation to fill.
+            Method to either aggregate grouped values (for down-sampling) or fill holes (for up-sampling)
+            in reindexed TimeSeries. For more information, see the `xarray DataArrayResample documentation
```
dennisbader, line 3387:
```suggestion
            in the reindexed TimeSeries. For more information, see the `xarray DataArrayResample documentation
```

dennisbader (changes requested):
Very nice PR, thanks and great job @jonasblanc 🚀  Just had some minor suggestions, after that we're ready to merge :)

darts/timeseries.py
```diff
@@ -3363,24 +3367,30 @@ def add_holidays(
...
+            <https://docs.xarray.dev/en/stable/generated/xarray.core.resample.DataArrayResample.html>`_.
+            Supported methods: ["all", "any", "asfreq", "backfill", "bfill", "count", "ffill",
+                                "first", "interpolate", "last", "max", "mean", "median", "min",
+                                "nearest", "pad", "prod", "quantile", "reduce", "std", "sum", "var"].
         kwargs
+            use method_args:dict[str, Any] to specify method specific arguments, otherwise
```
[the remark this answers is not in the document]
jonasblanc: nice catch, I forgot to update the docstring

madtoinou (approved):
LGTM, great first contribution @jonasblanc! 🚀

Issue #2638, opened 2025-01-07
[BUG] TCN model cannot be saved when used with callbacks


MarcBresson:
**Describe the bug**
For some reason, defining callbacks in pl_trainer_kwargs will make the torch.save function of TCN.save method save the lightning module. Because TCN contains Parametrized layers, torch will raise an error `RuntimeError: Serialization of parametrized modules is only supported through state_dict().`

Not setting any custom callbacks will not trigger the bug.

**To Reproduce**
```python
from darts.utils import timeseries_generation as tg
from darts.models.forecasting.tcn_model import TCNModel

def test_save(self):
    large_ts = tg.constant_timeseries(length=100, value=1000)
    model = TCNModel(
        input_chunk_length=6,
        output_chunk_length=2,
        n_epochs=10,
        num_layers=2,
        kernel_size=3,
        dilation_base=3,
        weight_norm=True,
        dropout=0.1,
        **{
            "pl_trainer_kwargs": {
                "accelerator": "cpu",
                "enable_progress_bar": False,
                "enable_model_summary": False,
                "callbacks": [LiveMetricsCallback()],
            }
        },
    )
    model.fit(large_ts[:98])

        model.save("model.pt")


import pytorch_lightning as pl
from pytorch_lightning.callbacks import Callback


class LiveMetricsCallback(Callback):
    def __init__(self):
        self.is_sanity_checking = True

    def on_train_epoch_end(
        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"
    ) -> None:
        print()
        print("train", trainer.current_epoch, self.get_metrics(trainer, pl_module))

    def on_validation_epoch_end(
        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"
    ) -> None:
        if self.is_sanity_checking and trainer.num_sanity_val_steps != 0:
            self.is_sanity_checking = False
            return
        print()
        print("val", trainer.current_epoch, self.get_metrics(trainer, pl_module))

    @staticmethod
    def get_metrics(trainer, pl_module):
        """Computes and returns metrics and losses at the current state."""
        losses = {
            "train_loss": trainer.callback_metrics.get("train_loss"),
            "val_loss": trainer.callback_metrics.get("val_loss"),
        }
        return dict(
            losses,
            **pl_module.train_metrics.compute(),
            **pl_module.val_metrics.compute(),
        )

```

will output

```
darts/models/forecasting/torch_forecasting_model.py:1679: in save
[...]

PR #2649, opened 2025-01-21
Allow to strip training data for storage and memory reason on save for GlobalForecastingModel

### LINKED ISSUE #1836: Trained Model - new param allowing to strip training data for storage and memory reasons
Hello, 

I would like to request a new parameter to a trained model which will strip all data before saving model for prediction purposes.

Similarly to statmodels's remove_data(), e.g.
https://www.statsmodels.org/dev/generated/statsmodels.regression.linear_model.RegressionResults.remove_data.html

Main reason is that many trained models have future use on prod environment where training data are irrelevant and only trained function used with prod data is needed to produce prediction. 

Best,
Michal

jonasblanc:

Fixes #1836 #2638 

### Summary

Add `clean` parameters to `ForecastingModel.save()` that allow saving without training specific attributes. This is done by implementing a private `_clean()` method returning the cleaned model. 

- For `GlobalForecastingModel` it removes `training_series`, `past_covariate_series` and `future_covariate_series` thus reducing the memory footprint and data leakage of the saved model.
- For `TorchForecastingModel` it additionally drops `trainer`, `trainer_params`, `model_params.pl_trainer_kwargs`, and the PyTorch model (which is actually saved via separate checkpoint). By removing these attributes before saving, it solves the issue caused by pickling callbacks set to the model (see #2638).
- Support for `clean` added to `EnsembleModel.save()` and `ConformalModel.save()`

These changes lead to a refactoring of `TorchForecastingModel.load()`:

- 🔴 Deprecate `location_map` parameter which had no effect.
-  Add `pl_trainer_kwargs` parameter to create a new PyTorch Lightning Trainer used for prediction, etc. Hardware location can be specified via `accelerator` key of  `pl_trainer_kwargs`.
- Load cleaned model on `cpu` per default.


darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -274,32 +275,44 @@ def test_save_model_parameters(self, config):
...
+        loaded_model = type(model).load(full_model_path_str)
+        loaded_model_drop_str = type(model).load(full_model_clean_path_str)
 
         assert model_prediction == loaded_model.predict(self.forecasting_horizon)
+        assert model_prediction == loaded_model_drop_str.predict(
+            self.forecasting_horizon, self.ts_pass_train
+        )
+
+        # test save with clean
+        assert loaded_model_drop_str.training_series is None
```
dennisbader, line 315: Could you add a check that the loaded clean model raises an error when calling `predict()` without passing a `series`?

darts/tests/models/forecasting/test_global_forecasting_models.py
dennisbader, line 1: nice tests :) 👍

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -291,6 +291,145 @@ def test_manual_save_and_load(self, tmpdir_fn):
         # compare chained load_from_checkpoint() save() with manual save
         assert model_chained_load_save.predict(n=4) == model_manual_save.predict(n=4)
 
+    def test_manual_save_and_load_with_drop_training(self, tmpdir_fn):
```
dennisbader, line 294: This test is almost a copy of the `test_manual_save_and_load` test.
Could you try to extend that test to also check for the clean safe?

And then as a last point it would be good to also check that the callbacks were removed.

dennisbader (changes requested):
Thanks @jonasblanc for this PR. It already looks like a great start 🚀 

I added some suggestions, mainly regarding how we could try to further unify the behavior of `_clean()` and ultimately `save()`.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1726,29 +1754,80 @@ def load(path: str, **kwargs) -> "TorchForecastingModel":
         path
             Path from which to load the model. If no path was specified when saving the model, the automatically
             generated path ending with ".pt" has to be provided.
+        pl_trainer_kwargs
+            By default :class:`TorchForecastingModel` creates a PyTorch Lightning Trainer with several useful presets
```
dennisbader, line 1758: I think we can simplify the description here. Maybe:

```
pl_trainer_kwargs
    Optionally, a set of kwargs to create a new PyTorch Lightning Trainer used to handle the underlying module. Check the `Lightning Trainer documentation <https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html>`_ for more information about the supported kwargs.
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1726,29 +1754,80 @@ def load(path: str, **kwargs) -> "TorchForecastingModel":
...
+            if reset_trainer_params:
+                dtype_to_precision = (
+                    {torch.float32: "32", torch.float64: "64"}
+                    if not pl_200_or_above
+                    else {torch.float32: "32-true", torch.float64: "64-true"}
+                )
+                model.trainer_params["precision"] = dtype_to_precision[
+                    model.model.dtype
+                ]
+                model.trainer_params["accelerator"] = model.model.device.type
```
dennisbader, line 1812: Actually, I believe the precision is already restored when loading the model checkpoint: https://github.com/unit8co/darts/blob/c48521ce9bbadbd9adde6df2dfb5797c60aad514/darts/models/forecasting/pl_forecasting_module.py#L548

In that case we could remove the code below. The accelerator they can set with the `pl_trainer_kwargs`

```suggestion
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1726,29 +1754,80 @@ def load(path: str, **kwargs) -> "TorchForecastingModel":
...
+        map_location = kwargs.get("map_location", None)
         with open(path, "rb") as fin:
-            model: TorchForecastingModel = torch.load(
-                fin, map_location=kwargs.get("map_location", None)
-            )
+            model: TorchForecastingModel = torch.load(fin, map_location=map_location)
+
+        reset_trainer_params = model.trainer_params is None
+        if reset_trainer_params:
+            model.trainer_params = {"callbacks": []}
```
dennisbader, line 1796:
```suggestion
        model.trainer_params = model.trainer_params or {}
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1726,29 +1754,80 @@ def load(path: str, **kwargs) -> "TorchForecastingModel":
...
+                ]
+                model.trainer_params["accelerator"] = model.model.device.type
         else:
             model._fit_called = False
             logger.warning(
                 f"Model was loaded without weights since no PyTorch LightningModule checkpoint ('.ckpt') could be "
                 f"found at {path_ptl_ckpt}. Please call `fit()` before calling `predict()`."
             )
+
+        if pl_trainer_kwargs is not None:
```
dennisbader, line 1820: If the user gives `pl_trainer_kwargs` then I would just use them and overwrite the _model_params and trainer_params (instead of enriching the existing ones)

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1643,12 +1643,30 @@ def _batch_collate_fn(batch: list[tuple]) -> tuple:
...
-    def save(self, path: Optional[str] = None) -> None:
+    def _clean(self):
+        """Returns a cleaned model, keeping only the necessary attributes for prediction."""
+        model = super()._clean()
+        # Copy from super()._clean() call __getstate__ which removes model and trainer
+        # a shallow copy is enough since we are only interested in removing pointers
+        model.model = copy.copy(self.model)  # keep the model for prediction
+        model._model_params = copy.copy(self._model_params)
+        model._model_params["pl_trainer_kwargs"] = None
+        model.trainer_params = None
```
dennisbader, line 1654: Maybe better to assign an empty here since when creating a new TorchForecastingModel, it will always have type dict.

```suggestion
        model.trainer_params = {}
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1726,29 +1754,80 @@ def load(path: str, **kwargs) -> "TorchForecastingModel":
...
+            parameter ``trainer`` in :func:`fit()` and :func:`predict()`.
         **kwargs
             Additional kwargs for PyTorch Lightning's :func:`LightningModule.load_from_checkpoint()` method,
             such as ``map_location`` to load the model onto a different device than the one from which it was saved.
             For more information, read the `official documentation <https://pytorch-lightning.readthedocs.io/en/stable/
             common/lightning_module.html#load-from-checkpoint>`_.
         """
-
         # load the base TorchForecastingModel (does not contain the actual PyTorch LightningModule)
+        map_location = kwargs.get("map_location", None)
```
dennisbader, line 1790: I'm actually not sure that `map_location` has any impact here, since the `model.model` is not retrieved when calling `torch.load`.
Can you check whether changing the "accelerator" in `pl_trainer_kwargs` automatically maps it to the desired location?

If that's the case, we can remove the parameter from here. That would require adding a "breaking change" entry in the CHANGELOG.md

dennisbader (changes requested):
Thanks for the updates @jonasblanc, looks like a great improvement 🚀 

I added a couple more suggestions. Let me know if you have any questions.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1643,12 +1643,30 @@ def _batch_collate_fn(batch: list[tuple]) -> tuple:
...
         return tuple(aggregated)
 
-    def save(self, path: Optional[str] = None) -> None:
+    def _clean(self):
+        """Returns a cleaned model, keeping only the necessary attributes for prediction."""
+        model = super()._clean()
+        # Copy from super()._clean() call __getstate__ which removes model and trainer
+        # a shallow copy is enough since we are only interested in removing pointers
+        model.model = copy.copy(self.model)  # keep the model for prediction
+        model._model_params = copy.copy(self._model_params)
```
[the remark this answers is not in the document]
jonasblanc: They are, but since we are doing a shallow copy, modifying `pl_trainer_kwargs` of `model._model_params` would modify the original `model._model_params` as well.

darts/models/forecasting/torch_forecasting_model.py
jonasblanc: Done

dennisbader (approved):
Everything looks great now, thanks a lot for this PR @jonasblanc 🚀 

I adapted some parts, mainly regarding:
- updating the documentation
- save/load torch model wrapper with pickle instead of torch

PR #2654, opened 2025-01-24
Feat/resample

### LINKED ISSUE #699: Extend `TimeSeries.resample()` to be more like pandas `resample()`

dennisbader:
Copy from #2643

Fixes #699 

### Summary

- Expose all aggregation functions of [xarray.DataArrayResample](https://docs.xarray.dev/en/stable/generated/xarray.core.resample.DataArrayResample.html) through the "method" argument of `TimeSeries.resample()`.
- Let users specify method specific arguments through `method_args`.
- Extend tests with at least one test per additional method.

### Other Information

- Not all methods make sense for either down/up-sampling. I chose not to restrict the use of such methods in situation where they are not expected to be used (ie. xarray is responsible for handling such case). However while testing I encountered this bug: [issue xarray asfreq](https://github.com/orgs/unit8co/projects/12/views/1?pane=issue&itemId=80891935&issue=unit8co%7Cdarts%7C699) which happens when `asfreq` is used for down-sampling.
- I chose to give access to `any()` and `all()` aggregation methods, to be consistent with `TimeSeries`, I cast `bool` back to `int` after the aggregation.
- I chose to let `pad` as the default method so as not to break the default behavior, even though it's not the most intuitive method for down-sampling.


dennisbader (commented):
Beautiful PR and tests, thanks a lot @jonasblanc 🚀

PR #2659, opened 2025-01-30
Catboost native multi-output with RMSE

### LINKED ISSUE #1306: Make CatboostModel use native Catboost multi-output when possible
See: https://gitter.im/u8darts/darts?at=6353986a27f328266d4e3d83

jonasblanc:

Fixes #1306

### Summary

Refactor logic handling native support for multi-output. 
- Move model specific logic into the newly defined `RegressionModel._native_support_multioutput()` (and sub-classes) to return if the underlying model does support multi-output natively.
- Refactor the logic leading to the decision to wrap a model into a `MultiOutputRegressor`.
- Switch from `sklearn` `_get_tags` to `__sklearn_tags__` as it will be required by `sklearn` 1.7. This leads to increasing both sklearn and XGBoost minimum supported version to 1.6 and 2.1.4 respectively. See [sklearn release notes](https://scikit-learn.org/stable/whats_new/v1.6.html#) for more information about tags and [XGBoost release notes](https://github.com/dmlc/xgboost/releases/tag/v2.1.4) for `sklearn` 1.6 compatibility.
- Improve `CatBoostModel` documentation by describing how to use native multioutput regression.


dennisbader (changes requested):
Thanks a lot @jonasblanc, this looks great and is going into the right direction 🚀 

I added a couple of suggestions. The major point would be that I think we should leave it to the user to decide whether to use native, or darts' multioutput regression.

requirements/core.txt
```diff
@@ -17,4 +17,4 @@ tbats>=1.1.0
 tqdm>=4.60.0
 typing-extensions
 xarray>=0.17.0
-xgboost>=1.6.0
+xgboost>=2.1.4
```
madtoinou, line 20: Since this lower limit is set on the xgboost dependency, all the logic around the `xgb_200_or_above` variable can be removed

madtoinou (approved):
Thanks a lot @jonasblanc for the PR, just minor changes but it looks very good to me!

dennisbader (approved):
Beautiful 😍  Thanks a lot for this great PR, and it also looks much more clean now with the refactoring to sklearn >= 1.6.0 🚀 

Made some last minor updates.

Issue #2635, opened 2025-01-02
Add `TimeSeries.from_polars`


baggiponte:
**Is your feature request related to a current problem? Please describe.**
I was trying to use darts with Polars, but casting a Polars DataFrame into a TimeSeries hardly works.

**Describe proposed solution**
There are two ways of doing this:

1. quick and dirty: add a `TimeSeries.from_polars` classmethod that handles conversion with mostly custom logic (this could potentially entail deprecating `from_dataframe` and renaming it into `from_pandas`: it's not the de facto dataframe library anymore).
2. Use [narwhals](https://github.com/narwhals-dev/narwhals) to rewrite `TimeSeries.from_dataframe` and support all dataframe backends that `narwhals` does.

To give a bit of context: narwhals is a new project that's grown a lot this year. It's a very thin, zero-dependency library for developers to write dataframe-agnostic code. It's quite used: plotly 6.0 will use polars for (AFAIK) _all_ its dataframe manipulation, and so do already libraries like altair, marimo, hierarchicalforecast by nixtla...

Glad to provide more details 😊

madtoinou:
Hi @baggiponte,

I really like the idea of re-implementing `from_dataframe` using `narwahls` so that darts can support both `pandas` and `polars`, it should not be too difficult... Let's see what @dennisbader thinks about it!

If this feature request makes it to the roadmap, would you have the time to contribute and open a PR for it?

baggiponte:
Ciao @madtoinou, thanks for the prompt reply! I would love to, but not sure if I have time this month. However, I am sure that @FBruzzesi and @marcogorelli might help (me) out - they are usually always available for narwhalification 🦄

On the implementation detail side, I guess the signature of the function could and should stay the same, am I right?

dennisbader:
Thanks for this feature request @baggiponte. I like the idea of adding support for polars (and others) to our factory methods.

I quickly checked out narwhals and it definitely looks interesting.

Just to clarify that we're all on the same page: We only want to add support for converting polars, ... DataFrames into TimeSeries (meaning `xarray`). So the TimeSeries itself will not run e.g. on polars for downstream tasks.

Solution 2. seems like a nice solution for me if we can guarantee that the performance doesn't suffer for the pandas support.

If we do that, then we could also add support for other methods:
- `from_series()`
- `from_group_dataframe()`

What do you think?

PR #2661, opened 2025-01-31
Narwhals implementation of `from_dataframe` and performance benchmark

authierj:

Fixes [#2635](https://github.com/unit8co/darts/issues/2635).

### Summary

A first draft of `from_dataframe` has been adapted to work with any dataframe. This is done using narwhals and the function is called `from_narwhals_dataframe`. In order to test the performance of the method, a file `narwhals_test_time.py` has been added to the pull request. 
With the latest commits, `from_narwhals_dataframe` is now as fast as `from_dataframe`.


darts/timeseries.py
```diff
@@ -746,6 +748,177 @@ def from_dataframe(
...
+            time_col_vals = df[time_col]
+
+            if time_col_vals.dtype == nw.String:
+                # Try to convert to integers if needed
+                try:
+                    time_col_vals = time_col_vals.cast(nw.Int64)
+                except Exception:
+                    pass
+
+            if time_col_vals.dtype == nw.Int64 or time_col_vals.dtype == np.integer:
```
MarcoGorelli, line 860: comparing to `np.integer` isn't supported, you may need to use `time_col_vals.dtype is in(nw.Int128, nw.Int64, nw.Int32, nw.Int16, nw.Int8, nw.UInt128, nw.UInt64, nw.UInt32, nw.UInt16, nw.UInt8`

🤔 we may need to introduce an easier way of doing this in Narwhals

darts/timeseries.py
```diff
@@ -746,6 +748,177 @@ def from_dataframe(
...
+                    )
+
+                # Temporarily use an integer Index to sort the values, and replace by a
+                # RangeIndex in `TimeSeries.from_xarray()`
+                time_index = time_col_vals.to_list()
+
+            elif time_col_vals.dtype == nw.String:
+                # The integer conversion failed; try datetimes
+                try:
+                    time_index = nw.Datetime(time_col_vals)
```
MarcoGorelli, line 876: you may be looking for `time_col_vals.str.to_datetime()`?

MarcoGorelli (commented):
thanks for giving this a go!

I've left a couple of comments

I suspect the `.to_list()` calls may be responsible for the slow-down. I'll take a look

authierj:
Hi @MarcoGorelli , 

Thanks for already looking at this and for your insights!

authierj:
Hi @MarcoGorelli,

I investigated the issue, and it appears that the `.to_list()` call is not responsible for the slowdown. However, the call `series_df.to_numpy()[:, :, np.newaxis]` on line `906` is very slow. The investigation is going on!

darts/timeseries.py
FBruzzesi: @authierj just as heads up, what Marco proposed here is now available (as of `narwhals>=1.25.0`)

narwhals_test_time.py
FBruzzesi, line 1: Just to make sure, this is the script you are using to estimate run time?

darts/timeseries.py
```diff
@@ -746,6 +748,177 @@ def from_dataframe(
...
+                isinstance(time_col_vals, VALID_INDEX_TYPES)
+                or np.issubdtype(time_col_vals.dtype, np.integer),
+                "If time_col is not specified, the DataFrame must be indexed either with "
+                "a DatetimeIndex, a RangeIndex, or an integer Index that can be converted into a RangeIndex",
+                logger,
+            )
+            time_index = time_col_vals.to_list()
+
+        xa = xr.DataArray(
+            series_df.to_numpy()[:, :, np.newaxis],
```
FBruzzesi, line 906: We do some additional check when converting from pandas to numpy. However I don't think that's enough to justify a 3x slow down.

FBruzzesi (commented):
Thanks @authierj for the effort on this! We really appreciate it! I left very non-relevant comments 😂

darts/timeseries.py
FBruzzesi: Edit: Well, it actually looks like that is the culprit.

@MarcoGorelli, I was able to pin it down to:

https://github.com/narwhals-dev/narwhals/blob/a7707077bc62f4491cac82ca3c609e83c3a87f23/narwhals/_pandas_like/dataframe.py#L823

which calls 

https://github.com/narwhals-dev/narwhals/blob/a7707077bc62f4491cac82ca3c609e83c3a87f23/narwhals/_pandas_like/dataframe.py#L341

for the majority of the slowdown

narwhals_test_time.py
authierj: yes exactly!

darts/timeseries.py
authierj: Thanks to you both, I just updated the code [here](https://github.com/unit8co/darts/pull/2661/commits/0041203af98b103b97ed91ecca651a299473e3d1)

darts/timeseries.py
authierj: Done [here](https://github.com/unit8co/darts/pull/2661/commits/0041203af98b103b97ed91ecca651a299473e3d1), thanks @MarcoGorelli

darts/timeseries.py
MarcoGorelli: interesting, thanks @FBruzzesi - I presume this is due to the `infer_dtype` call?

cuDF and Dask don't allow for storing arbirary objects, if they have `.dtype` `'object'` then `.schema` just reports it as `String`. Maybe we should just do the same for pandas? `.schema` should be quick

MarcoGorelli: This has been really useful, thanks @authierj for highlighting this area for improvement!

There's quite a few fastpaths we can use in Narwhals to reduce the overhead here

For a start, I hadn't realised how expensive `pandas.DataFrame.__getitem__` when the dataframe is backed by a 2D numpy array

Will make some perf improvements in Narwhals, then we can revisit

authierj: Hi @MarcoGorelli, thanks for the changes made in the narwhals library. With some additional improvements on top of yours, `from_narwhals_dataframe` seems to be as fast as `from_dataframe` !

MarcoGorelli: ooh nice, thanks for the update!

authierj:
To compare the performance of the methods `from_dataframe()` (which only accepts pandas DataFrames as input) and `from_narwhals_dataframe()` (which accepts all kinds of DataFrames), I implemented the script `narwhals_test_time.py`, [see here](https://github.com/unit8co/darts/pull/2661/commits/b08a74ffa360e028417c8fc73a3e16dcb34026fd). This script calls the two functions for a large number of different pandas DataFrames, with shuffled or unshuffled data, varying sizes, indices, and datetime formats.

Averaged over 10 runs, the processing times are as follows:

| method    | average processing time [s] |
| -------- | ------- |
| from_dataframe()  | 10.9718 |
| from_narwhals_dataframe() | 9.8564|

Therefore, `from_narwhals_dataframe()` is 1.1154 seconds faster than `from_dataframe()`, representing a 10.17% decrease in processing time on average.

As a consequence of this significant result, I will change the implementation of `from_dataframe()` and also modify `from_series()` to use the narwhals approach.

authierj:
The script used for testing the time performance of the methods is the following: 

```
import argparse
import json
import time
import warnings
from itertools import product

import numpy as np
import pandas as pd
from tqdm import tqdm

# Suppress all warnings
warnings.filterwarnings("ignore")


def test_from_dataframe(f_name: str):
    return getattr(TimeSeries, f_name)


def create_random_dataframes(
    num_rows: int = 10,
    num_columns: int = 3,
    index: bool = True,
    col_names_given: bool = True,
    start_date: str = "1900-01-01",
    freq: str = "D",
) -> tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]:
    """
    Create three pandas DataFrames with random data and dates as the index or as a column.

    Parameters:
    - num_rows (int): The number of rows in the DataFrames.
    - num_columns (int): The number of columns in the DataFrames.
    - index (bool): If True, the date is the index of the DataFrame. If False, the date is a column named 'date'.
    - start_date (str): The start date for the date range (used only if date_format is 'date').
    - freq (str): The frequency of the date range (used only if date_format is 'date').

    Returns:
    - tuple: A tuple containing three DataFrames (df_date, df_numpy, df_integer).
    """
    # Set a random seed for reproducibility
    np.random.seed(42)

    # Generate a date range or integer list based on the date_format parameter
    date_values = pd.date_range(start=start_date, periods=num_rows, freq=freq)
    integer_values = list(range(1, num_rows + 1))
    numpy_values = np.array(
        pd.date_range(start=start_date, periods=num_rows, freq=freq),
        dtype="datetime64[D]",
    )

    # Create random data for the DataFrames
    data = {f"col_{i}": np.random.randn(num_rows) for i in range(num_columns)}

    # Create the DataFrames
    df_date = pd.DataFrame(data)
    df_numpy = pd.DataFrame(data)
    df_integer = pd.DataFrame(data)

    if col_names_given:
        col_names = df_date.columns.values
    else:
        col_names = None

    # Set the date as index or as a column based on the index parameter
    if index:
        df_date.index = date_values
        df_numpy.index = numpy_values
        df_integer.index = integer_values
    else:
        df_date["date"] = date_values
        df_numpy["date"] = numpy_values
        df_integer["date"] = integer_values

    if index:
        time_col = None
    else:
        time_col = "date"

    return [
        [df_date, col_names, time_col],
        [df_numpy, col_names, time_col],
        [df_integer, col_names, time_col],
    ]


def test_dataframes() -> list:
    test_config = product(
        [10, 100, 1000, 10000, 100000],
        [100],
        [True, False],
        [True, False],
    )

    dataframes_list = [
        create_random_dataframes(
            num_rows=num_rows,
            num_columns=num_columns,
            index=index,
            col_names_given=col_names_given,
        )
        for num_rows, num_columns, index, col_names_given in test_config
    ]

    return dataframes_list


def calculate_processing_time(
    f_name: str,
    num_iter: int,
    save_path="data/",
):
    df_list = test_dataframes()
    df_func = test_from_dataframe(f_name)

    # Initialize dictionaries to store processing times
    times = {}

    # Initialize the progress bar
    total_iterations = (
        len(df_list) * 2 * 3
    )  # 2 iterations per dataframe configuration, 3 df per config
    progress_bar = tqdm(total=total_iterations, desc="Processing DataFrames")

    for df_config in df_list:
        for df, col_names, time_col in df_config:
            num_rows = len(df)
            dict_entry = str(num_rows)

            for i in range(2):
                # on the second run we shuffle the data
                if i == 1:
                    df = df.sample(frac=1)
                    dict_entry += "_shuffled"

                begin = time.time()
                for _ in range(num_iter):
                    _ = df_func(df, value_cols=col_names, time_col=time_col, freq=None)
                end = time.time()
                timer = (end - begin) / num_iter

                if dict_entry not in times:
                    times[dict_entry] = timer
                else:
                    times[dict_entry] += timer

                # Update the progress bar
                progress_bar.update(1)

    file_name = f_name + "_avg_time_" + str(num_iter) + "_iter.json"

    # Store the average times in separate JSON files
    with open(save_path + file_name, "w") as f:
        json.dump(times, f, indent=4)


if __name__ == "__main__":
    parser = argparse.ArgumentParser(
        description="The function to test and the number of iter can "
    )
    parser.add_argument(
        "--f_name", type=str, default="from_dataframe", help="method to time"
    )
    parser.add_argument(
        "--n_iter", type=int, default=100, help="number of function call"
    )

    args = parser.parse_args()

    f_name = args.f_name
    n_iter = args.n_iter

    calculate_processing_time(f_name, n_iter)

```

darts/timeseries.py
```diff
@@ -649,97 +651,112 @@ def from_dataframe(
...
+                if time_col_vals.dtype.time_zone is not None:
+                    logger.warning(
+                        "The provided Datetime data was associated with a timezone, which is currently not supported "
+                        "by xarray. To avoid unexpected behaviour, the tz information was removed. Consider calling "
+                        f"`ts.time_index.tz_localize({time_col_vals.dtype.time_zone})` when exporting the results."
+                        "To plot the series with the right time steps, consider setting the matplotlib.pyplot "
+                        "`rcParams['timezone']` parameter to automatically convert the time axis back to the "
+                        "original timezone."
+                    )
+                    time_col_vals = time_col_vals.dt.replace_time_zone(None)
```
dennisbader, line 711: This is duplicated here (from below). We can remember the time zone here and adapt further below to simplify things (see my comment below (C1) on what you can do):

```suggestion
                # remove and remember time zone here as polars converts to UTC
                time_zone = time_col_vals.dtype.time_zone
                if time_zone is not None:
                    time_col_vals = time_col_vals.dt.replace_time_zone(None)
                time_index = pd.DatetimeIndex(time_col_vals)
```

darts/tests/test_timeseries.py
dennisbader, line 1: nice rewriting of the tests :) 🚀 

As mentioned in the `requirements/core.txt` comment, we'll add the polars tests as part of our optional dependecies soon. For the meantime, could you remove "polars" from the parametrization, and remove the import polars?

As soon as the optional dependency checks are up, we can add the references back in another PR :)

dennisbader (changes requested):
Thanks for this great PR @authierj 🚀 

This will be a great addition to Darts :) 

I added a couple of suggestions, mainly on how to further simplify some things here and there, and improve the documentation

darts/timeseries.py
```diff
@@ -960,7 +977,7 @@ def from_group(static_cov_vals, group):
     @classmethod
     def from_series(
         cls,
-        pd_series: pd.Series,
+        pd_series: IntoSeries,
```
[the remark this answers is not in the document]
authierj: Done

hrzn:
This is very cool and I'm sure will make many users' lives easier!

darts/timeseries.py
```diff
@@ -649,91 +651,113 @@ def from_dataframe(
...
 
         # get values
         if value_cols is None:
-            series_df = df.loc[:, df.columns != time_col]
+            series_df = df.drop(time_col) if time_col else df
         else:
             if isinstance(value_cols, str):
                 value_cols = [value_cols]
-            series_df = df[value_cols]
+            series_df = df[value_cols]  # quite slow
```
FBruzzesi, line 662: A bit of an edge case, but pandas columns can also be integers, the safest way to deal with that would be:

```py
if isinstance(value_cols, (str, int)):
    value_cols = [value_cols]
series_df = df.select(nw.col(*value_cols))
```

darts/timeseries.py
```diff
@@ -649,91 +651,113 @@ def from_dataframe(
...
+                    )
                 # Temporarily use an integer Index to sort the values, and replace by a
                 # RangeIndex in `TimeSeries.from_xarray()`
                 time_index = pd.Index(time_col_vals)
 
-            elif np.issubdtype(time_col_vals.dtype, object):
+            elif time_col_vals.dtype == nw.String:
                 # The integer conversion failed; try datetimes
                 try:
                     time_index = pd.DatetimeIndex(time_col_vals)
```
FBruzzesi, line 692: Asking as I don't know: how would this behave with polars/pyarrow backed series?

As long as `time_col_vals` is a narwhals series, you can try to convert it to datetime via:

```py
time_col_vals.str.to_datetime()
```

we will try to infer the date format

darts/timeseries.py
```diff
@@ -649,91 +651,113 @@ def from_dataframe(
         TimeSeries
             A univariate or multivariate deterministic TimeSeries constructed from the inputs.
         """
+        df = nw.from_native(df)
```
FBruzzesi, line 654: Consider flagging this as 
```suggestion
        df = nw.from_native(df, eager_only=True, pass_through=False)
```

The first flag will make sure that the return type is a (eager) DataFrame (mostly for typing purposes) , the second will make sure to raise if the input is not a supported eager frame

darts/timeseries.py
```diff
@@ -649,91 +651,113 @@ def from_dataframe(
...
+            time_col_vals = nw.maybe_get_index(df)
+            if time_col_vals is None:
+                raise_log(
+                    ValueError(
+                        "No time column or index found in the DataFrame. `time_col=None` "
+                        "is only supported for pandas DataFrame which is indexed with one of the "
+                        "supported index types: a DatetimeIndex, a RangeIndex, or an integer "
+                        "Index that can be converted into a RangeIndex.",
+                    ),
+                )
```
FBruzzesi, line 729: Should you consider the value to be `np.arange(len(df))` or is that too big of an assumption?

darts/timeseries.py
```diff
@@ -649,91 +651,113 @@ def from_dataframe(
...
-            series_df = df[value_cols]
+            series_df = df[value_cols]  # quite slow
 
         # get time index
         if time_col:
             if time_col not in df.columns:
                 raise_log(AttributeError(f"time_col='{time_col}' is not present."))
 
-            time_index = pd.Index([])
             time_col_vals = df[time_col]
```
FBruzzesi, line 669: Similarly as above, if the time_col is an integer due to pandas allowing that, the safe route would be to use:

```py
time_col_vals = df.get_column(time_col)
```

darts/timeseries.py
```diff
@@ -1001,7 +1028,8 @@ def from_series(
         TimeSeries
             A univariate and deterministic TimeSeries constructed from the inputs.
         """
-        df = pd.DataFrame(pd_series)
+        nw_series = nw.from_native(pd_series, series_only=True)
```
FBruzzesi, line 1031: Similarly as above, if you want to already raise for non series objects, it's possible to do so by flagging `pass_through=False`

FBruzzesi (commented):
Hey @authierj , I left a few suggestions and considerations in the `from_*` functions! Hope they help :)

dennisbader:
> This is very cool and I'm sure will make many users' lives easier! I think it might be worth updating the docs / quickstart to maybe showcase an example for creating/exporting from/to polars?

Agreed @hrzn :) To any dataframe support will be added in another PR.

darts/timeseries.py
```diff
@@ -649,97 +651,112 @@ def from_dataframe(
...
                 "a DatetimeIndex, a RangeIndex, or an integer Index that can be converted into a RangeIndex",
                 logger,
             )
             # BUGFIX : force time-index to be timezone naive as xarray doesn't support it
             # pandas.DataFrame loses the tz information if it's not its index
-            if isinstance(df.index, pd.DatetimeIndex) and df.index.tz is not None:
+            if (
+                isinstance(time_col_vals, pd.DatetimeIndex)
+                and time_col_vals.tz is not None
+            ):
```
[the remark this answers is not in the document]
authierj: If moved to one lower indent, `isinstance(time_index, pd.DatetimeIndex)` will always be `true`. I moved the block but changed the clause a bit.

darts/timeseries.py
authierj: We do! The condition `np.issubdtype(time_index.dtype, np.integer)` is `True` if the index is `np.arange(len(df))` :)

dennisbader (changes requested):
Very nice @authierj 🚀 This looks great!

Also thanks @FBruzzesi for the additional comments.

Just some minor suggestions, then we're ready.

darts/timeseries.py
FBruzzesi: Yes exactly, non pandas cases would end up raising if `time_col` is not provided. It's a design choice you will have to make, but wanted to point out that that was the case :)

dennisbader (approved):

Issue #2679, opened 2025-02-17
Color choice for multivariate timeseries


authierj:
**Is your feature request related to a current problem? Please describe.**
Currently, it is not possible to choose the color of each component in a multivariate timeserie.  It is only possible to choose a color for all componenents of the timeserie.

**Describe proposed solution**
Enable the user to provide a list of colors (that must be a long as the number of components in the timeserie) to set the colors of the components in the plot.

**Describe potential alternatives**
Alternatively, one could provide the name of a colormap and darts would automatically sample colors from the colormap for the components of the timeserie.

PR #2680, opened 2025-02-17
Color choice for multivariate timeseries

authierj:

Fixes [#2679](https://github.com/unit8co/darts/issues/2679).

### Summary

The user is now able to provide a list of colors to specify the color of the components of the timeseries.


darts/timeseries.py
```diff
@@ -4207,6 +4213,23 @@ def plot(
...
+        else:
+            multi_color = isinstance(color, Sequence)
+
+        if multi_color:
+            raise_if_not(
+                len(color) == self.n_components,
+                "The color argument should have the same length as the number of plotted components "
+                f"({min(self.n_components, n_components_to_plot)}), only {len(color)} labels were provided",
+                logger,
+            )
```
dennisbader, line 4231: We could adapt this slightly to be consistent with how we handle the `label` (combine it with the suggestion further below).

```suggestion
        if not isinstance(color, (str, tuple)) and isinstance(color, Sequence):
            raise_if_not(
                len(color) == self.n_components,
                "The color argument should have the same length as the number of plotted components "
                f"({min(self.n_components, n_components_to_plot)}), only {len(color)} labels were provided",
                logger,
            )
            custom_colors = True
        else:
            custom_colors = False
```

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 ### For users of the library:
 
 **Improved**
+- Extended `TimeSeries.plot()` to accept list or tupple of colors to specify the color of the componenents of a TimeSeries. [#2680](https://github.com/unit8co/darts/pull/2680) by [Jules Authier](https://github.com/authierj)
```
dennisbader, line 13:
```suggestion

- `TimeSeries.plot()` now supports setting the color for each component in the series. Simply pass a list / sequence of colors with length matching the number of components as parameters "c" or "colors". [#2680](https://github.com/unit8co/darts/pull/2680) by [Jules Authier](https://github.com/authierj)
```

dennisbader (changes requested):
Thanks for this PR @authierj 🚀  Looks really good :) 

Just added some minor suggestions, after that we can merge.

darts/timeseries.py
```diff
@@ -4143,13 +4143,21 @@ def plot(
...
+            Optionally, set the line alpha for deterministic series, or the confidence interval alpha for
+            probabilistic series.
+        color
+            Can either be a single color or list of colors. Any matlplotlib colors is accepeted (string, hex string,
+            RGB/RGBA tuple). If a single color and the series only has a single component, it is used as the color
+            for that component. If a single color and the series has multiple components, it is used as the color
+            for each component. If a list of colors with length equal to the number of components in the series, the
+            colors will be mapped to the components in order.
+        c
+            An alias for color
```
madtoinou, line 4160: I would keep these argument "before" `args` and `kwargs`

authierj: I put them here because they are part not arguments of `TimeSeries.plot()` but in the `kwargs` dict. Therefore I thought making a distinction would make the whole cleaner.

I added an indent previously to show that these are inside `kwargs` but @dennisbader did not like it.

Every option is fine for me, let's just agree on one :)

darts/timeseries.py
```diff
@@ -4231,6 +4253,9 @@ def plot(
                     label_to_use = f"{label}_{comp_name}"
             kwargs["label"] = label_to_use
 
+            if custom_colors:
+                kwargs[color_key] = color[i]
```
dennisbader, line 4257:
```suggestion
            kwargs["c"] = color[i] if custom_colors else color
```

dennisbader (changes requested):

dennisbader (approved):
Looks great, thanks a lot @authierj 🚀 

While reviewing, I spotted some typos in the existing warning / error messages, so took the chance to fix those.

I slightly updated the (existing) logic as well to clean things up a bit :)

PR #2684, opened 2025-02-18
add public function untrained_model() in docs

### LINKED ISSUE #2677: [DOC] public function `untrained_model()` not documented, why?
I was looking for a function to reset a model for a while. I did not find anything in the documentation, but in the code there is `ForecastingModel.untrained_model(self)`, so every model should support it?

I'm wondering why it is not documented, as it appears to be an useful and public function? After close inspection, I found out that it is documented but only for the [BlockRNNModel](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.block_rnn_model.html?highlight=untrained_model#darts.models.forecasting.block_rnn_model.BlockRNNModel.untrained_model)

If this function is save to use and available for all model, I would like to make a Pull Request.

turbotimon:

Fixes #2677 

### Summary

Public function `ForecastingModel.untrained_model()` is no in docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
+- Made method `ForecastingModel.untrained_model()` public. Use this method to get a new (untrained) model instance created with the same parameters. [#2684](https://github.com/unit8co/darts/pull/2684) by [Timon Erhart](https://github.com/turbotimon)
+
 **Fixed**
 
 **Dependencies**
--- a/darts/models/forecasting/forecasting_model.py
+++ b/darts/models/forecasting/forecasting_model.py
@@ -2601,7 +2601,7 @@ class ForecastingModel(ABC, metaclass=ModelMeta):
         return model_params
 
     def untrained_model(self):
-        """Returns a new (untrained) model instance create with the same parameters."""
+        """Returns a new (untrained) model instance created with the same parameters."""
         return self.__class__(**copy.deepcopy(self.model_params))
 
     @property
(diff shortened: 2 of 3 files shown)


dennisbader:
Thanks for the PR @turbotimon 🚀 Did you locally install darts before generating the docs?

```bash
# from the repo root
pip install .
# then you can generate the docs
make --directory=./docs build-all-docs
```

turbotimon:
> Thanks for the PR @turbotimon 🚀 Did you locally install darts before generating the docs?
> 
> ```shell
> # from the repo root
> pip install .
> # then you can generate the docs
> make --directory=./docs build-all-docs
> ``
Yes i did, but then it's a problem on my site don't worry.

turbotimon:
@dennisbader I was able to build the docs successfully with a fresh venv. However, would you agree to these changes? I would then make a quick PR

https://github.com/unit8co/darts/blob/cda8f8148db6f28b0fef895f097927af3330fb4e/CONTRIBUTING.md?plain=1#L62-L71

- mention pandoc dependency
- change to editable mode (so no need to install it again and again)

```bash
# install pandoc. For other system see https://pandoc.org/installing.html
sudo apt-get install pandoc
# install the package editable
pip install -e .
# build the docs
make --directory=./docs build-all-docs
```

dennisbader:
@turbotimon, yes mentioning the pandoc dependency sounds good. Maybe just mention it in the comment to be system agnostic?

```
# make sure `pandoc` is installed (see https://pandoc.org/installing.html)
# then, install darts locally in develop mode
pip install -e .
# build the docs
make --directory=./docs build-all-docs
```

turbotimon:
@dennisbader yes it worked, I confirmed it in a fresh venv. You may try this as well or we leave it as it was (however, normally editable mode is very comfortable also for development).

Anyway, her is the PR #2698, i made pandoc deb system agnostic.

PR #2688, opened 2025-02-20
DOC fix install guide link in 00-quickstart.ipynb

turbotimon:
### Summary

`#installation-guide`  anchor does not exist. With the fix, it will point to the same target as in:

https://github.com/turbotimon/darts/blob/3a0f1f1474dc6b899daa5e4acc2a60b519794d3b/README.md?plain=1#L63


dennisbader (approved):
Thanks for this fix @turbotimon 🚀

Issue #2689, opened 2025-02-20
`to_dataframe(backend)` and `to_series(backend)` extension


authierj:
**Is your feature request related to a current problem? Please describe.**
[#2661](https://github.com/unit8co/darts/pull/2661/) makes Darts user inputs dataframe agnostic. It would be nice to close the loop and extend the current `TimeSeries.pd_dataframe()` and `TimeSeries.pd_series()` to any dataframe library (pandas, polars, arrow,. ..).

**Describe proposed solution**
Two new methods, `to_dataframe(backend)` and `to_series(backend)` where the `backend` is in ["pandas", "polars", "arrow"]. The most elegant solution would be using narwhals (https://github.com/narwhals-dev/narwhals) but handling the `pd.DatetimeIndex` may get a bit tricky!

**Describe potential alternatives**
A first naive approach would be to keep the whole processing in pandas and convert the dataframe to the desired backend.

PR #2701, opened 2025-02-27
`to_dataframe(backend)` and `to_series(backend)`

authierj:

Fixes [#2689](https://github.com/unit8co/darts/issues/2689).

### Summary

Created `to_dataframe(backend)` and `to_series(backend)` method, where the user can choose to receive a pandas/polars/pyarrow dataframe or series.

### Other Information

The proposed solution is heavily inspired by the proposition of @FBruzzesi provided [here](https://github.com/narwhals-dev/narwhals/issues/2056#issuecomment-2676143397)


darts/timeseries.py
```diff
@@ -1545,40 +1546,71 @@ def data_array(self, copy=True) -> xr.DataArray:
...
+        ----------
+        copy
+            Whether to return a copy of the series. Leave it to True unless you know what you are doing.
+
+        Returns
+        -------
+        pandas.Series
+            A Pandas Series representation of this univariate time series.
+        """
+        logger.warning("`pd_series` is deprecated, please use `to_series` instead")
```
dennisbader, line 1602:
```suggestion
        logger.warning("`TimeSeries.pd_series()` is deprecated and will be removed in a future version. Use `TimeSeries.to_series()` instead")
```

darts/timeseries.py
```diff
@@ -1606,36 +1650,62 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
+            Whether to return a copy of the dataframe. Leave it to True unless you know what you are doing.
+
+        Returns
+        -------
+        pandas.DataFrame
+            The Pandas DataFrame representation of this time series
+        """
+
+        logger.warning(
+            "`pd_dataframe` is deprecated, please use `to_dataframe` instead"
```
dennisbader, line 1701:
```suggestion
            "`TimeSeries.pd_dataframe()` is deprecated, and will be removed in a future version. Use `TimeSeries.to_dataframe()` instead"
```

darts/timeseries.py
```diff
@@ -1606,36 +1650,62 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
-                    columns=self._xa.get_index(DIMS[1]),
-                )
+            data = self._xa[:, :, 0].values
+            columns = self._xa.get_index(DIMS[1])
+
+        time_index = self._time_index
+
+        if time_as_index:
+            # special path for pandas with index
+            import pandas as pd
```
dennisbader, line 1666: pandas is already imported
```suggestion
```

darts/timeseries.py
```diff
@@ -1545,40 +1546,71 @@ def data_array(self, copy=True) -> xr.DataArray:
         """
         return self._xa.copy() if copy else self._xa
 
-    def pd_series(self, copy=True) -> pd.Series:
+    def to_series(
+        self,
+        copy: bool = True,
+        backend: Literal["pandas", "polars", "pyarrow"] = "pandas",
```
dennisbader, line 1552: Is it only these three backends that are supported? According to [their docs](https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.from_dict) it's more than those :) 

Also let's try update the parameter descriptions for backend here and in `to_dataframe()` similar to how we did it in the `from_*` methods.

darts/timeseries.py
```diff
@@ -1589,12 +1621,24 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
-        pandas.DataFrame
-            The Pandas DataFrame representation of this time series
+        DataFrame
+            A DataFrame representation of this time series.
         """
+
+        if time_as_index and backend != "pandas":
+            raise_log(
+                ValueError(
+                    "The `time_as_index` parameter is only supported for the pandas backend."
```
dennisbader, line 1637:
```suggestion
                    "`time_as_index=True` is only supported with `backend="pandas"`.
```

darts/timeseries.py
```diff
@@ -1606,36 +1650,62 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
-            else:
-                return pd.DataFrame(
-                    self._xa[:, :, 0].values,
-                    index=self._time_index,
-                    columns=self._xa.get_index(DIMS[1]),
-                )
+            data = self._xa[:, :, 0].values
+            columns = self._xa.get_index(DIMS[1])
+
+        time_index = self._time_index
```
dennisbader, line 1662: copy here similar as in the other PR

```suggestion
        time_index = self._time_index

        if copy:
            data = data.copy()
            time_index = time_index.copy()
```

darts/timeseries.py
```diff
@@ -1606,36 +1650,62 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
+        if time_as_index:
+            # special path for pandas with index
+            import pandas as pd
+
+            return pd.DataFrame(data=data, index=time_index, columns=columns)
+
+        data_ = {
+            time_index.name: time_index,  # set time_index as left-most column
+            **{col: data[:, idx] for idx, col in enumerate(columns)},
+        }
```
dennisbader, line 1673: you can overwrite `data`

```suggestion
        data = {
            time_index.name: time_index,  # set time_index as left-most column
            **{col: data[:, idx] for idx, col in enumerate(columns)},
        }
```

darts/timeseries.py
```diff
@@ -1545,40 +1546,71 @@ def data_array(self, copy=True) -> xr.DataArray:
...
+        """
+        logger.warning("`pd_series` is deprecated, please use `to_series` instead")
+        return self.to_series(copy=copy, backend="pandas")
+
+    def to_dataframe(
+        self,
+        copy: bool = True,
+        backend: Literal["pandas", "polars", "pyarrow"] = "pandas",
+        time_as_index: bool = False,
+        suppress_warnings: bool = False,
```
dennisbader, line 1610: can you add a param description for `suppress_warnings` (seems we forgot about that)

darts/timeseries.py
```diff
@@ -1545,40 +1546,71 @@ def data_array(self, copy=True) -> xr.DataArray:
...
+            A Pandas Series representation of this univariate time series.
+        """
+        logger.warning("`pd_series` is deprecated, please use `to_series` instead")
+        return self.to_series(copy=copy, backend="pandas")
+
+    def to_dataframe(
+        self,
+        copy: bool = True,
+        backend: Literal["pandas", "polars", "pyarrow"] = "pandas",
+        time_as_index: bool = False,
```
dennisbader, line 1609: I'd prefer to have `time_as_index=True` as default, since our users are used to `pd_dataframe()`, and that was the default there

dennisbader (commented):
Very nice again @authierj :) 

Looks already pretty good:
- had some minor suggestions regarding the docs to be more similar to the other PR
- let's merge the other PR and then:
  - add unit tests
  - add  changelog entry

darts/timeseries.py
authierj: sure!

authierj: would this be fine or do you want something more generic?

``` python
"""
suppress_warnings
            Whether to suppress the warning about transforming a stochastic TimeSeries
"""
```

darts/timeseries.py
authierj: That's right, they have more!

darts/timeseries.py
```diff
@@ -1612,12 +1648,26 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
+            Whether to suppress the warning about transforming a stochastic TimeSeries
         Returns
         -------
-        pandas.DataFrame
-            The Pandas DataFrame representation of this time series
+        DataFrame
+            A DataFrame representation of this time series.
         """
+
+        if time_as_index and backend != "pandas":
```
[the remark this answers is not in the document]
authierj: agreed!

darts/timeseries.py
```diff
@@ -1633,18 +1683,59 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
                 "_s".join((comp_name, str(sample_id)))
                 for comp_name, sample_id in itertools.product(comp_name, samples)
             ]
-            data = self._xa.stack(data=(DIMS[1], DIMS[2])).values
+            data = self._xa.stack(data=(DIMS[1], DIMS[2]))
```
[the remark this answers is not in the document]
authierj: indeed

darts/tests/test_timeseries.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy as np
 import pandas as pd
+import polars as pl
```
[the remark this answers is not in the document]
authierj: The syntax changes are a lot more complex than in the `from_*` PR because we handle different kind of dataframes (polars, pandas, etc). I was wondering how far we want to push this testing.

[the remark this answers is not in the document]
dennisbader: You are right we can keep these tests to a minimum and it should be covered by your `test_polars_creation`. 

We still need to handle the missing polars import here.

```
try:
    import polars as pl
    POLARS_AVAILABLE = False
except ModuleNotFoundError:
    pl = None
    POLARS_AVAILABLE = False
```

then in you can skip the `test_polars_creation` if it's not available. See my comment below

darts/timeseries.py
```diff
@@ -1599,9 +1619,40 @@ def pd_series(self, copy=True) -> pd.Series:
...
+            logger,
+        )
+        return self.to_series(copy=copy, backend="pandas")
+
+    def to_dataframe(
+        self,
+        copy: bool = True,
+        backend: ModuleType
+        | Implementation
+        | Literal["pandas", "polars", "pyarrow", "modin", "cudf"] = "pandas",
```
dennisbader, line 1650:
```suggestion
        backend: Union[ModuleType, Implementation, str] = "pandas",
```

darts/timeseries.py
```diff
@@ -1570,25 +1578,37 @@ def data_array(self, copy=True) -> xr.DataArray:
         """
         return self._xa.copy() if copy else self._xa
 
-    def pd_series(self, copy=True) -> pd.Series:
+    def to_series(
+        self,
+        copy: bool = True,
+        backend: ModuleType
+        | Implementation
+        | Literal["pandas", "polars", "pyarrow", "modin", "cudf"] = "pandas",
```
dennisbader, line 1586: using a literal means that we would have to keep track of what backends narwhals adds. Let's use a simple `str` instead (as they also do in their implementation). The curious user can read about it in the linked documentation

```suggestion
        backend: Union[ModuleType, Implementation, str] = "pandas",
```

darts/timeseries.py
```diff
@@ -1612,12 +1663,30 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
+            A DataFrame representation of this time series.
         """
+
+        if time_as_index and backend != "pandas":
+            raise_log(
+                Warning(
+                    '`time_as_index=True` is only supported with `backend="pandas"`.'
+                ),
+                logger,
+            )
```
dennisbader, line 1687: this would raise an error, we only want a warning

```suggestion
            logger.warning('`time_as_index=True` is only supported with `backend="pandas"`.')
            time_as_index = False
```

darts/timeseries.py
```diff
@@ -1599,9 +1619,40 @@ def pd_series(self, copy=True) -> pd.Series:
...
+        -------
+        pandas.Series
+            A Pandas Series representation of this univariate time series.
+        """
+        raise_deprecation_warning(
+            "`TimeSeries.pd_series()` is deprecated and will be removed in a future version. Use "
+            "`TimeSeries.to_series()` instead",
+            logger,
+        )
+        return self.to_series(copy=copy, backend="pandas")
```
dennisbader, line 1643:
```suggestion
        return self.to_series(copy=copy)
```

darts/timeseries.py
```diff
@@ -1637,14 +1706,56 @@ def pd_dataframe(self, copy=True, suppress_warnings=False) -> pd.DataFrame:
...
+            "`TimeSeries.pd_dataframe()` is deprecated, and will be removed in a future version. Use "
+            "`TimeSeries.to_dataframe()` instead",
+            logger,
+        )
+        return self.to_dataframe(
+            copy=copy,
+            backend="pandas",
+            time_as_index=True,
+            suppress_warnings=suppress_warnings,
+        )
```
dennisbader, line 1758:
```suggestion
        return self.to_dataframe(
            copy=copy,
            suppress_warnings=suppress_warnings
        )
```

darts/timeseries.py
```diff
@@ -4078,7 +4190,7 @@ def to_csv(self, *args, **kwargs):
                 )
             )
 
-        self.pd_dataframe().to_csv(*args, **kwargs)
+        self.to_dataframe().to_csv(*args, **kwargs)
```
dennisbader, line 4193: can you change `pd_dataframe()` in the above AssertionError?

dennisbader (approved):
Very nice @authierj and congrats for this great PR 🚀

PR #2710, opened 2025-03-03
Fix/inverse transform for static covariate with single category across series

### LINKED ISSUE #2705: [BUG] A StaticCovariatesTransformer fit to multiple categorical static covariates with the same constant value is uninvertable
**Describe the bug**
Multiple categorical static covariates with the same, single value are mapped to the same encoded covariate, which results in the inverse transformation failing. This seems to be because zero-cardinality static covariates are mapped to a static covariate with just the category name, rather than `{covariate_name}_{category_name}`, and therefore the column mapping collides. 

This may seem like a contrived scenario, but is more plausible than it looks at first glance and could produce a difficult-to-debug error when, for example, testing a pipeline on a small subset of time serieses. 

**To Reproduce**
```python
ts1 = TimeSeries.from_values(np.random.randn(10))
ts1 = ts1.with_static_covariates(pd.Series({
    "cov_a": 'foo',
    "cov_b": 'foo',
    "cov_c": 'foo',
}))
ts2 = TimeSeries.from_values(np.random.randn(10))
ts2 = ts2.with_static_covariates(pd.Series({
    "cov_a": 'foo',
    "cov_b": 'foo',
    "cov_c": 'bar',
}))

ohe = OneHotEncoder()
transformer = StaticCovariatesTransformer(transformer_cat=ohe)

transformer.fit([ts1, ts2])
ts1_enc, ts2_enc = transformer.transform([ts1, ts2])

ts1_enc.static_covariates, ts2_enc.static_covariates
# (static_covariates  foo  cov_c_bar  cov_c_foo
#  component                                   
#  0                  1.0        1.0        0.0,
#  static_covariates  foo  cov_c_bar  cov_c_foo
#  component                                   
#  0                  1.0        1.0        1.0)_

ts1_env_inv, ts2_enc_inv = transformer.inverse_transform([ts1_enc, ts2_enc])
# IndexError: boolean index did not match indexed array along dimension 1; dimension is 3 but corresponding boolean dimension is 4

```
**Expected behavior**
The transformation is invertible. 

- Python version 3.10.14
 - darts version [e.g. 0.33.0]

madtoinou:

Fixes #2705.

### Summary

- Each covariates is referred to by the combination of covariates name and category name, regardless of the number of categories (used to be for covariates with more than 1 category).

### Other Information

This might be a breaking changes for some as it can change the number of covariates in the series but very unlikely in my opinion.


dennisbader (approved):
Very nice fix, thanks a lot @madtoinou 🚀 🚀 🚀

Issue #2697, opened 2025-02-25
[QUESTION] Single lag values in linear regression causing trend and seasonal backtests to have different periods when retrain=False


gvas7:
I am trying to use RegressionModel forecasting to break out a 96 month period between Trend and Seasonality.  I forecast them separately, and then add back together.

The issue I am running into is that when I try to combine the backtests of the two (not sure this is the best approach but it's what I'm doing), I'm getting two different periods if I use a single lag of -12 and retrain=False.  If I use multiple lags, say [-1,-12] or I set retrain to true I'm ok.  My very basic code without all the bells and whistles is:

```python
## ------------ Univariate, Seasonality ---------------
#build model
linear_seasonal_model = RegressionModel(model=LinearRegression(), lags=[-12], add_encoders=None, lags_future_covariates=None)
#fit
linear_seasonal_model.fit(series=Y_seasonality_train,future_covariates=None)
#predict
linear_seasonal_pred = linear_seasonal_model.predict(n=12, future_covariates=None)
#backtest
# linear_seasonal_backtest = linear_seasonal_model.historical_forecasts(series=Y_seasonality_train, start=pd.Timestamp('2022-04-01'), retrain=False)
linear_seasonal_backtest = linear_seasonal_model.historical_forecasts(series=Y_seasonality_train, start=24, retrain=False, forecast_horizon=1)
## ----------------------------------------------------

## ------------- Multivariate, Trend ------------------
#build model
linear_trend_model = RegressionModel(model=LinearRegression(), lags=[-12], lags_future_covariates=[0], add_encoders=None)
#fit
linear_trend_model.fit(series=Y_trend_train, future_covariates=X_scaled)
#predict
linear_trend_pred = linear_trend_model.predict(n=12, future_covariates=X_scaled)
#backtest
# linear_trend_backtest = linear_trend_model.historical_forecasts(series=Y_trend_train, future_covariates=X_scaled, start=pd.Timestamp('2022-04-01'), retrain=False)
linear_trend_backtest = linear_trend_model.historical_forecasts(series=Y_trend_train, future_covariates=X_scaled, start=24, retrain=False, forecast_horizon=1)
## ----------------------------------------------------

## --------------- Combined Forecast ------------------
#combine forecasts
linear_forecast = linear_seasonal_pred + linear_trend_pred
#combine backtests
linear_backtest = linear_seasonal_backtest + linear_trend_backtest
#unscale
[...]

PR #2715, opened 2025-03-05
Fixes bug when using historical forecast with discrete lags < -1

authierj:

Fixes [#2697](https://github.com/unit8co/darts/issues/2697).

### Summary

When regression models had discrete lags < -1, the end of the window of the target series for the historical forecasts was not corrected correctly. I fixed this.

### Other Information

Nan


testing2.py
madtoinou, line 1: I think that you pushed this file by accident.

darts/utils/historical_forecasts/utils.py
```diff
@@ -1020,7 +1020,13 @@ def _get_historical_forecast_boundaries(
     hist_fct_tgt_start, hist_fct_tgt_end = historical_forecasts_time_index
     if min_target_lag is not None:
         hist_fct_tgt_start += min_target_lag * freq
-    hist_fct_tgt_end -= 1 * freq
+
+    # target lag has a gap between the max lag and the present
+    if hasattr(model, "lags") and model._get_lags("target") and not overlap_end:
```
madtoinou, line 1025: The docstring of `_adjust_historical_forecasts_time_index` is incorrect, we should remove the fact that it handles `overlap_end` (I was a bit worried that this argument was treated/taken into account twice but all good, we can keep your code as is).

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -132,6 +134,12 @@
         {},
         (6, 2),
     ),
+    (
+        LinearRegressionModel,
+        {"lags": [-3], "lags_past_covariates": 6, "lags_future_covariates": [0, 1]},
+        {},
+        (6, 2),
```
madtoinou, line 141: Did this model definition not pass the tests prior to your changes?

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
```diff
@@ -72,6 +72,8 @@
         {},
         (5, 3),
     ),
+    (LinearRegressionModel, {"lags": [-5]}, {}, (5, 1)),
+    (LinearRegressionModel, {"lags": [-5], "output_chunk_shift": 1}, {}, (5, 2)),
```
madtoinou, line 76: Nice!

madtoinou (commented):
Nice PR, a few adjustments to make and it should be ready.

> I don't really know how/where to test the overlap_end = True. A suggestion would be welcome!

You can create a new test, just for the `_get_historical_forecast_boundaries` method or eventually in one of the test for regression model optimized historical forecasts for example.

testing2.py
authierj: this is correct

darts/tests/utils/historical_forecasts/test_historical_forecasts.py
authierj: it did actually, good point. Thank you!

darts/utils/historical_forecasts/utils.py
authierj: it also doesn't handle `forecast_horizon` !

darts/utils/historical_forecasts/utils.py
```diff
@@ -782,8 +782,7 @@ def _adjust_historical_forecasts_time_index(
     show_warnings: bool,
 ) -> TimeIndex:
     """
-    Shrink the beginning and end of the historical forecasts time index based on the values of `start`,
-    `forecast_horizon` and `overlap_end`.
+    Shrink the beginning and end of the historical forecasts time index based on the value of `start`.
```
madtoinou, line 785: Perfect, can you please also update the comment above the call the `_adjust_historical_forecasts_time_index()` in `_get_historical_forecast_boundaries` as well?

dennisbader (approved):
Looks great, thanks a lot for the fix @authierj 🚀

PR #2757, opened 2025-04-03
Fix NLinear normalization

### LINKED ISSUE #2699: [BUG] Several issues with NLinear with `normalize=True`
**Describe the bug**

After reading the paper and having a closer look at the implementation of `DLinearModel`, I see several issues:

1. The `seq_last` is only detached, but needs also to be cloned as well. Otherwise, it it's values will be changed by line https://github.com/unit8co/darts/blob/8bea0713494e6d23321feaaec341fff4e6fb33dc/darts/models/forecasting/nlinear.py#L149 . This is because of the tensors still share the same underlying memory. The solution is: `seq_last = x[:, -1:, : self.output_dim].detach().clone()`

2. Why is the normalization disabled by default? That's the whole catch of this model, so it should be enabled by default. (I could only imagine that it performed so bad due to the bug, that it was disabled by default)

3. The normalization is only done for the target and not covariates (https://github.com/unit8co/darts/blob/8bea0713494e6d23321feaaec341fff4e6fb33dc/darts/models/forecasting/nlinear.py#L145). How I understand it, the original implementation from the paper does it also for the covariates (https://github.com/cure-lab/LTSF-Linear/blob/0c113668a3b88c4c4ee586b8c5ec3e539c4de5a6/models/NLinear.py#L27). And in my opinion, this makes totally sense. For example, it is done also 

**To Reproduce**

With normalize=True, NLinear currently fails to solve even very simple tasks:

```
from darts import TimeSeries, models
ts = linear_timeseries(start_value=0, end_value=10, length=10, freq='D')
model = models.NLinearModel(
    input_chunk_length=3,
    output_chunk_length=1,
    n_epochs=10,
    batch_size=1,
    const_init=True,
    normalize=True # Explore the difference when durning on normalize
    )
model.fit(ts)
print(model.predict(1))
```

**Expected behavior**

Output around ~10.0

 - Python version: 3.12
 - darts version 0.33.0

**Additional context**

turbotimon:

Fixes #2699.

### Summary

- Fixes the normalization also for the `shared_weighs=True` path (not covered by #2724)
- Uses `normalize=True` as the default now with an updates docstring.
- Past covariates are now also normalized. But no future covariates to allow further use for temporal characteristics (e.g. via add_encoders)
- Improved tests

Full discussion and details, see #2699.


turbotimon:
Some test still fail, and I'm not sure why. They can be fixed by changing the model to `normalize=False`, but I'm not sure why.

For example `test_lagged_training_data_unequal_freq` fails because the values have a differend order?:

```
expected_X.squeeze()
> array([23.33333333, 24.33333333, 25.33333333, 26.33333333, 25.55555556, 26.55555556, 27.55555556, 28.55555556])
X.squeeze()
> array([23.33333333, 24.33333333, 26.33333333, 25.33333333, 26.55555556, 25.55555556, 27.55555556, 28.55555556])

File: darts/models/forecasting/nlinear.py [hunk 1]
```diff
@@ -125,7 +126,7 @@ def forward(
             x = x.permute(0, 2, 1)  # (batch, out_dim, in_len)
 
             if self.normalize:
-                seq_last = x[:, :, -1:].detach()  # (batch, out_dim, 1)
+                seq_last = x[:, :, -1:].detach().clone()  # (batch, out_dim, 1)
```
dennisbader, line 129: thanks!

darts/models/forecasting/nlinear.py
```diff
@@ -171,10 +173,12 @@ def forward(
                     batch, self.output_chunk_length, self.output_dim * self.nr_params
                 )
 
-            x = x.view(batch, self.output_chunk_length, self.output_dim, self.nr_params)
             if self.normalize:
-                # model only forecasts target components, no need to slice
-                x = x + seq_last.view(seq_last.shape + (1,))
+                # Reverse the normalization for the target
```
dennisbader, line 177: 👍

darts/models/forecasting/nlinear.py
```diff
@@ -140,10 +141,11 @@ def forward(
             x = x.permute(0, 2, 1, 3)
         else:
             if self.normalize:
-                # get last values only for target features
-                seq_last = x[:, -1:, : self.output_dim].detach().clone()
-                # normalize the target features only (ignore the covariates)
-                x[:, :, : self.output_dim] = x[:, :, : self.output_dim] - seq_last
+                # get last values for all x but not future covariates
```
dennisbader, line 144: Thanks for giving this a go. However, I believe we should find a more general solution that we discussed in the issue.

Every change that we introduce here will be a "breaking" change for users that want to load the weights of a model stored with an earlier Darts version (in the sense that the weights were trained on an input with different value space).

So if we also want to add this for future covariates (the non-datetime attrs) then we would have to "break" it again.

In my opinion, either we find a robust solution that covers everything, or we leave the normalization only on the target for now.

By robust I'm thinking about the RINorm changes that we discussed in #2758. That would also allow to use the old behavior (target-only) and optionally also have fine-grained control over covariates normalization.

How does that sound?

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -134,10 +134,20 @@
...
         50.0,
     ),
+    (
+        NLinearModel,
+        {
+            "n_epochs": 10,
+            "normalize": True,
+            "pl_trainer_kwargs": tfm_kwargs["pl_trainer_kwargs"],
+        },
+        80.0,
```
dennisbader, line 149: I was curious why this would have larger error than the non-normalized one, so I tried it out (I ran a simplification of the test that was failing with `50`, e.g. `test_single_ts`, ...):


```
import numpy as np
from darts.models import NLinearModel

model = NLinearModel(
    input_chunk_length=24,
    output_chunk_length=12,
    random_state=0,
    n_epochs=100,
    normalize=True,
)
model.fit(series[:-36])
preds = model.predict(n=36, series=series[:-36])
series.plot()
preds.plot()
plt.show()
```

Seems like auto-regression is not performing as expected (this also happens in the current `master`):

image

darts/models/forecasting/nlinear.py
```diff
@@ -230,11 +234,15 @@ def __init__(
...
-            in subtracting the last value of the input sequence from the input sequence. Default: False.
+            in subtracting the last value of the input sequence from the input sequence. This will
+            be done for the target series and past covariates, but not future covariates because it would defeat the use
+            of encoders (see ``add_encoders``).
+            Without normalization, the models behaves like the simple Linear model proposed in the paper. Default: True.
 
             .. note::
                 This cannot be applied to probabilistic models.
             ..
+
```
dennisbader, line 245:
```suggestion
```

dennisbader (changes requested):
Thanks for PR and the fixes @turbotimon 🚀  I agree with setting `normalize=True` as default.

I left some comments regarding the past covariates normalization, where I believe we should try to find on a more long-term solution.

Also, it looks like there is still something fishing going on with `normalize=True` and auto-regression (forecast horizon `n` > `output_chunk_length`). This is already in the current `master` head, so not introduced by this PR. I left a comment in the tests. Would you mind having a look at that?

darts/models/forecasting/nlinear.py
turbotimon: I agree do some extent, but think we shouldn't overthink this model. It is, as the author says an "embarrassingly simple model" and not sure how many extra features it can handle. Enable the normalization for covariates is just closest to the paper. Except maybe for future (at least for the real future part, not historic future) covariates which are handled by a separate layer anyway in this implementation, so it's ok to handle it different. So I'm not sure if it's a good idea making the implementation more and more complicated if the model most likely can not handle it..

darts/tests/models/forecasting/test_global_forecasting_models.py
dennisbader: For me setting the role_size=12 works fine (still the issue though with auto-regression). Also, the role_size is to `output_chunk_length` if the user doesn't provide it. See here:

https://github.com/unit8co/darts/blob/836436d5e8f4e228d6dfcbb4b6bac12791cda7df/darts/models/forecasting/torch_forecasting_model.py#L1626

turbotimon: Great! Yes, now I see why it is necessary to clone x as well. Two things:
1. Maybe a comment for `x.clone()` is reasonable because we have seen this is not obvious?
2. I think `detach()` (even if it doesn't change the outcome) is a helpful optimization because it gets excluded from the differentiation graph. See https://pytorch.org/docs/stable/generated/torch.clone.html

dennisbader (approved):
Hi @turbotimon, I added the related to cloning `x`. 
Everything is ready now. Thanks again for the great PR and apologies that this took a while :) 🚀

PR #2789, opened 2025-04-23
Fix the RegressionEnsembleModel error with output_chunk_shift > 0

### LINKED ISSUE #2773: [QUESTION] Using EnsembleRegressionModel with output_chunk_shift
I'm trying to use EnsembleRegressionModel to improve the multi-step predictions. Right now, the two forecasting models I'm using are LightGBMModel and XGBoostModel. Their corresponding model and hyper paramters are listed below:

```python
PRICE_LGBM_PARAMS = {
    "lags": [-1, -12, -24],
    "lags_past_covariates": 2,
    "lags_future_covariates": [1, 3, 6, 9, 12],
    "output_chunk_length": 24,
    "output_chunk_shift": 13,
    "add_encoders": {
        "datetime_attribute": {
            "future": ["hour"],
            "past": ["hour"],
        },
        "tz": TIME_ZONE,
    },
    "n_estimators": 361,
    "learning_rate": 0.01,
    "max_depth": 3,
    "num_leaves": 61,
    "feature_fraction": 0.96,
    "verbose": -1,
}

PRICE_XGB_PARAMS = {
    "lags": 1,
    "lags_past_covariates": 1,
    "lags_future_covariates": [1],
    "output_chunk_length": 24,
    "output_chunk_shift": 13,
    "add_encoders": {
        "datetime_attribute": {"past": ["hour"]},
        "position": {"past": ["relative"], "future": ["relative"]},
        "transformer": Scaler(),
        "tz": TIME_ZONE,
    },
    "multi_models": True,
}
```

The next thing I did is following:

```python
models = [lgbm_model, xgb_model]
# The two models above have been fitted already
regression_model = LinearRegressionModel(lags=None, lags_past_covariates=None, lags_future_covariates=[0],
                                                 multi_models=True, output_chunk_length=24)

ensemble_model = RegressionEnsembleModel(
    forecasting_models=models,
    regression_model=regression_model,
    regression_train_n_points=-1,
    train_forecasting_models=False,
    train_using_historical_forecasts=True,
      )

ensemble_model.fit(
    series=processed_data["price_ts"],
    past_covariates=processed_data["past_covs"],
    future_covariates=processed_data["train_future_covs"],
)
```

The error message I got is 
```ValueError: Cannot perform auto-regression (n > output_chunk_length) with a model that uses a shifted output chunk (output_chunk_shift > 0)
```

I'm wondering how to fix this. Thanks for your help!

cnhwl:

Fixes #2773

### Summary
When forecasting model `output_chunk_shift` > 0 and RegressionEnsembleModel `regression_train_n_points` == -1 or some large number, it would occur the forecasting model predict error:
```
        if self.output_chunk_shift and is_autoregression:
            raise_log(
                ValueError(
                    "Cannot perform auto-regression `(n > output_chunk_length)` with a model that uses a "
                    "shifted output chunk `(output_chunk_shift > 0)`."
                ),
                logger=logger,
            )
```
Therefore, I try to limit the RegressionEnsembleModel `regression_train_n_points` to be not bigger than the forecasting model `output_chunk_length`. Moreover, not bigger than ((the forecasting model `output_chunk_length`) minus (the max lag of the forecasting model)).

### DIFF
--- a/darts/models/forecasting/regression_ensemble_model.py
+++ b/darts/models/forecasting/regression_ensemble_model.py
@@ -318,12 +318,36 @@ class RegressionEnsembleModel(EnsembleModel):
 
         # spare train_n_points points to serve as regression target
         is_single_series = isinstance(series, TimeSeries)
+
+        # check if the any forecasting model has output_chunk_shift > 0
+        has_output_chunk_shift = any([
+            model.output_chunk_shift > 0 for model in self.forecasting_models
+        ])
+
+        if has_output_chunk_shift:
+            # get the minimum output_chunk_length of the forecasting models
+            min_forecasting_model_output_chunk_length = min([
+                model.output_chunk_length for model in self.forecasting_models
+            ])
+
+        # ensure every train_n_points not greater than the output_chunk_length when has_output_chunk_shift
         if self.train_n_points == -1:
             if is_single_series:
-                train_n_points = [len(series)]
+                train_n_points = (
+                    [min(len(series), min_forecasting_model_output_chunk_length)]
+                    if has_output_chunk_shift
+                    else [len(series)]
+                )
             else:
                 # maximize each series usage
-                train_n_points = [len(ts) for ts in series]
+                train_n_points = (
+                    [
+                        min(len(ts), min_forecasting_model_output_chunk_length)
+                        for ts in series
+                    ]
+                    if has_output_chunk_shift
+                    else [len(ts) for ts in series]
+                )
 
             # shift by the forecasting models' largest input length
             all_shifts = []
@@ -373,6 +397,19 @@ class RegressionEnsembleModel(EnsembleModel):
             logger,
         )
 
+        raise_if(
+            has_output_chunk_shift
+            and self.regression_model.output_chunk_length
+            > min_forecasting_model_output_chunk_length - input_shift,
+            f"the `output_chunk_length` of the regression model {self.regression_model.output_chunk_length} is too big "
+            "when `output_chunk_shift` > 0 in the forecasting models "
+            "(must be strictly smaller than or equal to the min `output_chunk_length` of "
+            f"the forecasting models {min_forecasting_model_output_chunk_length} minus the max lag of "
+            f"the forecasting models {input_shift} which equals to "
+            f"{min_forecasting_model_output_chunk_length - input_shift})",
+            logger,
+        )
+
         if is_single_series:
             forecast_training = series[: -self.train_n_points]
             regression_target = series[-self.train_n_points :]
@@ -407,7 +444,7 @@ class RegressionEnsembleModel(EnsembleModel):
             num_samples=self.train_num_samples,
         )
 
-        if self.train_using_historical_forecasts:
+        if self.train_using_historical_forecasts and not has_output_chunk_shift:
             predictions = self._make_multiple_historical_forecasts(
                 train_n_points=self.train_n_points,
                 series=series,
--- a/darts/models/forecasting/ensemble_model.py
+++ b/darts/models/forecasting/ensemble_model.py
@@ -267,7 +267,9 @@ class EnsembleModel(GlobalForecastingModel):
         predictions = [
             model._predict_wrapper(
                 n=n,
-                series=series,
+                series=series[: -model.output_chunk_shift]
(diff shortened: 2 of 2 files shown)


dennisbader (changes requested):
Thanks for giving this a go @cnhwl. However, I think we need to adapt the proposed solution.

Here are some points:
- It should be possible to train an ensemble model on base forecasting models that use an output_chunk_shift. Requirements:
  - All models must use the same `output_chunk_shift` value.
  - All models must use the same `output_chunk_length` value.
  - In case of base models using `output_chunk_shift`, the actual `regression_model` (the ensemble model) must also use the same `output_chunk_shift`. In that case we need to check that the future covariates lags for `regression_model` are `{"future": [output_chunk_shift]}` (see [here](https://github.com/unit8co/darts/blob/21cfea2f4c53c5f241447ff9830f1d307def9191/darts/models/forecasting/regression_ensemble_model.py#L142))
- After that: the first `predict()` call in `RegressionEnsembleModel.fit()` (see [here](https://github.com/unit8co/darts/blob/21cfea2f4c53c5f241447ff9830f1d307def9191/darts/models/forecasting/regression_ensemble_model.py#L402)) should probably not be performed when we use historical fc to fit the model. This predict call is anyways only used to validate all series have the expected time index. Can we find another way to validate that all models have the required time frames? Maybe we can perform a check on the generated historical forecasts.
- Given all of the above, the model should be able to generate the desired forecasts

cnhwl:
> Thanks for giving this a go @cnhwl. However, I think we need to adapt the proposed solution.
> 
> Here are some points:
> 
> * It should be possible to train an ensemble model on base forecasting models that use an output_chunk_shift. Requirements:
>   
>   * All models must use the same `output_chunk_shift` value.
>   * All models must use the same `output_chunk_length` value.
>   * In case of base models using `output_chunk_shift`, the actual `regression_model` (the ensemble model) must also use the same `output_chunk_shift`. In that case we need to check that the future covariates lags for `regression_model` are `{"future": [output_chunk_shift]}` (see [here](https://github.com/unit8co/darts/blob/21cfea2f4c53c5f241447ff9830f1d307def9191/darts/models/forecasting/regression_ensemble_model.py#L142))
> * After that: the first `predict()` call in `RegressionEnsembleModel.fit()` (see [here](https://github.com/unit8co/darts/blob/21cfea2f4c53c5f241447ff9830f1d307def9191/darts/models/forecasting/regression_ensemble_model.py#L402)) should probably not be performed when we use historical fc to fit the model. This predict call is anyways only used to validate all series have the expected time index. Can we find another way to validate that all models have the required time frames? Maybe we can perform a check on the generated historical forecasts.
> * Given all of the above, the model should be able to generate the desired forecasts

Hi! @dennisbader I have completed the three requirements by checking the `output_chunk_shift` and `output_chunk_length`.
I still keep the code of making `self.regression_model.output_chunk_length > self.forecasting_models[0].output_chunk_length - input_shift` to avoid autoregression. If you have better ideas on series length assignment (`forecast_training `and `regression_target`), please let me know. 🤝

cnhwl:
Hi! @dennisbader I wonder if you have time to advance this pull request. Looking forward to your review.😊

dennisbader:
Hi @cnhwl and sorry for the long wait. While working on another PR I realized there are a couple of issues with `RegressionEnsembleModel` at the moment which I would like address before this PR.
So unfortunately, it will have to remain open for a bit longer.
I will let you know once we can get back to this PR. 
Sorry and hope it's okay 🙏

darts/models/forecasting/regression_ensemble_model.py
```diff
@@ -120,9 +120,38 @@ def __init__(
             show_warnings=show_warnings,
         )
 
+        expected_output_chunk_shift = 0
```
Jaco-Pastorius: Instead of initializing the ensemble model output_chunk_shift here, I'd maybe do so in a `RegressionEnsembleModel` property definition at the bottom of the `RegressionEnsembleModel` as its definition depends on other properties.

It might also keep things a bit cleaner by keeping it out of the `__init__` method.

dennisbader:
@cnhwl , we fix this now as part of #3041 :) Sorry for the long wait and thanks a lot for your PR. I'll close this one now 🚀

PR #2805, opened 2025-05-13
fix pass verbose to fit in historical_forecast

### LINKED ISSUE #2804: [QUESTION] Problem setting verbose=False with fit_kwargs in historical_forecasts
** Why can't I set `fit_kwargs={"verbose": False}` to `historical backtest`?**

When I try, I get an error:

Image

`ValueError: The following parameters cannot be passed in `fit_kwargs`: {'verbose'}. Make sure to pass them explicitly to the function/method.`

When I only set the `verbose`, from the `historical backtest` params, the progress bar is still shown

Image

My goal is to not have the progress bar shown during training, which normally can be done with setting `verbose` in fit:

Image

turbotimon:

Fixes #2804.

### Summary

verbose now get passed from historical_forecast onto fit (and not only predict) so progressbar etc are not shown during retrain


dennisbader (changes requested):
Thanks for giving this a go @turbotimon 🚀 . This fails because not all models support the `verbose` parameter in `fit()`.
We would have to change the API of `ForecastingModel.fit()` and subclasses to accept the `verbose` param even if it is not used.

On another thought, I think it could be valuable to make the `verbose` param of `historical_forecasts()` customizable to which level it should act on. For example, I assume that many users are interested in the global progress of the historical forecasts but not in the progress bars of fit and predict.

Basically verbose could accept either `bool` (show everything or nothing) and another type that specifies the level. Maybe an int in `{-1, 0, 1}` with `-1` meaning show nothing, `0` means show only historical forecasts progress, and `1` means show everything). Just an idea. Let me know what you think :)

dennisbader:
Alright, then go ahead with solution 1 :)

turbotimon:
@dennisbader I added solution 1 with the following comments:

- `verbose` is currently not passed to `super().fit` in most cases, because it also not done in most [`predict`](https://github.com/turbotimon/darts/blob/fix-hfcs-verbose-2804/darts/models/forecasting/exponential_smoothing.py#L146)
- `verbose` is only passed to the underlying `self.model.fit()` if the underlying model supports it
- a hack is needed for `CatBoostModel`, because it derived from `SKLearnModel` but SKLearnModel normally do not support `verbose` in their `fit` method (see code  changes for details)
- I added `verbose=False` to a couple of test to proof if it can be passed without errors. I opted for not implementing additional test just for that to not increase test time.

What do you think?

darts/models/forecasting/baselines.py
```diff
@@ -328,6 +328,7 @@ def fit(
         past_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
         future_covariates: Optional[Union[TimeSeries, Sequence[TimeSeries]]] = None,
         sample_weight: Optional[Union[TimeSeries, Sequence[TimeSeries], str]] = None,
+        verbose: bool = False,
```
dennisbader, line 331: propagate it to the _fit_wrapper of forecasting models (further below)

darts/models/forecasting/xgboost.py
```diff
@@ -278,7 +278,7 @@ def fit(
         val_sample_weight
             Same as for `sample_weight` but for the evaluation dataset.
         **kwargs
-            Additional kwargs passed to `xgb.XGBRegressor.fit()`
+            Additional kwargs passed to `xgb.XGBRegressor.fit()`, e.g. `verbose`.
```
dennisbader, line 281: let's make verbose an explicit parameter like in the other models

darts/tests/models/forecasting/test_sklearn_models.py
```diff
@@ -1098,7 +1098,9 @@ def test_fit(self, config):
         # auto-regression but past_covariates does not extend enough in the future
         with pytest.raises(ValueError):
             model_instance = model(lags=4, lags_past_covariates=4, multi_models=mode)
-            model_instance.fit(series=series, past_covariates=sine_multivariate1)
+            model_instance.fit(
```
dennisbader, line 1101: let's revert this change, since we want to check for a specific error, and verbose should not be part of it

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 - Fixed some issues in `NLinearModel` with `normalize=True` that resulted in decreased predictive accuracy. Using `shared_weights=True` and auto-regressive forecasting now work properly. [#2757](https://github.com/unit8co/darts/pull/2757) by [Timon Erhart](https://github.com/turbotimon).
 - Fixed a bug when training a `TorchForecastingModel`, where using certain `torchmetrics` that require a 2D model output (e.g. R2Score) raised an error. [He Weilin](https://github.com/cnhwl).
+- Fixed a bug in `historical_forecast`: Argument `verbose` now gets passed `fit` during retrain. [#2805](https://github.com/unit8co/darts/pull/2805) by [Timon Erhart](https://github.com/turbotimon).
```
dennisbader, line 40: Some minor rephrasing. Could you also add the following line to bottom of the "Improved" section?

```
- Added parameter `verbose` to `ForecastingModel.fit()` that allows to control the verbosity for model fitting. Ignored if the underlying model does not support it. [#2805](https://github.com/unit8co/darts/pull/2805) by [Timon Erhart](https://github.com/turbotimon).
```


```suggestion
- Fixed a bug in `historical_forecast` where the `verbose` parameter was not propagated to the the model's `fit()` method. [#2805](https://github.com/unit8co/darts/pull/2805) by [Timon Erhart](https://github.com/turbotimon).
```

dennisbader (changes requested):
Thanks for the updates @turbotimon, it looks great now 🚀 

Just had a couple minor suggestions, after that it's ready to be merged :)

darts/tests/models/forecasting/test_sklearn_models.py
turbotimon: Ok, but I added it to test_models_runnability as I think it is important to check if the reflection/inspection "hack" is tested for the SkLearnModels

turbotimon:
@dennisbader sorry for the silence, I was heavily occupied, but i finally had time to implemented your suggestions

dennisbader (changes requested):
Thanks for the great updates @turbotimon and also sorry for the waiting - I was on holidays :) Everything looks good. 

I quickly checked the new behavior for a TorchForecastingModel and now I have to admit that your solution 3. from a while ago seems like the preferred option.

> 3. But another option could be to allow passing verbose explicit in fit_kwargs and predict_kwargs and it would override the verbosity received from historical_forecast for these methods. This would basically solve 2. but with more flexibility (e.g. only fit but not predict). But this solution would presume 1. 

I also think that it should be possible to deactivate the fit progress bar while still showing the historical forecast progress (as you mentioned in the original issue).

Would it be okay for you to apply this suggestion? I could also do the changes if you prefer that :) 

Sorry for the back and forth. I hope it doesn't inflict too much extra effort 🙏

turbotimon:
@dennisbader I'm also in favor of solution 3 to be honest :) But unfortunately, I probably won't have time for that at the moment, so please go one.

I would continue to propose that ‘verbose’ remain included in all fit/predict methods for all models (even if it has no effect for some of them). This would make the interfaces more consistent across models and could be beneficial in the future. (If not, at least consistent for model families like e.g. GlobalForecastingModel)

dennisbader:
@turbotimon perfext, and I agree with keeping the params in the API.

Sounds good, I'll take it over. Thanks again for the great work!

dennisbader:
@turbotimon I implemented the changes now :) Here are the updates:
- all fit methods have the verbose: Optional[bool] = None
- all predict methods have the verbose: Optional[bool] = None
- historical forecasts allows to set `verbose` (for the historical forecast progress) `fit_params={"verbose": ...}` (verbose that will be passed to the model's fit method), `predict_params={"verbose": ...}` (verbose that will be passed to the model's predict method)

turbotimon:
Nice! I will try it out as soon as I can!

Issue #1779, opened 2023-05-17
Deterministic evaluation of probabilistic models (e.g. nbeats)


mrtn37:
When predicting a time series, one can turn some models probabilistic by defining a likelihood parameter.
E.g.: 
NBEATSModel(
    ...
    likelihood=GaussianLikelihood(),
    random_state=42
) 

If done this way, the prediction is not deterministic any more. Two subsequent calls to 
predict(series=.., n=1, num_samples=1)
yield different results.

For developing purposes, it would be nice to be able to fix the random state of the likelihood as well so that, independent of the num_samples defined, the result of the prediction always stays the same.

dennisbader:
Hi @mrtn37 and thanks for writing.
The model predictions are reproducible depending on the order of calls.

- If you create two models with identical parameters (and same random state) and train & predict, you will get the same results (assuming that both models called predict() the same number of times)
- If you save a model and load it, it will give identical predictions as the original model (assuming that both objects called predict() the same number of times).

The model's internal random state changes based on the number of times you called methods with the @random_method decorator. It is intended that two consecutive predict calls give different results when probabilistic.

What you suggest is adding an optional parameter to `predict()` to enforce that the random_state remains unchanged and therefore gives reproducible results upon successive calls?

mrtn37:
Thank you for the explanation. That makes sense :)
Even though: keeping in mind how often predict was called, resp. saving copies of the model might be a bit cumbersome during development.

> What you suggest is adding an optional parameter to `predict()` to enforce that the random_state remains unchanged and therefore gives reproducible results upon successive calls?

Yes, that sounds like a good proposal. Just to clarify: the parameter would (re-)set the random_state to a certain value, such that  two consequtive calls: predict(series=.., n=, num_samples=, random_state=) would produce the same two nummerical arrays.

PR #2808, opened 2025-05-23
Deterministic sampling for all probabilistic models

### LINKED ISSUE #1835: Simplify deterministic sampling for probabilistic models in production
First of all, thanks for this amazing library! It has made my work a lot easier.

I'm wrapping a probabilistic forecasting model in a FastAPI endpoint for access by end users. I'm using `QuantileRegression` and the results look great. However, I noticed that the forecasts differ subtly between invocations when feeding in the same input timeseries. This is because the model instance uses its own RNG state that cannot be conveniently seeded from the outside using e.g.  `torch.manual_seed(seed_int)` or `np.random.seed(seed_int)`. If I understand correctly, this is managed through the `@random_method` decorator defined here:

https://github.com/unit8co/darts/blob/8e93720e6277726089058cf4d039b97de0f57c34/darts/utils/torch.py#L85

My model definition looks like the below.

```python
model = NLinearModel(
    input_chunk_length=24*5,
    output_chunk_length=24,
    likelihood=QuantileRegression(quantiles=[0.01, 0.05, 0.2, 0.5, 0.8, 0.95, 0.99]),
    random_state=42,
    pl_trainer_kwargs={'deterministic': True}
)
```

The below is what I'm currently doing to re-seed the RNG before every invocation of `.predict()`.

It would be great if there was a cleaner way to prevent `_random_instance` from being re-seeded on every invocation of `predict()` with a new random seed, i.e. to disable the effect of `@random_method`.

```python
with torch.random.fork_rng():
    # Re-seed the hidden RNG state of the model instance to ensure deterministic sampling.
    model._random_instance = np.random.RandomState(seed_int)
    prediction = model.predict(
        series=ts,
        n=24,
        num_samples=500,
    )
```

Jaco-Pastorius:

Fixes #1835, #1779.

### Summary
**Main feature** : when specifying `random_state` in `predict` function, the output is always the same. This was implemented in the `random_method` decorators in the  darts.utils.utils.py and darts.utils.torch.py files. 

All the probabilistic models now have a unified randomization both at model initialization and prediction. Models concerned :
- Models inheriting from StatsForecastModel
- Models inheriting from TorchForecastingModel
- Models inheriting from SKLearnModel
- Models inheriting from ConformalModel
- Additional models : VARIMA, ARIMA, Prophet, KalmanForecaster, ExponentialSmoothing

### DIFF
--- a/darts/models/forecasting/arima.py
+++ b/darts/models/forecasting/arima.py
@@ -19,7 +19,7 @@ if sys.version_info >= (3, 10):
 else:
     from typing_extensions import TypeAlias
 
-import numpy as np
+from sklearn.utils import check_random_state
 from statsmodels.tsa.arima.model import ARIMA as staARIMA
 
 from darts.logging import get_logger
@@ -27,6 +27,7 @@ from darts.models.forecasting.forecasting_model import (
     TransferableFutureCovariatesLocalForecastingModel,
 )
 from darts.timeseries import TimeSeries
+from darts.utils.utils import random_method
 
 logger = get_logger(__name__)
 
@@ -35,6 +36,7 @@ IntOrIntSequence: TypeAlias = Union[int, Sequence[int]]
 
 
 class ARIMA(TransferableFutureCovariatesLocalForecastingModel):
+    @random_method
     def __init__(
         self,
         p: IntOrIntSequence = 12,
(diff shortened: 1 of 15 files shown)


dennisbader (changes requested):
Really good job @Jaco-Pastorius, thanks a lot 🚀 
This will be a great feature for Darts :) 

I added a couple of suggestions here and there, but we're already very close to merging 💯

PR #2812, opened 2025-05-28
Fix tuple type annotation in TimeSeries

ajyoon:
- [ ] ~~Mentioned all issues that this PR fixes or addresses.~~ (N/A)

### Summary
- `TimeSeries.shape` was incorrectly annotated `tuple[int]`; this is changed to `tuple[int, int, int]`.

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -1928,130 +1928,6 @@ class TimeSeries:
             schema = {k: deepcopy(v) for k, v in schema.items()}
         return schema
 
-    def quantile_df(self, quantile=0.5) -> pd.DataFrame:
-        """
-        Return a Pandas DataFrame containing the single desired quantile of each component (over the samples).
-
-        Each of the series components will appear as a column in the DataFrame. The column will be named
-        "<component>_X", where "<component>" is the column name corresponding to this component, and "X"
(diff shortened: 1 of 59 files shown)


darts/timeseries.py
```diff
@@ -2019,7 +2019,9 @@ def quantile_timeseries(self, quantile=0.5, **kwargs) -> Self:
 
         return self.__class__(new_xa)
 
-    def quantiles_df(self, quantiles: tuple[float] = (0.1, 0.5, 0.9)) -> pd.DataFrame:
```
dennisbader: here I think a `Optional[list[float]]` would actually make more sense. Could you adapt it?


E.g.

```py

    def quantiles_df(self, quantiles: Optional[list[float]] = None) -> pd.DataFrame:
        quantiles = quantiles or [0.1, 0.5, 0.9]
        return pd.concat(
            [
                self.quantile_timeseries(quantile).to_dataframe()
                for quantile in quantiles
            ],
            axis=1,
        )

ajyoon:
@dennisbader apologies for my delay - looks like you beat me on the quantiles change, so I've merged upstream and removed those changes. This PR is now just the annotation fix on `TimeSeries.shape`

PR #2821, opened 2025-06-11
fix: change default shap method of extra trees and random forest models

rijkvandermeulen:
### Summary

I noticed that retrieving shapley values for certain tree-based models (e.g., `ExtraTreeRegressor`) took significantly longer when the `shap_method` argument was not specified explicitly. When this argument is unspecified a mapping based on the underlying ML model happens in the background. I noticed there that the `ExtraTreeRegressor` and `RandomForestRegressor` mapped to the permutation method rather than the (faster and hence more applicable default) tree method. Notice that the `ExtraTreeRegressor`  model in fact had 2 entries in the dict (one mapping to the tree method and one mapping to permutation). Due to the sequence this will always evaluate to permutation.

This PR changes the default for both models to `_ShapMethod.TREE`

### DIFF
--- a/darts/explainability/shap_explainer.py
+++ b/darts/explainability/shap_explainer.py
@@ -529,12 +529,11 @@ class _RegressionShapExplainers:
         # Tree models
         "DecisionTreeRegressor": _ShapMethod.TREE,
         "ExtraTreeRegressor": _ShapMethod.TREE,
+        "RandomForestRegressor": _ShapMethod.TREE,
         # Ensemble model
         "AdaBoostRegressor": _ShapMethod.PERMUTATION,
         "BaggingRegressor": _ShapMethod.PERMUTATION,
-        "ExtraTreesRegressor": _ShapMethod.PERMUTATION,
         "HistGradientBoostingRegressor": _ShapMethod.PERMUTATION,
-        "RandomForestRegressor": _ShapMethod.PERMUTATION,
         "RidgeCV": _ShapMethod.PERMUTATION,
         "Ridge": _ShapMethod.PERMUTATION,
         # Linear models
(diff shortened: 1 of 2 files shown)


darts/explainability/shap_explainer.py
```diff
@@ -529,12 +529,11 @@ class _RegressionShapExplainers:
         # Tree models
         "DecisionTreeRegressor": _ShapMethod.TREE,
         "ExtraTreeRegressor": _ShapMethod.TREE,
+        "RandomForestRegressor": _ShapMethod.TREE,
         # Ensemble model
         "AdaBoostRegressor": _ShapMethod.PERMUTATION,
         "BaggingRegressor": _ShapMethod.PERMUTATION,
-        "ExtraTreesRegressor": _ShapMethod.PERMUTATION,
```
dennisbader, line 535: Thanks for this PR @rijkvandermeulen 🚀 

Sklearn has two separate models `ExtraTreeRegressor` and `ExtraTreesRegressor` (Tree and Tree**s**).

Would changing PERMUTATION to TREE make sense for other models here as well? It seems the `HistGradienBosstingRegressor` also supports it (see [here](https://github.com/shap/shap/issues/1028)). Maybe others as well?

rijkvandermeulen: @dennisbader good catch haha! And you're right: `HistGradienBosstingRegressor` is better off with TREE iso PERMUTATION as well. Did a quick check on the other models, but I think their default makes sense. I guess that some models could be used with TREE depending on the choice of base learner (e.g., `AdaBoostRegressor`) but as it won't always work, I think it would be better to leave them on PERMUTATION.

PR #2829, opened 2025-06-23
let lightning handle moving preds to cpu

dennisbader:

Fixes #.

### Summary

- attempt to fix cuda sync issues for windows.

### DIFF
--- a/darts/models/forecasting/pl_forecasting_module.py
+++ b/darts/models/forecasting/pl_forecasting_module.py
@@ -4,7 +4,6 @@ This file contains abstract classes for deterministic and probabilistic PyTorch 
 
 import copy
 from abc import ABC, abstractmethod
-from collections.abc import Sequence
 from functools import wraps
 from typing import Any, Optional, Union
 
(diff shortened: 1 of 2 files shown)


darts/models/forecasting/pl_forecasting_module.py
```diff
@@ -325,7 +323,7 @@ def predict_step(
         batch: TorchInferenceBatch,
         batch_idx: int,
         dataloader_idx: Optional[int] = None,
-    ) -> Sequence[TimeSeries]:
+    ) -> tuple[torch.Tensor, Any, Any]:
```
madtoinou: Wouldn't it be possible to have something more informative about the type of `batch_series_schemas` and `batch_pred_starts`?

madtoinou (approved):
Just a minor comment, otherwise it looks good to me.

darts/models/forecasting/pl_forecasting_module.py
dennisbader: I'll do it like this (I want to avoid having to import pandas in this torch module only for pd.Timestamp):

```
-> tuple[
        torch.Tensor,
        Sequence[Optional[dict[str, Any]]],
        Sequence[Any],
    ]
```

PR #2845, opened 2025-07-18
Feat/add future cov for block rnn

### LINKED ISSUE #2563: Add future covariates support to `BlockRNN`
While RNN model only supports future covariates BlockRNN only supports the past covariates.

[https://unit8co.github.io/darts/generated_api/darts.models.forecasting.rnn_model.html?highlight=rnn#darts.models.forecasting.rnn_model.RNNModel](url)

I did not know there was two different model called BlockRNN and RNN. How does this two differs in a sense mathematically so that one of them is support past covariates and the other is not ?

Jaco-Pastorius:

Fixes #2563.

### Summary

Add of future and static covariates support for BlockRNN.

### Other Information

Change of RNN output used for prediction. The predictions now use `out` which corresponds to the RNN output for each timestep in the input sequence from the last RNN layer.

We were previously using `hidden` which corresponds to the hidden state of the last timestep for each layer.


darts/models/forecasting/block_rnn_model.py
```diff
@@ -453,17 +482,32 @@ def encode_year(idx):
...
         hidden_fc_sizes = [] if self.hidden_fc_sizes is None else self.hidden_fc_sizes
-
+        future_cov_dim = (
+            future_covariates.shape[1] if future_covariates is not None else 0
+        )
+        static_cov_dim = (
+            static_covariates.shape[0] * static_covariates.shape[1]
+            if static_covariates is not None
+            else 0
+        )
```
madtoinou, line 532: These variables could be reused instead of the loop line 492-498, to simplify a bit the logic.

darts/models/forecasting/block_rnn_model.py
```diff
@@ -178,18 +197,27 @@ def __init__(
...
         """ Here, we apply the FC network only on the last output point (at the last time step)
         """
-        if self.name == "LSTM":
-            hidden = hidden[0]
-        predictions = hidden[-1, :, :]
+        out = out.permute(0, 2, 1)
+        out = self.upsampling_out(out)
+        predictions = out.permute(0, 2, 1)
+        if x_future is not None:
+            predictions = torch.cat([predictions, x_future], dim=-1)
```
madtoinou: Is it intentional to completely get ride of the behavior specific to `self.name == "LSTM"`?

examples/04-RNN-examples.ipynb
madtoinou, line 1: I don't think that this example should be completely deleted, no?

madtoinou (changes requested):
Nice contribution, just a couple of comments to simplify a bit the code and it should be ready for merging.

darts/models/forecasting/block_rnn_model.py
Jaco-Pastorius: Yes, this was intended since `hidden` (the second thing returned by the rnn torch module) was used for predictions before. Its dimensions change depending on the RNN model used (GRU/RNN or LSTM). We now instead use `out` whose dimension doesn't depend on what model we use. We therefore don't need the specific LSTM-related logic anymore.

examples/04-RNN-examples.ipynb
Jaco-Pastorius: Indeed, I'll put it back next commit, sorry about that.

dennisbader (approved):
Thanks a lot for this great improvement @Jaco-Pastorius, it looks really good 🚀 

I made some final adaptions:
- removed unused variables
- only uses upsample when input_chunk_length < output_chunk_length.
- updated the comments so that we’ll have it easier in the future to debug (e.g. input shape -> output shape)
- used the original notebook and reran to avoid the large diffs from the previous commits

Issue #2755, opened 2025-04-02
[BUG] darts Trainer torch precision


neutron-blast:
**Describe the bug**
The provided TorchForecastingModel code from Darts 0.34.0 explicitly restricts precision to ["64-true", "32-true"] (for Lightning 2.x), rejecting "16-true" outright.

**To Reproduce**

> def generate_torch_kwargs():
>     return {
>         "pl_trainer_kwargs": {
>             "accelerator": "gpu",
>             "gradient_clip_val": 0.1,  # Adjust based on your training stability
>             "precision": "16-true",  # Use mixed precision if supported; remove or set to 32 otherwise
>             "callbacks": [
>                 TFMProgressBar(enable_train_bar_only=True)
>             ],
>         }
>     }

**Expected behavior**
16 bit precision should be accepted.

**System (please complete the following information):**
 - Python version: [e.g. 3.12]
 - darts version [e.g. 0.34.0]

**Additional context**

dennisbader:
Hi @neutron-blast, and thanks for raising this issue. We can indeed now add support for 16-true (I vastly remember some limitations back in the days which was why we didn't add it then).

I'll add it to our backglog.

Would you be interested in contributing?

Joelius300:
Thank you, I would also like to see support for this. Here's another relevant issue regarding mixed precision: https://github.com/unit8co/darts/issues/2484

breznak:
This should be closed per https://github.com/unit8co/darts/pull/2883 , thanks everyone!

dennisbader:
Indeed, thanks @breznak :)

Issue #2484, opened 2024-07-31
Mixed precision support for Torch Models


Joelius300:
**Is your feature request related to a current problem? Please describe.**

Currently, the only valid precisions when training a torch model are `32-true` and `64-true` (see [here](https://github.com/unit8co/darts/blob/479656cb0293a640dc1b99edd14cd34768e37f24/darts/models/forecasting/torch_forecasting_model.py#L487)). Although it seems to be possible to load weights for models with lower precision, namely `float16` since #2046, I didn't find a way to train one from scratch using these lower precisions. Given limited hardware resources, it would be awesome to speed up training with `bf16-mixed`, `16-mixed` or any of the [others supported by Lightning](https://lightning.ai/docs/fabric/stable/fundamentals/precision.html).

**Describe proposed solution**

It would be nice to be able to specify the precision like now but with more options available (all that work I guess).

**Describe potential alternatives**

I have not found an alternative besides not training through darts.

**Additional context**

I'm sure there's a reason only 32 and 64 bit are supported but I didn't find much in the docs or other issues. The disclaimer was written in #1651 including tests that 16 bit is not supported but I was not able to figure out what exactly is the issue with e.g. `16-mixed`.

I also found a number of issues mentioning precision, e.g. #2344, #1987, but none of them seem to request supporting 16 bit still, so I thought I'd open this issue; apologies if I missed something relevant.

Especially #1987 seems to have talked about it, but then it was closed because the bug that reset the precision after loading was fixed. There's also the old #860, which was not revisited IIRC, but it mentions that it would break functionality for Scalers.

MichaelVerdegaal:
Also interested in this

Joelius300:
Completed in #2883

PR #2883, opened 2025-08-26
Add mixed precision and 16-bit support to `TorchForecastingModel`

daidahao:

Addresses #2484 #2755

### Summary

Previously, the `"precision"` option set in `pl_trainer_kwargs` is ignored in favour of precision of the time series data type. That led to the lack of support for mixed precision training and 16-bit true precision despite `pytorch-lightning.Trainer` native support.

I added support for mixed precision and any other precision options supported by Lightning in `TorchForecastingModel`.

Here, by default when the `"precision"` option is not set in `pl_trainer_kwargs` or a custom trainer, the model precision is determined by the data type. When it is set, we can assume that the user understand the ramifications of that option and the option is always passed along to the trainer, even when there is a mismatch between the model precision and the data type.

Sometimes the mismatch is necessary when the data type is 32-bit but the model can be trained in mixed precision. The other times the mismatch would cause an error, say, training 32-bit model on 64-bit data.

**Precision is an advanced option for `TorchForecastingModel` training and users should only use it at their own benefits and risks!**

### Other Information

N/A

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -20,7 +20,6 @@ import copy
 import datetime
 import inspect
 import os
-import re
 import shutil
 import sys
 from abc import ABC, abstractmethod
@@ -82,13 +81,7 @@ from darts.utils.utils import _build_tqdm_iterator, _parallel_apply
 tokens = pl.__version__.split(".")
 pl_200_or_above = int(tokens[0]) >= 2
 
-if pl_200_or_above:
-    from pytorch_lightning.callbacks import ProgressBar
-    from pytorch_lightning.tuner import Tuner
-else:
-    from pytorch_lightning.callbacks import ProgressBarBase as ProgressBar
-    from pytorch_lightning.tuner.tuning import Tuner
-
+from pytorch_lightning.tuner.tuning import Tuner
 
 DEFAULT_DARTS_FOLDER = "darts_logs"
 CHECKPOINTS_FOLDER = "checkpoints"
@@ -446,53 +439,42 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
         model = self._create_model(self.train_sample)
         self._module_name = model.__class__.__name__
 
+        # we should determine the precision based on time series data type
+        # however if user has defined a precision, we should follow that
         precision = None
-        dtype = self.train_sample[0].dtype
-        if np.issubdtype(dtype, np.float32):
-            logger.info("Time series values are 32-bits; casting model to float32.")
-            precision = "32" if not pl_200_or_above else "32-true"
-        elif np.issubdtype(dtype, np.float64):
-            logger.info("Time series values are 64-bits; casting model to float64.")
-            precision = "64" if not pl_200_or_above else "64-true"
-        else:
-            raise_log(
-                ValueError(
-                    f"Invalid time series data type `{dtype}`. Cast your data to `np.float32` "
-                    f"or `np.float64`, e.g. with `TimeSeries.astype(np.float32)`."
-                ),
-                logger,
-            )
-        precision_int = int(re.findall(r"\d+", str(precision))[0])
-
         precision_user = (
             self.trainer_params.get("precision", None)
             if trainer is None
             else trainer.precision
         )
+        dtype = self.train_sample[0].dtype
         if precision_user is not None:
-            # currently, we only support float 64 and 32
-            valid_precisions = (
-                ["64", "32"] if not pl_200_or_above else ["64-true", "32-true"]
-            )
-            if str(precision_user) not in valid_precisions:
-                raise_log(
-                    ValueError(
-                        f"Invalid user-defined trainer_kwarg `precision={precision_user}`. "
-                        f"Use one of ({valid_precisions})"
-                    ),
-                    logger,
+            logger.info(f"Using user-defined precision: {precision_user}")
+            precision = precision_user
+        elif np.issubdtype(dtype, np.float32):
+            logger.info("Time series values are 32-bits; casting model to float32.")
+            precision = "32" if not pl_200_or_above else "32-true"
+        elif np.issubdtype(dtype, np.float64):
+            logger.info("Time series values are 64-bits; casting model to float64.")
+            precision = "64" if not pl_200_or_above else "64-true"
+        elif np.issubdtype(dtype, np.float16):
+            if pl_200_or_above:
+                logger.info("Time series values are 16-bits; casting model to float16.")
+                precision = "16-true"
+            else:
+                logger.info(
+                    "Time series values are 16-bits; casting model to float32 "
+                    "but trained in mixed precision."
                 )
-            precision_user_int = int(re.findall(r"\d+", str(precision_user))[0])
+                precision = 16
         else:
-            precision_user_int = None
-
-        raise_if(
-            precision_user is not None and precision_user_int != precision_int,
-            f"User-defined trainer_kwarg `precision='{precision_user}'` does not match dtype: `{dtype}` of the "
-            f"underlying TimeSeries. Set `precision` to `{precision}` or cast your data to `{precision_user}"
-            f"` with `TimeSeries.astype(np.float{precision_user_int})`.",
-            logger,
-        )
+            raise_log(
+                ValueError(
+                    f"Invalid time series data type `{dtype}`. Cast your data to `np.float32` "
+                    f"or `np.float64` or `np.float16`, e.g. with `TimeSeries.astype(np.float32)`."
+                ),
+                logger,
+            )
         self.trainer_params["precision"] = precision
 
         # we need to save the initialized TorchForecastingModel as PyTorch-Lightning only saves module checkpoints
@@ -516,9 +498,13 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
             return trainer
 
         trainer_params = {key: val for key, val in self.trainer_params.items()}
-        has_progress_bar = any([
-            isinstance(cb, ProgressBar) for cb in trainer_params.get("callbacks", [])
-        ])
+        is_progress_bar = lambda cb: cb.__class__.__name__ in [
+            "ProgressBar",
+            "ProgressBarBase",
+        ]
+        has_progress_bar = any(
+            map(is_progress_bar, trainer_params.get("callbacks", []))
+        )
         # we ignore `verbose` if `trainer` has a progress bar, to avoid errors from lightning
         if verbose is not None and not has_progress_bar:
             trainer_params["enable_model_summary"] = (
--- a/darts/tests/models/forecasting/test_ptl_trainer.py
+++ b/darts/tests/models/forecasting/test_ptl_trainer.py
@@ -23,8 +23,10 @@ class TestPTLTrainer:
     series = linear_timeseries(length=100).astype(np.float32)
     pl_200_or_above = int(pl.__version__.split(".")[0]) >= 2
     precisions = {
+        16: None if not pl_200_or_above else "16-true",
         32: "32" if not pl_200_or_above else "32-true",
         64: "64" if not pl_200_or_above else "64-true",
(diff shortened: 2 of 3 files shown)


dennisbader:
Thanks a lot for the PR @daidahao  🚀 Tomorrow I'll have time to review :)

dennisbader:
Ahh got it, thanks @daidahao. I'll push the update in a few minutes.

dennisbader (changes requested):
Thanks for this PR @daidahao, it's looking like a great start 🚀 

I tested it out a little and I do have some concerns / findings:

- I'm testing on a Apple M1 Pro, 2021, macOS 15.3.2 (tested CPU and MPS (GPU)).
  - 16-true: forecasts contain **only NaN values** regardless of input series dtype or CPU / GPU. Not exactly sure exactly if it's a hardware issue or code. How does it look on your side?
  - bf16-true: **works** but sadly we have to convert to np.float32 in the end 😢 
  - bfloat16-mixed and float16-mixed: 
    - On MPS (GPU):
      - When input has dtype np.float16: **Kernel dies**
      - When input has dtype np.float32: **Works** (but converts output to np.float32).
     - On CPU: 
        - When input has dtype np.float16: **Raises error** `RuntimeError: Unexpected floating ScalarType in at::autocast::prioritize` (since `torch.cat()` doesn't support float16 on CPU).
        - When input has dtype np.float32: **Works** (but converts output to np.float32).

Looking at the above, the only configurations that seems to work are:

- bf16-true regardless of input `series` data type (float16, float32, float64).
- bf16-mixed and 16-mixed when input `series` data type is float32.

Given all of the above it seems like there are a lot of exceptions happening which can be quite tricky for the user to understand. So my ideas:

- Should we for now only handle these and otherwise raise exceptions from our side? This could help users a lot (even though you are correct that it is an advanced feature)

daidahao:
Hi Dennis,

I've tested it again on both MacBook Pro (M4 Pro 48G) and a GPU server (Tesla V100 16G). I have dependencies close to yours.

Here are my testing results with TSMixerModel:
- 16-true: NaN losses and predictions regardless of the input series type.
- 16-mixed:
   - MPS: disabled by Lightning and using 32-true due to not using CUDA
   - CPU: switched to bf16-mixed instead by Lightning
   - GPU (cuda): works with both 16-bit and 32-bit input.
- bf16-true: works with both 16-bit and 32-bit input.
- bf16-mixed: 
  - MPS: not supported for either 16-bit or 32-bit input.
  - CPU: same as yours, error for 16-bit input and works with 32-bit input
  - GPU (cuda): works with both 16-bit and 32-bit input.
 
Lightning documentation mentioned that BFloat16 is more stable than 16-true and the results here seem to confirm that. Given the results, can I suggest defaulting to bf16-true and producing a warning in `_init_model()` when input is 16-bit and precision not set by the user? I've noticed there is a warning from creating a 16-bit TimeSeries, perhaps there could be a similar warning here for the model.

Conversion from bfloat16 to np.float32 is probably fine since numpy does not have an equivalent of bfloat16. We could produce a warning to user when `predict_from_dataset()` is called on 16-bit input so they know that the data type has changed from the input.

To you last point, if we handle some precision options for the user, the support from Lightning might change/expand in the future and we could end up with more nested conditionals to deal with possible combinations of device, input type, and precision. I am in favor of a hands-free approach when the precision option is set by user, since it might need to be set based on the hardware. That said, we could produce a warning to the user when it is set and to suggest options that are applicable to most platforms like bf16-true or 16-mixed.

Finally, I've noticed that some models do not support any mixed precision like TFT due to masking operations.

Here is my script for testing the precision option.

```python
import numpy as np
import pandas as pd
from darts import TimeSeries
from darts.models import TSMixerModel

tfm_kwargs = {
    "batch_size": 16,
    "n_epochs": 1,
    "random_state": 42,
    "pl_trainer_kwargs": {
        "accelerator": "mps",
        "precision": "16-mixed",
        "enable_progress_bar": False,
        "enable_model_summary": False,
    },
}
length = 10000

times = pd.date_range("20130101", periods=length, freq="s")
pd_series = pd.Series(np.linspace(0, 1, length), index=times)
series = TimeSeries.from_series(pd_series).astype(np.float32)
train_series, val_series = series[:-int(length*0.8)], series[-int(length*0.2):]

# model = TFTModel(100, 100, add_relative_index=True, **tfm_kwargs)
model = TSMixerModel(100, 100, **tfm_kwargs)
model.fit(train_series, val_series=val_series, verbose=True)
model.predict(100)
```

dennisbader:
@daidahao 

> can I suggest defaulting to bf16-true and producing a warning in _init_model() when input is 16-bit and precision not set by the user

Yes, that sounds good (including the warning). It would be nice to warn if bf16-true/mixed is used in any case (user-defined precision or not). I think it's okay to leave the warning only within the `_init_model` scope to avoid spamming the user during prediction.

> To you last point, if we handle some precision options for the user, the support from Lightning might change/expand in the future and we could end up with more nested conditionals to deal with possible combinations of device, input type, and precision. I am in favor of a hands-free approach when the precision option is set by user, since it might need to be set based on the hardware. That said, we could produce a warning to the user when it is set and to suggest options that are applicable to most platforms like bf16-true or 16-mixed.

I see the point about having to adapt when lightning makes changes. I agree with you in general, only that in Darts we follow the idea that what we offer/expose to the user should also work. Since setting the precision is however a more advanced feature and dependent on many things at once (input dtype, device, platform) I think it's okay to be more lenient.
Maybe we could inform / direct the user to some source that explains the limitations when they set a precision? Even this PR could already help.

daidahao:
@dennisbader 

Your suggestions are most reasonable. I will try to implement the changes we've discussed soon.

> I think it's okay to leave the warning only within the `_init_model` scope to avoid spamming the user during prediction.

Another issue that might be related but perhaps best addressed in another PR. Is there a need for `predict_from_dataset()` to return the same dtype as the input? I imagine many users would expect the output dtype to be the same as the input, regardless of the model being used. What do you think?

dennisbader:
@daidahao 

> Another issue that might be related but perhaps best addressed in another PR. Is there a need for predict_from_dataset() to return the same dtype as the input? I imagine many users would expect the output dtype to be the same as the input, regardless of the model being used. What do you think?

Yes, that's what I was worried a bit about in the beginning as well where I mentioned that we should handle maybe handle a only subset of scenarios where the output dtype would match the input dtype. You mean both `predict()` and `predict_from_dataset()`, right?

The user can also simply cast the input series dtype to match the precision for it to be availbale. E.g. cast `series = series.astype(np.float64)` would make `precision="64-true"` available.

I'm happy to discuss further though :)

daidahao:
@dennisbader 

I've implemented the changes as discussed. We are producing warnings whenever a user-defined precision is provided or input time series are 16-bit. 

I've also expanded `test_mixed_precision_training()` to cover "16-mixed" and "bf16-mixed" options. I've also updated `test_auto_precision_casting()` to allow for 32-bit predictions when inputs are 16-bit and similarly `test_builtin_extended_trainer()`.

> Yes, that's what I was worried a bit about in the beginning as well where I mentioned that we should handle maybe handle a only subset of scenarios where the output dtype would match the input dtype. You mean both predict() and predict_from_dataset(), right?

Yes, I think `predict()` depends on `predict_from_dataset()`. I see it as a separate issue because the current implementation of `predict_from_dataset()` and related functions such as `residuals()` do not explicitly guarantee the return data type is the same. I notice that calling `residuals()` on `ConformalQRModel` returns `np.float64` regardless of the input data type. I could raise this as a separate issue if you think they should be the same.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -436,51 +435,47 @@ def _init_model(self, trainer: Optional[pl.Trainer] = None) -> PLForecastingModu
...
-        if np.issubdtype(dtype, np.float32):
+        if precision_user is not None:
+            logger.warning(
+                f"Using user-defined precision: {precision_user}. "
+                "For mixed precision training, recommended options are 'bf16-mixed' "
+                "and '16-mixed'. See https://github.com/unit8co/darts/pull/2883 for "
+                "a discussion on low precision options across hardware platforms and "
+                "https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.trainer.trainer.Trainer.html#lightning.pytorch.trainer.trainer.Trainer.params.precision"
+                " for a full list of options in PyTorch-Lightning."
+            )
```
dennisbader: I would suggest to log an info with general info about user-defined precision, and a warning only when a '16' -like precision is used. E.g.

```suggestion
            logger.info(
                f"Using user-defined precision: {precision_user}. The model output will have the same dtype. If you "
                f"encounter issues, it's usually due to a conflict between input series data type and precision, or an "
                f"unsupported precision for the given device or model. For more information, see "
                f"https://github.com/unit8co/darts/pull/2883 for a discussion on low precision options across hardware "
                f"platforms."
            )
            if "16" in str(precision_user):
                logger.warning(
                    "Detected user-defined float16-like precision. For mixed precision training, recommended "
                    "options are 'bf16-mixed' and '16-mixed'."
                )
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -436,51 +435,47 @@ def _init_model(self, trainer: Optional[pl.Trainer] = None) -> PLForecastingModu
...
+            precision = precision_user
+        elif np.issubdtype(dtype, np.float32):
             logger.info("Time series values are 32-bits; casting model to float32.")
             precision = "32-true"
         elif np.issubdtype(dtype, np.float64):
             logger.info("Time series values are 64-bits; casting model to float64.")
             precision = "64-true"
+        elif np.issubdtype(dtype, np.float16):
+            logger.warning(
+                "Time series values are 16-bits; casting model to bfloat16. "
```
dennisbader:
```suggestion
                "Time series values are 16-bits; casting model to bfloat16 and model output will have dtype float32. "
```

darts/tests/models/forecasting/test_ptl_trainer.py
```diff
@@ -80,21 +85,55 @@ def test_prediction_custom_trainer(self):
     def test_custom_trainer_setup(self):
         model = RNNModel(12, "RNN", 10, 10, random_state=42, **tfm_kwargs)
 
-        # trainer with wrong precision should raise ValueError
+        # no error with correct precision
```
dennisbader, line 97: Can we handle the valid precision (16, 32, 64, and "mixed") tests in one parametrized test? (e.g. with @pytest.mark.parametrize("precision", [16, 32, 64, "mixed"])

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -1127,6 +1127,46 @@ def test_train_from_10_n_epochs_20_fit_15_epochs(self):
...
+                "RNN",
+                10,
+                10,
+                n_epochs=20,
+                **tfm_kwargs,
+            )
+            model.fit(self.series.astype(dtype), epochs=1)
+            assert 1 == model.epochs_trained
+
+            preds = model.predict(n=10)
```
dennisbader: Could you add a check here and in the next test that there are no NaNs in the prediction?

darts/tests/models/forecasting/test_ptl_trainer.py
daidahao: It is being handled in b8a93fe3e6616ede8383a676117a832c363facd4

darts/tests/models/forecasting/test_torch_forecasting_model.py
daidahao: Yes, it is being added in b8a93fe3e6616ede8383a676117a832c363facd4.

daidahao:
@dennisbader Thanks, Dennis! I think I have addressed all your suggestions. Let me know what you think.

dennisbader (approved):
Thanks a lot for the updates @daidahao and this PR @daidahao 🚀 
All is ready now to be merged after the tests have passed 💯

PR #2895, opened 2025-09-11
fix: properly save naive model

### LINKED ISSUE #2839: [BUG] Cannot save global baseline models
**Describe the bug**

Trying to save a global baseline model (GlobalNaiveSeasonal, GlobalNaiveAggregate, GlobalNaiveDrift) throws the following error.

> AttributeError: Saving a checkpoint is only possible if a model is attached to the Trainer. Did you call Trainer.save_checkpoint() before calling Trainer.{fit,validate,test,predict}?

**To Reproduce**

```python
from darts.models import LinearRegressionModel, RNNModel, GlobalNaiveSeasonal, GlobalNaiveAggregate, GlobalNaiveDrift
from darts.models.forecasting.forecasting_model import GlobalForecastingModel
from darts.models.forecasting.torch_forecasting_model import TorchForecastingModel

passengers = AirPassengersDataset().load()

# model = LinearRegressionModel(lags=1, output_chunk_length=1)  # works, global but not torch
# model = RNNModel(input_chunk_length=1, n_epochs=10)  # works, torch
# model = GlobalNaiveSeasonal(input_chunk_length=input_steps, output_chunk_length=1)  # doesn't work, torch without weights
# model = GlobalNaiveAggregate(input_chunk_length=1, output_chunk_length=1)  # doesn't work, torch without weights
model = GlobalNaiveDrift(input_chunk_length=1, output_chunk_length=1)  # doesn't work, torch without weights

model.fit(passengers)

path = "model"
model.save(path)  # <---- fails here

# make sure loading works too.
# Unfortunately, you cannot use GlobalForecastingModel.load for torch models :(
# model = GlobalForecastingModel.load(path)
model = TorchForecastingModel.load(path)
```

**Expected behavior**

The model is saved with all required information so that it can be loaded with TorchForecastingModel.load (or even better GlobalForecastingModel.load).

- Python version: 3.12.9
- darts version: 0.35.0

agysi:

Fixes #2839.

### Summary

This PR fixes the save for naive models. Previously, naive models were not properly saved, resulting in an error when performing the following:

```python
model = GlobalNaiveDrift(input_chunk_length=1, output_chunk_length=1)  
model.fit(passengers)
model.save(path)  # <---- fails here
```
This caused some incoherence with non-naive models where this kind of code works.

### Other Information
This issue came from the `TorchForecastingModel.save`. Internally, it will try to save the trainer strategy, which is not initialized when fitting a naive model. If predict is called before saving however, this initializes properly the trainer strategy.
To solve this issue, the fit method of naive models has been adjusted to perform a simple predict to initialize the model, so that the save can be done properly.

### DIFF
--- a/darts/tests/models/forecasting/test_global_forecasting_models.py
+++ b/darts/tests/models/forecasting/test_global_forecasting_models.py
@@ -348,9 +348,7 @@ class TestGlobalForecastingModels:
             load_kwargs = {"pl_trainer_kwargs": {"accelerator": "cpu"}}
         else:
             load_kwargs = {}
-        loaded_model_clean_str = type(model).load(
-            full_model_clean_path_str, **load_kwargs
-        )
+        loaded_model_clean = type(model).load(full_model_clean_path_str, **load_kwargs)
         loaded_model_no_predict = type(model).load(
             full_model_path_no_predict_str, **load_kwargs
         )
@@ -370,11 +368,11 @@ class TestGlobalForecastingModels:
         )
 
         # Training data is not stored in the clean model
-        assert loaded_model_clean_str.training_series is None
+        assert loaded_model_clean.training_series is None
 
         # The serie to predict need to be provided at prediction time
         with pytest.raises(ValueError) as err:
-            loaded_model_clean_str.predict(self.forecasting_horizon)
+            loaded_model_clean.predict(self.forecasting_horizon)
         if isinstance(model, TorchForecastingModel):
             assert str(err.value) == (
                 "Input `series` must be provided. This is the result either from fitting on multiple series, "
@@ -389,7 +387,7 @@ class TestGlobalForecastingModels:
             )
 
         # When the serie to predict is provided, the prediction is the same
-        assert model_prediction == loaded_model_clean_str.predict(
+        assert model_prediction == loaded_model_clean.predict(
             self.forecasting_horizon, series=self.ts_pass_train, **cov_kwargs
         )
 


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
+- Fixed a bug when saving a naive model directly after fitting it (without performing prediction). [#2839](https://github.com/unit8co/darts/issues/2839), by [Alain Gysi](https://github.com/Kurokabe)
```
dennisbader: we always point at the PR directly instead of the issue :) 

```suggestion
- Fixed a bug when saving a `GlobalNaiveModel` directly after fitting it (without performing prediction). [#2895](https://github.com/unit8co/darts/pull/2895), by [Alain Gysi](https://github.com/Kurokabe)
```

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -291,6 +291,16 @@ def test_save_model_parameters(self, config):
...
+            GlobalNaiveDrift(
+                input_chunk_length=4,
+                output_chunk_length=3,
+                **tfm_kwargs,
+            ),
+            GlobalNaiveAggregate(
+                input_chunk_length=4,
+                output_chunk_length=3,
+                **tfm_kwargs,
+            ),
```
dennisbader: Testing only `GlobalNaiveSeasonal` is fine since they all use the same base model logic behind the hood

```suggestion
```

darts/tests/models/forecasting/test_global_forecasting_models.py
```diff
@@ -376,6 +386,69 @@ def test_save_load_model(self, tmpdir_fn, model):
...
+        ],
+    )
+    def test_save_load_naive_torch_model(self, tmpdir_fn, model):
```
dennisbader: I would suggest that instead of adding a complete new test here, you could just extend `test_save_load_model()` to store an additional model save right after `fit()`, and then load it and compare it to the other predictions. It's good to test this for all the other models as well :)

darts/models/forecasting/global_baseline_models.py
```diff
@@ -182,7 +182,20 @@ def fit(
         self
             Fitted model.
         """
-        return super().fit(series, past_covariates, future_covariates, *args, **kwargs)
+        output = super().fit(
```
dennisbader: Nice, this is definitely a solution that would work :) 

There is a downside to this though that we would have to run the prediction on all time series (it can be thousands of series).

To avoid this, I would suggest we try out something `trainer.strategy.connect(model)` to connect the model to the trainer's strategy.

You could put it in the `else:` condition [here](https://github.com/unit8co/darts/blob/8821f5109a59605f25a0a259c3f36ac31d094ac8/darts/models/forecasting/torch_forecasting_model.py#L1316).

What do you think?

PR #2898, opened 2025-09-15
Feature/ Skip resampling in TFT to provide speed-up

### LINKED ISSUE #2897: Significant bottleneck in TFT model training & inference
There is a significant time bottleneck in the current implementation of TFT. The bottleneck is most significant when the training set is large (>10K samples). On NVIDIA V100 GPU, we observed 90% overheads from `_TimeDistributedInterpolation`. That module is responsible for resampling on embedding dimension in `_ResampleNorm` and `_AddNorm` when the input size does not match the output size. However, we observed that resampling can be skipped or replaced by linear projection without affecting the accuracy of the TFT. 

On our interval datasets, skipping the resampling operation leads to 10x speedup on V100 GPU and 30x speedup on H100 GPU.

Provide a `skip_resampling` option for TFT. When set to `True`, the behaviours of TFT are changed as follows:
- Resampling in `flattened_grn` of `_VariableSelectionNetwork` is skipped.
- `_ResampleNorm` for `single_variable_grns` of `_VariableSelectionNetwork` is replaced by a `_ProjectNorm`.
    - `_ProjectNorm` replaces the resampling in `_ResampleNorm` with linear projection.
- Resampling in `_GatedResidualNetwork` for `single_variable_grns` of `_VariableSelectionNetwork` is skipped.

By default, `skip_resampling` is set to `False` and the old behaviours (sampling) of TFT are retained.

Skipping the resampling operation by default, but that might lead to compatibility issues with previously trained TFT checkpoints.

**Additional context**
A Python script for running the TFT training & inference is provided.
```python
import logging
import time

import numpy as np
import pandas as pd

from darts import TimeSeries

logging.basicConfig(level=logging.INFO)

# Generate the dataset and preprocess it
def generate_series(length: int, num_covariates: int):
    times = pd.date_range("20130101", periods=length, freq="s")
    series = TimeSeries.from_times_and_values(
        times=times,
        values=np.cos(np.linspace(0, np.pi * 10, length, dtype=np.float32)).reshape(-1, 1)
    )
    future_covariates = TimeSeries.from_times_and_values(
        times=times,
        values=np.random.rand(length, num_covariates).astype(np.float32)
    )
    return series, future_covariates

series, future_covariates = generate_series(length=10000, num_covariates=15)

# Split the dataset into training and validation sets
train_series, val_series = series.split_after(0.8)

model_kwargs = {
    "batch_size": 512,
    "n_epochs": 1,
    "pl_trainer_kwargs": {
        "accelerator": "gpu",
    },
    "optimizer_kwargs": {"lr": 1e-2},
}
fit_kwargs = {
    "dataloader_kwargs": {
        "num_workers": 0,
    }
}

# Define the TFT model
model = TFTModel(
    input_chunk_length=100,
    output_chunk_length=100,
    # skip_resampling=False,
    **model_kwargs
)

# Train the TFT model
start_time = time.time()
model.fit(
    train_series,
    val_series=val_series,
    val_future_covariates=future_covariates,
    verbose=True,
    **fit_kwargs
)
print(f"Training time: {time.time() - start_time:.4f} seconds")
```

daidahao:

Fixes #2897.

### Summary

Resampling in TFT's VariableSelectionNetwork has introduced overheads
for training due to slow `interpolate()` implementation in PyTorch. I've
added an option `skip_interpolation` to replace such operations in TFT
while accuracy are largely not affected.

- `skip_interpolation` defaults to `False` and TFT would retain the old
  behaviour of applying interpolation on feature embeddings.
- When set to `True`, all interpolation operations are replaced by 
  projection (in `_ResampleNorm` or `_GatedResidualNetwork`).
- Quite a few typing errors are fixed in TFT.
- Fix a TFT kernel error on MPS device.
- Add TFT tests for MPS devices and `skip_interpolation` option.

### Other Information

A Python script for running the TFT training & inference is provided.

```py
import logging
import time

import numpy as np
import pandas as pd

from darts import TimeSeries

logging.basicConfig(level=logging.INFO)

# Generate the dataset and preprocess it
def generate_series(length: int, num_covariates: int):
    times = pd.date_range("20130101", periods=length, freq="s")
    series = TimeSeries.from_times_and_values(
        times=times,
        values=np.cos(np.linspace(0, np.pi * 10, length, dtype=np.float32)).reshape(-1, 1)
    )
    future_covariates = TimeSeries.from_times_and_values(
        times=times,
        values=np.random.rand(length, num_covariates).astype(np.float32)
    )
    return series, future_covariates

series, future_covariates = generate_series(length=10000, num_covariates=15)

# Split the dataset into training and validation sets
train_series, val_series = series.split_after(0.8)

model_kwargs = {
    "batch_size": 512,
    "n_epochs": 1,
    "pl_trainer_kwargs": {
        "accelerator": "gpu",
    },
    "optimizer_kwargs": {"lr": 1e-2},
}
fit_kwargs = {
    "dataloader_kwargs": {
        "num_workers": 0,
    }
}

# Define the TFT model
model = TFTModel(
    input_chunk_length=100,
    output_chunk_length=100,
    skip_interpolation=False, # Change to `True` for speed-up
    **model_kwargs
)

# Train the TFT model
start_time = time.time()
model.fit(
    train_series,
    val_series=val_series,
    val_future_covariates=future_covariates,
    verbose=True,
    **fit_kwargs
)
print(f"Training time: {time.time() - start_time:.4f} seconds")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Added `add_regressor_configs` parameter to the `Prophet` model, enabling component-specific control over `prior_scale`, `mode`, and `standardize` for the future covariates. [#2882](https://github.com/unit8co/darts/issues/2882) by [Ramsay Davis](https://github.com/RamsayDavisWL).
 - 🔴 Increased the decimal places for quantile component names from 2 to 3 for more precise quantiles. (e.g. `component_name_q0.500` for quantile 0.5). This affects quantile forecasts as well as quantiles computed with `TimeSeries.quantile()`. [#2887](https://github.com/unit8co/darts/pull/2786) by [He Weilin](https://github.com/cnhwl).
 - Added parameter `load_best` to `TorchForecastingModel.fit()` and `fit_from_dataset()` which, when `True`, will automatically load (and use) the best model on the validation set at the end of the training process. [#2903](https://github.com/unit8co/darts/pull/2903) by [He Weilin](https://github.com/cnhwl).
-- Added model creation parameters `random_errors` and `error` to `ExponentialSmoothing` that give control over how probabilistic forecasts are generated. [#2290491](https://github.com/unit8co/darts/pull/2904) by [Jakub Chłapek](https://github.com/jakubchlapek)
+- Added model creation parameters `random_errors` and `error` to `ExponentialSmoothing` that give control over how probabilistic forecasts are generated. [#2904](https://github.com/unit8co/darts/pull/2904) by [Jakub Chłapek](https://github.com/jakubchlapek)
 - Added parameter `val_length` to `ForecastingModel.historical_forecasts()`, `backtest()` and `residuals()` which will extract a validation set of length `val_length` after the end of each training set when `retrain=True`. The validation set is then used to fit the underlying forecasting model if it supports it. This is especially useful for early stopping mechanisms to reduce overfitting and / or training times. [#2894](https://github.com/unit8co/darts/pull/2894) by [Dennis Bader](https://github.com/dennisbader).
 - 🔴 Renamed the `RegressionEnsembleModel` ensemble model attribute from `regression_model` to `ensemble_model` to make it more clear that this model is used to combine the predictions of the base models. [#2894](https://github.com/unit8co/darts/pull/2894) by [Dennis Bader](https://github.com/dennisbader).
 
@@ -31,6 +31,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 **Dependencies**
 
 - We raised the minimum pytorch-lightning version to `pytorch-lightning>=2.0.0`. [#2888](https://github.com/unit8co/darts/pull/2888) by [Dennis Bader](https://github.com/dennisbader).
+- We made the Darts core and `torch` packages lighter by removing XGBoost and StatsForecast from the dependencies. All our forecasting models wrapping around these libraries will still be supported. To use them simply install the packages manually or via `u8darts[notorch]` and `u8darts[all]`. [#2906](https://github.com/unit8co/darts/pull/2906) by [Jakub Chłapek](https://github.com/jakubchlapek)
 
 ### For developers of the library:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,7 +3,7 @@
 Below, we detail how to install Darts using either `conda` or `pip`.
 
 ## From PyPI
-Install Darts with all models except the ones from optional dependencies (Prophet, LightGBM, CatBoost, see more on that [here](#enabling-optional-dependencies)): `pip install darts`.
+Install Darts with all models except the ones from optional dependencies (Prophet, LightGBM, CatBoost, XGBoost, StatsForecast see more on that [here](#enabling-optional-dependencies)): `pip install darts`.
 
 If this fails on your platform, please follow the official installation
 guide for [PyTorch](https://pytorch.org/get-started/locally/), then try installing Darts again.
@@ -12,8 +12,8 @@ As some dependencies are relatively big or involve non-Python dependencies,
 we also maintain the `u8darts` package, which provides the following alternate lighter install options:
 
 * Install Darts with all available models: `pip install "u8darts[all]"`
-* Install core only (without neural networks, Prophet, LightGBM and Catboost): `pip install u8darts`
-* Install core + Prophet + LightGBM + CatBoost: `pip install "u8darts[notorch]"`
+* Install core only (without neural networks, Prophet, LightGBM, Catboost, XGBoost and StatsForecast): `pip install u8darts`
+* Install core + Prophet + LightGBM + CatBoost + XGBoost + StatsForecast: `pip install "u8darts[notorch]"`
 * Install core + neural networks (PyTorch): `pip install "u8darts[torch]"` (equivalent to `pip install darts`)
 
 ## From conda-forge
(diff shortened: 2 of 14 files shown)


daidahao:
Hi @dennisbader would you mind reviewing this PR when you are free?

daidahao:
<del>Seems like the error came from modifying a shallow copy of `tfm_kwargs` in `test_on_mps`. I will fix that in the next commit.</del>

<del>All the tests should be fixed now (running fine on my MacBook).</del>

Remove the MPS test because MPS is not available on GitHub Actions.

darts/models/forecasting/tft_model.py
```diff
@@ -368,11 +376,11 @@ def decoder_variables(self) -> list[str]:
         return self.variables_meta["model_config"]["time_varying_decoder_input"]
 
     @staticmethod
-    def expand_static_context(context: torch.Tensor, time_steps: int) -> torch.Tensor:
+    def expand_static_context(context: torch.Tensor) -> torch.Tensor:
         """
         add time dimension to static context
         """
-        return context[:, None].expand(-1, time_steps, -1)
+        return context.unsqueeze(1).contiguous()
```
dennisbader, line 384: Nice one 👍 the static cov handling looks much cleaner now :)

darts/models/forecasting/tft_submodels.py
```diff
@@ -236,13 +241,46 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return output
 
 
+class _ProjectNorm(nn.Module):
```
dennisbader: Could you integrate this into `_ResampleNorm` instead of having a dedicated module? In the end, the code is identical. Only that this one uses nn.Linear instead of _TimeDistributedInterpolation.

A parameter could then control which one should be used. (also some documentation / description would be nice that explains why one would choose projection over td interpolation

darts/models/forecasting/tft_submodels.py
```diff
@@ -275,32 +313,40 @@ class _GateAddNorm(nn.Module):
...
         self.glu = _GatedLinearUnit(
             self.input_size, hidden_size=self.hidden_size, dropout=self.dropout
         )
-        self.add_norm = _AddNorm(
-            self.hidden_size,
-            skip_size=self.skip_size,
-            trainable_add=trainable_add,
-            norm=layer_norm,
-        )
+        if not self.skip_add:
```
dennisbader: It looks to me like this would skip the residual connection from the GRN (see top left in the image below).

Couldn't we also swap the `_TimeDistributedInterpolation` resampling in `_AddNorm`  with the linear projection (like you do in `_VariableSelectionNetwork`) to still be able to make the residual connection when `skip_resample=True`?

I'm just thinking out loud here and it has been a while I haven't deep-dived into TFTModel :D let me know if I missed something.

image

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
+- Added hyperparameter `skip_resampling` in `TFTModel` to skip resampling operations on feature embeddings which may cause time bottlenecks. When set to `True`, it could provide substantial training and inference speed-up. [#2897](https://github.com/unit8co/darts/issues/2897) by [Zhihao Dai](https://github.com/daidahao).
```
dennisbader:
```suggestion
- Added hyperparameter `skip_resampling` to `TFTModel` that will skip resampling operations on feature embeddings. When `True`, it can greatly increase training and inference efficiency while predictive accuracy remains largely unaffected. [#2898](https://github.com/unit8co/darts/pull/2898) by [Zhihao Dai](https://github.com/daidahao).
```

CHANGELOG.md
```diff
@@ -19,6 +20,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
+- Fixed a bug causing crashes when running `TFTModel` on MPS devices (macOS with GPUs). [#2897](https://github.com/unit8co/darts/issues/2897) by [Zhihao Dai](https://github.com/daidahao).
```
dennisbader: Very nice :) I'm curious, what was actually the part that caused the crashed?

daidahao: Aha it took me a while to debug but `expand_static_context()` was the reason for crashing on MPS.

darts/models/forecasting/tft_submodels.py
daidahao: That was actually my initial implementation. My concern was that if we integrate projection into `_ResampleNorm`, this may cause confusion from the naming. When resampling is replaced here, we are still building a connection between  input and output, but using linear projection here.

darts/models/forecasting/tft_submodels.py
daidahao: It's mostly for efficiency. Although `_GatedResidualNetwork` is used throughout TFT, but resampling within it could only happen at two locations:
- In `flattened_grn` when `input_size_total` != `num_inputs`.
- In `single_variable_grns` of a variable when its `input_size` != TFT's `hidden_size`.

Because `flattened_grn` is for computing variable weights and `single_variable_grns` for embedding each variable, I don't think 1D resampling is appropriate here in the first place. You are right that residual connection via projection could have some benefits here, at some costs of efficiency.

If you would like, I could implement a linear projection here.

darts/models/forecasting/tft_submodels.py
dennisbader: I think the name _ResampleNorm fits for both time distributed interpolation or linear projection.
Maybe instead we could rename the flag that you added to `skip_interpolation` or something?

darts/models/forecasting/tft_submodels.py
dennisbader: Yes, I think we should try adding the projection here as well and test it. In the end it would be nice to have similar logic applied for both cases. What do you think?

darts/models/forecasting/tft_submodels.py
daidahao: Great. I've renamed the flag to `skip_interpolation` to make it more clear. I've also integrated `_ProjectNorm` into `_ResampleNorm`.

darts/models/forecasting/tft_submodels.py
daidahao: I've updated the code d9a5b2e to replace all resampling interpolation with linear projection when `skip_interpolation` is `True`, instead of removing the resampling all together.

daidahao:
So now I've updated the code d9a5b2e to replace all interpolation resampling with linear projection when `skip_interpolation` is `True`. The speed-up is a bit less compared to the initial PR, about 5.6x on V100 now (6.2x before). But it preserves the residual connections in the original TFT paper.

dennisbader (approved):
Looks beautiful now, thanks for this great PR and another contribution @daidahao 🚀 

I tested and it looks like everything is working well!

Also btw, I'm not sure if you already knew, but it seems like your PR also fixed TFT support for mixed precision :)

daidahao:
Very nice! Yes, TFT seems to work fine on mixed precision now haha.

PR #2318, opened 2024-04-11
Auto batch size for torch model

BohdanBilonoh:
Checklist before merging this PR:
- [x] Mentioned all issues that this PR fixes or addresses.
- [x] Summarized the updates of this PR under **Summary**.
- [x] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md). 

### Summary

Auto batch size finding for `TorchForecastingModel`. This is just a wrapper for `lightning.pytorch.tuner.tuning.Tuner.scale_batch_size`

PR #2905, opened 2025-09-18
Feature/ Find largest batch size for `TorchForecastingModel`

daidahao:

Closes #2318.

### Summary

 **Add `scale_batch_size()` to `TorchForecastingModel`:**
  
  - A wrapped around Lightning Tuner's method of the same name, `scale_batch_size()` finds the largest batch size before out-of-memory error for training or prediction.
  - Options for Tuner method are supported, including `mode`, `steps_per_trial`, `init_val`, and `max_trials`.
  - Tuner requires a `batch_size` attribute within `LightningDataModule` or model and disallows previous `train_loader` and `val_loader`. Because of that, I implemented `_CustomDataModule` to return dataloaders as per `batch_size`.
  - The previous behaviours of `dataloader_kwargs` are being preserved with the new datamodules.
  - Update `_setup_for_train()`, `_train()`, `fit_from_dataset()`, `lr_find()` methods to use datamodules instead of direct data loaders.

**Testing:**
  - Add `test_scale_batch_size` for validating `scale_batch_size()` method.
  - Add `test_scale_batch_size_no_updates` for validating that batch size scaling does not update model weights.
  - Update `test_dataloader_kwargs_setup` to validate `datamodule` instead of `train_dataloaders` and `val_dataloaders` due to changes.
  - Update `helper_check_val_set` used in `test_val_set` to again validate `datamodule`.

### Other Information

Should we remove arguments like `val_*`, `sample_weight` from `scale_batch_size()`? They do not affect the batch size scaling and removing them could simplify the method call.


dennisbader:
Thanks for this PR @daidahao . I remember when experimenting with the other PR and the feature in general, that I wasn't fully convinced of the functionality.
Even if it might give us the largest batch size possible, I'm not sure that training on that batch size is always what we want in the end (on most use cases it just ended up in having all training data in one batch). 

Have you used it yourself and if yes, what's your perspective on it? I'm happy to discuss :)

daidahao:
Hi Dennis @dennisbader , I think it would depend on the uses cases, particularly the dataset size and the hardware. The main benefit of scaling the batch size is to maximise the GPU usage, when the dataset is too large to fit into one batch and could take very long time to train. In those cases, we often scale the batch size manually by power of 2 until the GPU is fully utilised (~100%) to speed up training, similar to what the feature could do.

You are right in that in many cases, scaling batch size is less helpful because many datasets are simply too small to see the benefits. Even in cases of larger datasets, the user might not benefit from this functionality if they are using only CPUs or less powerful GPUs. A small batch size would easily saturate the hardware.

The issue is that there is no tuner from Lightning that could scale the batch size by the CPU/GPU utilisation. This tuner provides a close approximation by reaching the out-of-memory error.

To your second question, we are now using Darts on a large dataset with 1M+ time points and often find ourselves manually tuning the batch size whenever a hyper-parameter has been changed. I could see the implementation of this feature benefiting our model training process greatly. But unlike mixed precision or skip-interpolation TFT, we could not easily overwrite the methods on an older version of Darts (from conda). I will try to update to a newer version and test this feature in the coming days.

Publicly, what would be sensible is perhaps to find a large public dataset for testing this feature and best reproducibility. I reckon that datasets like `AirPassengersDataset` is only a few hundred time points and too small for testing. What would be your suggestions on this?

We could reuse the script from #2898 for benchmarking.

```python
import logging
import time

import numpy as np
import pandas as pd

from darts import TimeSeries

logging.basicConfig(level=logging.INFO)

# Load the dataset
series, future_covariates = ...

# Split the dataset into training and validation sets
train_series, val_series = series.split_after(0.8)

model_kwargs = {
    "batch_size": 512,
    "n_epochs": 1,
    "pl_trainer_kwargs": {
        "accelerator": "gpu",
    },
    "optimizer_kwargs": {"lr": 1e-2},
}
fit_kwargs = {
    "dataloader_kwargs": {
        "num_workers": 0,
    }
}

# Define the TFT model
model = TFTModel(
    input_chunk_length=100,
    output_chunk_length=100,
    skip_interpolation=True, # Change to `True` for speed-up
    **model_kwargs
)

# Train the TFT model
start_time = time.time()
model.fit(
    train_series,
    val_series=val_series,
    val_future_covariates=future_covariates,
    verbose=True,
    **fit_kwargs
)
print(f"Training time: {time.time() - start_time:.4f} seconds")

# Test the TFT model
# TODO
```

dennisbader:
@daidahao, this could come in handy now actually with the foundation models :) I will need some time come back to reviewing it though, as I would like to prioritize the other PRs at the moment. I am aiming at doing a new release within the next two weeks 🚀

daidahao:
> @daidahao, this could come in handy now actually with the foundation models :) I will need some time come back to reviewing it though, as I would like to prioritize the other PRs at the moment. I am aiming at doing a new release within the next two weeks 🚀

Indeed, I could see the benefits of it for "large" models like Chronos-2. There is no need to rush this and we can wait until after the next release.

dennisbader:
@daidahao, I would like to take this up again :) After locally updating the darts foundation paper repo for the variable input chunk length support, I do see that infering the largest batch size can be very useful.

daidahao:
@dennisbader Yes, absolutely! Let me know how I can help.

dennisbader (approved):
@daidahao, all good thanks :) I've finalized the PR now and added support for finding the largest batch size for prediction / inference as well. Should be good to go now 🚀 Thanks again for this PR 💯

PR #2908, opened 2025-09-21
Improved parameter validation for `TorchForecastingModel`

### LINKED ISSUE #2843: Modified parameter validation to support better subclassing
Current parameter validation for TorchForecasting models seems to break when using subclassing of the existing models because it is using hardcoded TorchForecastingModel/PLForecastingModule/cls combination:
https://github.com/unit8co/darts/blob/2391d8bb1c8290c9b940d5658c47072fe3de0ca8/darts/models/forecasting/torch_forecasting_model.py#L375C9-L375C31

Using the following logic that inspects the whole chain of arguments seems to work well

```    
    @classmethod
    def _validate_model_params(cls, **kwargs):
        valid_kwargs = set(
            inspect.signature(PLForecastingModule.__init__).parameters.keys()
        )

        for base in inspect.getmro(cls):
            if base is object:
                break
            try:
                sig = inspect.signature(base.__init__)
                valid_kwargs.update(sig.parameters.keys())
            except (ValueError, TypeError):
                # In case a built-in class throws or __init__ is not introspectable
                continue

        # Remove 'self' and 'args/kwargs' from consideration
        valid_kwargs.discard("self")
        valid_kwargs.discard("args")
        valid_kwargs.discard("kwargs")

        invalid_kwargs = [kwarg for kwarg in kwargs if kwarg not in valid_kwargs]

        raise_if(
            len(invalid_kwargs) > 0,
            f"Invalid model creation parameters. Model `{cls.__name__}` has no args/kwargs `{invalid_kwargs}`",
            logger=logger,
        )
```

tRosenflanz:
Fixes #2843 

### Summary

Simple improvement to consider parameters of the full class inheritance of the checked class instead of specific classes for `TorchForecastingModel`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 ### For developers of the library:
 
+ - Reworked the `_validate_model_params` function of `TorchForecastingModel` to support more complicated cases of class inheritance. [#2908](https://github.com/unit8co/darts/pull/2908) by [Tim Rosenflanz](https://github.com/trosenflanz).
+
 ## [0.37.1](https://github.com/unit8co/darts/tree/0.37.1) (2025-08-18)
 
 ### For users of the library:
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -362,14 +362,24 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
 
     @classmethod
     def _validate_model_params(cls, **kwargs):
-        """validate that parameters used at model creation are part of :class:`TorchForecastingModel`,
-        :class:`PLForecastingModule` or cls __init__ methods.
-        """
-        valid_kwargs = (
-            set(inspect.signature(TorchForecastingModel.__init__).parameters.keys())
-            | set(inspect.signature(PLForecastingModule.__init__).parameters.keys())
-            | set(inspect.signature(cls.__init__).parameters.keys())
+        # initiate with PLForecastingModule params that isn't part of the base class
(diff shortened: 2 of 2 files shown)


tRosenflanz:
Do you want me to add a test for this? I can use the example from the original issue for it

dennisbader (changes requested):
Hi @tRosenflanz and thanks a lot for the PR 🚀 

Yes, a unit test would be great :) Could you also:
- try to improve the code coverage with the new tests (from the `codecov` comment above). 
- add an entry to `CHANGELOG.md` describing your change

tRosenflanz:
Can I rework the [`test_wrong_model_creation_params`](https://github.com/unit8co/darts/blob/e1067715fc729377ec4775e763b7a325227eb541/darts/tests/models/forecasting/test_torch_forecasting_model.py#L1480) using a RnnLambda class or would you prefer a separate test? They will be testing essentially the same behavior so I reworked the existing test but had to change the params  because inheritance changes the argument order

Also I will remove that uncovered Try/except block - it only comes into play with very exotic cases that should probably be forbidden anyway like using double inheritance with C implemented types like `list`

dennisbader:
Hi @tRosenflanz, removing the try / except sounds good. Probably better to let it fail then continue silently and we won't know what went wrong.

I would prefer a new test. You can pretty much copy-paste the `test_wrong_model_creation_params`, and adapt it to use a Subclass that adds one additional parameter. You can then add this param to the existing `valid_kwarg` and test that it works. `invalid_kwarg` should still raise an exception.

dennisbader (approved):
Perfect, thanks a lot @tRosenflanz, this really simplifies subclassing from TFM 🚀 

I pushed a minor change to also test that passing the positional arguments works.

PR #2920, opened 2025-10-04
[Feature]: implement median sampling for quantile regression

### LINKED ISSUE #2912: Feature request: allow `predict_likelihood_parameters` for auto-regressive `QuantileRegression`
Currently, sampling from torch-based, auto-regressive `QuantileRegression` models is slow and consumes a great amount of memory.
Allowing `predict_likelihood_parameters` for auto-regressive `QuantileRegression` could skip Monte Carlo sampling entirely and speed up forecasting. 

The issues has been explored previously in #2731, but I would like to re-open the issue here. I think a slight alternation to the solution might lead to better results.

1. `model.predict()` allows `predict_likelihood_parameters=True` and `n>output_chunk_length` only when `num_samples=1` and `model.likelihood` is `QuantileRegression` (for now, other likelihoods could be explored in the future).
2. Inside `PLForecastingModule.predict_step()` when making auto-regressive, batch prediction, appends the **median ($q=0.5$)** prediction to the target.

This is also a common sampling approach with some of the latest deep learning models, such as Chronos-Bolt, TimesFM, etc.

We could also set `num_samples=len(quantiles)` for `model.predict()`. When appending the prediction, we first get new quantiles from such those samples and use those new quintiles as the new samples for target. However, this may add great complexity to the implementation since the samples might be spread across different batches.

**Additional context**
I tried replicating the experiment in #2731 (the best I could) and am getting better convergence than before. However, my model might not have the same hyper-parameters as @dennisbader, so please take that with a grain of salt.
- `input_chunk_length=24`
- `output_chunk_length=12`
- `quantiles=[0.1, 0.5, 0.9]`
- "forecast sampled" is plotted with `[0.1, 0.9]` quantiles.

On the other hand and fundamentally, this is a different sampling method than Monte Carlo. Divergence should be small but is expected over longer auto-regressive horizons.

Image

daidahao:

Fixes #2912.

### Summary
**This is a DRAFT PR and an experimental implementation.**

Different to Monte Carlo sampling, median sampling consumes the median quantile of last output chunk as target input for making autoregressive forecasts (when horizon > output chunk length). It allows predicting likelihood parameters (quantiles) with a single sample, as opposed to using $\gg 1$ samples in MC.

Implement median sampling for torch-based quantile regression `.predict()` when the following conditions are all met:
- `model.likelihood` is `QuantileRegression`,
- `num_samples == 1`,
- `predict_likelihood_parameters=True`.

Additionally, we implement median sampling for conformal model `ConformalQRModel` when the above conditions and the below are all met:
-  `quantiles == model.likelihood.quantiles`.

See #2912 for previous discussions.

### DIFF
--- a/darts/models/forecasting/conformal_models.py
+++ b/darts/models/forecasting/conformal_models.py
@@ -18,6 +18,7 @@ from darts.utils.likelihood_models.base import (
     LikelihoodType,
     quantile_names,
 )
+from darts.utils.likelihood_models.torch import QuantileRegression
 
 try:
     from typing import Literal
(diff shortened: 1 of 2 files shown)


daidahao:
@dennisbader Let me know how these play out in other scenarios!

In terms of implementation, I think we could provide an additional option, say, `enable_median_sampling` for `.predict()`, `historical_forecasts()` methods to explicitly enable median sampling, avoid sanity check conflicts, and differentiate from MC.

PR #2931, opened 2025-10-17
Changes to ts representation

jakubchlapek:

### Summary

Changed the TimeSeries representation both for terminal and notebook html.

### Other Information

The representation no longer relies on xarray, and also should be more intuitive as well as provide a better overview over the data. 

#### New HTML representation:

image

#### New string representation:

image

### DIFF
--- a/darts/datasets/datasets.py
+++ b/darts/datasets/datasets.py
@@ -530,7 +530,7 @@ class ElectricityDataset(DatasetLoaderCSV):
 
     References
     ----------
-    .. [1] https://archive.ics.uci.edu/ml/datasets/ElectricityLoadDiagrams20112014
+    .. [1] https://archive-beta.ics.uci.edu/dataset/321/electricityloaddiagrams20112014
 
     """
 
@@ -552,7 +552,8 @@ class ElectricityDataset(DatasetLoaderCSV):
         super().__init__(
             metadata=DatasetLoaderMetadata(
                 "Electricity.csv",
-                uri="https://archive.ics.uci.edu/static/public/321/electricityloaddiagrams20112014.zip",
+                uri="https://cdn.uci-ics-mlr-prod.aws.uci.edu/321/electricityloaddiagrams20112014.zip",
+                # uri="https://archive.ics.uci.edu/static/public/321/electricityloaddiagrams20112014.zip",
                 hash="acfe6783eea43905e510f537add940fd",
                 header_time="Unnamed: 0",
                 format_time="%Y-%m-%d %H:%M:%S",
--- a/darts/tests/utils/historical_forecasts/test_historical_forecasts.py
+++ b/darts/tests/utils/historical_forecasts/test_historical_forecasts.py
@@ -427,14 +427,14 @@ class TestHistoricalforecast:
 
     ts_past_cov_train_range = tg.gaussian_timeseries(
         length=len(ts_pass_train_range),
-        freq=ts_pass_train_range.freq_str,
+        freq=ts_pass_train_range.freq,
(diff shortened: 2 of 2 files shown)


darts/tests/utils/test_formatting.py
```diff
@@ -0,0 +1,117 @@
...
+        assert truncate_key("a" * 20, max_len=20) == "a" * 20
+        assert truncate_key("a" * 21, max_len=20) == "a" * 17 + "..."
+        assert len(truncate_key("a" * 30, max_len=20)) == 20
+
+    def test_format_dict_empty(self):
+        """Test format_dict with empty dict."""
+        assert format_dict({}) == "    <empty>"
+        assert format_dict({}, render_html=True) == "&lt;empty&gt;"
+
+    def test_format_dict_items(self):
```
dennisbader, line 37: Would be more robust and easier to interpret if we compared against the actual complete expected string instead of "in"

darts/tests/utils/test_formatting.py
```diff
@@ -0,0 +1,117 @@
...
+        """Test format_dict with items."""
+        d = {"key1": "value1", "key2": "value2"}
+        result = format_dict(d)
+        assert "    key1" in result
+        assert "value1" in result
+        assert "\n" in result
+
+    def test_format_dict_truncation(self):
```
dennisbader, line 51: same comment here and in other tests below (you can also pick lower "max" display values to check against shorter expected strings

darts/tests/utils/test_formatting.py
dennisbader, line 1: also here, I would suggest to compare against the actual expected representation. This will allow a future developer to spot some issues more efficiently

darts/timeseries.py
```diff
@@ -5482,24 +5489,123 @@ def __ge__(self, other) -> np.ndarray:
...
+            values_str += "\n\nInfo: only sample median was displayed"
+
+        freq_str = self._freq_str if self._freq_str is not None else str(self.freq)
+
+        # create a dict for consistent formatting with other sections
+        info_dict = {
+            "Shape": f"(times: {self.n_timesteps}, components: {self.n_components}, samples: {self.n_samples})",
+            "Time frame": f"({self.start_time()}, {self.end_time()}, {freq_str})",
+            "Size": format_bytes(self._values.nbytes),
+        }
```
dennisbader: This is almost identical to the code in _repr_html_ except:

- df.to_string vs df.to_html (but this we can make dynamic via `getattr(df, repr_method)()`)
-and the probabilistic info `"\n\nInfo: only sample median was displayed"` has some differences with `"\n"`

Can we try to refactor this into a function used by both methods? :)

darts/timeseries.py
```diff
@@ -5482,24 +5489,123 @@ def __ge__(self, other) -> np.ndarray:
...
+                open_by_default=not hierarchy_empty,
+            )
+            + make_collapsible_section(
+                "Metadata",
+                f"{format_dict(self.metadata, render_html=True)}",
+                open_by_default=not metadata_empty,
+            )
         )
 
+    def _repr_dataframe(self, max_rows: int, max_cols: int) -> pd.DataFrame:
```
dennisbader: I added this to display any timeseries with O(1) complexity. I quickly tested where before it took a couple of seconds to display a sampled series with 1 mio rows :)

darts/utils/_formatting.py
dennisbader, line 1: Can you add parameter descriptions to all the functions? Even though we hide this module in the user documentation, for future devs it will be useful :)

Issue #2933, opened 2025-10-20
Discussion: Foundational / Pre-trained Models in Darts


daidahao:
**Is your feature request related to a current problem? Please describe.**
Amazon has just released [Chronos 2](https://github.com/amazon-science/chronos-forecasting), a pre-trained Transformer model for forecasting. Unlike previous Chronos, it has direct multivariate and (future and past) covariate support.

I wonder if there is any interest from the Darts side to implement and support it inside Darts? If so, I am happy to try implementing a prototype in Darts APIs.

**Describe proposed solution**
A prototype implementation of pre-trained Chronos 2 in Darts torch model APIs. Some additional dependencies may be required (from a glance of [Amazon code](https://github.com/amazon-science/chronos-forecasting/tree/main/src/chronos/chronos2)):
- einops (maybe replaced by torch operations).

The user would also need to download pre-trained model weights (120M parameters, around 450MB) from Amazon S3 bucket.

**Describe potential alternatives**
A clear and concise description of any alternative solutions or existing features that might solve it.

**Additional context**

JuanCruzC97:
It would be great to have the darts interface available to evaluate this kind of pretrained models.

dennisbader:
Hi @daidahao and @JuanCruzC97, and thanks for raising this issue. We are currently re-evaluating whether we could / should interface some of these models.

One or two years ago I was against adding them, but now that the models also support covariates and multivariate forecasting it looks different :)

Before implementing anything, I would first like to think about a couple of high level topics.
It would be very nice to get your feedback on this. We could also shape the plan for this together, if you like :)

- Where would we have this interface? I'm worried specifically about unit testing, where each time our CI/CD pipeline is run, it would have to re-download the models.
  - Would we keep it in the main `darts` repo?
  - Or would create a new repo such as `darts-fm`?
  - Or is there even a better solution which wouldn't burden the main repo too much?
- What models should we add? 
  - In my opinion, all models must be built on `pytorch`. Otherwise, we won't add them.
  - Chronos 2 looks good with rich covariates-, and multivariate forecasting support
  - TimesFM2.5, maybe? There is currently a PR open #2934, even though that one might be a bit too early for me (since we haven't yet made a clear plan)
  - ...?
- API wise, the models must inherit from `GlobalForecastingModel` (I wonder if we could wrap some of the pytorch-based models with our TorchForecastingModel)
  - I hope not much more is needed, except of the downloading capabilities
[...]

PR #2944, opened 2025-11-06
Feature/ Foundation model and Chronos-2

### LINKED ISSUE #2943: [New Model] Chronos 2 zero shot forecasting
Just want to create a sub-issue of #2933 here so when I create a PR for Chronos-2, it won't accidentally close #2933.

Amazon has just released [Chronos 2](https://github.com/amazon-science/chronos-forecasting), a pre-trained Transformer model for forecasting. Unlike previous Chronos, it has direct multivariate and (future and past) covariate support.

I wonder if there is any interest from the Darts side to implement and support it inside Darts? If so, I am happy to try implementing a prototype in Darts APIs.

A prototype implementation of pre-trained Chronos 2 in Darts torch model APIs. Some additional dependencies may be required (from a glance of [Amazon code](https://github.com/amazon-science/chronos-forecasting/tree/main/src/chronos/chronos2)):
- einops (maybe replaced by torch operations).

The user would also need to download pre-trained model weights (120M parameters, around 450MB) from HuggingFace.

**Additional context**

daidahao:

Fixes #2943.

### Summary
This PR adds base `FoundationModel` and `Chronos2Model` to Darts forecasting models.

<details open>
<summary>Major Changes</summary>

- Add `FoundationModel` as a base class for all foundation forecasting models. All foundation models like Chronos-2 should inherit the base to make use of PyTorch datasets, optimized historical forecasting, Lightning APIs for model training (fine-tuning), checkpointing, etc.
- Add `HuggingFaceModelMixin` as a mixin class for foundation models that require downloading model configuration and weights from HuggingFace Hub. The class provides methods for downloading model config, weight files, and loading them into a `PLForecastingModule` instance.
- Add `Chronos2Model` for zero-shot forecasting using Amazon's pre-trained checkpoint. It supports past and future covariates and can converts `quantiles` from Chronos-2 to Darts `QuantileRegression` likelihood model. By default, it is deterministic by outputting only the median quantile. It can be probabilistic with user-selected quantiles from pre-trained quantiles. Fine-tuning is not supported for now.
- Add dependencies for `huggingface-hub` and `safetensors`, the former for downloading model files and the other for loading model weights. Both should be lightweight enough to ship with Darts.
- Raise PyTorch to `>2.0.0` for SDPA and other performance benefits.
</details>

<details>
<summary>Quickstart</summary>

chronos2

```python
import numpy as np
import pandas as pd
from matplotlib import pyplot as plt

from darts.models import Chronos2Model

# maximum prediction length w/o triggering auto-regression where the results
# would diverge from the original implementation due to different sampling methods
prediction_length = 1024
# whether to produce probabilistic forecasts
probabilistic = True
# quantiles to be used for probabilistic forecasts
quantiles = [0.1, 0.5, 0.9]

# convert to float32 due to MPS not supporting float64
ts_energy = ElectricityConsumptionZurichDataset().load().astype(np.float32)

# extract temperature, solar irradiation and rain duration
ts_weather = ts_energy[["T [°C]", "StrGlo [W/m2]", "RainDur [min]"]]

ts_energy = ts_energy[["Value_NE5", "Value_NE7"]]

# create train and validation splits
validation_cutoff = pd.Timestamp("2022-07-01")
ts_energy_train, ts_energy_val = ts_energy.split_after(validation_cutoff)

# load model
model = Chronos2Model(
    input_chunk_length=8192,  # maximum context length
    output_chunk_length=prediction_length,  # maximum prediction length w/o AR
    likelihood=(
        QuantileRegression(quantiles) if probabilistic else None
    ),
)
# fit model w/o fine-tuning
model.fit(ts_energy_train, future_covariates=ts_weather)

# predict on the validation inputs w/ covariates
pred = model.predict(
    n=prediction_length,
    future_covariates=ts_weather,
    predict_likelihood_parameters=probabilistic,
)

# plot results
plt.style.use('ggplot')
fig, ax = plt.subplots(figsize=(15, 5))
ts_energy_val[:prediction_length].plot(label='actual', ax=ax, c='k')
if pred.n_components == ts_energy_val.n_components:
    pred.plot(label='forecast', ax=ax)
else:
    for component in ts_energy_val.components:
        for q in quantiles:
            pred[f"{component}_q{q:.3f}"].plot(label=f'forecast_{component}_{int(q*100)}%', ax=ax)

plt.show()
```
</details>

<details>
<summary>Chronos-2 Adaptation</summary>

Chronos-2 was ported from https://github.com/amazon-science/chronos-forecasting/commit/c23d34cd887b889c302ca7b6df3fa0bca96d78a9 and I have since made some changes to integrate within Darts.
- Remove dependencies on `transformers` and `einops` libraries.
- Load model config and weights from HuggingFace Hub using `HuggingFaceModelMixin`.
- Remove `output_attentions` option from forward pass.
- Integrate likelihood model and loss computation with Darts `QuantileRegression`, and remove original loss computation in forward pass.
- Replace `*Output` return type with direct `torch.Tensor` to comply with Darts `PLForecastingModule` interface.
- Replace `einops` rearrange operations with native PyTorch tensor operations.

The key principle here is to introduce as little dependencies as possible and implement it fully in PyTorch. I think using `chronos-forecasting` library is convenient but has the risk of conflicts with Darts dependencies in the future.

</details>

<details>
<summary>Fidelity Tests</summary>

For validation, I use the `ElectricityConsumptionZurichDataset` to generate forecasts with Darts implementation and the original. See `test_chronos2.py` for details. Because TSFMs might be slower than other torch models, I limit the fidelity tests to 2 (probabilistic or median).

**Important Notes:** Due to differences in probabilistic sampling methods, zero-shot forecasts obtained here would differ from those obtained using the original implementation when prediction horizon `n` is larger than 1024.

</details>

<details open>
<summary>What is Missing</summary>

- Documentation for `FoundationModel` (using `Chronos2Model` as prime example) either as part of Quickstart or User Guide.
- <strike>Example notebook for `Chronos2Model` to showcase Chronos-2 capabilities.</strike>
- <strike>CHANGELOG.</strike>
- <strike>README model catalogue update.</strike>
- <strike>Unit tests for `Chronos2Model`, maybe using a tiny mock Chronos-2 configuration.</strike>
- <strike>Unit tests for `FoundationModel`.</strike>
- <strike>Example code in `Chronos2Model` docstring.</strike>
- (Optional) more fidelity tests for `Chronos2Model` using different datasets?
- Some more I haven't thought of...

</details>

<details open>
<summary>Discussions</summary>

There are a few points I would like to discuss here, following the discussions in #2933:
- <strike>`FoundationModel` inherits from `MixedCovariatesTorchModel` and thus `TorchForecastingModel`. This is to allow optimized historical forecasts, PyTorch data loaders, Lightning APIs to be used. Would that be confusing since it also introduces a lot more parameters from `TorchForecastingModel` and `PLForecastingModule`?</strike> -> Using `TorchForecastingModel` as base for `FoundationModel`.
- <strike>Chronos-2 has its own `RINorm` routine, different than Darts `RINorm` in `io_processor()`. See my notes in `_Chronos2Module.forward()` for details. What would be the best way to integrate both and to ensure fine-tuning support w/ normalized loss?</strike> -> Not integrated for now until fine-tuning is supported.
- What should be included as test cases for Chronos-2, like target-only, target+past, target+future, etc.?
- <strike>I understand fine-tuning is not a priority for now. But since Chronos-2 can be trained using the quantile loss, should we allow fine-tuning Chronos-2 for our users?</strike> -> **Fine-tuning not supported for now.**
- <strike>SDPA is used default by Chronos-2 but only introduced from [PyTorch 2.0.0](https://docs.pytorch.org/tutorials/intermediate/scaled_dot_product_attention_tutorial.html). Should we raise `torch` to 2.0.0 or differentiate between using `torch<2.0.0` and `torch>2.0.0`?</strike> -> **Raised to `>2.0.0`.**

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 **Improved**
 
 - 🚀🚀 Added new forecasting model `Chronos2Model`: Amazon's pre-trained 120M-parameter foundational model for zero-shot forecasting. `Chronos2Model` can be used for univariate, multivariate, and multiple time series forecasting without any training. It supports past and future covariates and can output deterministic or probabilistic forecasts. Check out the new [Chronos-2 Foundation Model Notebook](https://unit8co.github.io/darts/examples/25-Chronos-2-examples.html) for examples and usage of Chronos-2. [#2944](https://github.com/unit8co/darts/pull/2944) by [Zhihao Dai](https://github.com/daidahao).
-- Added an example notebook demonstrating how to use Chronos-2: [Chronos-2 Foundation Model Examples](https://unit8co.github.io/darts/examples/25-Chronos-2-examples.html). [#2944](https://github.com/unit8co/darts/pull/2944) by [Zhihao Dai](https://github.com/daidahao).
 - 🔴 Improved the performance of the `TimeSeries.map()` method for functions that take two arguments. The mapping is now applied on the entire time index and values array which requires users to reshape the time index explicitly within the function. See more information in the `TimeSeries.map()` method documentation. [#2911](https://github.com/unit8co/darts/pull/2911) by [Jakub Chłapek](https://github.com/jakubchlapek)
 
 **Fixed**


daidahao:
@dennisbader Could you please add @abdulfatir as a reviewer? The PR is not fully ready yet due to missing parts and discussion points (see above), but we can ask @abdulfatir for review once those are completed.

dennisbader:
@daidahao, is @abdulfatir not able to review already? I believe every user should be able to add a review by default.

daidahao:
> @daidahao, is @abdulfatir not able to review already? I believe every user should be able to add a review by default.

Aha, my mistakes. I thought we would need to add someone explicitly to review the changes.

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,407 @@
...
+from darts.models.forecasting.torch_forecasting_model import (
+    MixedCovariatesTorchModel,
+    TorchForecastingModel,
+)
+from darts.utils.data.torch_datasets.utils import TorchTrainingSample
+
+logger = get_logger(__name__)
+
+
+class HuggingFaceModelMixin:
```
abdulfatir: Is there a reason why this custom Mixin is needed? `huggingface_hub` already contains [`PyTorchModelHubMixin`](https://huggingface.co/docs/huggingface_hub/en/package_reference/mixins#huggingface_hub.PyTorchModelHubMixin) with similar functionality.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+        d_model: int = 512,
+        d_kv: int = 64,
+        d_ff: int = 2048,
+        num_layers: int = 6,
+        num_heads: int = 8,
+        dropout_rate: float = 0.1,
+        layer_norm_epsilon: float = 1e-6,
+        feed_forward_proj: str = "relu",
+        rope_theta: float = 10000.0,
+        attn_implementation: Literal["eager", "sdpa"] | None = None,
```
abdulfatir, line 66: I saw discussion somewhere on whether SDPA would require changes to torch versions. Note that the benefit that SDPA provides for Chronos-2 is relatively minor, so if needed the default may be changed to eager. See: https://github.com/amazon-science/chronos-forecasting/pull/331

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+        # slice the last num_output_patches hidden states to be input into the output_patch_embedding
+        forecast_embeds = hidden_states[:, -num_output_patches:]
+        quantile_preds: torch.Tensor = self.output_patch_embedding(forecast_embeds)
+
+        quantile_preds = quantile_preds.view(batch_size, -1)
+        quantile_preds = self.instance_norm.inverse(quantile_preds, loc_scale)
+
+        return quantile_preds
+
+    # TODO: fine-tuning support w/ normalized loss
```
abdulfatir, line 457: Would it make sense to only support zero-shot inference in the initial phase? I think fine-tuning support comes with a lot more design decisions such as finding and setting good defaults.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+
+        # reshape quantile_preds to (batch, time, vars, quantiles)
+        quantile_preds = quantile_preds.view(
+            batch_size,
+            n_variables,
+            num_output_patches,
+            self.num_quantiles,
+            self.chronos_config.output_patch_size,
+        )
+        quantile_preds = quantile_preds.permute(0, 2, 4, 1, 3).reshape(
```
abdulfatir, line 539: I think previously this was done with einops. Adding a comment would be helpful here.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+
+        if not self.probabilistic:
+            quantile_preds = quantile_preds[:, :, :, [self.median_idx]]
+
+        return quantile_preds
+
+
+class Chronos2Model(FoundationModel, HuggingFaceModelMixin):
+    _repo_id = "amazon/chronos-2"
+    _repo_commit = "18128c7b4f3fd286f06d6d4efe1d252f1d2a9a7c"
```
abdulfatir: This is not strictly required as we do not plan to make changes to the model artifact on the HF repo. Only README will be updated which does not affect the actual model.

darts/models/components/chronos2_submodels.py
dennisbader, line 1: Can we move this into `darts/models/components`? I think it would be cleaner to have submodules there in the future (it's also excluded from the generated documentation). The reason why it's still there for TFTModel is that we want to avoid some breaking changes due to import errors.

darts/tests/models/forecasting/test_chronos2.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # load the Chronos-2 pipeline
+        pipeline: Chronos2Pipeline = BaseChronosPipeline.from_pretrained("amazon/chronos-2")
+
+        # make predictions
+        quantiles, mean = pipeline.predict_quantiles(
+            inputs=[
+                {
+                    "target": ts_energy_train.values().T,
+                    "past_covariates": {
```
dennisbader, line 137: It looks like here you used past covariates but later on in the Darts implementation you only provide future covariates. Is this intended?

requirements/torch.txt
```diff
@@ -1,3 +1,5 @@
 pytorch-lightning>=2.0.0,<2.5.3
 tensorboardX>=2.1
 torch>=1.8.0
+huggingface-hub>=0.35.3
+safetensors>=0.6.2
```
dennisbader, line 5: I wonder whether we should add these dependencies to our torch flavor (they will always be installed together with `darts` or `u8darts[torch]`, `u8darts[all]`.
We could add them to the optional dependencies that are only installed for unit testing.
Then we can notify the user to install the dependencies if they aren't installed yet

darts/models/forecasting/chronos2_model.py
daidahao: @abdulfatir  Great insight from https://github.com/amazon-science/chronos-forecasting/pull/331 and I  followed it too when it was first posted. Regarding the torch version, I am in favour of raising `torch` to `>=2.0.0` since it was released more than two years ago and could bring performance benefits for Darts users including sdpa, `torch.compile()`, etc. I leave it as is for now and hear what @dennisbader would like to say.

darts/models/forecasting/chronos2_model.py
dennisbader: Yeah, I'm with you on this one. For the beginning we could only support zero-shot which would also simplify adding other models in the future.
Like this we can get the models out to the users rather soon, and listen to their feedback / how desired this feature is.

daidahao: @abdulfatir

I agree. Fine-tuning will be turned off for Chronos-2 for now (`_allows_finetuning=False`).

BTW, I saw a fine-tuning tutorial in `chronos-2-quickstart.ipynb`. Are there any official recommended settings for fine-tuning Chronos-2 now? If not, should we expect them to be released sometime soon?

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+
+    _allows_finetuning = True
+
+    def __init__(
+        self,
+        input_chunk_length: int,
+        output_chunk_length: int,
+        output_chunk_shift: int = 0,
+        probabilistic: bool = True,
```
dennisbader: It would be great to hide/hadle this behavior from/for the user. My idea:

- By default (`likelihood=None, loss_fn=None`) the model will be deterministic. Behind the hood we set `self.probabilistic = False`
- To make the model probabilistic, let the user provide a `likelihood=QuantileRegression(quantiles=user_choice)`. The user can pick any quantiles of the available ones from Chronos2. E.g. `quantiles=[0.1, 0.5, 0.9]`. Like this, users can control which quantiles they would like to see when using `predict_likelihood_parameters=True`. Behind the hood, we will simply extract the corresponding quantiles from the model output in `forward`.

What do you think?

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+        enable_finetuning: bool = False,
+        loss_fn: Optional[nn.Module] = None,
+        likelihood: Optional[TorchLikelihood] = None,
+        local_dir: Optional[Union[str, os.PathLike]] = None,
```
dennisbader, line 571: Could we find a way that the users does not need to control this?
What I would expect is that the first time I'm creating the model, it would download the weights from HF. Creating the model a second time will detect that the model has been loaded already and then load from local.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+
+        .. note::
+            Due to differences in probabilistic sampling methods, zero-shot forecasts obtained here would differ from
+            those obtained using the original implementation when prediction horizon `n` is larger than 1024.
+        """
+        self.local_dir = local_dir
+
+        # validate output_chunk_shift
+        raise_if_not(
+            output_chunk_shift == 0,
```
dennisbader: Is there a reason we shouldn't allow this? I would imagine that we could simply extract part of the model output. Something along the line of `output = output[..., output_chunk_shift:]` in `forward()`.

darts/models/forecasting/chronos2_model.py
daidahao: Thanks. I will add some comments here. Addressed by ecc3c960c5c078843e46638b5a874f613df9f29c.

darts/models/forecasting/chronos2_model.py
daidahao: @abdulfatir Thanks for the assurance here😊. While I am not worried about changes on Chronos-2 model weights, I have seen other TSFMs like TimesFM2.5 changing model weights and causing state_dict mismatch. So I enforce `_repo_commit` here for all TSFMs to prevent unexpected CI fails.

darts/models/forecasting/chronos2_model.py
daidahao: Fixed by 5c967ba61b5caa3a2832d174ed4fb699eeca1a3b.

darts/models/components/chronos2_submodels.py
daidahao: Sure thing, I will move this into `darts/models/components` once all comments on the file have been addressed.

darts/models/forecasting/chronos2_model.py
abdulfatir: > Are there any official recommended settings for fine-tuning Chronos-2 now? If not, should we expect them to be released sometime soon?

As noted in that notebook, fine-tuning support is experimental. What we mean by that is that it's not been rigorously tested and the defaults may not be ideal. Chronos-2 with fine-tuning will be available in the next version of AutoGluon (sometime in December), so we plan to iron out the defaults by then.

darts/models/forecasting/chronos2_model.py
abdulfatir: Good point and the design choice makes sense in this case! I can only assure from our side haha. 😄

darts/models/forecasting/chronos2_model.py
abdulfatir: > Creating the model a second time will detect that the model has been loaded already and then load from local.

This is, by the way, the default behavior in hugging face libraries (`huggingface_hub`, `transformers`).

requirements/torch.txt
daidahao: I would recommend adding these dependencies to torch flavour for two reasons:
Most open-source TSFMs are hosted on Huggingface and stored in `safetsensors` format. If we want future TSFMs to work out-of-the-box, they are the absolute minimum for loading TSFMs.

Also, both libraries are rather lightweight. On my macOS 15 with Python 3.13, adding these two libraries Darts would install 11 additional libraries. All those together is only 13MB:

```
 + anyio==4.11.0
 + click==8.3.0
 + h11==0.16.0
 + hf-xet==1.2.0
 + httpcore==1.0.9
 + httpx==0.28.1
 + huggingface-hub==1.1.2
 + safetensors==0.6.2
 + shellingham==1.5.4
 + sniffio==1.3.1
 + typer-slim==0.20.0
```

darts/models/forecasting/chronos2_model.py
daidahao: Very excited for the coming updates!

darts/models/forecasting/chronos2_model.py
daidahao: By default, `huggingface_hub` will download the files to a local cache folder like `~/.cache/huggingface/hub/models--amazon--chronos-2`. Creating the model a second time would not redownload the model files! See [Download files from the Hub](https://huggingface.co/docs/huggingface_hub/en/guides/download) here.

I provide a `local_dir` option here in case the user wants to load model config and weights from another directory.

darts/models/forecasting/foundation_model.py
daidahao: That's a great point. For `HuggingFaceModelMixin`, we only need the download files capabilities `hf_hub_download` from `huggingface_hub` and not much else. `PyTorchModelHubMixin` is very useful but we might not need `push_to_hub` and `save_pretrained` methods for TSFMs in Darts. Those methods also have very similar names to some Darts methods like `load()` and `save()` and may cause some confusion.

Also, the model dimension is a bit implicit with `PyTorchModelHubMixin` style as every hyperparameter is stored in `config`. I prefer being a bit more explicit with the model dimension, dropout, etc. in the model constructor.

darts/tests/models/forecasting/test_chronos2.py
daidahao: The definitions of covariates in Chronos-2 are different to Darts. Here is the mapping between Chronos-2 and Darts:
- past covariates = past cov + historical future covariates
- future covariates = future future covariates

Targets are the same in both. See "(Advanced) Numpy/torch API" section in [chronos-2-quickstart.ipynb](https://github.com/amazon-science/chronos-forecasting/blob/main/notebooks/chronos-2-quickstart.ipynb) here.

So the inputs here are actually equal to providing target + future covariates and no past covariates in Darts.

darts/models/forecasting/chronos2_model.py
daidahao: Great suggestion! Should we throw an error when the `user_quantiles` are not part of Chronos-2 quantiles?

Update: The model will throw an error when `user_quantiles` are not a subset of Chronos-2 quantiles.

darts/models/forecasting/chronos2_model.py
daidahao: Sorry! I was not familiar with `output_chunk_shift` property here so I just disabled it. I now read that autoregression is disabled for `output_chunk_shift>0`. Is there any sanity check to enforce that in `TorchForecastingModel`?

darts/models/components/chronos2_submodels.py
daidahao: Fixed by 1f817fed948651f448c42a4db453e8f1bd87a9b9

darts/models/forecasting/foundation_model.py
abdulfatir: I would still argue that for future-proofing things and reducing the amount of new code, it would better to use the mixin from the library instead of creating a custom mixin. That way things are cached in the standard way and new choices are minimized.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,872 @@
...
+        quantile_preds = quantile_preds[:, :, : self.n_targets, :]
+
+        if not self.probabilistic:
+            quantile_preds = quantile_preds[:, :, :, [self.median_idx]]
+
+        return quantile_preds
+
+
+class Chronos2Model(FoundationModel, HuggingFaceModelMixin):
+    _repo_id = "amazon/chronos-2"
```
abdulfatir: Actually, taking another look at this, and also based on my comment about the HF Mixin, it would be better if this is a customizable parameter. This way the default can still be `amazon/chronos-2` but the user would be able specify other model IDs to potentially use different models compatible with the Chronos-2 architecture. Consider the case where Amazon releases another variant of the model or someone fine-tunes the model and puts it on HF. The flexibility would prove really useful here. Note also that if you use the `PyTorchModelHubMixin`, `from_pretrained` would support hugging face model IDs and local paths which would trivially enable cases where users have local checkpoints/variants of the model.

darts/models/forecasting/chronos2_model.py
abdulfatir: @daidahao Please check my comment above on the `_repo_id`.

darts/models/forecasting/chronos2_model.py
daidahao: Fixed by 35e43f8efa832c7d1ecc1d82661fbf8440e39e5a

darts/models/forecasting/chronos2_model.py
daidahao: Fixed by d307373da2183c5a82e754833bb262567afe3465

darts/tests/models/forecasting/test_chronos2.py
dennisbader: Thanks for the clarification. Could you then extend this test to also use one of the covariates as "past only" (e.g. as `past_covariates` in Darts), and the others as `future_covariates`?
Want to make sure that using `past_covariates` leads to the expected results.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        d_kv: int = 64,
+        d_ff: int = 2048,
+        num_layers: int = 6,
+        num_heads: int = 8,
+        dropout_rate: float = 0.1,
+        layer_norm_epsilon: float = 1e-6,
+        feed_forward_proj: str = "relu",
+        rope_theta: float = 10000.0,
+        attn_implementation: Literal["eager", "sdpa"] | None = None,
+        chronos_config: dict[str, Any] = {},
```
dennisbader: Should avoid to default mutable object
```suggestion
        chronos_config: Optional[dict[str, Any]] = None,
        
        # afterwards below
        ...
        chronos_config = chronos_config or {}
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        quantile_preds = self.instance_norm.inverse(quantile_preds, loc_scale)
+
+        return quantile_preds
+
+    # TODO: fine-tuning support w/ normalized loss
+    # Currently, Darts own `RINorm` is not used as Chronos-2 has its own implementation. Major differences
+    # 1. Chronos-2 `RINorm` normalizes both target and covariates, while Darts normalizes target only.
+    # 2. Chronos-2 `RINorm` additionally applies `arcsinh` transformation after standardization
+    # 3. Chronos-2 uses normalized values for loss computation, while Darts uses denormalized values.
+    # We need to think about how best to implmement Chronos-2 `RINorm` in `io_processor()` without
```
dennisbader:
```suggestion
    # We need to think about how best to implement Chronos-2 `RINorm` in `io_processor()` without
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        """
+        self.local_dir = local_dir
+
+        # load model config for validation
+        config = self._load_config()
+        chronos_config = config["chronos_config"]
+
+        # validate `input_chunk_length` against model's context_length
+        context_length = chronos_config["context_length"]
+        raise_if(
```
dennisbader: Can you use `raise_log` everything instead of `raise_if` and `raise_if_not`?
We want to remove `raise_if*` at some point, as it has some overhead compared to `raise_log` + only with `raise_log` it is clear that always an exception is raised

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        References
+        ----------
+        .. [1] A. Ansari, O. Shchur, J. Küken et al. "Chronos-2: From Univariate to Universal Forecasting", 2025.
+                arXiv https://arxiv.org/abs/2510.15821.
+        .. [2] "Introducing Chronos-2: From univariate to universal forecasting", 2025. Amazon Science Blog.
+                https://www.amazon.science/blog/introducing-chronos-2-from-univariate-to-universal-forecasting
+
+        Examples
+        --------
+        >>> # TODO: add example code here
```
dennisbader: putting a comment not to forget about it

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+    # Fine-tuning is turned off for now pending proper fine-tuning support
+    # and configuration.
+    _allows_finetuning = False
+
+    def __init__(
+        self,
+        input_chunk_length: int,
+        output_chunk_length: int,
+        output_chunk_shift: int = 0,
+        enable_finetuning: bool = False,
```
dennisbader: Let's remove this then, since we won't allow fine-tuning for now?

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+                set(user_quantiles).issubset(quantiles),
+                f"The quantiles for QuantileRegression likelihood {user_quantiles} "
+                f"must be a subset of Chronos-2 quantiles {quantiles}.",
+                logger,
+            )
+
+        super().__init__(enable_finetuning=enable_finetuning, **kwargs)
+
+    @property
+    def user_quantiles(self) -> list[float]:
```
dennisbader: let's remove this property (see my comment below)

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+            if self.probabilistic and self.likelihood
+            else [0.5]
+        )  # pyright: ignore[reportAttributeAccessIssue]
+
+    def _create_model(self, train_sample: TorchTrainingSample) -> PLForecastingModule:
+        pl_module_params = self.pl_module_params or {}
+
+        # additional params
+        additional_params = {
+            "user_quantiles": self.user_quantiles,
```
dennisbader: there is no need to pass this to the _Chronos2Module. The likelihood is passed along in `pl_module_params` and will be assigned already in `_Chronos2Module.__init__()` and accessible with `self.likelihood`.
You can directly get the quantiles from the likelihood there.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        enable_finetuning: bool = False,
+        likelihood: Optional[QuantileRegression] = None,
+        local_dir: Optional[Union[str, os.PathLike]] = None,
+        **kwargs,
+    ):
+        """Chronos-2 Model for Zero-Shot and Fine-Tuned Time Series Forecasting.
+
+        This is an implementation of Amazon's Chronos-2 model [1]_, [2]_, ported from
+        `amazon-science/chronos-forecasting <https://github.com/amazon-science/chronos-forecasting>`_
+        and adapted for Darts APIs. From the original authors:
```
dennisbader:
```suggestion
        with adaptions to use the Darts API. From the original authors:
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        Chronos-2 produces multi-step-ahead quantile forecasts and uses a group attention mechanism for efficient
+        in-context learning across related series and covariates. Trained on a combination of real-world and large-scale
+        synthetic datasets, it achieves state-of-the-art zero-shot accuracy among public models on fev-bench, GIFT-Eval,
+        and Chronos Benchmark II. Chronos-2 is also highly efficient, delivering over 300 time series forecasts per
+        second on a single A10G GPU and supporting both GPU and CPU inference."
+
+        This model supports past covariates (known for `input_chunk_length` points before prediction time),
+        and future covariates (known for `output_chunk_length` points after prediction time).
+
+        Using this model would automatically download the pre-trained model from HuggingFace Hub (amazon/chronos-2).
```
dennisbader:
```suggestion
        By default, using this model will automatically download the pre-trained model from HuggingFace Hub (amazon/chronos-2).
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        Alternatively, you can specify a local directory containing the model config and weights using the
+        ``local_dir`` parameter.
+
+        By default, this model is deterministic and outputs only the median (0.5 quantile). To enable probabilistic
+        forecasts, pass a :class:`darts.utils.likelihood_models.torch.QuantileRegression` instance to the ``likelihood``
```
dennisbader: adding `~` will abbreviate in the rendered docs to only the `QuantileRegression` part

```suggestion
        forecasts, pass a :class:`~darts.utils.likelihood_models.torch.QuantileRegression` instance to the ``likelihood``
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        parameter. The quantiles used must be a subset of those used during Chronos-2 pre-training, see below for
+        details. It is recommended to call :func`predict()` with ``num_samples >> 1`` to get meaningful results.
```
dennisbader:
```suggestion
        details. It is recommended to call :func`predict()` with  ``predict_likelihood_parameters=True`` or
        ``num_samples >> 1`` to get meaningful results.
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+
+        Fine-tuning of Chronos-2 is not supported at the moment. Setting ``enable_finetuning=True`` will raise an error.
```
dennisbader: Let's simply not add the parameter until we actually support it

```suggestion
        Fine-tuning of Chronos-2 is not supported at the moment.
```

darts/models/forecasting/chronos2_model.py
dennisbader: Yes, you can go ahead an raise torch to `>=2.0.0` 👍

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+        -------
+        torch.Tensor
+            the output tensor in the shape of `(n_samples, n_time_steps, n_targets, n_quantiles)` for
+            probabilistic forecasts, or `(n_samples, n_time_steps, n_targets, 1)` for
+            deterministic forecasts (median only).
+        """
+        x_past, x_future, _ = x_in
+        # According to `self._process_input_batch()` in `PLForecastingModule`,
+        # x_past is a stack of [past_target, past_covariates, historic_future_covariates],
+        # while x_future is just future_covariates.
```
dennisbader: It's true :) 
```suggestion
        # x_past is a stack of [past_target, past_covariates, historic_future_covariates],
        # x_future is just future_covariates.
```

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+    max_output_patches: int = 1
+    time_encoding_scale: int | None = None
+
+
+class _Chronos2Module(PLForecastingModule):
+    """
+    Chronos2 module
+    """
+
+    quantiles: torch.Tensor
```
dennisbader: Do we need to keep this (and the handling in `__init__`? Seems like we never really use it

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,869 @@
...
+    use_reg_token: bool = False
+    use_arcsinh: bool = False
+    max_output_patches: int = 1
+    time_encoding_scale: int | None = None
+
+
+class _Chronos2Module(PLForecastingModule):
+    """
+    Chronos2 module
+    """
```
dennisbader:
```suggestion
```

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,402 @@
...
+            **pl_module_params,
+            **additional_params,
+        )
+        self._load_model_weights(module)
+        return module
+
+
+class FoundationModel(MixedCovariatesTorchModel):
+    _allows_finetuning: bool = False
+    _probabilistic: bool = False
```
dennisbader: We can probably infer everything under the hood dynamically, using `likelihood` and `TorchForecastingModel.supports_probabilistic_prediction`

```suggestion
```

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,402 @@
...
+        **kwargs,
+    ):
+        """Foundation Forecasting Model with PyTorch Lightning backend.
+
+        This class is meant to be inherited to create a new foundation forecasting model.
+        It governs the interactions between:
+            - Darts forecasting models (module) :class:`PLTorchForecastingModel`
+            - Darts integrated PL Lightning Trainer :class:`pytorch_lightning.Trainer` or custom PL Trainers
+            - Dataset loaders :class:`TorchTrainingDataset` and :class:`TorchInferenceDataset` or custom Dataset
+              Loaders.
```
dennisbader: Required for rendering docs properly

```suggestion
        
        - Darts forecasting models (module) :class:`PLTorchForecastingModel`
        - Darts integrated PL Lightning Trainer :class:`pytorch_lightning.Trainer` or custom PL Trainers
        - Dataset loaders :class:`TorchTrainingDataset` and :class:`TorchInferenceDataset` or custom Dataset
          Loaders.
        
```

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,402 @@
...
+        raise_if(
+            len(invalid_kwargs) > 0,
+            f"Invalid model creation parameters. Model `{cls.__name__}` has no args/kwargs `{invalid_kwargs}`",
+            logger=logger,
+        )
+
+    @property
+    def probabilistic(self) -> bool:
+        """Whether the model produces probabilistic forecasts."""
+        return self._probabilistic
```
dennisbader: Same comment as above, should be handled with 

`TorchForecastingModel.supports_probabilistic_prediction`

```suggestion
```

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,402 @@
...
+    @property
+    def allows_finetuning(self) -> bool:
+        """Whether fine-tuning is allowed for this foundation model."""
+        return self._allows_finetuning
+
+    @property
+    def enable_finetuning(self) -> bool:
+        """Whether fine-tuning is enabled for this foundation model. When enabled, calling `fit()`
+        will update the model weights. When disabled, calling `fit()` will not update the model weights."""
+        return self._enable_finetuning
```
dennisbader: I think we can drop these properties and leave them as private attributes (I'm working currently on some Documentation updates, and want to reduce a bit was gets displayed to the user).

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,402 @@
...
+
+    @property
+    def _requires_training(self) -> bool:
+        return self.enable_finetuning
+
+    @abstractmethod
+    def _create_model(self, train_sample: TorchTrainingSample) -> PLForecastingModule:
+        """Just like in `TorchForecastingModel`, subclasses must implement this method."""
```
dennisbader: You should be able to remove this when making the class abstract (adding ABC):

`class FoundationModel(MixedCovariatesTorchModel, ABC)`

```suggestion
```

dennisbader (commented):
Great progress @daidahao 🚀 I added some more suggestions.

Regarding the HuggingFaceModelMixin, I will still need to think a bit more about it. @abdulfatir raises some good points about using the [PyTorchModelHubMixin](https://huggingface.co/docs/huggingface_hub/en/package_reference/mixins#huggingface_hub.PyTorchModelHubMixin) instead.

On the other hand I also agree with @daidahao about reducing what is public to the user. 🤔

darts/models/forecasting/chronos2_model.py
daidahao: Done by 32ae5c0f0ca11440d5e617fba430d5e3aba94516

darts/models/forecasting/chronos2_model.py
daidahao: Thanks. I will do some investigations about `PyTorchModelHubMixin` implementation.

darts/models/forecasting/chronos2_model.py
daidahao: Now that we are raising torch to `>=2.0.0`, would it be okay to remove `attn_implementation` option (and any eager implementation) as the user would not set it from `Chronos2Model` and it is always defaulting to sdpa anyway? @abdulfatir

requirements/torch.txt
daidahao: Checking the [GIFT-EVAL](https://huggingface.co/spaces/Salesforce/GIFT-Eval) benchmark, most open-source and open-weights TSFMs are hosted on HuggingFace and stored in `safetensors` format.
- google/timesfm-2.5-200m-pytorch
- ibm-research/flowstate
- Salesforce/moirai-2.0-R-small
- mldi-lab/Kairos_50m
- Datadog/Toto-Open-Base-1.0
- thuml/sundial-base-128m

Many are univariate, but it is a good indicator of the dependencies we would need for future TSFMs. Including `huggingface-hub` and `safetensors` means users can use them out of the box once implemented.

Let me know what you think on this. :) @dennisbader

darts/models/forecasting/chronos2_model.py
daidahao: @dennisbader

These are two separate but related issues here: one for whether we should support future "variants" of Chronos-2, the other being the usage of `PyTorchModelHubMixin`. While the second one is still under investigations, I would like to start some discussions here on the first issue. Should we allow the user to specify another variant of Chronos-2 by changing the repo_id?

If so, we would expose the option `repo_id` to `Chronos2Model` so that a different repo id can be specified. I am a bit neutral on the issue here: on one hand, it allows a "fine-tuned" / "better" variant to be used. On the other hand, it adds to the option count and has no immediate value (no other variant of Chronos-2 exists yet). What do you think? @dennisbader

darts/models/forecasting/foundation_model.py
daidahao: @abdulfatir @dennisbader
I've done diggings into `PyTorchModelHubMixin` by `huggingface_hub` and now think it might not be a good replacement for our `HuggingFaceModelMixin` for two main reasons:
1. Parameter validation: Our current logic of TSFM `__init__()` include hyper-parameter validation like `output_chunk_length` and `quantiles` against model configuration. At this stage, our `HuggingFaceModelMixin` would only download `config.json` file and cache it for later use. But `PyTorchModelHubMixin` does not provide configuration loading and calling `from_pretrained()` would instantiate the `nn.Module` directly.
2. Ambiguous parameter: The `pretrained_model_name_or_path` is quite ambiguous because it can be either a `repo_id` or a local directory. So if a user has a local directory called `amazon/chronos-2` then the behaviour is guaranteed. With our `HuggingFaceModelMixin`, we separate them out into `local_dir` and  `repo_id`, and `local_dir` always overwrite the `repo_id`.

**About Caching:**
Under the hood, both mixin classes use `hf_hub_download` to download model configuration and weight files. So to clarify about caching, our `HuggingFaceModelMixin` shares the same caching mechanism. Model files would not be downloaded twice using either mixins. If the user has downloaded `amazon/chronos-2` using another library like `chronos-forecasting`, those files would not be redownloaded either.

Our `HuggingFaceModelMixin` is actually very lightweight with a few lines of actual core code, mainly for checking the file existence and downloading if not, as well as creating a model instance. The other lines are just there for reference for developers and never shown to the users. (Maybe I could remove the public properties as well to hide them from users in Documentation?)

In terms of future proofing, `HuggingFaceModelMixin` would work as long as `hf_hub_download` accepts `repo_id`, `filename`, and `revision`.

darts/models/forecasting/chronos2_model.py
daidahao: I've added example code here.

CHANGELOG.md
dennisbader, line 1: Let's add an entry to the changelog

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+        num_layers
+            Number of Chronos-2 encoder layers.
+        num_heads
+            Number of attention heads in each encoder block.
+        dropout_rate
+            Dropout rate of the model.
+        layer_norm_epsilon
+            Epsilon value for layer normalization layers.
+        feed_forward_proj
+            Acctivation of feed-forward network.
```
dennisbader:
```suggestion
            Activation of feed-forward network.
```

darts/models/forecasting/chronos2_model.py
dennisbader: @daidahao yes, allowing to use other model ids compatible with the Chronos2 architecture sounds great :) Users could try out other variants without out us having to add them.
This is good parameter to have for other foundation models as well. Apart from fine-tuned models, I'm also thinking of "tiny" models which can simplify unit testing or running on smaller machines.

dennisbader: And I assume this would also include repo commit?

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,388 @@
...
+)
+from darts.models.forecasting.torch_forecasting_model import (
+    MixedCovariatesTorchModel,
+    TorchForecastingModel,
+)
+
+logger = get_logger(__name__)
+
+
+class HuggingFaceModelMixin:
```
dennisbader: How about we convert this mixin to an independent hugging face connector that the model can instantiate and use to load the model? This would:

- reduce the amount of inheritance that we already have
- remove the need for properties and setters
- allow to keep the connector methods public for calling it from other modules, while still not showing anything to the user on documentation side.

I rewrote this class a little, see the code here:

<details>

<summary>HuggingFaceConnector</summary>

```py
class HuggingFaceConnector:
    def __init__(
        self,
        repo_id: str,
        repo_commit: str,
        local_dir: Optional[str] = None,
        config_file: str = "config.json",
        model_file: str = "model.safetensors",
    ):
        """HuggingFaceConnector enables loading a model configuration and weights from HuggingFace Hub.

        This class provides methods to download the model configuration and weights from a specified HuggingFace
        repository, or to load them from a local directory, if provided.

        Optionally, the local directory where pre-downloaded model files are stored can be set using the `local_dir`
        parameter.

        This class provides methods to load the model configuration and weights:
        - load_config() : Load the model configuration from a JSON file.
        - load_model_weights(module: PLForecastingModule) : Load the model weights into the given PyTorch module.
        - load_model(module_class: type[PLForecastingModule], pl_module_params: dict) :
          Load the model by creating an instance of the given module class and loading the weights.

        Parameters
        ----------
        repo_id
             The HuggingFace repository ID where the model is stored.
        repo_commit
             The commit ID of the model in the HuggingFace repository. Must be a specific commit hash to ensure
             non-changing model files.
        local_dir
             The local directory where the pre-downloaded model files are stored.
        config_file
             The name of the configuration file. Default is "config.json".
        model_file
             The name of the model weight file. Default is "model.safetensors".
        """
        if local_dir is not None:
            local_dir_path = Path(local_dir)
            if not local_dir_path.exists():
                raise_log(ValueError(f"`local_dir` directory `{local_dir}` does not exist."), logger)
            if not local_dir_path.is_dir():
                raise_log(ValueError(f"`local_dir` path `{local_dir}` is not a directory."), logger)
            local_dir = local_dir_path

        self.repo_id = repo_id
        self.repo_commit = repo_commit
        self.local_dir = local_dir
        self.config_file = config_file
        self.model_file = model_file

    def load_config(self) -> dict:
        """Load the model configuration from a JSON file.

        Returns
        -------
        dict
            The model configuration.
        """
        config_path = self._get_file_path(self.config_file)
        with open(config_path) as f:
            config = json.load(f)
        return config

    def load_model_weights(
        self,
        module: PLForecastingModule,
    ) -> None:
        """Load the model weights from a safetensors file.

        Parameters
        ----------
        module
            The PyTorch module to load the weights into.
        """
        module_path = self._get_file_path(self.model_file)
        state_dict = load_file(module_path)
        module.load_state_dict(state_dict)

    def load_model(
        self,
        module_class: type[PLForecastingModule],
        pl_module_params: dict,
        additional_params: Optional[dict] = None,
    ) -> PLForecastingModule:
        """Load the model by creating an instance of the given module class and loading
        the weights. Some configuration files might contain external parameters that
        are not part of the module class constructor like `architectures`. They are filtered
        out before instantiating the module.

        Parameters
        ----------
        module_class
            The class of the PyTorch Lightning module to instantiate.
        pl_module_params
            The parameters of the PyTorch Lightning module to instantiate.
        additional_params
            Additional parameters to pass to the `module_class` constructor when instantiating.

        Returns
        -------
        PLForecastingModule
            The loaded PyTorch Lightning module.
        """
        additional_params = additional_params or {}
        config = self.load_config()
        module_params = self._extract_module_params(module_class, config)
        module = module_class(
            **module_params,
            **pl_module_params,
            **additional_params,
        )
        self.load_model_weights(module)
        return module

    def _get_file_path(
        self,
        filename: str,
    ) -> os.PathLike:
        """Get the path to a file either from a local directory or by downloading it from HuggingFace.

        Parameters
        ----------
        filename
            The name of the file to retrieve.

        Returns
        -------
        os.PathLike
            The path to the requested file.
        """
        if self.local_dir is not None:
            path = Path(self.local_dir) / filename
            if not path.exists():
                raise FileNotFoundError(
                    f"File {filename} not found in {self.local_dir}"
                )
            if not path.is_file():
                raise ValueError(f"Path {path} is not a file")
            return path
        else:
            file_path = hf_hub_download(
                repo_id=self.repo_id,
                filename=filename,
                revision=self.repo_commit,
            )
            return Path(file_path)

    @staticmethod
    def _extract_module_params(
        module_class: type[PLForecastingModule],
        config: dict,
    ):
        """Extract params from `config` to set up the given `module_class`."""
        get_params = list(inspect.signature(module_class.__init__).parameters.keys())
        get_params.remove("self")
        return {kwarg: config.get(kwarg) for kwarg in get_params if kwarg in config}
```

</details>

Now in the actual `Chronos2Model` we can remove the attributes `_repo_id`, `_repo_commit`, setting `self.local_dir`, ...
Instead, we instantiate the connector. See the updated code below (I removed all the docs to make for visibility)

<details>

<summary>Chronos2Model</summary>

```py
class Chronos2Model(FoundationModel):
    # Fine-tuning is turned off for now pending proper fine-tuning support
    # and configuration.
    _allows_finetuning = False

    def __init__(
        self,
        input_chunk_length: int,
        output_chunk_length: int,
        output_chunk_shift: int = 0,
        likelihood: Optional[QuantileRegression] = None,
        repo_id = "amazon/chronos-2",
        repo_commit = "18128c7b4f3fd286f06d6d4efe1d252f1d2a9a7c",
        local_dir: Optional[Union[str, os.PathLike]] = None,
        **kwargs,
    ):
        hf_connector = HuggingFaceConnector(
            repo_id=repo_id,
            repo_commit=repo_commit,
            local_dir=local_dir,
        )

        # load model config for validation
        config = hf_connector.load_config()
        chronos_config = config["chronos_config"]

        # validate `input_chunk_length` against model's context_length
        context_length = chronos_config["context_length"]
        if input_chunk_length > context_length:
            raise_log(
                ValueError(
                    f"`input_chunk_length` {input_chunk_length} cannot be greater than "
                    f"model's context_length {context_length}"
                ),
                logger,
            )

        # validate `output_chunk_length` and `output_chunk_shift` against model's prediction length
        prediction_length = (
            chronos_config["output_patch_size"] * chronos_config["max_output_patches"]
        )
        if output_chunk_length + output_chunk_shift > prediction_length:
            raise_log(
                ValueError(
                    f"`output_chunk_length` {output_chunk_length} plus `output_chunk_shift` {output_chunk_shift} "
                    f"cannot be greater than model's maximum prediction length {prediction_length}"
                ),
                logger,
            )

        quantiles = chronos_config["quantiles"]
        # by default (`likelihood=None`), model is deterministic
        # otherwise, only QuantileRegression likelihood is supported and quantiles must be
        # a subset of Chronos-2 quantiles
        if likelihood is not None:
            if not isinstance(likelihood, QuantileRegression):
                raise_log(
                    ValueError(
                        f"Only QuantileRegression likelihood is supported for Chronos2Model in Darts. "
                        f"Got {type(likelihood)}."
                    ),
                    logger,
                )
            user_quantiles: list[float] = likelihood.quantiles
            if not set(user_quantiles).issubset(quantiles):
                raise_log(
                    ValueError(
                        f"The quantiles for QuantileRegression likelihood {user_quantiles} "
                        f"must be a subset of Chronos-2 quantiles {quantiles}."
                    ),
                    logger,
                )

        self.hf_connector = hf_connector
        super().__init__(enable_finetuning=False, **kwargs)

    def _create_model(self, train_sample: TorchTrainingSample) -> PLForecastingModule:
        pl_module_params = self.pl_module_params or {}
        return self.hf_connector.load_model(
            module_class=_Chronos2Module,
            pl_module_params=pl_module_params,
        )
```

</details>

darts/utils/timeseries_generation.py
```diff
@@ -43,7 +43,7 @@ def constant_timeseries(
     length: Optional[int] = None,
     freq: Union[str, int] = None,
     column_name: Optional[str] = "constant",
-    dtype: np.dtype = np.float64,
+    dtype: type[np.float32 | np.float64] = np.float64,
```
dennisbader: Can we exclude all these changes that are not related to this PR? :)
(Also the one in timeseries.py)

We can gladly open a new pr that fixes type hint issues for dtypes

darts/models/forecasting/chronos2_model.py
dennisbader: @daidahao, I still think that it's not working fully as intended.

I would assume:
- When I set `local_dir` which is currently empty, I expect the model to be downloaded to that directory
- When I create the model a second time with `local_dir`, it should use the already downloaded local files

Currently it raises an error when `local_dir` is empty.

darts/models/forecasting/chronos2_model.py
```diff
@@ -0,0 +1,865 @@
...
+    # and configuration.
+    _allows_finetuning = False
+
+    def __init__(
+        self,
+        input_chunk_length: int,
+        output_chunk_length: int,
+        output_chunk_shift: int = 0,
+        likelihood: Optional[QuantileRegression] = None,
+        local_dir: Optional[Union[str, os.PathLike]] = None,
```
dennisbader, line 571: I'm a bit concerned about the new parameter names (`local_dir`, `repo_id`, `repo_commit`)

- local_dir: as a bit ambiguous with the existing TFM `work_dir`, can we make it more clear that this is about loading the foundation models?
- `repo_id` simply from the name it is not obvious to a new user what this could relate to (e.g. hub_model_id or something? really just a suggestion, I hope we can find something better)
- `repo_commit`: this also differs from the hf parameter `revision` (maybe `hub_model_version` or `hub_model_hash`, `hub_model_revision`, ...)

darts/models/forecasting/chronos2_model.py
daidahao: Yes, that would include repo commit.

daidahao: In terms of "tiny" models, I have added a tiny dummy Chronos-2 for unit testing, see "Dummy Tests" in `test_chronos2.py`.

darts/models/forecasting/chronos2_model.py
daidahao: I've renamed `repo_id` and `repo_commit` to `hub_model_name` and `hub_model_revision`. I've also added more descriptions to `local_dir` to differentiate from `work_dir`.

Hope this would make things more clear.

darts/models/forecasting/foundation_model.py
daidahao: Thank you for the code! I've placed connector in `darts/models/components/huggingface_connector.py`.

darts/models/forecasting/foundation_model.py
daidahao: Thank you for the suggestions here. Following @dennisbader advice, we are now replacing `HuggingFaceModelMixin` with a simplified `HuggingFaceConnector`.

requirements/torch.txt
dennisbader: Alright, could you then at least move `Chronos2Model` import in `darts/models/__init__.py` into a dedicated try / except ModuleNotFoundError? No need to warn there.

```py
try:
    from darts.models.forecasting.chronos2_model import Chronos2Model
except ModuleNotFoundError:
    Chronos2Model = NotImportedModule(module_name="(Py)Torch", warn=False)
```

Like this, users could still install `u8darts` (core) + torch + lightning and use all torch models except the foundation models.

darts/tests/models/forecasting/test_chronos2.py
dennisbader, line 1: kudos, very nice tests :) Could you add a last one that checks whether the multiple multivariate series (list of multivariate series) forecasting works (can be deterministic)?

darts/models/forecasting/foundation_model.py
daidahao: It allows an optional `model_revision`, which will be set to `None` for Chronos-2. But for other foundation models, I would strongly recommend setting `model_revision` to specific commit ID without assurances from model authors.

darts/models/forecasting/chronos2_model.py
daidahao: Aha previously, I assumed that `local_dir` should contain the model files already. Now I've changed to logic to download model files into `local_dir` if empty.

darts/tests/models/forecasting/artefacts/chronos2/tiny_chronos2/model.safetensors
dennisbader, line 1: can we move these files into an `artefacts` folder and rename them to make it more clear

Renaming:
- `dummy/chronos2` -> `tiny_chronos/chronos2`
- `fidelty` -> `chronos_prediction`


Then move the files into
- tests/models/forecasting
   - artefacts (new folder for any future test files)
     - chronos
       - tiny_chronos/chronos2
       - chronos_prediction

darts/models/forecasting/chronos2_model.py
abdulfatir: @dennisbader @daidahao 

I will say this once again 😬 : 

The HFMixin is kind of an industry standard and they have handled a bunch of these cases. People know what kind of arguments to provide to load a model from HF or local. If darts is anyway taking a dependency on HF-hub, it makes little sense to reinvent what HF already provides stably.

darts/tests/models/forecasting/test_chronos2.py
daidahao: Fixed in 60f94e5ff45bb96ff75fd575714d4eb91b21f24a

darts/tests/models/forecasting/artefacts/chronos2/tiny_chronos2/model.safetensors
daidahao: Fixed in 6c4af791bb20b020a25854b40f6b231b0fb96de2

darts/models/forecasting/chronos2_model.py
daidahao: Okay, I tried to implement `_Chronos2Module` as a subclass of both `PLForecastingModule` and `PyTorchModelHubMixin`. Unfortunately, that did not work because `PyTorchModelHubMixin.from_pretrained()` will pass parameters like `architectures`, `transformers_version` to `PLForecastingModule` which the latter does not accept.

Another problem with `.from_pretrained()` is that it does not accept `local_dir` but only `cache_dir`, which will use a local directory for caching and would have a different directory structure. 

Finally, using `from_pretrained()` means that we can only do parameter validation inside `_create_model()`, i.e., when the model called `model.fit()` rather than in initialization.

I wish there was a easy way to integrate `PyTorchModelHubMixin` but it seems incompatible with Darts APIs at the moment.

daidahao:
@dennisbader Alongside the CHANGELOG, I might need your help on the documentation. Should we add a section on Foundation model in Quickstart and/or User Guide? Also, should we add an example notebook for using Chronos-2 in Darts in Examples? Where do you think we may start?

dennisbader:
> @dennisbader Alongside the CHANGELOG, I might need your help on the documentation. Should we add a section on Foundation model in Quickstart and/or User Guide? Also, should we add an example notebook for using Chronos-2 in Darts in Examples? Where do you think we may start?

@daidahao A small example notebook should be enough for now. Make sure to keep it concise, and keep computation time low (max a minute to compute the notebook). You don't need to cover every aspect such as backtesting, historical forecasts and so on. Instead just reference to another resource and mention that the same features are supported as for regular TorchForecastingModels.

Some things that would be nice to explain:
- How the user can control the lookback & output windows with input/output chunk length & output chunk shift parameters
- The new hub related parameters & loading
- no fine-tuning supported for now

daidahao:
@dennisbader I've created a new example notebook for Chronos-2 and updated the CHANGELOG. Let me know what you think and what might still be missing.

I feel that we are very close to the finish line 🏁🚀

abdulfatir:
@daidahao notebook looks great!

A small comment: the example only shows past covariates, and for this specific split, using past-only covariates actually worsens the results slightly. This will be very confusing for the the reader. I would recommend showing the known-future covariates scenario in the notebook instead. That is generally the more popular setting. Of course, you should put the caveat about what things could be known in the future or could be reasonably predicted (weather) to be supplied as future covariates.

daidahao:
@abdulfatir Thank you for the suggestions. I noticed the worse results too when creating the notebook. Unfortunately, there are no future covariates available from this particular dataset as the weather measurements came from a weather station and not a forecast. Creating future covariates like time of day is possible and but may not be as interesting as using covariates from the data source.

For the sake of demonstration, I will update the notebook to use weather measurements as future covariates instead. But I will also put up a caveat about one should use weather forecasts in practice. What do you think?

abdulfatir:
> For the sake of demonstration, I will update the notebook to use weather measurements as future covariates instead. But I will also put up a caveat about one should use weather forecasts in practice. What do you think?

Yes, this makes sense and is also a common approach in practice.

darts/models/components/huggingface_connector.py
```diff
@@ -0,0 +1,193 @@
...
+        if self.local_dir is not None:
+            path = Path(self.local_dir) / filename
+            if not path.exists():
+                logger.warning(
+                    f"File {filename} not found in `local_dir` {self.local_dir}. "
+                    f"Attempting to download from HuggingFace Hub instead and "
+                    f"save it to the `local_dir`."
+                )
+            elif not path.is_file():
+                raise ValueError(f"Path {path} is not a file")
```
dennisbader: use `raise_log` to raise the exception

darts/models/forecasting/foundation_model.py
```diff
@@ -0,0 +1,209 @@
...
+                    f"Fine-tuning is not supported for {self.__class__.__name__}."
+                    " Please set `enable_finetuning=False`."
+                ),
+                logger,
+            )
+
+        self._enable_finetuning = enable_finetuning
+
+    @classmethod
+    def _validate_model_params(cls, **kwargs):
```
dennisbader: Once #2908 is merged, we'll be able to remove this 😍 (no need to do it in this PR though :) )

darts/tests/models/forecasting/test_chronos2.py
```diff
@@ -0,0 +1,460 @@
...
+from darts.datasets import ElectricityConsumptionZurichDataset
+from darts.tests.conftest import TORCH_AVAILABLE, tfm_kwargs
+from darts.utils.likelihood_models import GaussianLikelihood
+from darts.utils.timeseries_generation import (
+    gaussian_timeseries,
+    linear_timeseries,
+    sine_timeseries,
+)
+
+if not TORCH_AVAILABLE:
```
dennisbader, line 10: we need to skip these tests before importing any torch-related files (e.g. GaussianLikelihood). These no-torch flavors will only be tested once the PR was merged.

The easiest is to just skip before importing any darts modules.

darts/tests/models/forecasting/test_chronos2.py
```diff
@@ -0,0 +1,460 @@
...
+    pytest.skip(
+        f"Torch not available. {__name__} tests will be skipped.",
+        allow_module_level=True,
+    )
+
+from darts.models import Chronos2Model
+from darts.utils.likelihood_models import QuantileRegression
+
```
dennisbader, line 26: Since we're using randomness in this test (gaussian, and sampling) let's fix it here. Randomness can sometimes cross-contaminate tests from other modules.

```suggestion

np.random.seed(42)

```

CHANGELOG.md
```diff
@@ -11,15 +11,21 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
+- 🚀🚀 Added **Chronos-2**, the **first foundation model** for time series forecasting. `Chronos2Model` can be used for univariate, multivariate, and multiple time series forecasting without any training. It supports past and future covariates and can output determinstic or probabilistic forecasts. Check out the new [Chronos-2 Foundation Model Notebook](https://unit8co.github.io/darts/examples/24-SKLearnClassifierModel-examples.html) for examples and usage of Chronos-2. [#2944](https://github.com/unit8co/darts/pull/2944) by [Zhihao Dai](https://github.com/daidahao). Many thanks to [Dennis Bader](https://github.com/dennisbader) and [Abdul Fatir](https://github.com/abdulfatir) for their reviews.
+- Added an example notebook demonstrating how to use Chronos-2: [Chronos-2 Foundation Model Examples](https://unit8co.github.io/darts/examples/25-Chronos-2-examples.html). [#2944](https://github.com/unit8co/darts/pull/2944) by [Zhihao Dai](https://github.com/daidahao).
```
dennisbader: 🥳 🚀 

```suggestion
- 🚀🚀 Added new forecasting model `Chronos2Model`: Amazon's pre-trained 120M-parameter foundational model for zero-shot forecasting. `Chronos2Model` can be used for univariate, multivariate, and multiple time series forecasting without any training. It supports past and future covariates and can output deterministic or probabilistic forecasts. Check out the new [Chronos-2 Foundation Model Notebook](https://unit8co.github.io/darts/examples/25-Chronos-2-examples.html) for examples and usage of Chronos-2. [#2944](https://github.com/unit8co/darts/pull/2944) by [Zhihao Dai](https://github.com/daidahao).
```

examples/25-Chronos-2-examples.ipynb
dennisbader, line 1: Let's run this notebook as part of our PR merge checks. You can add the notebook name to the end of the list here:

https://github.com/unit8co/darts/blob/2c666f882bc4ae3612a5b06d3a9b1cdf505cc924/.github/workflows/merge.yml#L88

dennisbader (changes requested):
Really nice updates, thanks a lot @daidahao
The example notebook looks great as well!

I added some last few minor suggestions, after that everything should be good to go 🚀

examples/25-Chronos-2-examples.ipynb
```diff
@@ -0,0 +1,572 @@
+{
```
dennisbader, line 1: <p>Really great example, kudos again <span class="mention" data-value="daidahao" data-denotation-char="@">﻿<span contenteditable="false">@daidahao</span>﻿</span> :)</p><p> </p><p>I only added some minor suggestions, we're almost there!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/2944/discussion/#comment-2531865931'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='0'/>

dennisbader, line 1: <pre class="ql-syntax" spellcheck="false">Line #15.    plt.style.use("ggplot")
</pre><p>Let's use the default style to comply with our other example notebooks. This will mean, that the order of plotting the series will have to be adjusted a bit.</p><p> </p><p>Usually, we plot the actual series first (black color), and then the predictions (blues, purples, etc.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/2944/discussion/#comment-2531865938'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

dennisbader, line 1: <p>I would prefer if we showcased the example with <code>predict_likelihood_parameters=True</code> . Sampling with the "large" models can be quite computationally expensive, and only has benefits if auto-regression is used.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/2944/discussion/#comment-2531865942'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='23'/>

dennisbader, line 1: <p>The interval coverage  <code>mic</code>  (e.g. the share of actuals that are inside the prediction interval, we expect 80% to be inside) and interval width <code>miw</code> (the height / width of the interval) could be nice to show here :) I find the quantile loss not always very intuitive.</p><p> </p>

<pre class="ql-syntax" spellcheck="false">from darts.metrics import mic, miw

mic_prob = mic(val_energy, pred_cov, q_interval=(0.1, 0.9))
miw_prob = miw(val_energy, pred_cov, q_interval=(0.1, 0.9))
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/2944/discussion/#comment-2531865948'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='26'/>

darts/models/forecasting/foundation_model.py
dennisbader: We just merged #2908, so we can actually remove this methods  🥳

examples/25-Chronos-2-examples.ipynb
daidahao: Great advice. I've applied a few changes to the notebook, mainly:
- Replace `mql` with `mic` and `miw`.
- Use the default style and order (actual first, forecast next).
- Call `predict()` with `predict_likelihood_parameters=True` instead of sampling.

daidahao:
@dennisbader Can I suggest ignoring files like `CHANGELOG.md` from triggering workflows. No need to waste money on my silly typos😊

```yaml
name: darts PR workflow

on:
  pull_request:
    branches:
      - master
    paths-ignore:
      - 'CHANGELOG.md'

```

dennisbader:
> @dennisbader Can I suggest ignoring files like `CHANGELOG.md` from triggering workflows. No need to waste money on my silly typos😊
> 
> ```yaml
> name: darts PR workflow
> 
> on:
>   pull_request:
>     branches:
>       - master
>     paths-ignore:
>       - 'CHANGELOG.md'
> ```

Yes that sounds like a good idea :) Could we do this in a separate PR?

dennisbader:
Beautiful 😍 Thanks again for this great work, and also impressive how quickly you could set this up! 🚀 

This will allow us to add new foundation models efficiently! Really nice to see this, kudos 🔥

dennisbader (approved):

abdulfatir:
Indeed, great work @daidahao!

daidahao:
@abdulfatir @dennisbader Thank you both for your code reviews. Learnt a lot from this experience 😊

dennisbader:
Yes, thank you too @abdulfatir for the helpful reviews!

PR #2947, opened 2025-11-14
add os.Path support for load/save torch models

turbotimon:

### Summary

Currently, the save/load methdos for TorchForecastingModel do not support [os.PathLike](https://docs.python.org/3/library/os.html#os.PathLike) object for the path argument. Other model-families (conformal, ensemble, forecasting) already do so. This PR changes that, improving the usability as well as make the interface more similar.

No additional test are needed. As the path argument is transformed to a pathlib.Path object at the beginning it is [guaranteed to work with str as well as any os.PathLike objects](https://docs.python.org/3/library/pathlib.html#pure-paths)

The other Model-family also support file-like objects (IObuffer). But since [pl.Trainer.save_checkpoint](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.trainer.trainer.Trainer.html#lightning.pytorch.trainer.trainer.Trainer.save_checkpoint) does not supports it, i didn't add this support as well. Torch.save would support it, however.

### DIFF
--- a/darts/models/forecasting/torch_forecasting_model.py
+++ b/darts/models/forecasting/torch_forecasting_model.py
@@ -26,6 +26,7 @@ import sys
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from glob import glob
+from pathlib import Path
 from typing import Any, Callable, Literal, Optional, Union
 
 if sys.version_info >= (3, 11):
@@ -1933,7 +1934,7 @@ class TorchForecastingModel(GlobalForecastingModel, ABC):
 
     def save(
         self,
-        path: Optional[str] = None,
+        path: Optional[Union[str, os.PathLike]] = None,
         clean: bool = False,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -2000,7 +2004,9 @@ def save(
 
     @staticmethod
     def load(
-        path: str, pl_trainer_kwargs: Optional[dict] = None, **kwargs
+        path: Optional[Union[str, os.PathLike]] = None,
```
dennisbader: this should not be optional

```suggestion
        path: Union[str, os.PathLike],
```

darts/models/forecasting/torch_forecasting_model.py
dennisbader, line 1: Would be great to add this to all other save / load methods:

- model construction param: `work_dir`
- `to_onnx()`
- `load_weights()`
- `work_dir` in `load_from_checkpoint()`
- `work_dir` in `load_weights_from_checkpoint()`

dennisbader (changes requested):
Thanks a lot for this PR @turbotimon 🚀 

Adding support for Path-like objects sounds great. In this case there are some more places where we should allow this, to provide a unified usage over all methods.

darts/models/forecasting/torch_forecasting_model.py
turbotimon: yes that makes sense. Will try to do so as soon as i find time

daidahao:
@turbotimon

Would you like to continue the work on this? I think both supporting `os.PathLike` and file-like object in `save()`/`load()` would be useful for use cases, like `pathlib` or in-memory operations.

While `TorchForecastingModel` stores two files (`pt` and `pt.ckpt`), we could still support file-like objects by bundling both in a compressed format. I am thinking using [tarfile](https://docs.python.org/3/library/tarfile.html) or [zipfile](https://docs.python.org/3/library/zipfile.html) as solutions. What do you think?

Thanks to @dennisbader for leading me here.

turbotimon:
@daidahao thanks for asking and sorry for the late reply

> Would you like to continue the work on this?

Actually, yes, but I'm pretty busy right now… So I don't mind if someone else takes care of it; otherwise, I'll take another look at it as soon as I have some free time.

> While TorchForecastingModel stores two files (pt and pt.ckpt), we could still support file-like objects by bundling both in a compressed format. I am thinking using [tarfile](https://docs.python.org/3/library/tarfile.html) or [zipfile](https://docs.python.org/3/library/zipfile.html) as solutions. What do you think?

I do not think that is related to "file-like object", it's just a representative of a path, not a real file. And .ckpt gets added by convention. Bundling can, however, be advantageous for other reasons, but I don't have a strong opinion on that.

PR #2951, opened 2025-11-16
Workflows: ignore certain files

daidahao:

Fixes #.

### Summary
Ignore certain files like `CHANGELOG.md` and `README.md` in PR workflows. There is no need to run workflows on changes like typos or formatting. Feel free to suggest any files that can be ignored.


dennisbader:
Hi @daidahao, I just checked whether updating the Changelog will not re-trigger the checks, but it seems they were triggered. So this change is not effective yet 🤔

daidahao:
@dennisbader 

It turned out that `paths-ignore` rules apply to git diff between the last commit in the PR branch and the last sync commit in the master branch, see https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#git-diff-comparisons

What that means is that the workflow would still be triggered as long as some non-ignored files are changed in the PR, like `develop.yml` here☹️. But if the PR only changes the ignored files, the workflow would not be triggered. 

From GitHub documentation, I could not find options that can stop triggering the workflows between commits in the same PR.

dennisbader:
@daidahao, I'll close this PR then. Feel free to re-open if the there is another.

PR #2980, opened 2025-12-26
[New Model] TimesFM-2.5 Foundation Model

### LINKED ISSUE #2976: [New Model] TimesFM-2.5 Foundation Model
A sub issue of #2933.

In September, Google released [TimesFM 2.5](https://github.com/google-research/timesfm), a 200M-parameter pre-trained time series model, on GitHub and [HuggingFace](https://huggingface.co/collections/google/timesfm-release). 

The model itself supports univariate series only. Google has recently added covariate support by using [xreg](https://github.com/google-research/timesfm/blob/master/src/timesfm/utils/xreg_lib.py) (a wrapper for ridge regression at inference time).

TimesFM2.5 currently ranks second on GIFT-EVAL among single open-source models, after Chronos-2 (implemented in #2943 #2961). If implementing a TimesFM 2.5 is of interest of Darts, I am happy to contribute a PR.

Porting TimesFM-2.5 to `FoundationModel` should be easy and requires no additional dependencies than those already included (`huggingface_hub` for downloading weights, `safetensors` for loading weights, and `torch`). However, supporting covariates could be a challenge:

In Google's repo, there are two modes for adding covariate support, aka, `xreg_mode`:
- `"xreg + timesfm"` (default): fits a model on the residuals of the TimesFM forecast
- `"timesfm + xreg"`: fits a model on the targets then forecasts on the residuals via TimesFM

Either mode involves solving ridge regression on targets and covariates **at inference time** (see `forecast_with_covariates()` in `timesfm_2p5_base.py`). We might not be able to do so in Darts because this might cause data leakage in the same batch (e.g., when calling `historical_forecasts()`).

Instead, I suggest we use a linear head and a MSE loss with L2 regularization to train that head and mimic ridge regression.

Another potential challenge is of course fine-tuning #2964. Since fine-tuning would require a deterministic loss or a Quantile Regression, this may not be compatible with the MSE loss. I haven't thought a way to address the conflict. Perhaps we get rid of the MSE loss when fine-tuning is enabled but this would make the model behave differently than not fine-tuning.

**Additional context**
Overall, a few discussions points I can think of:
- Should we support covariates for TimesFM-2.5?
- If so, should we add a linear head and a MSE loss with L2 to train the head? Are there better ways of doing this without risking data leakage?
- What modes would the model support (`"xreg + timesfm"`, `"timesfm + xreg"`, or both) and which is the default?
- Should the model support fine-tuning? If so, would it be compatible with the MSE loss? 

I've tried to reach out to Google about collaborations but they haven't responded. See https://github.com/google-research/timesfm/issues/348. I will invite TimesFM-2.5 co-authors for discussions here.

daidahao:

Fixes #2976.

### Summary
This PR adds `TimesFM2p5Model` to Darts forecasting models. TimesFM 2.5 is Google's latest time series pre-trained foundation model from the TimesFM family. It supports univariate, multivariate, and multiple time series forecasting without training and can output deterministic or probabilistic forecasts.

<details>
<summary>Quickstart</summary>

```python
import numpy as np

from darts import set_option
from darts.metrics import mae
from darts.models import TimesFM2p5Model

# use darts plotting style
set_option("plotting.use_darts_style", True)

# convert to float32 as TimesFM 2.5 works with float32 input
data = ElectricityConsumptionZurichDataset().load().astype(np.float32)
ts_energy = data["Value_NE5"]
# split into train and validation sets by last 7 days
train_energy, val_energy = ts_energy.split_before(len(ts_energy) - 7 * 24 * 4)

# use last 30 days of data to predict next 1 day
model = TimesFM2p5Model(
    output_chunk_length=1 * 24 * 4,
)

# fit model (no training is performed)
model.fit(
    series=train_energy,
    verbose=True,
)

# predict next 7 days
pred = model.predict(
    n=7 * 24 * 4,
    series=train_energy,
)
val_energy.plot(label="actual")
pred.plot(label="forecast")

# compute MAE on validation set
mae_val = mae(val_energy, pred)
print(f"MAE on validation set: {mae_val:.2f}")
# MAE on validation set: 603.00
```

image

</details>

<details open>
<summary>TODO</summary>

- [x] Update CHANGELOG.
- [x] Update README model catalogue and `index.rst`.
- [x] Update `covariates.md`.
- [x] Complete unit tests for `TimesFM2p5Model `.

</details>


daidahao:
@dennisbader 

Let me know what you think of the current implementation. I've restricted `TimesFM2p5Model` to be a univariate model (though covariate support is possible with an external regressor like you suggested) and disabled fine-tuning for now.

The univariate nature of TimesFM 2.5 raises an interesting question here: should `FoundationModel` inherit from `TorchForecastingModel` rather than `MixedCovariatesTorchModel`? I understand that we want to prioritise models with covariate support. However, I imagine some future models could be supporting future covariates only and thus not satisfying "MixedCovariates". What do you think?

darts/models/__init__.py
```diff
@@ -89,8 +89,10 @@
 
 try:
     from darts.models.forecasting.chronos2_model import Chronos2Model
+    from darts.models.forecasting.timesfm2p5_model import TimesFM2p5Model
 except ModuleNotFoundError:
     Chronos2Model = NotImportedModule(module_name="(Py)Torch", warn=False)
+    TimesFM2p5Model = NotImportedModule(module_name="(Py)Torch", warn=False)
```
dennisbader, line 95: Add the model to `__all__` at the bottom.

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
+"""
+TimesFM 2.5
+-----------
+
+For detailed examples and tutorials, see:
+
+* `TimesFM 2.5 Model Examples
+  <https://unit8co.github.io/darts/examples/xxx.html>`__
```
dennisbader: I don't think we need a dedicated notebook for TimesFM but rather direct the user to the chronos example and mention that TimesFM only supports univariate forecasts without covariates

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+            base class.
+        """
+
+        super().__init__(**kwargs)
+
+        # default model parameters (config.json is ignored)
+        self.input_patch_len = self.config.input_patch_len  # 32
+        self.output_patch_len = self.config.output_patch_len  # 128
+        self.num_layers = self.config.stacked_transformers.num_layers  # 20
+        self.num_quantiles_plus_one = len(self.config.quantiles) + 1  # 10
```
dennisbader, line 117: would be nice to explain here why + 1

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+        input_embeddings = self.tokenizer(tokenizer_inputs)
+
+        output_embeddings = input_embeddings
+        for _, layer in enumerate(self.stacked_xf):
+            output_embeddings = layer(
+                output_embeddings,
+                masks[..., -1],
+            )
+        output_ts = self.output_projection_point(output_embeddings)
+        # output_quantile_spread = self.output_projection_quantiles(output_embeddings)
```
dennisbader: is this skipped intentionally? The residual block for it is defined in `__init__` but never used

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+        hf_connector = HuggingFaceConnector(
+            model_name=hub_model_name,
+            model_revision=hub_model_revision,
+            local_dir=local_dir,
+        )
+
+        # As per the original implementation, the model config is ignored and default
+        # parameters are used instead.
+        config = _TimesFM2p5_200M_Definition()
```
dennisbader, line 568: Wouldn't that mean that we will never be able to support other flavors of this model? E.g. no "small" / "large" models / checkpoints with different param config?

I don't really like the fact that _TimesFM2p5 module object is initialized without the model parameters, and instead a constant config is loaded. What is your opinion on that?

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+            (n, mu, sigma), _ = _update_running_stats(
+                n, mu, sigma, patched_x_past[:, i], patched_mask[:, i]
+            )
+            patch_mu.append(mu)
+            patch_sigma.append(sigma)
+        context_mu = torch.stack(patch_mu, dim=1)
+        context_sigma = torch.stack(patch_sigma, dim=1)
+
+        # normalize inputs and apply mask
+        normed_inputs = _revin(patched_x_past, context_mu, context_sigma, reverse=False)
```
dennisbader, line 276: It looks like, next to the running statistics, they also allow revin on the raw input (e.g. like our use_reversible_instance_norm) see [here](https://github.com/google-research/timesfm/blob/6bd8044275f8b76cdc9554f2fecccac5f31a156c/src/timesfm/timesfm_2p5/timesfm_2p5_torch.py#L401). We could activate it here as well through the `@ioprocessor` decorator on `forward()`.

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+        torch.Tensor
+            the output tensor in the shape of `(n_samples, n_time_steps, n_targets, n_quantiles)` for
+            probabilistic forecasts, or `(n_samples, n_time_steps, n_targets, 1)` for
+            deterministic forecasts (median only).
+        """
+        x_past, _, _ = x_in
+
+        # TimesFM 2.5 is a univariate model and its inputs does not have a variable dimension,
+        # so here we reshape x_past to (n_samples * n_variables, n_time_steps)
+        x_past = x_past.permute(0, 2, 1).reshape(-1, self.input_chunk_length)
```
dennisbader, line 236: Something we have started recently (e.g. in BlockRNNModel, see [here](https://github.com/unit8co/darts/blob/892c16d9d3382249672ce2497661f37aab560816/darts/models/forecasting/block_rnn_model.py#L210)), is to comment the input / output shapes of each operation for interpretability. I find it very useful, when coming back to work on any model, to know how the tensor shapes change throughout the model.

Would you be open to add this information?

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+        # but the mean is not being used even in deterministic setting.
+        # Instead, the median (0.5 quantile) is used as the deterministic output.
+        self.user_quantile_indices = [
+            self.config.quantiles.index(q) + 1 for q in user_quantiles
+        ]
+
+    def _forward(
+        self,
+        inputs: torch.Tensor,
+        masks: torch.Tensor,
```
dennisbader, line 157: I see that you removed the `decode_chaches` input and logic, compared to the source implementation. Was this intentionally?

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+        patched_x_past = x_past.unfold(1, self.input_patch_len, self.input_patch_len)
+        patched_mask = mask.unfold(1, self.input_patch_len, self.input_patch_len)
+
+        # determine batch size and number of input patches after patching
+        batch_size, num_input_patches, _ = patched_x_past.shape
+
+        # running stats
+        n = torch.zeros(batch_size, device=patched_x_past.device)
+        mu = torch.zeros(batch_size, device=patched_x_past.device)
+        sigma = torch.zeros(batch_size, device=patched_x_past.device)
```
dennisbader: shouldn't we also specify the dtype for proper casting of the model?

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,595 @@
...
+        **kwargs,
+    ):
+        """TimesFM 2.5 Model for zero-shot forecasting.
+
+        This is an implementation of Google's TimesFM 2.5 model, ported from
+        `google-research/timesfm <https://github.com/google-research/timesfm>`_ with adaptations to use the Darts API.
+        It is an updated version of the original TimesFM model [1]_, [2]_, with a larger context length (16,384 vs 512)
+        and better predictive accuracy.
+
+        This model supports either univariate or multivariate time series, but does not support covariates.
```
dennisbader, line 329: maybe mention here that the model itself is not really a multivariate model, but we achieve it by forecasting the target components independently

darts/models/components/timesfm2p5_submodels.py
```diff
@@ -0,0 +1,492 @@
...
+    mask: torch.Tensor,
+) -> tuple[
+    tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+    tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+]:
+    """Updates the running stats."""
+    is_legit = torch.logical_not(mask).to(x.dtype)
+    inc_n = torch.sum(is_legit, dim=-1)
+
+    inc_mu_numerator = torch.nansum(x, dim=-1)
```
dennisbader, line 461: In the original implementation it multiplies by `is_legit`, was this removed intentionally?
```suggestion
    inc_mu_numerator = torch.nansum(x * is_legit, dim=-1)
```

darts/models/components/timesfm2p5_submodels.py
```diff
@@ -0,0 +1,492 @@
...
+    inc_n_safe = torch.where(inc_n == 0, 1.0, inc_n)
+    inc_mu = inc_mu_numerator / inc_n_safe
+    inc_mu = torch.where(inc_n == 0, 0.0, inc_mu)
+
+    inc_var_numerator = torch.nansum((x - inc_mu.unsqueeze(-1)) ** 2, dim=-1)
```
dennisbader, line 466: same here
```suggestion
    inc_var_numerator = torch.nansum(
        ((x - inc_mu.unsqueeze(-1)) ** 2) * is_legit, dim=-1
    )
```

darts/models/components/timesfm2p5_submodels.py
```diff
@@ -0,0 +1,492 @@
...
+    return output
+
+
+class _PerDimScale(nn.Module):
+    """Per-dimension scaling."""
+
+    def __init__(self, num_dims: int):
+        super().__init__()
+        self.num_dims = num_dims
+        self.per_dim_scale = nn.Parameter(torch.zeros(num_dims))
```
dennisbader, line 249: is there no need to give the model device and dtype to `zeros`?

darts/tests/models/forecasting/test_timesfm2p5.py
```diff
@@ -0,0 +1,165 @@
...
+if not TORCH_AVAILABLE:
+    pytest.skip(
+        f"Torch not available. {__name__} tests will be skipped.",
+        allow_module_level=True,
+    )
+
+from darts import TimeSeries
+from darts.datasets import ElectricityConsumptionZurichDataset
+from darts.models import TimesFM2p5Model
+from darts.tests.conftest import TORCH_AVAILABLE, tfm_kwargs
```
dennisbader: unused at this stage

```suggestion
from darts.tests.conftest import tfm_kwargs
```

darts/tests/models/forecasting/test_timesfm2p5.py
dennisbader, line 1: as you mentioned in the PR description, some more tests would be great. Especially:
- test that univariate case works (no fidelity required)
- test no covariates support

darts/models/components/timesfm2p5_submodels.py
daidahao: There is no need to specify device in `__init__()` as the device of any parameters are handled by Lightning. As for the dtype, the default floating point is float32 in [PyTorch](https://docs.pytorch.org/docs/stable/generated/torch.get_default_dtype.html)

darts/models/components/timesfm2p5_submodels.py
daidahao: Yes, I removed it because `mask` is essentially `torch.isnan(x)` so `torch.nansum(x * is_legit, dim=-1)` is equivalent to `torch.nansum(x, dim=-1)`.

darts/models/components/timesfm2p5_submodels.py
daidahao: Same as above, I simplified it because it is equivalent.

darts/models/forecasting/timesfm2p5_model.py
daidahao: It was not specified in the original impl. Also, the default dtype would be float32.

darts/models/forecasting/timesfm2p5_model.py
daidahao: Yes, we could. I would also caution that their revin is different to `RINorm` in `ioprocessor`, especially for `stdev` calculation. An exact match and validation would not be possible.

darts/models/forecasting/timesfm2p5_model.py
daidahao: I don't like that either. But the original impl. uses a constant config and does not support any other flavors. Seems like the source impl. as of now would not support a future flavor.

darts/models/forecasting/timesfm2p5_model.py
daidahao: Yes. There are two output heads in TimesFM, one for `output_patch_len` (128) horizon, the other for `output_quantile_len` (1,024) horizon. The latter is disabled by default (`use_continuous_quantile_head=False`) and used to "avoid quantile collapsing" when the forecast horizon is longer than 128. It is not needed here because Darts uses Monte Carlo sampling (`output_chunk_length` time points at a time) when horizon is longer than 128. Quantile collapsing should be mitigated when number of samples is reasonably large.

daidahao: <del>

Perhaps we could add `use_continuous_quantile_head` as a paramater to model definition. When enabled, the maximum `output_chunk_length + shift` would be 1,024 and the latter head would be used as in the original impl.:
```py
      if fc.use_continuous_quantile_head:
        for quantile_index in [1, 2, 3, 4, 6, 7, 8, 9]:
          full_forecast[:, :, quantile_index] = (
            quantile_spreads[:, : fc.max_horizon, quantile_index]
            - quantile_spreads[:, : fc.max_horizon, 5]
            + full_forecast[:, : fc.max_horizon, 5]
          )
```

</del>

Apologies, just realised the above would not be feasible because `full_forecast` (from the first head) can only predict 128 points at a time in Darts. Instead, we could do:

When `use_continuous_quantile_head=True`, maximum `output_chunk_length + shift` would be 1,024 and `full_forecast` would just be `quantile_spreads`.

darts/models/forecasting/timesfm2p5_model.py
dennisbader: I think that is fine. Thanks for adding io_processor

darts/models/components/timesfm2p5_submodels.py
dennisbader: 👍

darts/models/forecasting/timesfm2p5_model.py
dennisbader: @daidahao, just wanted to check in again about this comment (whether we do not need `decode_caches`)

darts/models/forecasting/timesfm2p5_model.py
dennisbader: Alright, I guess we can leave it for now 👍

darts/models/forecasting/timesfm2p5_model.py
dennisbader: Hmm I'm not sure I fully understand why they have two output heads. Is the first one with `output_patch_len` (128) somehow better than the other for smaller horizons? Or was it just that they added the second head with `output_patch_len` (1024) at a later stage and for backwards compatibility they made it optional?

If the second one is as good as the first one, then I would say we should use that one by default, or dynamically whenever `output_chunk_length + output_chunk_shift > 128`. 

What do you think?

darts/models/forecasting/timesfm2p5_model.py
daidahao: @dennisbader 

I've taken a second look at this. `decode_caches` only benefits autoregressive forecasting, which is handled by `PLForecastingModule._get_batch_prediction()`. The method then calls `.forward()` which in turn expects a tuple consisting of `(x_past, x_future, x_static)`. Passing `decode_caches` is not possible without changing the upstream `_get_batch_prediction()` logic.

darts/models/forecasting/timesfm2p5_model.py
dennisbader: I did some tests based on the Chronos2 example notebook to compare the performance of both output heads for shorter horizon (output_chunk_length <= 128). The predictions are almost identical, see below (head 1 is the small head, head 2 is the large head).

image

image

From this, I strongly favor only using the larger head `output_patch_len` (1024) in general, as it's also great to support longer horizons. For now we can deactivate the first head. We can always add support for it at a later stage if users really need the first head.

darts/models/forecasting/timesfm2p5_model.py
```diff
@@ -0,0 +1,627 @@
+"""
+TimesFM 2.5
+-----------
+
+For detailed examples and tutorials, see:
+
+* `TimesFM 2.5 Model Examples
+  <https://unit8co.github.io/darts/examples/xxx.html>`__
```
dennisbader:
```suggestion
TimesFM 2.5 can be used the same way as other foundation models (e.g. Chronos2), with the exception
that it does not support any type of covariates.

For detailed examples and tutorials, check out the Chronos2 notebook:

* `Chronos-2 Foundation Model Examples
  <https://unit8co.github.io/darts/examples/25-Chronos-2-examples.html>`__
```

daidahao:
> Hi @daidahao, I see you re-requested a review? It looks like we're still using the smaller output head. Once you have the suggestions applied from [here](https://github.com/unit8co/darts/pull/2980#discussion_r2746049175), I'll review again :)

Hi @dennisbader, as I've said in the comments, using the second head would fail the fidelity tests because the original impl. uses either 1) only the first head, or 2) the first head + the second head.

In that case, I would recommend using only the first head to keep the fidelity tests. But if you prefer, I could remove the fidelity tests and use only the second head.

Let me know how we may proceed with the second head. :)

dennisbader:
@daidahao ahh I see. So it basically performs:

- with `use_continuous_quantile_head=False`:
  - Use only the smaller head for all quantiles. 
  - If `horizon > 128`: perform auto-regression with a 128 output steps until it reaches `horizon`
- with `use_continuous_quantile_head=True`:
  - Perform the same as with `use_continuous_quantile_head=False` to get the median forecasts
  - Add quantiles from the larger head (e.g. the new `i`th quantile forecast `q_i_final` is defined as `q_i_final = q_i_lh - q_0.5_lh + q0.5_sh`), where:
    - `q_i_lh` is the `i`th quantile forecast of the large head
    - `q_0.5_lh` is the median forecast of the large head
    - `q_0.5_sh` is the median forecast of the small head

I wonder why they don't just use the larger head for this 🤔 See also the discussion [here](https://github.com/google-research/timesfm/issues/306) about the crossing quantiles. To me it sounds like the larger head is more stable.

From my perspective, I would just use the larger head in Darts. It simplifies the code and allows users to use define larger `output_chunk_length`. You could add a note in the model docstring that states that we only use that output.

For the fidelity tests, you could check that the normalized deviation is not larger than some expected value. Similar to [the example](https://github.com/unit8co/darts/pull/2980#discussion_r2746049175) above.

How does that sound?

daidahao:
@dennisbader 

I just pushed several commits to:
1. Use larger head only for longer horizon (1,024) in Darts.
2. Replace fidelity test "ground truths" to set `use_continuous_quantile_head=True` and `horizon=1024` when using the original impl.
3. Test for `normalized_deviation` calculated as `abs(pred - original)/original_mean` in Darts and check that all deviations are smaller than 1e-3.

Unfortunately, the new fidelity tests would fail as the deviation could be as high as 26% when the original impl. is using BOTH heads and a different sampling strategy for `horizon=1024`.

daidahao:
@dennisbader 

My advice is to use ONLY the small head for provability and simplicity until we find a better solution.

dennisbader (approved):
Looks great, thanks a lot for another great PR @daidahao 🚀 
As you say, regarding the larger head, we can also add this in the future.

daidahao:
@dennisbader 

Thank you for the reviews! Cheers! 🤗

PR #2984, opened 2026-01-07
Fix/smape

### LINKED ISSUE #2983: [BUG] sMAPE raises misleading (arguably spurious) error message when denominator is zero
**Describe the bug**
Calling smape where the actuals *and* predictions are zero at a point raises an error (incorrectly) claiming "`actual_series` must be strictly positive to compute the sMAPE."

**To Reproduce**
```
ts_actuals = TimeSeries.from_values(np.array([0, 2, 3, 4, 5])) 
ts_preds = (TimeSeries.from_values(np.array([0, 2, 3, 4, 5]) )) 

print(np.all(ts_actuals.values() == ts_preds.values()))  # True

smape(ts_actuals, ts_preds)  # Would expect 0.0, 40%, or raising ValueError depending on implementation
# ValueError: `actual_series` must be strictly positive to compute the sMAPE.
```

**Expected behavior**
I would expect that, as `smape` is defined for points of zero actual *or* prediction, it would evaluate to zero for points of exactly zero actual and exactly zero prediction. It does not seem practical to raise an error in this very specific situation that intuitively represents a point of perfect prediction. 

eg, I would edit: 

```
    def sape(...)
    ... 
-    if not np.logical_or(y_true != 0, y_pred != 0).all():
-        raise_log(
-            ValueError(
-                "`actual_series` must be strictly positive to compute the sMAPE."
-            ),
-            logger=logger,
-        )
-    return 200.0 * np.abs(y_true - y_pred) / (np.abs(y_true) + np.abs(y_pred))

+   numerator = np.abs(y_true - y_pred) 
+   denominator = (np.abs(y_true) + np.abs(y_pred))
+   return 200 * np.where(denomenator == 0, 0.0, numerator / denominator)
```

Failing that, I would expect the error raised to correctly state that actuals and predictions must not *both* be zero. 

 - Python version: [3.10
 - darts version 0.40.0

**Additional context**

eschibli:

Fixes #2983 https://github.com/unit8co/darts/issues/2983

### Summary

Updated s(m)ape to not raise a ValueError when actuals and predictions are zero for the same timestep, and instead report 0 error for that point.


darts/metrics/metrics.py
```diff
@@ -1889,12 +1889,7 @@ def smape(
         passed as input, parallelising operations regarding different ``TimeSeries``. Defaults to `1`
         (sequential). Setting the parameter to `-1` means using all the available processors.
     verbose
-        Optionally, whether to print operations progress.
-
-    Raises
-    ------
-    ValueError
-        If the `actual_series` and the `pred_series` contain some zeros at the same time index.
+        Optionally, whether to print operations progress.s
```
brunnedu:
```suggestion
        Optionally, whether to print operations progress.
```

darts/metrics/metrics.py
brunnedu:
```suggestion
```

brunnedu:
```suggestion
   When :math:`\\left| y_t \\right| + \\left| \\hat{y}_t \\right| = 0` for some :math:`t` (i.e., both actual and 
   prediction are zero), the error for that time step is 0.
```

brunnedu:
```suggestion
   When :math:`\\left| y_t \\right| + \\left| \\hat{y}_t \\right| = 0` for some :math:`t` (i.e., both actual and 
   prediction are zero), the error for that time step is 0.
```

brunnedu:
Hi @eschibli, I see that you updated the docstrings, but the CI is still flagging some formatting issues.
To make this easier and avoid these errors in the future, I recommend setting up `pre-commit` locally. For this you can simply run the following commands from the root of the repo:
```bash
pip install pre-commit && pre-commit install
pre-commit run --all-files
```
The `run --all-files` command should identify and fix the remaining docstring issues for you. Once done, just git add, commit, and push again!

eschibli:
> Hi @eschibli, I see that you updated the docstrings, but the CI is still flagging some formatting issues. To make this easier and avoid these errors in the future, I recommend setting up `pre-commit` locally. For this you can simply run the following commands from the root of the repo:
> 
> ```shell
> pip install pre-commit && pre-commit install
> pre-commit run --all-files
> ```
> 
> The `run --all-files` command should identify and fix the remaining docstring issues for you. Once done, just git add, commit, and push again!


Right. Sorry for hammering your pipeline, I always forget how to get pre-commit working on Windows!

PR #2989, opened 2026-01-13
Fix: StaticCovariatesTransformer cols_cat order differing from column order

### LINKED ISSUE #2988: [BUG] `StaticCovariatesTransformer` creates wrong one-hot column names when `cols_cat` order differs from data
**Describe the bug**

The `StaticCovariatesTransformer` creates incorrect one-hot encoded column names when the order specified in the `cols_cat` parameter differs from the actual column order in the data. This results in column names that combine the wrong feature names with the wrong category values (e.g., `City_US` and `Country_New York` instead of `City_New York` and `Country_US`).

**To Reproduce**

```python
import pandas as pd
import numpy as np
from sklearn.preprocessing import OneHotEncoder
from darts import TimeSeries
from darts.dataprocessing.transformers import StaticCovariatesTransformer

# 1. Create data with columns in order: [Country, City]
df = pd.DataFrame({"Country": ["US", "China"], "City": ["New York", "Beijing"]})
series = [TimeSeries.from_values(np.ones((5, 1)), static_covariates=df.iloc[[i]]) for i in range(2)]

# 2. Initialize transformer with different order: [City, Country]
transformer = StaticCovariatesTransformer(
    transformer_cat=OneHotEncoder(sparse_output=False),
    cols_cat=["City", "Country"]  # This order differs from df.columns
)

# 3. Fit and Transform
transformed = transformer.fit_transform(series)

print("Actual Data Values: Country=US, City=New York")
print(f"Resulting Columns:  {transformed[0].static_covariates.columns.tolist()}")
print("\nTransformed Values:")
print(transformed[0].static_covariates.to_string())
```

**Expected behavior**

For the first series with `Country="US"` and `City="New York"`, the expected one-hot encoded columns should be:
- `City_New York = 1.0`
- `City_Beijing = 0.0`
- `Country_US = 1.0`
- `Country_China = 0.0`

**Actual behavior**

```
Actual Data Values: Country=US, City=New York
Resulting Columns:  ['Country_Beijing', 'Country_New York', 'City_China', 'City_US']

Transformed Values:
static_covariates  Country_Beijing  Country_New York  City_China  City_US
0                              0.0               1.0         0.0      1.0
```

The column names are **incorrect**:
- `Country_New York = 1.0` (should be `City_New York`)
- `City_US = 1.0` (should be `Country_US`)

The "Country" feature has city names, and the "City" feature has country names.

**System**
 - Python version: 3.13.7
 - darts version: 0.40.0

brunnedu:

Fixes #2988.

### Summary

Fixed a bug in `StaticCovariatesTransformer` where one-hot encoded column names were incorrectly generated when the `cols_cat` parameter order differed from the column order in the input data.

**Example of the issue:**
```python
# Data has columns in order: ["Country", "City"]
df = pd.DataFrame({"Country": ["US"], "City": ["New York"]})
series = TimeSeries.from_values(np.ones((5, 1)), static_covariates=df)

# User specifies different order: ["City", "Country"]
transformer = StaticCovariatesTransformer(
    transformer_cat=OneHotEncoder(),
    cols_cat=["City", "Country"]
)
transformed = transformer.fit_transform(series)

# Before fix: Column names were wrong (e.g., "City_US" instead of "Country_US")
# After fix: Column names are correct ("City_New York", "Country_US")
```

**Changes:**
- Modified `StaticCovariatesTransformer` so that the `cols_cat` parameter respects the column order in the input data when applying transformations, ensuring consistent and correct one-hot encoded column names
- Added test `test_cols_cat_order_different_from_data` to verify correct behavior when `cols_cat` order differs from data column order

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 **Fixed**
 
 - Fixed bug in `StaticCovariatesTransformer` where one-hot encoded column names were incorrectly assigned when the order of columns specified in `cols_cat` differed from the actual data column order. This caused silent data corruption where column names combined wrong feature names with wrong category values (e.g., `City_US` instead of `Country_US`). [#2989](https://github.com/unit8co/darts/pull/2989) by [Dustin Brunner](https://github.com/brunnedu).
+- Fixed a bug in `TorchTrainingDataset` where `max_samples_per_ts` was not acting as an upper bound on the number of samples per time series. Now `max_samples_per_ts` correctly acts as an upper bound, capping the dataset size at the actual number of samples that can be extracted from the longest series. [#2987](https://github.com/unit8co/darts/pull/2987) by [Dustin Brunner](https://github.com/brunnedu).
+- Updated s(m)ape to not raise a ValueError when actuals and predictions are zero for the same timestep. [#2984](https://github.com/unit8co/darts/pull/2984) by [eschibli](https://github.com/eschibli).
 
 **Dependencies**
 
--- a/darts/metrics/metrics.py
+++ b/darts/metrics/metrics.py
@@ -1744,8 +1744,8 @@ def sape(
     .. math::
         200 \\cdot \\frac{\\left| y_t - \\hat{y}_t \\right|}{\\left| y_t \\right| + \\left| \\hat{y}_t \\right|}
 
-    Note that it will raise a `ValueError` if :math:`\\left| y_t \\right| + \\left| \\hat{y}_t \\right| = 0` for some
-    :math:`t`. Consider using the Absolute Scaled Error (:func:`~darts.metrics.metrics.ase`)  in these cases.
+    When :math:`\\left| y_t \\right| + \\left| \\hat{y}_t \\right| = 0` for some :math:`t` (i.e., both actual and
+    prediction are zero), the error for that time step is defined as 0.
 
     If :math:`\\hat{y}_t` are stochastic (contains several samples) or quantile predictions, use parameter `q` to
     specify on which quantile(s) to compute the metric on. By default, it uses the median 0.5 quantile
@@ -1785,11 +1785,6 @@ def sape(
     verbose
         Optionally, whether to print operations progress.
 
-    Raises
-    ------
-    ValueError
-        If `actual_series` and `pred_series` contain some zeros at the same time index.
-
     Returns
     -------
     float
@@ -1822,14 +1817,14 @@ def sape(
         remove_nan_union=True,
         q=q,
     )
-    if not np.logical_or(y_true != 0, y_pred != 0).all():
-        raise_log(
-            ValueError(
-                "`actual_series` must be strictly positive to compute the sMAPE."
-            ),
-            logger=logger,
-        )
-    return 200.0 * np.abs(y_true - y_pred) / (np.abs(y_true) + np.abs(y_pred))
+    numerator = 200 * np.abs(y_true - y_pred)
+    denominator = np.abs(y_true) + np.abs(y_pred)
+    return np.divide(
+        numerator,
+        denominator,
+        out=np.zeros_like(numerator, dtype=y_true.dtype),
+        where=denominator != 0,
+    )
 
 
 @multi_ts_support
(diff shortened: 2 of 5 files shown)


brunnedu:
@jakubchlapek Thanks for the review! I double-checked this, and the fix is only required for `cols_cat`, not for `cols_num`. Here's why:

**For categorical columns:** The `_create_category_mappings` function uses `cols_cat_ordered` to construct new column names when applying one-to-many encoders like `OneHotEncoder` (e.g., `"City_NewYork"`, `"City_LA"`). The column names are created by zipping feature names with the learned categories, so the order must match the actual data order to pair them correctly.

**For numerical columns:** Transformers like `MinMaxScaler` preserve the original column names (1-to-1 mapping). The column names come directly from `series.static_covariates.columns` when adding them back, not from the `cols_num` parameter. The `cols_num` parameter is only used to identify which columns to scale (via `.isin(cols_num)`), so its order doesn't affect the output.

darts/dataprocessing/transformers/static_covariates_transformer.py
dennisbader, line 1: Looks great @brunnedu! I agree with @jakubchlapek that it would be more robust for the future if we performed the same operation for both `cols_num` and `cols_cat`. I would suggest to combine `_infer_static_cov_dtypes()` and `_create_component_masks()` into one method (and remove the old ones) that returns both processed column names and masks.

e.g. something like below:

```
    @staticmethod
    def _process_static_cov_columns(
        stat_covs: pd.DataFrame,
        cols_num: Optional[Sequence[str]],
        cols_cat: Optional[Sequence[str]],
    ) -> tuple[list[str], list[str], np.ndarray, np.ndarray]:
        """
        Extracts numerical and categorical static covariate (component / columns) names and their component masks 
        in order of the input data.
        """
        if cols_num is None:
            mask_num = stat_covs.columns.isin(
                stat_covs.select_dtypes(include=np.number).columns
            )
        else:
            mask_num = stat_covs.columns.isin(cols_num)
        cols_num = stat_covs.columns[mask_num].tolist()

        if cols_cat is None:
            mask_cat = stat_covs.columns.isin(
                stat_covs.select_dtypes(exclude=np.number).columns
            )
        else:
            mask_cat = stat_covs.columns.isin(cols_cat)
        cols_cat = stat_covs.columns[mask_cat].tolist()
        return cols_num, cols_cat, mask_num, mask_cat
```

That would simplify things a bit and we can also revert all changes below

darts/tests/dataprocessing/transformers/test_static_covariates_transformer.py
```diff
@@ -211,6 +211,51 @@ def test_zero_cardinality_multi_series(self):
...
+        ]
+
+        transformer = StaticCovariatesTransformer(
+            transformer_cat=OneHotEncoder(sparse_output=False),
+            cols_cat=["City", "Country"],
+        )
+
+        transformed = transformer.fit_transform(series)
+
+        expected_columns = {
```
dennisbader: Instead of comparing two sets of columns, we should rather check that both series have the expected column order (e.g. compare lists)

darts/tests/dataprocessing/transformers/test_static_covariates_transformer.py
```diff
@@ -211,6 +211,51 @@ def test_zero_cardinality_multi_series(self):
...
+            "Country_US",
+        }
+        actual_columns = set(transformed[0].static_covariates.columns.tolist())
+        assert actual_columns == expected_columns
+
+        first_static_covs = transformed[0].static_covariates
+        assert first_static_covs["Country_US"].iloc[0] == 1.0
+        assert first_static_covs["Country_China"].iloc[0] == 0.0
+        assert first_static_covs["City_New York"].iloc[0] == 1.0
+        assert first_static_covs["City_Beijing"].iloc[0] == 0.0
```
dennisbader: These could also be simplified to (might be easier to compare the two series)

```suggestion
    assert first_static_covs.iloc[0].tolist() == [1.0, 0.0, 1.0, 0.0]
```

darts/dataprocessing/transformers/static_covariates_transformer.py
brunnedu: I agree, thanks for the suggestion! 👍

PR #2994, opened 2026-01-22
Feat/autc score metric

### LINKED ISSUE #2970: Add overlay_dx() metric to darts.metrics: tolerance-sweep “visual alignment” score (normalized AUC)
When evaluating forecasts, common error metrics (MAE/RMSE/MAPE/…) don’t reflect visual/operational alignment across tolerance thresholds. In practice, stakeholders often care about “how often are we within X%?” across multiple X values.

We propose Overlay-dx: compute the fraction of points within tolerance bands across a range of tolerances (defined as % of target range), then compute the normalized AUC of that curve to get a [0, 1] score (higher is better).

Darts metrics are exposed in `darts.metrics.metrics` with a standard signature accepting `actual_series`, `pred_series`, `intersect`, optional quantile selection q, and reduction functions (component/series reduction, parallelism). 

We’d like Overlay-dx to be available like other “aggregated over time” metrics, so users can call it:
```
from darts.metrics import overlay_dx

score = overlay_dx(actual_series, pred_series)
```
Implementation:
```
@multi_ts_support
@multivariate_support
def overlay_dx(
    actual_series,
    pred_series,
    intersect: bool = True,
    *,
    q=None,
    max_percentage: float = 100.0,
    min_percentage: float = 0.1,
    step: float = 0.1,
    component_reduction=np.nanmean,
    series_reduction=None,
    n_jobs: int = 1,
    verbose: bool = False,
):
    """Overlay-DX (normalized AUC of tolerance-sweep coverage curve)."""
```

**Additional context**
- Repository: https://github.com/Smile-SA/overlay_dx
- Pypi: https://pypi.org/project/overlay-dx/0.1.0/

jakubchlapek:

Fixes #2970 .

### Summary

Implements a new metric called `autc` (area under tolerance curve), which computes the fraction of points within tolerance bands across a range of tolerances (defined as % of target range), then computes the normalized AUC of that curve to get a [0, 1] score.

Additionally adds helper plotting functions to display the curves visually with support for multivariate timeseries, and along with unit tests.

Example plot of ExponentialSmoothing model on AirPassengers
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 - Fixed bug in `StaticCovariatesTransformer` where one-hot encoded column names were incorrectly assigned when the order of columns specified in `cols_cat` differed from the actual data column order. This caused silent data corruption where column names combined wrong feature names with wrong category values (e.g., `City_US` instead of `Country_US`). [#2989](https://github.com/unit8co/darts/pull/2989) by [Dustin Brunner](https://github.com/brunnedu).
 - Fixed a bug in `TorchTrainingDataset` where `max_samples_per_ts` was not acting as an upper bound on the number of samples per time series. Now `max_samples_per_ts` correctly acts as an upper bound, capping the dataset size at the actual number of samples that can be extracted from the longest series. [#2987](https://github.com/unit8co/darts/pull/2987) by [Dustin Brunner](https://github.com/brunnedu).
 - Updated s(m)ape to not raise a ValueError when actuals and predictions are zero for the same timestep. [#2984](https://github.com/unit8co/darts/pull/2984) by [eschibli](https://github.com/eschibli).
+- Updated outdated links to [holidays package](https://holidays.readthedocs.io/en/latest/#available-countries) in `TimeSeries`, `holidays_timeseries()`, and `Prophet`. [#2999](https://github.com/unit8co/darts/pull/2999) by [Zhihao Dai](https://github.com/daidahao).
 
 **Dependencies**
 
--- a/darts/models/forecasting/prophet_model.py
+++ b/darts/models/forecasting/prophet_model.py
@@ -97,13 +97,9 @@ class Prophet(FutureCovariatesLocalForecastingModel):
             ..
         country_holidays
             An optional country code, for which holidays can be taken into account by Prophet.
+            See `holidays package <https://holidays.readthedocs.io/en/latest/#available-countries>`__ for available
+            country codes.
 
-            See: https://github.com/dr-prodigy/python-holidays
-
-            In addition to those countries, Prophet includes holidays for these
-            countries: Brazil (BR), Indonesia (ID), India (IN), Malaysia (MY), Vietnam (VN),
-            Thailand (TH), Philippines (PH), Turkey (TU), Pakistan (PK), Bangladesh (BD),
-            Egypt (EG), China (CN), and Russia (RU).
         cap
             Parameter specifying the maximum carrying capacity when predicting with logistic growth.
             Mandatory when `growth = 'logistic'`, otherwise ignored.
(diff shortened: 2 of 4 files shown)


darts/utils/statistics.py
```diff
@@ -1090,3 +1090,166 @@ def plot_residuals_analysis(
...
+    ValueError
+        If the actual series has zero range (max == min).
+
+    Examples
+    --------
+    >>> from darts import TimeSeries
+    >>> from darts.utils.statistics import plot_tolerance_curve
+    >>> actual = TimeSeries.from_values([1.0, 2.0, 3.0, 4.0, 5.0])
+    >>> pred = TimeSeries.from_values([1.1, 2.2, 2.9, 4.1, 5.0])
+    >>> plot_tolerance_curve(actual, pred)
```
brunnedu: To maintain consistency with the other `plot_*` functions in `statistics.py` you could remove the "Examples" and incomplete "Raises" sections. Also, if there is a relevant external source for this consider adding a References section to match our standard format.

darts/metrics/metrics.py
```diff
@@ -2414,6 +2415,136 @@ def coefficient_of_variation(
...
+        The (sequence of) actual series.
+    pred_series
+        The (sequence of) predicted series.
+    intersect
+        For time series that are overlapping in time without having the same time index, setting `True`
+        will consider the values only over their common time interval (intersection in time).
+    min_tolerance
+        The minimum tolerance level as a fraction of the series range. Default is 0.0 (0%).
+    max_tolerance
+        The maximum tolerance level as a fraction of the series range. Default is 1.0 (100%).
```
brunnedu: Wouldn't "half-range" instead of "range" be more correct here based on the implementation and the description at the top of the docstring?

darts/utils/statistics.py
```diff
@@ -1090,3 +1090,166 @@ def plot_residuals_analysis(
...
+
+    See Also
+    --------
+    :func:`~darts.metrics.metrics.autc` : Compute the Area Under Tolerance Curve as a single score.
+    """
+    from darts.metrics.utils import _get_tolerances_and_coverages
```
brunnedu: Is there a specific reason for this inline import here? Otherwise consider moving to top-level imports for consistency.

darts/utils/statistics.py
```diff
@@ -1090,3 +1090,166 @@ def plot_residuals_analysis(
...
+    tolerances, coverages = _get_tolerances_and_coverages(
+        y_true, y_pred, min_tolerance, max_tolerance, step
+    )
+
+    n_components = coverages.shape[1]
+    is_univariate = n_components == 1
+
+    if axis is None:
+        plt.figure(figsize=fig_size)
+        axis = plt
```
brunnedu:
```suggestion
        fig, axis = plt.subplots(figsize=fig_size)
```
As we use `axis.set_xlabel` further down this currently results in an `AttributeError: module 'matplotlib.pyplot' has no attribute 'set_xlabel'`. This modification should fix it.

darts/metrics/utils.py
```diff
@@ -1034,3 +1034,57 @@ def _compute_score(
...
+            "components to compute the AUTC. Found zero range for at least one component."
+        )
+
+    # compute absolute errors normalized by half the range
+    abs_errors = np.abs(y_true - y_pred)
+    half_range = y_range / 2
+    normalized_errors = abs_errors / half_range
+
+    # get tolerance levels
+    tolerances = np.arange(min_tolerance, max_tolerance + step, step)
```
brunnedu:
```suggestion
    tolerances = np.linspace(min_tolerance, max_tolerance, num_steps)
```
Suggestion: Use `np.linspace` for better robustness against numerical instability. `np.arange` often has unpredictable endpoint inclusion when working with floats; for example, `np.arange(1.0, 1.2+0.1, 0.1)` results in `[1.0, 1.1, 1.2, 1.3]` due to floating-point drift. Switching to `np.linspace` with a defined `num_steps` is much more predictable and ensures the endpoints are strictly respected.

jakubchlapek: np.arange(0, 0.3+0.1, 0.1) results in [0.0, 0.1, 0.2, 0.3]. As its end-exclusive we add the step size. Is there sometihng else im missing? The current version also fits overlay_dx api, but that's the second priority

PR #3000, opened 2026-01-28
Feat/add to group dataframe() function

### LINKED ISSUE #2965: [FEATURE] Add function `to_group_dataframe()` to generate long DataFrame from list of `TimeSeries`
Add function `to_group_dataframe()` to `darts/timeseries.py` that takes a list of TimeSeries and generates a "long" DataFrame (with the desired "backend") with static covariates (optional with default True, allow column select) and metadata (optional with default False, allow column select) added as columns. 
For reference, the output should allow to be used by `TimeSeries.from_group_df()` to re-create the same TimeSeries.

Jaco-Pastorius:

Fixes #2965.

### Summary

- Added `to_group_dataframe()`function that takes a list of series as an input and returns a long format dataframe.
- Added support for static covariates and metatada in the `TimeSeries.to_dataframe()` method. Those can be added as columns in the result dataframe.
- Two new arguments for the `TimeSeries.to_dataframe()` method control the addition of metadata and static covariates.
- Component level static covariates are also supported in `to_group_dataframe()` and `TimeSeries.to_dataframe()`

### Other Information

A new feature request should be added to add component level static covariate in `from_group_dataframe()`

### DIFF
--- a/darts/timeseries.py
+++ b/darts/timeseries.py
@@ -828,7 +828,7 @@ class TimeSeries:
                     "`pandas.RangeIndex(len(df))`. If this is not desired consider adding a time column "
                     "to your `DataFrame` and defining `time_col`."
                 )
-            # if we are here, the dataframe was pandas
+            # if we are here, the DataFrame was pandas
             elif not (
                 isinstance(time_index, VALID_INDEX_TYPES)
                 or np.issubdtype(time_index.dtype, np.integer)
@@ -1802,37 +1802,38 @@ class TimeSeries:
         backend: Union[ModuleType, Implementation, str] = Implementation.PANDAS,
         time_as_index: bool = True,
         suppress_warnings: bool = False,
-        add_static_cov: Optional[Union[list[str], str, bool]] = False,
-        add_metadata: Optional[Union[list[str], str, bool]] = False,
+        add_static_covariates: Union[bool, str, list[str]] = False,
+        add_metadata: Union[bool, str, list[str]] = False,
     ):
         """Return a DataFrame representation of the series in a given `backend`.
 
         Each of the series components will appear as a column in the DataFrame.
-        If the series is stochastic, the samples are returned as columns of the dataframe with column names
+        If the series is stochastic, the samples are returned as columns of the DataFrame with column names
         as 'component_s#' (e.g. with two components and two samples:
         'comp0_s0', 'comp0_s1' 'comp1_s0' 'comp1_s1').
 
         Parameters
         ----------
         copy
-            Whether to return a copy of the dataframe. Leave it to True unless you know what you are doing.
+            Whether to return a copy of the DataFrame. Leave it to True unless you know what you are doing.
         backend
             The backend to which to export the `TimeSeries`. See the `narwhals documentation
             <https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.from_dict>`__ for all supported
             backends.
         time_as_index
-            Whether to set the time index as the index of the dataframe or in the left-most column.
+            Whether to set the time index as the index of the DataFrame or in the left-most column.
             Only effective with the pandas `backend`.
         suppress_warnings
             Whether to suppress the warnings for the `DataFrame` creation.
-        add_static_cov
-            Whether to add static covariates from the time series as columns in the resulting dataframe (one column per
-            component-static covariate pair). Can be a bool in case all the static covariates should be added, or a
-            string/list of string in case only a subset are needed.
+        add_static_covariates
+            Whether to add the series' static covariates to the resulting DataFrame (one column per component-static
+            covariate pair). If a bool, controls whether to add all static covariates or none. If a string, or list of
+            strings, specifies the subset of static covariate columns / names to add.
         add_metadata
-            Whether to add metadata from the time series as columns in the resulting dataframe (one column per
-            metadata). Can be a bool in case all the metadata should be added, or a string/list of string in case only
-            a subset are needed.
+            Whether to add the series' metadata to the resulting DataFrame (one column per metadata entry). If a bool,
+            controls whether to add all metadata entries or none. If a string, or list of strings, specifies the subset
+            of metadata keys / names to add.
+
         Returns
         -------
         DataFrame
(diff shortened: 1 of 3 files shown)


darts/timeseries.py
```diff
@@ -1834,23 +1843,82 @@ def to_dataframe(
...
+            for static_cov_col in static_cov_cols:
+                for comp in components:
+                    value = static_covs.loc[comp, static_cov_col]
+                    data_col = np.full(data.shape[0], value)
+                    if len(components) > 1:
+                        column = "_".join((comp, static_cov_col))
+                    else:
+                        column = static_cov_col
+                    data_dict[column] = data_col
+                    columns = columns.append(pd.Index([column]))
```
jakubchlapek: while this works for deterministic timeseries as columns will be a pd.Index, in the stochastic branch columns is set to a list. since .append returns None this will fail. I suggest removing this line since i don't see use for it, and if possible normalizing columns to always be a list

darts/timeseries.py
```diff
@@ -1834,23 +1843,82 @@ def to_dataframe(
...
+                        logger=logger,
+                    )
+            else:
+                raise_log(
+                    ValueError("add_metadata must be of type bool, str or list[str]"),
+                    logger=logger,
+                )
+            for metadata_col in metadata_cols:
+                data_col = np.full(data.shape[0], metadata[metadata_col])
+                columns = columns.append(pd.Index([metadata_col]))
```
jakubchlapek: same case as above, should be save to remove

darts/timeseries.py
```diff
@@ -1834,23 +1843,82 @@ def to_dataframe(
...
+                    raise_log(
+                        ValueError("All values in add_static_cov must be of type str"),
+                        logger=logger,
+                    )
+            else:
+                raise_log(
+                    ValueError("add_static_cov must be of type bool, str or List[str]"),
+                    logger=logger,
+                )
+            for static_cov_col in static_cov_cols:
```
jakubchlapek, line 1897: would be nice to also add some validation whether the static covariate names provided by user as a list exist in the timeseries and provide an explanation error message

darts/timeseries.py
```diff
@@ -1834,23 +1843,82 @@ def to_dataframe(
...
+                    raise_log(
+                        ValueError("All values in add_metadata must be of type str"),
+                        logger=logger,
+                    )
+            else:
+                raise_log(
+                    ValueError("add_metadata must be of type bool, str or list[str]"),
+                    logger=logger,
+                )
+            for metadata_col in metadata_cols:
```
jakubchlapek, line 1929: same case with input validation as in the static covariates block

darts/timeseries.py
```diff
@@ -1834,23 +1843,82 @@ def to_dataframe(
...
+                )
+                if not all(isinstance(x, str) for x in static_cov_cols):
+                    raise_log(
+                        ValueError("All values in add_static_cov must be of type str"),
+                        logger=logger,
+                    )
+            else:
+                raise_log(
+                    ValueError("add_static_cov must be of type bool, str or List[str]"),
+                    logger=logger,
```
jakubchlapek: just typo in 'list' casing its whatever
```suggestion
                    ValueError("add_static_cov must be of type bool, str or list[str]"),
                    logger=logger,
```

darts/timeseries.py
```diff
@@ -6054,6 +6122,81 @@ def slice_intersect(series: Sequence[TimeSeries]) -> list[TimeSeries]:
...
+        A grouped DataFrame representation of the input `TimeSeries`(s) in the specified `backend`.
+        The DataFrame is obtained by concatenating the individual DataFrames generated from each series.
+    """
+
+    dfs = []
+
+    if isinstance(series, TimeSeries):
+        series = [series]
+
+    for idx, serie in enumerate(series):
```
jakubchlapek: can get rid of enumerate here
```suggestion
    for serie in series:
```

darts/timeseries.py
```diff
@@ -6054,6 +6122,81 @@ def slice_intersect(series: Sequence[TimeSeries]) -> list[TimeSeries]:
...
+            suppress_warnings,
+            add_static_cov,
+            add_metadata,
+        )
+        _df = nw.from_native(_df)
+        dfs.append(_df)
+
+    df = nw.concat(dfs)
+    df = df.to_native()
+    if backend == "pandas" and not time_as_index:
```
jakubchlapek: since we also accept Implementation we should normalise here as Implementation.PANDAS will not pass this check, although it should.
maybe we can do
```
backend = Implementation.from_backend(backend)
```
at the top since it accepts both Implementations and strings and will handle it on their side.
```suggestion
    if backend.is_pandas() and not time_as_index:
```

darts/timeseries.py
```diff
@@ -6054,6 +6122,81 @@ def slice_intersect(series: Sequence[TimeSeries]) -> list[TimeSeries]:
...
+
+    for idx, serie in enumerate(series):
+        _df = serie.to_dataframe(
+            copy,
+            backend,
+            time_as_index,
+            suppress_warnings,
+            add_static_cov,
+            add_metadata,
+        )
```
jakubchlapek: personal preference/nitpick, but nice to have these be keyword arguments instead of positional
```suggestion
        _df = serie.to_dataframe(
            copy=copy,
            backend=backend,
            time_as_index=time_as_index,
            suppress_warnings=suppress_warnings,
            add_static_cov=add_static_cov,
            add_metadata=add_metadata,
        )
```

Jaco-Pastorius:
Hey @jakubchlapek, thanks a lot for your review it all makes sense to me !

I've also added some tests in my latest commit as some lines weren't covered. 

I also added an argument in `to_group_dataframe` called `add_group_col` that gives the user the option to add a group column in the resulting long format data frame in case they don't have any static covariates or metadata that could be used for groups. Let me know what you think about it !

Cheers

darts/timeseries.py
```diff
@@ -1859,23 +1868,102 @@ def to_dataframe(
...
+                        logger=logger,
+                    )
+                missing_cols = [
+                    col for col in metadata_cols if col not in metadata.keys()
+                ]
+                if missing_cols:
+                    raise_log(
+                        ValueError(
+                            f"The following metadata to add via `add_metadata` do not exist: {missing_cols}."
+                            f"Available static covariates are: {list(metadata.keys())}"
```
jakubchlapek: typo :)

```suggestion
                            f"Available metadata are: {list(metadata.keys())}"
```

darts/timeseries.py
```diff
@@ -6093,6 +6181,89 @@ def slice_intersect(series: Sequence[TimeSeries]) -> list[TimeSeries]:
...
+        should be added, or a string/list of strings in case only a subset is needed. True by default as static
+        covariates should be provided to specify groups.
+    add_metadata
+        Whether to add metadata from the time series as columns in the resulting DataFrame
+        (one column per metadata entry). Can be a bool in case all metadata should be added,
+        or a string/list of strings in case only a subset is needed.
+    add_group_col
+        Whether to add a group column in the resulting long format dataframe. The values of that group column will go
+        from 1 to the number of time series in the input list
+
```
jakubchlapek: based on the implementation and my testing the group numbering starts from 0 (0 to N-1)
```suggestion
        from 0 to the number of time series minus 1 in the input list

```

PR #3002, opened 2026-01-29
[Feature] NeuralForecast Adapter

### LINKED ISSUE #658: [New model] Informer
See: https://github.com/unit8co/darts/issues/467

Informer:

https://arxiv.org/pdf/2012.07436.pdf

https://paperswithcode.com/paper/informer-beyond-efficient-transformer-for

https://github.com/zhouhaoyi/Informer2020

### LINKED ISSUE #937: [New model] Autoformer
https://github.com/thuml/Autoformer

### LINKED ISSUE #1029: [New model] FEDformer
https://github.com/DAMO-DI-ML/ICML2022-FEDformer

### LINKED ISSUE #1872: Add PathTST
Link to repository: https://github.com/yuqinie98/PatchTST

**Additional context**
The model promises efficiency and MAE and MSE better than transformer and non-transformer models.

### LINKED ISSUE #1943: NBEATS and NBEATSx handling exogenous variables
I am quite new and I am still trying to understand how NBEATS is handling exogenous variables, I know the architecture is not designed for that, and this issue is overcome in NBEATx  

How are handled in Darts NBEATS version external variables/covariates?

> In addition to the univariate version presented in the paper, our implementation also supports multivariate series (and covariates) by flattening the model inputs to a 1-D series and reshaping the outputs to a tensor of appropriate dimensions. Furthermore, it also supports producing probabilistic forecasts (by specifying a likelihood parameter).

Are you planning to implemente NBEATSx? 
Here NBEATSx Paper: [https://arxiv.org/pdf/2104.05522.pdf)](https://arxiv.org/pdf/2104.05522.pdf)
Here a Python implementation: [https://github.com/cchallu/nbeatsx/blob/main/nbeatsx_example.ipynb](https://github.com/cchallu/nbeatsx/blob/main/nbeatsx_example.ipynb)

### LINKED ISSUE #2138: Add iTransformer (Inverted Transformer) to DNN models
It appears that DNN models are advancing on a monthly basis. The iTransofmer paper from Dec 2023 claims to be a more resource efficient model with better accuracy than TiDE, PatchTST, TimesNet and other recent DNNs for time series.

Consider adding iTransformer to the darts DNN model set.

**Additional context**
iTransformer paper:
https://arxiv.org/pdf/2310.06625.pdf

iTransformer implementation:
https://github.com/thuml/iTransformer

### LINKED ISSUE #2164: Adding NPTS and Seasonal NPTS as baseline methods
It would be useful to add the Non-Parametric Time Series forecaster (NPTS) and seasonal NPTS as baseline methods (example : https://docs.aws.amazon.com/forecast/latest/dg/aws-forecast-recipe-npts.html). There is also a global model that was developed DeepNPTS. NPTS and DeepNPTS are strong baseline and DeepNPTS are competitive to complex models. You can find the paper here (https://arxiv.org/pdf/2312.14657.pdf) and the models are included in GluonTS.

### LINKED ISSUE #2379: Add KAN (Kolmogorov-Arnold Network)  to Darts
There has been a recent boom of positive comments on the KAN (Kolmogorov-Arnold Network) model. Are you thinking about incorporating it? Repo: https://github.com/KindXiaoming/pykan

### LINKED ISSUE #2395: New model - TimeMixer
Latest and greatest model yet again. 

https://github.com/kwuking/TimeMixer

### LINKED ISSUE #2424: New model - xLSTM
New model that may revive RNN architecture: 

https://github.com/NX-AI/xlstm

### LINKED ISSUE #2425: New model - SOFTS
Hello there! It would be great to see the [SOFTS](https://github.com/Secilia-Cxy/SOFTS) model in some next darts future release.

### LINKED ISSUE #2800: [feature request] Adding new model: TimeXer?
https://github.com/thuml/TimeXer

"TimeXer: Empowering Transformers for Time Series Forecasting with Exogenous Variables" (NeurIPS 2024)

https://raw.githubusercontent.com/thuml/TimeXer/main/figures/Introduction.png

This performance chart looks decent.

Just want to compare with other models in darts, under one framework.

### LINKED ISSUE #2844: Neural Forecast
Good morning, 

So, I understand based on the feedback about Neural Prophet that you would need more dependencies. But I decided to ask about finding a way to incorporate the Neural Forecast package in Darts as well. I decided to make this a separate issue since my original question was resolved. 

https://github.com/Nixtla/neuralforecast

Such as STEM-GNN, FedFormer, their original LSTM like structure, and DeepNTS. 

You had done integration with StatsForecast already.

daidahao:
- [x] Unit tests for various NF models.
- [x] Unit tests for basic TFM functionality.
- [x] README model catalogue and `index.rst`.
- [x] Covariates documentation.
- [x] Clarify dependencies.

Fixes #2844, fixes #2379, fixes #2164, fixes #1872, fixes #1029, fixes #937, fixes #658, fixes #2800, fixes #2425, fixes #2424, fixes #2395, fixes #2138, fixes #1943

### Summary
Adds a new `NeuralForecastModel` integration (new `nf_model.py`) to interface `NeuralForecast` with optional dependency setup, import-free model creation, and model wiring into the PyTorch forecasting module stack.

It supports the following via Darts `TorchForecastingModel`:
- Univariate forecasting (any NF base model, univariate or multivariate).
- Multivariate forecasting (any NF model).
- Past/future/static covariates (any NF model with support).
- Multiple time series forecasting (any NF model).
- Loss function (any NF model).
- Probabilistic forecasting (any NF model).
- Output chunk shift (any NF model).
- All other functionality supported in `TorchForecastingModel`: auto-regressive forecasting, historical forecasts, model loading/saving, and a lot more.

`NeuralForecastModel` is inspired by `StatsForecastModel` #2770. It can take any `BaseModel` instance from NeuralForecast and convert it into a `PLForecastingModule`.

**Design Rationale**:
Instead of interfacing the core `NeuralForecast` from NeuralForecast, I decided to interface the  `BaseModel` which all torch models inherit from. That is because core `NeuralForecast` does not expose a PyTorch-native `forward()` API, making TFM integration difficult if not impossible. Interfacing the `BaseModel` which provides PyTorch-native interface is admittedly a bit clumsy but allows existing Darts torch infrastructure to be used.

**Limitations**:
- Recurrent models are not supported due to different input/output shapes. Besides, almost all NeuralForecast's recurrent models--DeepAR, GRU, RNN, LSTM--are implemented in Darts natively.
- Core `NeuralForecast` are not supported due to its lack of torch-native APIs.

**API Reference**:
- [NeuralForecast Map](https://github.com/Nixtla/neuralforecast/blob/c9c53deacb33a21857850234c436d84d7ae6cc21/nbs/docs/api-reference/neuralforecast_map.ipynb) gives a high-level overview of NeuralForecast design and flow.
- [Adding Models to NeuralForecast](https://github.com/Nixtla/neuralforecast/blob/c9c53deacb33a21857850234c436d84d7ae6cc21/nbs/docs/tutorials/adding_models.ipynb) defines the tensor shape of `forward()` inside a **univariate*** `BaseModel`.

*Surprisingly, I could not find the shape definitions for a multivariate `BaseModel`. They are inferred from `TSMixerx` and other multivariate models.

<details>

<summary><strong>Code Example</strong> (Adapted from Chronos-2 notebook)</summary>

```py
import numpy as np

from darts import set_option
from darts.models.forecasting.nf_model import NeuralForecastModel


def prepare_data(val_n_days: int):
    # convert to float32 as Chronos-2 works with float32 input
    data = ElectricityConsumptionZurichDataset().load().astype(np.float32)
    ts_energy = data[["Value_NE5"]]
    # extract temperature, solar irradiation and rain duration
    ts_weather = data[["T [°C]", "StrGlo [W/m2]", "RainDur [min]"]]
    # split into train and validation sets by last 7 days
    train_energy, val_energy = ts_energy.split_before(len(ts_energy) - val_n_days * 24 * 4)
    return train_energy, val_energy, ts_weather

# prepare data
train_energy, val_energy, ts_weather = prepare_data(val_n_days=7)

# define a NeuralForecast model (30 days input, 7 days output)
model = NeuralForecastModel(
    model="TSMixerx",
    output_chunk_length=7 * 24 * 4,
    likelihood=QuantileRegression(quantiles=[0.1, 0.5, 0.9]),
    n_epochs=10,
)
# train model
model.fit(
    series=train_energy,
    future_covariates=ts_weather,
    max_samples_per_ts=1000,
    verbose=True,
)

# predict 7 days ahead
pred_prob = model.predict(
    n=7 * 24 * 4,
    num_samples=100,
)

# use darts plotting style
set_option("plotting.use_darts_style", True)
val_energy.plot(label="actual")
pred_prob.plot(label="forecast");
```

</details>

<figure>
KAN forecasts
<figcaption><p>KAN forecasts</p></figcaption>
</figure>

<figure>
TSMixerx forecasts
<figcaption><p>TSMixerx forecasts</p></figcaption>
</figure>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
-- 🚀🚀 Added new forecasting model `NeuralForecastModel` to convert any NeuralForecast base model into a Darts `TorchForecastingModel`. Like all torch models, it supports univariate, multivariate, probabilistic forecasting, and more. Depending on the base model, it also supports past, future, and static covariates. [#3002](https://github.com/unit8co/darts/pull/3002) by [Zhihao Dai](https://github.com/daidahao)
+- 🚀🚀 Added new forecasting model `NeuralForecastModel` to convert any of the 30+ NeuralForecast base model into a Darts `TorchForecastingModel`. This includes models such as NBEATSx, PatchTST, TimeXer, KAN, and many more. Like all Darts torch models, it supports univariate, multivariate, probabilistic forecasting, optimized backtesting and more. Depending on the base model, it also supports past, future, and static covariates. [#3002](https://github.com/unit8co/darts/pull/3002) by [Zhihao Dai](https://github.com/daidahao)
 
 **Fixed**
 


daidahao:
There are several models mentioned in Issues but not currently implemented in Darts. A NeuralForecast adapter could become an intermediary solution for those requests.

**Univariate:**
KAN #2379 
DeepNPTS #2164 
PatchTST #1872 
FEDformer #1029 
Autoformer #937 
Informer #658 

**Multivariate:**
TimeXer #2800 
SOFTS #2425 
xLSTM #2424 
TimeMixer #2395 
iTransformer #2138 
NBEATSx #1943

daidahao:
Hi @dennisbader, any thoughts on the proposed adapter here?

dennisbader:
@daidahao, thanks a lot for this PR and I'm definitely in favor of adding support for neuralforecast models! 

I also like the full-integration into our torch framework.

I'm currently wrapping up things for the next Darts release. Therefore, it'll take me some days to come back to this and give a real review.

I'll let you know once I can get to it :)

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,653 @@
...
+            hist_exog = x_past[:, :, self.past_slice]
+            if self.is_multivariate:
+                # -> (B, X, L, 1)
+                hist_exog = hist_exog.transpose(1, 2).unsqueeze(-1)
+                # -> (B, X, L, C)
+                hist_exog = hist_exog.repeat(1, 1, 1, self.n_targets)
+
+        # process future covariates if supported and provided
+        if x_future is not None:
+            # `futr_exog`: (B, L + H, F)
```
jakubchlapek: to keep in line with the branch above and safeguard is self.future_slice is None
```suggestion
        if self.future_slice is not None:
            # `futr_exog`: (B, L + H, F)
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,653 @@
...
+
+
+class _WindowBatch(TypedDict):
+    insample_y: torch.Tensor
+    insample_mask: torch.Tensor
+    # outsample_y: Optional[torch.Tensor]
+    # outsample_mask: Optional[torch.Tensor]
+    hist_exog: Optional[torch.Tensor]
+    futr_exog: Optional[torch.Tensor]
+    stat_exog: Optional[torch.Tensor]
```
jakubchlapek: just cleanup, probably can be removed now, also would be nice to provide a short docstring explaining the usecase
```suggestion
class _WindowBatch(TypedDict):
    insample_y: torch.Tensor
    insample_mask: torch.Tensor
    hist_exog: Optional[torch.Tensor]
    futr_exog: Optional[torch.Tensor]
    stat_exog: Optional[torch.Tensor]
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,653 @@
...
+    "lr_scheduler",
+    "lr_scheduler_kwargs",
+    "dataloader_kwargs",
+}
+
+
+class _PseudoLoss(BasePointLoss):
+    def __init__(self, likelihood: Optional[TorchLikelihood]):
+        n_likelihood_params = likelihood.num_parameters if likelihood is not None else 1
+        super().__init__(outputsize_multiplier=n_likelihood_params)
```
jakubchlapek, line 122: would be nice to have this documented also

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,653 @@
...
+            logger.warning(
+                "Multivariate NeuralForecast models require static covariates to be the same "
+                "across time series, but may be different across target components. "
+                "If you have multiple time series, setting `use_static_covariates=True` "
+                "will use the static covariates of the first sample in each batch, instead of "
+                "providing different static covariates per time series."
+            )
+
+        # consider static covariates if supported by `nf_model_class`
+        self._considers_static_covariates = use_static_covariates
```
jakubchlapek, line 657: is there a reason to not use `self.nf_model_class.EXOGENOUS_STAT` as the basis for `self._considers_static_covariates`? maybe then we could add the option to explicitly not take them into account based on user param

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,653 @@
...
+            Under the hood, a new base model instance will be created with the relevant parameters from ``model``.
+            That means that ``model`` itself will not be trained or updated.
+            Use :func:`nf_model` to access the new base model instance.
+        .. warning::
+            For compatibility, when static covariates are enabled for a multivariate base model, Darts will use the
+            static covariates of the first sample in each batch as the static covariates for the entire batch.
+            This may cause issues if you have multiple time series with different static covariates.
+            Please consider setting ``use_static_covariates=False`` to disable support or
+            setting ``batch_size=1`` to ensure that each batch only contains one time series.
+        """
```
jakubchlapek, line 610: input validation here, maybe for output_chunk_shift >= 0 as well
```suggestion
        if not isinstance(model, BaseModel):
            raise ValueError(
                "`model` must be a NeuralForecast model imported from `neuralforecast.models`."
            )
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,653 @@
...
+            is_multivariate=self.supports_multivariate,
+            **pl_module_params,
+        )
+
+    @property
+    def nf_model(self) -> BaseModel:
+        if not isinstance(self.model, _PLForecastingModule):
+            raise_log(
+                ValueError(
+                    "The underlying NeuralForecast model has not been created yet."
```
jakubchlapek: the error could be a bit more helpful here, maybe suggest to call .fit

jakubchlapek (commented):
Hey @daidahao thanks for the PR! I also think adding support for neuralforecast models is a great idea. In the meantime while waiting for dennis's review I have also a few suggestions and questions, but mostly small things. Overall I agree with the majority of design decisions so it's hard to not nitpick, the code and the documentation looks good :)

daidahao:
> Hey @daidahao thanks for the PR! I also think adding support for neuralforecast models is a great idea. In the meantime while waiting for dennis's review I have also a few suggestions and questions, but mostly small things. Overall I agree with the majority of design decisions so it's hard to not nitpick, the code and the documentation looks good :)

@jakubchlapek

Thanks for your kind comments! I might be busy for the next couple days but will catch up once I can.

darts/models/forecasting/nf_model.py
daidahao: @jakubchlapek 

Since `output_chunk_shift` is a `PLForecastingModule` attribute, I think it might be best to validate that in `PLForecastingModule` or `TorchForecastingModel`.

darts/models/forecasting/nf_model.py
daidahao: @jakubchlapek

Check out https://github.com/daidahao/darts/blob/8a301a366dbfd4f00842201da7b35cb07352bbbe/darts/models/forecasting/torch_forecasting_model.py#L1022 and TSMixer which I used as reference. Whereas `self.supports_static_covariates` defines the capability, `self._considers_static_covariates` decides whether the model uses static covariates.

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+    N: likelihood parameters
+
+In NeuralForecast, `BaseModel.forward()` takes a single argument which is a dictionary
+containing all inputs. See `BaseModel._parse_windows()` and `BaseModel.training_step()`
+to see how these inputs are being built and used.
+
+We thus define the expected keys and their types below:
+- `insample_y`: (B, L, C), historical target values in the input window.
+- `insample_mask`: (B, L), binary mask indicating available target values in `insample_y`.
+- `hist_exog`: (B, L, X) for univariate models or (B, X, L, C) for multivariate models,
```
dennisbader, line 27: Is this order of `(B, X, L, C)` correct? e.g. `L` changes position with `X`?

Also, is it really the case that in the multivariate model, each target component has its own multivariate past covariates?

Or is this simply the translation to our multiple time series case (a list of target series which can each be uni- or multivariate)?

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+        super().__init__(outputsize_multiplier=n_likelihood_params)
+
+
+class _NFModel(BaseModel):
+    """This serves as a protocol for expected NeuralForecast BaseModel API."""
+
+    def forward(self, window_batch: _WindowBatch) -> torch.Tensor: ...
+
+
+class _PLForecastingModule(PLForecastingModule):
```
dennisbader: just to conform with the module naming of other models

```suggestion
class _NeuralForecastModule(PLForecastingModule):
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+        # NeuralForecast models do not use `output_chunk_shift`, but we still allow users
+        # to set it to create a time gap. From the base model's perspective, it is trained
+        # to predict the shifted output chunk as if it immediately follows the input chunk.
+        if output_chunk_shift > 0:
+            logger.warning(
+                f"NeuralForecast base models natively do not use `output_chunk_shift`. "
+                f"Setting `output_chunk_shift={output_chunk_shift}` will create a time gap "
+                f"that is imperceptible to the base model. The model is trained to predict "
+                f"the shifted output chunk as if it immediately follows the input chunk."
+            )
```
dennisbader: Is this warning really required? The same applies to our Darts torch models, they are not really aware that there is a gap, it's simply a matter of what future values of the target are passed to compute the loss.

In my opinion we can remove it :) 


```suggestion
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+
+        if self.nf_model_class.RECURRENT:
+            raise_log(
+                NotImplementedError(
+                    "Recurrent NeuralForecast models are currently not supported."
+                ),
+                logger,
+            )
+        if self.supports_multivariate and use_static_covariates:
+            logger.warning(
```
dennisbader, line 648: This here I don't fully understand conceptually. For me, static covariates are only useful, whenever we deal with multiple time series (e.g. a list of time series, each uni or multivariate). So, using the same static covariates for all input series would not have any benefits 🤔 

I'm probably missing something 😅

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+        )
+        # remove ignored params
+        if len(ignored_params_in_use) > 0:
+            logger.info(
+                f"The following NeuralForecast model parameters will be ignored "
+                f"as they are either managed by Darts or not relevant: {ignored_params_in_use}"
+            )
+            for param in ignored_params_in_use:
+                self.nf_model_params.pop(param)
+        if self.nf_model_params.get("use_norm", False) and use_reversible_instance_norm:
```
dennisbader: Is norm/denorm only applied within the forward scope? If the logic is applied outside, then we have to ignore the param in general.
If the logic is applied inside, then do we really have to intervene? Worst case it would perform a sequential norm/denormalization, no?

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+    "h_train",
+    "inference_input_size",
+    "step_size",
+    "num_lr_decays",
+    "early_stop_patience_steps",
+    "scaler_type",
+    "futr_exog_list",  # prepared by Darts
+    "hist_exog_list",  # prepared by Darts
+    "stat_exog_list",  # prepared by Darts
+    "exclude_insample_y",  # TODO: check if this should be ignored
```
dennisbader: would be great if we could support `exclude_insample_y`. As far as I understand it allows some models to only be trained on the covariates. That enables support for some great use cases, where you just want to predict the target based on other features (e.g. predictive maintenance where you try to model the normal operating range of a signal based on measurements of other sensors)

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+    "drop_last_loader",
+    "random_seed",  # TODO: check if this should be ignored
+    "alias",  # TODO: check if this should be ignored
```
dennisbader: yes can be ignored, handled by our `model_name`

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+
+class _WindowBatch(TypedDict):
+    insample_y: torch.Tensor
+    insample_mask: torch.Tensor
+    hist_exog: Optional[torch.Tensor]
+    futr_exog: Optional[torch.Tensor]
+    stat_exog: Optional[torch.Tensor]
+
+
+IGNORED_NF_MODEL_PARAM_NAMES = {
```
dennisbader: We should probably also add "revin" to the list

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+            future_covariates,
+            static_covariates,
+            future_target,
+        ) = train_sample
+
+        # TODO: sanity checks on covariate support of the nf_model_class
+
+        # validate number of target components
+        n_targets = future_target.shape[1]
+        if n_targets != 1 and not self.supports_multivariate:
```
dennisbader: is this not something that should be dynamically checked in earlier in `fit()` or fit_from_dataset when validating the input series / tensors?

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+        return _PLForecastingModule(
+            nf_model=nf_model,  # pyright: ignore[reportArgumentType]
+            n_past_covs=n_past_covs,
+            n_future_covs=n_future_covs,
+            is_multivariate=self.supports_multivariate,
+            **pl_module_params,
+        )
+
+    @property
+    def nf_model(self) -> BaseModel:
```
dennisbader: is this required?

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,676 @@
...
+        # -> (B, H, C, N)
+        y_pred = y_pred.unflatten(-1, (self.n_targets, -1))
+
+        return y_pred
+
+
+class NeuralForecastModel(MixedCovariatesTorchModel):
+    def __init__(
+        self,
+        model: BaseModel,
```
dennisbader: Thanks for this @daidahao and I really like how this PR looks. Also nice that you took our StatsForecastModel as a guide line.

Now for the neuralforecast models, I think we might be better off not to expect an instance of the nf model, but rather the model class name, or model class. The init signature could look something like this:

```
class NeuralForecastModel(MixedCovariatesTorchModel):
    def __init__(
        self,
        input_chunk_length: int,
        output_chunk_length: int,
        output_chunk_shift: int = 0,
        use_static_covariates: bool = False,
        model: Union[str, type[BaseModel]] = "SomeDefaultNFModel",
        model_kwargs: Optional[dict] = None,
        **kwargs,
    ):
```

Why now this instead of how we did for statsforecast (SF) models?
- We could simplify the lifes of our Darts users that are used to the input/output chunk vocabulary. For SF models, we do not have any of these "fixed" parameter names, therefore I ddin't care too much about expecting to model instance directly
- The main reason against the model instance is that we store (and deep copy) all model creation parameter values under `ForecastingModel.model_params`. That means, we would have a deep copy of the neural network lying in our model_params dict. I really want to avoid this
- PyTorch Lightning complains when passing a nn.Module instance as an input parameter to a LightningModule. We can also fix this by moving the NF module creation into `_PLForecastingModule`
- Hyperparameter will be explicitely passed as `model_kwargs`. This allows us to:
  - update the hparams for the user based on the input data that we get in `_create_model`, and also some of our Darts parameters (e.g. input_chunk_length -> input_size, random_state -> random_seed, n targets -> num_series, ...)
- Hyperparameter logging (e.g. later with Mlflow is nicer, since we don't have the model instance as hparam)
- The user can switch the NF models without having to import them

```
model = NeuralForecastModel(
    input_chunk_length=24,
    output_chunk_length=12,
    model="NBEATSx",
)
```

darts/models/forecasting/nf_model.py
daidahao: @dennisbader

You are right. I was thinking in the pattern of foundation models where the models are aware of the gap. I've removed it now.

darts/models/forecasting/nf_model.py
daidahao: Removing it now.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader

Surprisingly, this is not being checked by TFM. It appears all TFMs support multivariate and so this is not being checked.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader

I fully agree with your suggestions here. Recording NF model and kwargs in `ForecastingModel.model_params` prevents a deep copy of the NF model and enable explicit hyperparameter logging.

So far, I've updated the init signature and moved NF module creation inside `_NeuralForecastModule`. It might take me extra days to update the docstring and address other comments. I will let you know once everything is ready.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader
The latter case is true: the RINorm logic is applied inside the forward scope when `use_norm` or `revin` is set to `True`.

What I discovered is that models like `TimeXer` might crash when Darts RINorm and the NF model RINorm are performed sequentially and can only be avoided by turning off either of them:
```
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [MPSFloatType [32, 7, 1]], which is output 0 of DivBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

After you have pointed out the `revin` parameter, I also found that `RMoK` model always has RINorm enabled.

Overall, I agree with your sentiment that we should not intervene RINorm inside the forward scope. However, I think warnings are necessary. I've updated the handling logic to:
- If `use_norm` or `revin` is set to `True`, warn the users to set either the parameter or `use_reversible_instance_norm` to `False`.
- If neither parameter is in NF signature, but `revin_affine` is present, warn the users to set `use_reversible_instance_norm` to `False`.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader
Looking at NF `TSMixerx` implementation, I conclude that in the NF multivariate model, each target component has its own past/future covariates, rather than being shared across target components.

In NF [multivariate forecasting example](https://nixtlaverse.nixtla.io/neuralforecast/docs/tutorials/multivariate_tsmixer.html), we also see time series data being stored (`Y_df`) in a long format. The DataFrame would have `unique_id` (target series id, but actually target component in Darts), `ds` (timestamp), and `y` (value) columns. Any other columns are considereed exogenous variables (covariates) for the target series. In the case of multivariate forecasting, `Y_df` would consist of target component chunks repeating over `ds`. As far as I know, there is no support for multiple series in NF.

PS: Even though NF allows for different covariates per target component, the multivariate example `ETTm2` in fact has covariates repeated for all targets:
```py
# calculate std within 'ex_1', 'ex_2', 'ex_3', 'ex_4' for each time step
Y_std = Y_df.groupby('ds')[['ex_1', 'ex_2', 'ex_3', 'ex_4']].std()
print(Y_std)
"""
                     ex_1  ex_2  ex_3  ex_4
ds                                         
2016-07-01 00:00:00   0.0   0.0   0.0   0.0
2016-07-01 00:15:00   0.0   0.0   0.0   0.0
2016-07-01 00:30:00   0.0   0.0   0.0   0.0
2016-07-01 00:45:00   0.0   0.0   0.0   0.0
2016-07-01 01:00:00   0.0   0.0   0.0   0.0
...                   ...   ...   ...   ...
2018-02-20 22:45:00   0.0   0.0   0.0   0.0
2018-02-20 23:00:00   0.0   0.0   0.0   0.0
2018-02-20 23:15:00   0.0   0.0   0.0   0.0
2018-02-20 23:30:00   0.0   0.0   0.0   0.0
2018-02-20 23:45:00   0.0   0.0   0.0   0.0

[57600 rows x 4 columns]
"""
```

darts/models/forecasting/nf_model.py
daidahao: @dennisbader

See my [above comment](https://github.com/unit8co/darts/pull/3002#discussion_r2810024666), there are no multiple series in NF. Static covariates may be different per target component, so there could be some (limited) benefits here, say if the model assumes channel independence but uses static covariates as conditional.

It is a rather interesting design, IMHO, to take in `(B, C, S)` tensor shape in a univariate model, but `(C, S)` in a multivariate model.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader

Yes, that would a great feature if supported. NF currently applies zeroing on `insample_y` to "exclude" it. We could implement a similar logic in Darts TFM, but it means 1) the input chunk of target components are still accessed, even though never looked at in `module.forward()`, 2) the input chunk length must still be positive to access past/future covariates.

I would suggest that feature be controlled by a TFM or PLFM-level parameter, so best not to interfere here.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader 

Rather than exception here, perhaps we could support multivariate forecasting in a univariate NF model by converting multivariate series into multiple examples of univariate series:
- Fold target components into batch dimension.
- Duplicate covariates to match target batch dimension if covariates are supported.

What do you think?

dennisbader: @daidahao , yes that would be a nice addition to at least support the multivariate case for all models :) We should make it clear in the model description how we handle this case though (that this differs from the source implementation)

darts/models/forecasting/nf_model.py
dennisbader: Are you sure that there are no "multiple series" in NF? Isn't their dataset handling that case via some series "id_col" (e.g. to identify which df row belongs to which series), where different series are collected along the batch dimension? The same is the case for Darts.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader 
This has been added with the latest commit.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader 
Sorry, I should have been more specific here. There are no multiple series in NF multivariate forecasting, but there could be multiple series in univariate forecasting.

There are three columns in a `Y_df` training set: `unique_id` (time series id), `ds` (timestamp), and `y` (value). In the univariate forecasting, there could be multiple univariate time series and the `unique_id` is the id for each univariate time series. In the multivariate forecasting, `unique_id` is the id of the target component. But where is the time series id? The answer is that, IMHO, there is none. Here is why:

If we define a `Y_df` with two multivariate series, each with three components. We can train a multivariate model like `MLPMultivariate` perfectly.
```py
import pandas as pd

Y_df = pd.DataFrame({
    "unique_id": ["A"] * 8 + ["B"] * 8 + ["C"] * 8 + ["A"] * 8 + ["B"] * 8 + ["C"] * 8,
    "ds":
        pd.date_range(start="2026-01-01", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-01-01", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-01-01", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-02-02", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-02-02", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-02-02", periods=8, freq="d").to_list(),
    "y": list(range(48)),
})
model = MLPMultivariate(
    h=2, 
    input_size=2,
    n_series=3,
    max_steps=10,
)

fcst = NeuralForecast(
    models=[model],
    freq='d'
)
fcst.fit(df=Y_df, val_size=12)
forecasts = fcst.predict()
```

However, if we define the two series like this with timestamp misalignment, the model can still be trained and predictions would start from 2026 for "A" and "B", but 2027 for "C". 
```py
Y_df = pd.DataFrame({
    "unique_id": ["A"] * 8 + ["B"] * 8 + ["C"] * 8 + ["A"] * 8 + ["B"] * 8 + ["C"] * 8,
    "ds":
        pd.date_range(start="2026-01-01", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-01-01", periods=8, freq="d").to_list() +
        pd.date_range(start="2027-01-01", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-02-02", periods=8, freq="d").to_list() +
        pd.date_range(start="2026-02-02", periods=8, freq="d").to_list() +
        pd.date_range(start="2027-02-02", periods=8, freq="d").to_list(),
    "y": list(range(48)),
})
```

In Darts language, the NF "multivariate series" is in fact a multiple of univariate series and they do not need to be temporally aligned. While we might think there are two "multivariate series" here, that is in fact not true. Because under the hood, NF would concatenate values for each target component. Each component would have exactly one series:
```py
from neuralforecast.tsdataset import TimeSeriesDataset

dataset = TimeSeriesDataset.from_df(Y_df)[0]
dataset[0]
"""
{'temporal': tensor([[ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7., 24., 25., 26., 27., 28., 29.,
          30., 31.],
         [ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
           1.,  1.]]),
 'temporal_cols': Index(['y', 'available_mask'], dtype='object'),
 'static': None,
 'static_cols': None,
 'y_idx': 0}
"""
```

Multivariate forecasting in NF should be understood as cross learning--individual univariate series would have dependencies on other series, but they do not have to be aligned temporally.

daidahao: @dennisbader 
How would this affect our `NenuralForecastModel`? My suggestion is that it should not. Looking at multivariate models `MLPMultivariate` and `TSMixerx` in NF, there is no batch dimension for `stat_exog` input.

In fact, `stat_exog` is duplicated in `forward()` to match the batch dimension in both models. NF multivariate models assume all examples in a batch to have the same static covariates, which is how we interface them currently.

dennisbader:
@daidahao thanks a lot for the updates 🚀 I'll review today. Yes, let's only add these models as optional dependencies (e.g. not part of any of the darts package options, they have to be installed independently).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 ### For users of the library:
 
 **Improved**
+- 🚀🚀 Added new forecasting model `NeuralForecastModel` to convert any NeuralForecast base model into a Darts `TorchForecastingModel`. Like all torch models, it supports univariate, multivariate, probabilistic forecasting, and more. Depending on the base model, it also supports past, future, and static covariates. [#3002](https://github.com/unit8co/darts/pull/3002) by [Zhihao Dai](https://github.com/daidahao)
```
dennisbader: just a new line

```suggestion

- 🚀🚀 Added new forecasting model `NeuralForecastModel` to convert any NeuralForecast base model into a Darts `TorchForecastingModel`. Like all torch models, it supports univariate, multivariate, probabilistic forecasting, and more. Depending on the base model, it also supports past, future, and static covariates. [#3002](https://github.com/unit8co/darts/pull/3002) by [Zhihao Dai](https://github.com/daidahao)
```

docs/source/index.rst
```diff
@@ -651,6 +651,12 @@ Our regression models are designed to predict continuous numerical values, makin
      - ✅ ✅
      - ✅
      - `TSMixer paper <https://arxiv.org/pdf/2303.06053.pdf>`_, `PyTorch Implementation <https://github.com/ditschuk/pytorch-tsmixer>`_
+   * - `NeuralForecastModel <https://unit8co.github.io/darts/generated_api/darts.models.forecasting.nf_model.html#darts.models.forecasting.nf_model.NeuralForecastModel>`_
+     - ✅ ✅
+     - ✅ ✅ ✅
+     - ✅ ✅
+     - ✅
+     - `NeuralForecast Documentation <https://nixtlaverse.nixtla.io/neuralforecast>`_
```
dennisbader: the nixtla url doesn't work, can you update it here and in other docs (probably readme as well)

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+            model's covariate support).
+        output_chunk_shift
+            Optionally, the number of steps to shift the start of the output chunk into the future (relative to the
+            input chunk end). This will create a gap between the input and output. If the model supports
+            `future_covariates`, the future values are extracted from the shifted output chunk. Predictions will start
+            `output_chunk_shift` steps after the end of the target `series`. If `output_chunk_shift` is set, the model
+            cannot generate autoregressive predictions (`n > output_chunk_length`). Default: ``0``.
+        model_kwargs
+            A dictionary of architectural parameters to initialize the `NeuralForecast` base model. The expected
+            parameters depend on the base model used. See NeuralForecast documentation for details. Default: ``None``.
```
dennisbader:
```suggestion
            parameters depend on the base model used. See the NeuralForecast documentation for details. Default: ``None``.
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+            instance to be used for probabilistic forecasting.
+
+        - **Output chunk shift**: Supported for any base model, univariate or multivariate.
+
+          - Simply set ``output_chunk_shift`` to create a time gap between the input and output.
+
+        Parameters
+        ----------
+        model
+            Name or class of the NeuralForecast base model to be used from ``neuralforecast.models``,
```
dennisbader: adding the link here again for the nf model table would be nice :)

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+        return y_pred
+
+
+class NeuralForecastModel(MixedCovariatesTorchModel):
+    def __init__(
+        self,
+        model: str | type[BaseModel],
+        input_chunk_length: int,
+        output_chunk_length: int,
+        output_chunk_shift: int = 0,
```
dennisbader: Can we make `model` a kwarg by finding a good default model? It would allow to use NeuralForecastModel as any other torch models (for example for testing, we wouldn't have to initialize all nf models with model="model").

```suggestion
        input_chunk_length: int,
        output_chunk_length: int,
        output_chunk_shift: int = 0,
        model: str | type[BaseModel] = "TiDE",
```

Now we can call these models in the same way when testing.

```
from darts.models import NeuralForecastModel, TiDEModel

model = NeuralForecastModel(24, 12)
model = TiDEModel(24, 12)
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+        model_kwargs: dict | None = None,
+        use_static_covariates: bool = False,
+        **kwargs,
+    ):
+        """NeuralForecast Model.
+
+        Can be used to fit any `NeuralForecast` univariate or multivariate base model.
+        For a list of available base models,
+        see `NeuralForecast package <https://nixtlaverse.nixtla.io/neuralforecast/docs/capabilities/overview.html>`__.
```
dennisbader:
```suggestion
        see the `NeuralForecast package <https://nixtlaverse.nixtla.io/neuralforecast/docs/capabilities/overview.html>`__.
```

darts/tests/models/forecasting/test_nf_models.py
```diff
@@ -0,0 +1,663 @@
...
+                **kwargs,
+            )
+            model.fit(series=self.univariate_series)
+            pred = model.predict(n=4)
+            assert isinstance(pred, TimeSeries)
+            assert pred.n_timesteps == 4
+            assert pred.n_components == 1
+
+    @pytest.mark.parametrize("model_name, model_kwargs", ALL_MODELS)
+    def test_probabilistic_forecasting(
```
dennisbader, line 531: would be great to also test the case for multivariate target series

darts/tests/models/forecasting/test_nf_models.py
```diff
@@ -0,0 +1,663 @@
...
+    def test_multiple_series(self, model_name: str, model_kwargs: dict | None):
+        model = NeuralForecastModel(
+            model=model_name,
+            input_chunk_length=10,
+            output_chunk_length=8,
+            model_kwargs=model_kwargs,
+            **kwargs,
+        )
+        model.fit(series=self.multiple_series)
+        pred = model.predict(n=4, series=self.multiple_series[0])
```
dennisbader, line 403: would be great to have one test that checks that calling `predict()` with a multiple series also gives the expected results (list of predictions)

darts/tests/models/forecasting/test_nf_models.py
dennisbader, line 1: Nice tests, thanks a lot :) Actually, we already have test_torch_forecasting_models.py and test_global_forecasting_models.py which should cover most of the test cases. Can we at least add one of the models (e.g. TiDE?) to the model lists of those tests as well? (ideally we would have less duplicate tests, but maybe it's too tedious / more secure to leave your new tests here as well)

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+
+class _NeuralForecastModule(PLForecastingModule):
+    def __init__(
+        self,
+        nf_model_class: type[_NFModel],
+        nf_model_params: dict,
+        n_past_covs: int,
+        n_future_covs: int,
+        n_stat_covs: int,
+        is_multivariate: bool,
```
dennisbader: Is this parameter necessary? We could infer this directly within `__init__` since we have access to `self.n_targets`

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+            slice(
+                self.n_targets + n_past_covs,
+                self.n_targets + n_past_covs + n_future_covs,
+            )
+            if n_future_covs > 0
+            else None
+        )
+
+    @property
+    def converts_to_multivariate(self) -> bool:
```
dennisbader: we could store this as a model attribte in `__init__` to  avoid having to compute this condition in each forward call

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+        ignored_params_in_use = _NF_MODEL_IGNORED_PARAMS.intersection(
+            self.nf_model_params.keys()
+        )
+        if len(ignored_params_in_use) > 0:
+            logger.info(
+                f"The following NeuralForecast model parameters will be ignored "
+                f"as they are either managed by Darts or not relevant: {ignored_params_in_use}"
+            )
+            for rinorm_param in ignored_params_in_use:
+                self.nf_model_params.pop(rinorm_param)
```
dennisbader: here, params are not related to rin

```suggestion
            for param in ignored_params_in_use:
                self.nf_model_params.pop(param)
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+            if rinorm_param is None:
+                continue
+            if self.nf_model_params.get(rinorm_name, rinorm_param.default):
+                logger.warning(
+                    f"NeuralForecast model's `{rinorm_name}=True` may be incompatible with "
+                    f"`PLForecastingModule`'s `use_reversible_instance_norm=True` since they "
+                    f"both apply reversible instance normalization to the target series. "
+                    f"If you experience issues, please consider setting one of them to `False`."
+                )
+            return
```
dennisbader: why would we return already here?

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+        # import and validate the NeuralForecast base model class
+        if isinstance(model, str):
+            self.nf_model_class = self._import_nf_model_class(model)
+            self._validate_nf_model_class(name=f"neuralforecast.models.{model}")
+        else:
+            self.nf_model_class = model
+            self._validate_nf_model_class()
+
+        # extract, validate, and update the NeuralForecast base model parameters
+        self.nf_model_params = model_kwargs or {}
```
dennisbader: can we create a copy here of the model kwargs? Otherwise we might mutate the user's `model_kwargs`

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+    This allows the NeuralForecast base model to output the correct number of parameters for probabilistic forecasting.
+
+    """
+
+    def __init__(self, likelihood: TorchLikelihood | None):
+        n_likelihood_params = likelihood.num_parameters if likelihood is not None else 1
+        super().__init__(outputsize_multiplier=n_likelihood_params)
+
+
+class _NFModel(BaseModel):
```
dennisbader: I would say, either you actually make this a protocol with

```
from typing import Protocol

class _NFModel(Protocol):
    # define the methods and attributes required here
```

Or you simply annotate `nf_model_class: type[BaseModel],`  later on.

This would also avoid the need for the pyright ignore in the `NeuralForecastModel` `nf_model_class=self.nf_model_class,  # pyright: ignore[reportArgumentType]`.

darts/models/forecasting/nf_model.py
dennisbader: Thanks for the clarification :) Let's leave it then as is 👍

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+                "across time series, but may be different across target components. "
+                "If you have multiple time series, setting `use_static_covariates=True` "
+                "will use the static covariates of the first sample in each batch, instead of "
+                "providing different static covariates per time series."
+            )
+
+        # consider static covariates if supported by `nf_model_class`
+        self._considers_static_covariates = use_static_covariates
+
+    def _import_nf_model_class(self, model: str) -> type[BaseModel]:
```
dennisbader:
```suggestion
    @staticmethod
    def _import_nf_model_class(model: str) -> type[BaseModel]:
```

darts/models/forecasting/nf_model.py
```diff
@@ -0,0 +1,806 @@
...
+            )
+
+    def _update_nf_model_params(
+        self,
+        input_chunk_length: int,
+        output_chunk_length: int,
+    ) -> None:
+        # set `input_size` and `h` to match the `input_chunk_length` and `output_chunk_length` of the PL module
+        self.nf_model_params["input_size"] = input_chunk_length
+        self.nf_model_params["h"] = output_chunk_length
```
dennisbader, line 751: should we also have to set the random seed?

docs/source/index.rst
daidahao: @dennisbader 
This was actually the link on NF GitHub repo. I would update it.

daidahao:
@dennisbader 
Thank you for the review once again. I always find your review to be most comprehensive and meticulous. I will let you know once all the comments are addressed and the PR is ready.

I also appreciate the suggestion that we create a small notebook for using `NeuralForecastModel` and comparing against Darts natives. It should a great showcase of NF expanded capabilities via `TorchForecastingModel` in Darts, and to clarify design choices and caveats and offer comparisons.

Because I might not have extra bandwidth at the moment, I encourage anyone interested to start a PR for `NeuralForecastModel` notebook in the future. I am happy to provide support and review. My hope is that this PR (and possibly notebook) can be ready in time for next release.

BTW, congratulations on hitting **3000** issues/PRs! It is another milestone for this repository and I am proud to a small part of it.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader
Okay. Removing `_NFModel` now. Previously, I was following the `_SFModel` pattern in `sf_model.py`. It is not actually needed here, because we actually interface `BaseModel.__call__()` (`nn.Module`) rather than `forward()` of `BaseModel`

darts/models/forecasting/nf_model.py
daidahao: @dennisbader
This is a property for whether the **NF base model** is multivariate, but not for **_NeuralForecastModule**. NF multivariate base models expect different tensor shapes in `forward()` than univariate base models.

daidahao: Renaming it to `is_multivariate_base` to prevent confusion.

darts/models/forecasting/nf_model.py
daidahao: @dennisbader
If `nf_model_params[rinorm_name]` is set to `False`, RINorm is not enabled inside `forward()`, so we would return here.

darts/tests/models/forecasting/test_nf_models.py
daidahao: @dennisbader
Just added NF TiDE (univariate) and TSMixerx (multivariate) to `test_torch_forecasting_models.py` and `test_global_forecasting_models.py`. Also found that `test_torch_forecasting_models.py` assume all models to enable static covariates if supported. So I changed `use_static_covariates` default to `True`.

dennisbader:
@daidahao, perfect, thanks a lot for the updates 💯 

I pushed some last minor changes, it's ready now to be merged 🥳  🚀 

I also updated the PR description, to close the "new model" requests / issues that you referenced above. With the new `NeuralForecastModel` we have these supported in Darts :) 

Thanks again for the great work on this. It's an awesome feature!

dennisbader (approved):

PR #3003, opened 2026-01-30
Enhance fine-tuning capabilities for foundation models

### LINKED ISSUE #2964: [Feature] Chronos-2 fine-tuning support
It might be a good time to think about fine-tuning support of Chronos-2 since it has been supported in [Amazon's repo](https://github.com/amazon-science/chronos-forecasting/) and [autogluon](https://github.com/abdulfatir/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/chronos/chronos2.py).

In Amazon's repo, the default fine-tuning strategy is `"full"` so all model weights will be updated. But in autogluon (Amazon's time series library), the default fine-tuning strategy is [LoRA](https://huggingface.co/docs/peft/en/package_reference/lora) which means only parts of the model will be updated and in fact some low-rank matrices would injected into the network as weights. PEFT is required for LoRA but not for `"full"`.

There might be a challenge of implementing lora in Darts since 1) it would add `peft` as a dependency and 2) it is unclear whether peft is compatible with Lightning.

Enable the `_allows_finetuning` in `Chronos2Model` for initial testing. However, there might be a problem with the normalized training loss as noted in the comments:

```
    # TODO: fine-tuning support w/ normalized loss
    # Currently, Darts own `RINorm` is not used as Chronos-2 has its own implementation. Major differences
    # 1. Chronos-2 `RINorm` normalizes both target and covariates, while Darts normalizes target only.
    # 2. Chronos-2 `RINorm` additionally applies `arcsinh` transformation after standardization
    # 3. Chronos-2 uses normalized values for loss computation, while Darts uses denormalized values.
    # We need to think about how best to implement Chronos-2 `RINorm` in `io_processor()` without
    # breaking existing behavior, while also allowing fine-tuning with normalized loss.
```

**Additional context**

Kurokabe:

Fixes #2964 

### Summary
This PR implements native support for full and partial fine-tuning of foundation models (e.g., `Chronos2Model`) and adds advanced integration capabilities for external libraries like `peft`. 

1.  **Foundation Model Enhancements**:
    *   Updated `FoundationModel` base class to accept `enable_finetuning`, `freeze_patterns`, and `unfreeze_patterns`.
    *   Automatic injection of `LayerFreezeCallback` when fine-tuning is enabled with specific patterns.
    *   Added `internal_model` property to provide direct access to the underlying `nn.Module`, facilitating advanced use cases like PEFT/LoRA.

2.  **Callback Improvements**:
    *   Ensured `PeftCallback` correctly handles adapter merging during checkpointing, allowing models trained with LoRA to be saved and reloaded as standard Darts models.

3.  **Documentation & Examples**:
    *   Added a new example notebook `26-Chronos-2-finetuning-examples.ipynb` demonstrating full fine-tuning, partial fine-tuning with layer freezing, and LoRA integration.
    *   Included performance evaluation and persistence (save/load) examples for each method.

4.  **Testing**:
    *   Expanded tests in `test_foundation.py` covering all new fine-tuning scenarios and ensuring correct model state after saving/loading.

### How Has This Been Tested?
- [x] Added unit tests for `FoundationModel` fine-tuning logic.
- [x] Verified LoRA integration and weight merging via `PeftCallback`.
- [x] Manual verification of the example notebook.
- [x] All added tests in `test_foundation.py` pass.

### DIFF
--- a/darts/tests/models/forecasting/test_foundation.py
+++ b/darts/tests/models/forecasting/test_foundation.py
@@ -9,6 +9,7 @@ import pytest
 
 from darts import TimeSeries, concatenate
 from darts.tests.conftest import TORCH_AVAILABLE, tfm_kwargs
+from darts.utils.likelihood_models import QuantileRegression
 from darts.utils.timeseries_generation import linear_timeseries
 
 if not TORCH_AVAILABLE:
@@ -17,7 +18,7 @@ if not TORCH_AVAILABLE:
         allow_module_level=True,
     )
 
-from darts.models import Chronos2Model
+from darts.models import Chronos2Model, TimesFM2p5Model
 
 
 def generate_series(n_variables: int, length: int, prefix: str):
@@ -408,3 +409,50 @@ class TestFoundationModel:
             )
 
             model.fit(self.series)
+
+    @pytest.mark.parametrize(
+        "config",
+        [
+            (
+                TimesFM2p5Model,
+                "output_projection_point.hidden_layer.weight",
+                "google/timesfm-2.5-200m-pytorch",
+            ),
+            (Chronos2Model, "output_patch_embedding.*", "autogluon/chronos-2-small"),
+        ],
+    )
+    def test_finetuning_all_models(self, config):
+        """Tests fine-tuning with user-quantiles that are different to the ones the model was trained on."""
+        model_cls, pattern, model_revision = config
+        quantiles = [0.1, 0.5, 0.9]
+
+        model = model_cls(
+            input_chunk_length=12,
+            output_chunk_length=6,
+            enable_finetuning={"unfreeze": [pattern]},
+            n_epochs=1,
+            likelihood=QuantileRegression(quantiles),
+            hub_model_name=model_revision,
+            **tfm_kwargs,
+        )
+
+        # fit model with validation series (training quantile loss is different from evaluation quantile loss)
+        model.fit(self.series, val_series=self.series)
+
+        # Check requires_grad status
+        unfrozen_found = False
+        frozen_found = False
+
+        for name, param in model.model.named_parameters():
+            if pattern.replace("*", "") in name:
+                assert param.requires_grad is True
+                unfrozen_found = True
+            else:
+                assert param.requires_grad is False
+                frozen_found = True
+
+        assert unfrozen_found
+        assert frozen_found
+
+        preds = model.predict(n=6, predict_likelihood_parameters=True)
+        assert preds.shape == (6, self.series.n_components * len(quantiles), 1)


daidahao:
Hi @Kurokabe Thank you for this PR and your efforts at making fine-tuning work for foundation models! Here are my suggestions:

### Nested Model Attribute
After reviewing the code, I have some worries as to the nested `model` attribute of `FoundationModel`. From my perspective (having written `FoundationModel` and implemented Chronos-2 and TimesFM in Darts), I would raise two concerns:
- It adds a new layer to new model implementation, e.g., `FoundationModel` -> `FoundationPLModule` -> `nn.Module`, and creates confusion for developers, with limited benefits, i.e., PEFT support.
- It makes the model checkpoint (aka, `ckpt` file), incompatible with original checkpoints, because of the `model.*` prefix.

Even if we want PEFT support for foundation models, I wonder if we can do so without running into a nested `model.model.model` situation via more straightforward method overrides:

```py
class FoundationModel(MixedCovariatesTorchModel, ABC):

    @abstractmethod
    def _create_original_model(self, train_sample: TorchTrainingSample) -> PLForecastingModule:
        """Create the original PyTorch Lightning forecasting module without any PEFT adapters."""

    def _create_model(self, train_sample: TorchTrainingSample) -> PLForecastingModule:
        model = self._create_original_model(train_sample)
        if self._enable_finetuning and self.peft_config is not None:
            from peft import get_peft_model
            model = get_peft_model(model, self.peft_config)
        return model
```

We then override the `save()` method to ensure the PEFT-merged checkpoint is being saved when called:
```py
    def save(
        self,
        path: Optional[str] = None,
        clean: bool = False,
    ) -> None:
        if self._enable_finetuning and self.peft_config is not None:
            self.model.merge_adapter()
        super().save(path=path, clean=clean)
```

That way, we could avoid implementing additional `ModelTransformCallback` and `PeftCallback` which IMHO are a bit opaque to use and maintain.

I also argue that we might not need adapter merge for training checkpoints as it adds overheads and those checkpoints do not need to be compatible. Instead, we could suggest the users call `save()` at the end of training to get portable model weights.

### Fine-tuning Hyperparameters

Like I said in #2964, I recommend exposing the fine-tuning hyper-parameters to users rather than the callback. This allows direct control of fine-tuning behaviours.

```py
model_lora = Chronos2Model(
    input_chunk_length=24,
    output_chunk_length=6,
    enable_finetuning=True,
    n_epochs=50,
    unfreeze_patterns=unfreeze_patterns,
    peft_config=peft_config,
)
```

For partial fine-tuning, please also consider:
- Removing `freeze_patterns` as it is redundant to `unfreeze_patterns` that is more common than the former.
- Using `fnmatch` or suffix (`.endswith()`) to match model weights rather than prefix-only. Users might want match `*.self_attention.q.weight` rather than a prefix like `encoder.block.0.layer.1`.
- Raising an error when any pattern is not matched in `unfreeze_patterns` to prevent silent fails.
- Would it also be possible to combine `enable_finetuning` and `unfreeze_patterns` into one parameter `enable_finetuning` for shared semantics?

For PEFT fine-tuning, please consider:
- Exposing `peft_config` as a model hyper-parameter to directly configure PEFT.

Those are merely my suggestions for your considerations. Feel free to ignore them if you disagree.

Many thanks.

darts/models/forecasting/foundation_model.py
```diff
@@ -24,6 +27,8 @@ class FoundationModel(MixedCovariatesTorchModel, ABC):
     def __init__(
         self,
         enable_finetuning: bool = False,
+        freeze_patterns: list[str] | None = None,
```
dennisbader: We could combine the entire logic into a single parameter `enable_fine_tuning` to simplify things for the user. Here is an idea of how it could look like:

```
enable_fine_tuning: Optional[Union[bool, dict[str, list[str]]]] = None

# parameter description could look something like
"""
Parameters
----------
enable_fine_tuning
    Enables model fine-tuning. Only effective, if not `None`.
    If a bool, specifies whether to perform full fine tuning / training (all parameters are updated) or keep all parameters frozen.
    If a dict, specifies which parameters to fine tune. Must only contain one key-value record. Can be used to: 
    
    - Unfreeze specific parameters, while keeping everything else frozen: `{"unfreeze": ["patterns.to.freeze"]}`
    - Freeze specific parameters, while keeping everything else unfrozen: `{"freeze": ["patterns.to.freeze"]}`
    
    (TODO: add some words about the allowed patterns for parameter matching) 
"""
```

Like this, we can support fine-tuning in general for all our torch models (not only the foundation models), and move the parameter and handling into `TorchForecastingModel`.

And then we should add the param description to all torch models

darts/models/forecasting/foundation_model.py
```diff
@@ -173,8 +179,83 @@ def encode_year(idx):
...
+            else:
+                pl_trainer_kwargs = dict(pl_trainer_kwargs)
+
+            callbacks = pl_trainer_kwargs.get("callbacks")
+            if callbacks is None:
+                callbacks = []
+            else:
+                callbacks = list(callbacks)
+
+            callbacks.append(
```
dennisbader: As @daidahao mentioned, the parameter freezing/unfreezing could (should) be performed without the need of a callback. You could move that logic into a function that is called somewhere after `_create_model(...)`

darts/models/forecasting/foundation_model.py
dennisbader, line 1: For fine-tuning the foundation models, we should make sure that during training we use a `QuantileRegression(quantiles)` with all `quantiles` that the original weights were trained on.

The user should still be able specify some different quantiles when creating the model with `likelihood=QuantileRegression(other_quantiles)`. These quantile will only be used for prediction.

examples/26-Chronos-2-finetuning-examples.ipynb
```diff
@@ -0,0 +1,957 @@
+{
```
dennisbader: <p>Very nice notebook, thanks a lot <span class="mention" data-value="Kurokabe" data-denotation-char="@">﻿<span contenteditable="false">@Kurokabe</span>﻿</span> !</p><p> </p><p>Let's make sure we also update the notebook after having applied the code suggestions :)</p><p> </p><p>Here are some suggestions for the notebook:</p><ul><li>Since the new full/partial fine tuning support will be added to all torch models, we could rename it to Torch &amp; Foundation Model Fine-Tuning</li><li class="ql-indent-1">Foundation model fine-tuning can be one section</li><li class="ql-indent-1">Regular torch model fine-tuning can be another section</li><li class="ql-indent-2">Show the approach from <a href="https://unit8co.github.io/darts/userguide/torch_forecasting_models.html#re-training-or-fine-tuning-a-pre-trained-model" target="_blank">here</a> including the new fine tuning support</li><li>No need to compare the output of a loaded model to the original one</li><li>Remove the PEFT related parts</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/3003/discussion/#comment-2803345388'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='0'/>

dennisbader: <p>using <code>output_chunk_length=6</code> and forecast horizon <code>n=24</code> will perform auto-regression. Maybe it would be better to avoid this since we should focus on the output window that the model was fine-tuned on. You can use <code>output_chunk_length=12</code> and use a shorter val set</p><p> </p><p><code>train_passengers, val_passengers = data[:-12], data[-12:]</code>&nbsp;</p><p> </p><p>Also, I think it would be nice to use a QuantileRegression in the example, so we can show that the quantiles were also fine-tuned properly.</p><p> </p><p>Here's the model setup I used, which gave some nice results</p>

<pre class="ql-syntax" spellcheck="false">full_finetuned_model = Chronos2Model(
&nbsp; &nbsp; input_chunk_length=24,
&nbsp; &nbsp; output_chunk_length=12,
    use_reversible_instance_norm=True,
&nbsp; &nbsp; likelihood=QuantileRegression([0.1, 0.5, 0.9]),
&nbsp; &nbsp; enable_finetuning=True,
    random_state=42,
&nbsp; &nbsp; n_epochs=100,
)

# ... later predict with predict_likelihood_parameters=True
pred_full_finetuned = full_finetuned_model.predict(
&nbsp; &nbsp; n=len(val_passengers),
&nbsp; &nbsp; series=train_passengers,
    predict_likelihood_parameters=True,
)

# ... metrics can still be computed against the median
mape(data, pred_full_finetuned, q=0.5)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/unit8co/darts/pull/3003/discussion/#comment-2803345396'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

darts/models/forecasting/foundation_model.py
daidahao: @dennisbader 

I love this suggestion! Instead of enabling fine-tuning for foundation models only, we could enable it at the TFM level to benefit all torch models. This could deliver huge benefits to many training + fine-tuning pipelines.

dennisbader: Yep :)

dennisbader:
Looks very nice, thanks a lot @Kurokabe, this will be great 🚀 We're very close 🥳 

I pushed some changes:
- the main change was to make the user quantile loss and training loss independent for the foundation models
- some docs updates
- I've started some changes on the example notebook as well, but couldn't finish yet. Will have to wait then for when I'm back :)

dennisbader (approved):
Managed to finish the example notebook now 🚀 
- I shortened it a bit
- Added the recommended way to fine-tune a regular TFM (e.g. with `load_weights`)
- Only cover the partial fine-tuning one, as the "full" fine-tuning example can be easily derived from that.

Everything is ready now to be merged. Thanks a lot for the great work @Kurokabe 💯

Kurokabe:
That's awesome, thanks @dennisbader ! 🚀

PR #2802, opened 2025-05-04
Improve TorchForecastingModel prediction efficiency

dennisbader:
Checklist before merging this PR:
- [x] Mentioned all issues that this PR fixes or addresses.
- [x] Summarized the updates of this PR under **Summary**.
- [x] Added an entry under **Unreleased** in the [Changelog](../CHANGELOG.md).

Fixes #2780.

### Summary

- Drastically improves prediction efficiency of `TorchForecastingModel`.
  - Moved forecast TimeSeries creation out of the batch prediction process and instead apply it once on the total predictions. Using multi-processing (`n_jobs>1`) created a huge overhead when performing it on the batch-level
  - Using multi-processing (`n_jobs>1`) now boosts the forecast `TimeSeries` creation proportionally!
  - Instead of passing the entire `TimeSeries` as part of the prediction batches, we now only pass the schema of the `TimeSeries`.
- Added parameter `values_only: bool = True` to `TorchForecastingModel.predict_from_dataset()`:
  - Whether to return the predicted values only. If `False`, will return `TimeSeries` objects. Otherwise, will
     return a tuple of `(np.ndarray, list[dict[str, Any]], list[Union[pd.Timestamp, int]])`. The first element
     represents the predictions with shape `(num_predictions, n, columns, num_samples)`. The second element
     represents the schemas of forecasted target `TimeSeries`. The third element represents the prediction start
     times.
- Added `TimeSeries.schema(copy: bool = True)` to extract the schema from a series. This allows to create new series with identical schema.

Some comparisons:

| process | boost (time old / time new)|
|---------|----------------------|
| predict() or historical_forecast() (last_points_only=False) n_jobs=1 | 1.2 |
| predict() or historical_forecast() (last_points_only=False) n_jobs=-1 | 1.9 |
| historical_forecast() (last_points_only=True) n_jobs=1 | 5.4 |
| historical_forecast() (last_points_only=True) n_jobs=-1 | 5.4 |

PR #3021, opened 2026-02-18
Rethinking Typing for Darts

mizeller:

- [x] Summarized the updates of this PR under Summary.
- [x] Added an entry under Unreleased in the [Changelog](https://github.com/unit8co/darts/CHANGELOG.md).

Addresses no particular issue.

### Summary

This PR introduces two new types, `TimeIndex` and `TimeSeriesLike`, to make the code a bit more readable, maintainable & compatible with modern type checkers. 

In the future, common `TypeAliases` can be defined at a centralised location, i.e. `darts/typing.py`, and used throughout the whole code-base. 

### Other Information

Example: See image. 

image (8)


mizeller:
- `TimeIndex` → `pd.DatetimeIndex | pd.RangeIndex`

- `TimeSeriesLike` → `TimeSeries | Sequence[TimeSeries]` (inspired by `np.typing.ArrayLike`)

- added optional None default type everywhere we have `None` default value → `... | None = None`  (essentially to get rid of most [invalid-parameter-default](https://docs.astral.sh/ty/reference/rules/#invalid-parameter-default) errors)

- added ty to pyproject + some useful defaults to typecheck the codebase efficiently

```
$ ty check | wc -l
2038
```

mizeller:
I initially started the `feat: TimeSeriesSchema` commit because I wanted to get rid of the  `Any` type annotation everywhere...turned out a bit more complex than expected; especially for the `schema: dict[str, Any]` annotations.

Generally, I found the output tuples of the `TorchForecastingModel` and the associated tests pretty hard to interpret: 

```
self._assert_eq(ds[1][1:], (None, None, None, None, self.cov_st2, self.target2.schema()))        
``` 

PS: This commit essentially extends #2802

mizeller:
Turns out creating `TimeSeriesSchema` objects creates some massive overhead in terms of performance compared to using the previous dict-based method (~2x slower, no matter the size n_rows or n_Iter).

I'll revert these commits for now.

darts/typing.py
```diff
@@ -0,0 +1,18 @@
+"""DARTS TYPES
+
+tldr;   This file defines reusable types for the Darts Eco-System.
```
dennisbader: This will be visible in our documentation

```suggestion
"""Darts types
```

dennisbader (changes requested):
Looks great thanks @mizeller 🚀 

I added one comment. Can you also update the PR description to reflect the current state of this PR (e.g. remove the `ty` related things since it wasn't added here)?

PR #3023, opened 2026-02-19
Feature: Allow specifying a subset of columns to apply the data transformer on

### LINKED ISSUE #3020: [Feature] DataTransformer: Allow specifying a subset of columns to apply the data transformer on
Currently our data transformers (except StaticCovariatesTransformer) do not allow selecting only a subset of columns to apply the transformation on.

We should add a new parameter `columns` (or other name) to DataTransformer creation that specifies the subset of TimeSeries columns / components to apply the transformer on.

Why would this be useful?

For example in a `Pipeline` we would like to apply multiple transformers sequentially, but not all of them should act on the same columns.

We already have the logic pretty much covered by `component_mask` in methods `fit()`, `transform()`, ...
It would be more user (and API) friendly, if we specified this during transformer creation with the new parameter.

u8-krpeczek:

Fixes #3020

### Summary

This PR introduces a user-friendly `columns` argument to the data transformer base classes and out-of-the-box data transformers. It allows users to specify a subset of string column names to transform in a multivariate TimeSeries, leaving the unselected components unmodified.

### Other Information

Behaviour:
  - `columns=["A", "B"]`: Transforms only components "A" and "B".
 - `columns=[]`: Modifies nothing.
 - `columns=None`: Transforms all components (preserves default historical behaviour).

Interaction with `component_mask` and `mask_components`:
 - If `columns` is specified during transform creation passing a `component_mask` will raise an error.
 - If `columns` is left as `None` (default), `component_mask` can be passed. This should preserve historical behaviour. 
 - If `columns` is specified, but `mask_components` is set to `False`, the transformer constructor will raise an exception. 

Tests:
 - Checks for default behaviour.
 - Checks for different subsets of columns.

### DIFF
--- a/darts/dataprocessing/transformers/base_data_transformer.py
+++ b/darts/dataprocessing/transformers/base_data_transformer.py
@@ -180,15 +180,6 @@ class BaseDataTransformer(ABC):
 
         self._mask_components = mask_components
         self._columns = [columns] if isinstance(columns, str) else columns
-        if self._columns is not None and not self._mask_components:
-            raise_log(
-                ValueError(
-                    "Contradictory arguments: `columns` was provided, but `mask_components` "
-                    "is set to False. If you want to transform specific columns, "
-                    "`mask_components` must be True."
-                ),
-                logger=logger,
-            )
 
     def set_verbose(self, value: bool):
         """Set the verbosity status.
--- a/darts/dataprocessing/transformers/diff.py
+++ b/darts/dataprocessing/transformers/diff.py
@@ -28,6 +28,7 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
         name: str = "Diff",
         n_jobs: int = 1,
         verbose: bool = False,
+        columns: str | list[str] | None = None,
     ):
         r"""Differencing data transformer.
 
@@ -74,6 +75,11 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
             required amount of time.
         verbose
             Whether to print operations progress
+        columns
+            Optionally, a string or list of strings specifying the names of the components (columns)
+            to transform. If specified, only these components will be transformed, and the remaining
+            components will be kept untouched. For more information refer to the `BaseDataTransformer`
+            documentation.
 
         Examples
         --------
@@ -115,11 +121,12 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
             n_jobs=n_jobs,
             verbose=verbose,
             mask_components=False,
+            columns=columns,
         )
 
     @staticmethod
     def ts_fit(series: TimeSeries, params: Mapping[str, Any], **kwargs) -> Any:
-        lags, dropna = params["fixed"]["_lags"], params["fixed"]["_dropna"]
+        lags, _ = params["fixed"]["_lags"], params["fixed"]["_dropna"]
         lags_sum = sum(lags)
         if series.n_timesteps <= lags_sum:
             raise_log(
@@ -130,16 +137,20 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
                 ),
                 logger,
             )
-        component_mask = Diff._get_component_mask(kwargs, dropna)
-        vals = Diff.apply_component_mask(series, component_mask, return_ts=False).copy()
-        # First `lags_sum` values of time series will be 'lost' due to differencing;
-        # need to remember these values to 'undifference':
-        start_vals = vals[:lags_sum, :, :]
-        diffed = start_vals
+        component_mask = kwargs.get("component_mask")
+
+        # store start values for later undifferencing (inverse transform)
+        # `start_vals` store all columns, also the ones that are not diffed
+        start_vals = series.all_values(copy=True)[:lags_sum]
+        # `diffed` focus on columns that are actually diffed; first `lags_sum` values of time series will be 'lost'
+        diffed = Diff.apply_component_mask(series, component_mask, return_ts=False)[
+            :lags_sum
+        ]
         cutoff = 0
+        component_mask_ = component_mask if component_mask is not None else slice(None)
         for lag in lags:
             # Store first `lag` values of current differencing step:
-            start_vals[cutoff:, :, :] = diffed
+            start_vals[cutoff:, component_mask_, :] = diffed
             diffed = diffed[lag:, :, :] - diffed[:-lag, :, :]
             cutoff += lag
         return start_vals, component_mask, series.start_time(), series.freq
@@ -149,7 +160,7 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
         series: TimeSeries, params: Mapping[str, Any], **kwargs
     ) -> TimeSeries:
         lags, dropna = params["fixed"]["_lags"], params["fixed"]["_dropna"]
-        component_mask = Diff._get_component_mask(kwargs, dropna)
+        component_mask = kwargs.get("component_mask")
         diffed = Diff.apply_component_mask(series, component_mask, return_ts=True)
         for lag in lags:
             diffed = diffed.diff(n=1, periods=lag, dropna=dropna)
@@ -184,7 +195,7 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
                 ),
                 logger,
             )
-        component_mask = Diff._get_component_mask(kwargs, dropna)
+        component_mask = kwargs.get("component_mask")
         if np.any(fit_component_mask != component_mask):
             raise_log(
                 ValueError(
@@ -194,10 +205,21 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
                 logger,
             )
         if dropna:
-            nan_shape = (sum(lags), series.n_components, series.n_samples)
-            nan_vals = np.full(nan_shape, fill_value=np.nan)
-            series = series.prepend_values(nan_vals)
+            start_shape = (sum(lags), series.n_components, series.n_samples)
+            start_fill_vals = np.full(start_shape, fill_value=np.nan)
+
+            # fill start values with components that were not diffed
+            if component_mask is not None:
+                start_fill_vals[:, ~component_mask, :] = start_vals[
+                    :, ~component_mask, :
+                ]
+            series = series.prepend_values(start_fill_vals)
+
+        # from this point, only look at columns that were actually diffed
         vals = Diff.apply_component_mask(series, component_mask, return_ts=False).copy()
+        if component_mask is not None:
+            start_vals = start_vals[:, component_mask, :]
+
         if vals.shape[1] != start_vals.shape[1]:
             raise_log(
                 ValueError(
@@ -230,17 +252,3 @@ class Diff(FittableDataTransformer, InvertibleDataTransformer):
             copy=False,
             **series._attrs,
         )
-
-    @staticmethod
-    def _get_component_mask(kwargs, dropna):
-        component_mask = kwargs.get("component_mask", None)
-        if dropna and (component_mask is not None):
-            raise_log(
-                ValueError(
-                    "Cannot specify `component_mask` with `dropna = True`, "
-                    "since differenced and undifferenced components will be "
-                    "of different lengths."
-                ),
-                logger,
-            )
-        return component_mask
--- a/darts/tests/dataprocessing/transformers/test_base_data_transformer.py
+++ b/darts/tests/dataprocessing/transformers/test_base_data_transformer.py
@@ -151,23 +151,6 @@ class TestBaseDataTransformer:
 
         assert test_input == test_input_copy
 
-    def test_columns_validation(self):
-        """
-        Tests that string columns are converted to lists, and that
-        passing a contradictory `mask_components=False` and `columns` argument raises a ValueError.
-        """
-        mock_string = self.DataTransformerMock(scale=2, translation=2, columns="A")
-        assert mock_string._columns == ["A"]
-
-        with pytest.raises(ValueError) as exc_info:
-            # Should raise a ValueError warning the user that `mask_compoents` needs to be `True`
-            # for the `columns` argument to work.
-            self.DataTransformerMock(
-                scale=2, translation=2, columns=["A"], mask_components=False
-            )
-
-        assert "Contradictory arguments" in str(exc_info.value)
-
     def test_generate_component_mask(self):
         """
         Test if all behaviours of `_generate_component_mask` are correct.
--- a/darts/tests/dataprocessing/transformers/test_invertible_fittable_data_transformer.py
+++ b/darts/tests/dataprocessing/transformers/test_invertible_fittable_data_transformer.py
@@ -222,7 +222,7 @@ class TestLocalFittableInvertibleDataTransformer:
 
         transformed = mock.fit_transform(test_input)
 
-        if "A" in col_names:
+        if col_names is None or "A" in col_names:
             assert transformed["A"] == constant_timeseries(
                 value=12, length=10, column_name="A"
             )
@@ -231,7 +231,7 @@ class TestLocalFittableInvertibleDataTransformer:
                 value=1, length=10, column_name="A"
             )
 
-        if "B" in col_names:
+        if col_names is None or "B" in col_names:
             assert transformed["B"] == constant_timeseries(
                 value=14, length=10, column_name="B"
             )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
-- Added a `columns` parameter to data transformers to allow specifying a subset of columns to apply the transformation on. [#3023](https://github.com/unit8co/darts/pull/3023) by [Krzsztof Pęczek](https://github.com/u8-krpeczek)
+- Added parameter `columns` to all data transformers (e.g. `Scaler`, `Diff`, `BoxCox`, ...) to apply the transformations only on subset of components. This makes it much easier to build multivariate pipelines where different components require different transformations. [#3023](https://github.com/unit8co/darts/pull/3023) by [Krzsztof Pęczek](https://github.com/u8-krpeczek)
 - `TFTExplainer` plotting improvements: [#3039](https://github.com/unit8co/darts/pull/3039) by [ReinerBRO](https://github.com/ReinerBRO).
   - `plot_variable_selection()` now returns the matplotlib figures for downstream usage (saving, editing, ...). It returns a single figure when explaining a single TimeSeries. Otherwise, it returns a list of figures.
   - `plot_variable_selection()` now accepts a `show_plot: bool = True` parameter that allows to suppress showing the plot.
--- a/darts/tests/dataprocessing/test_pipeline.py
+++ b/darts/tests/dataprocessing/test_pipeline.py
@@ -12,6 +12,7 @@ from darts.dataprocessing.transformers import (
     InvertibleDataTransformer,
     InvertibleMapper,
     Mapper,
+    Scaler,
 )
 from darts.typing import TimeSeriesLike
 from darts.utils.timeseries_generation import constant_timeseries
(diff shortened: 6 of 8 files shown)


darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -167,11 +168,19 @@ def __init__(
...
         self._verbose = verbose
         self._n_jobs = n_jobs
 
+        self._mask_components = mask_components
+        self._columns = [columns] if isinstance(columns, str) else columns
+        if self._columns is not None and not self._mask_components:
+            raise_log(
+                "Contradictory arguments: `columns` was provided, but `mask_components` "
+                "is set to False. If you want to transform specific columns, "
+                "`mask_components` must be True."
```
dennisbader: `raise_log` expects an input of type `Exception`, so it should look like below. Let's also update the other similar changes


```suggestion
                ValueError(
                    "Contradictory arguments: `columns` was provided, but `mask_components` "
                    "is set to False. If you want to transform specific columns, "
                    "`mask_components` must be True."
                )
```

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         verbose: bool = False,
         parallel_params: bool | Sequence[str] = False,
         mask_components: bool = True,
+        columns: str | list[str] | None = None,
```
dennisbader, line 68: The parameter description is missing here and in other files. You can add them to the list below

jakubchlapek, line 68: missing param in the docstring for columns with the mask_components/columns interaction

the same case for the transformer subclasses, but I won't mark each one to not clutter too much

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -168,11 +169,19 @@ def __init__(
...
-        self._mask_components = mask_components
         self._name = name
         self._verbose = verbose
         self._n_jobs = n_jobs
 
+        self._mask_components = mask_components
+        self._columns = [columns] if isinstance(columns, str) else columns
+        if self._columns is not None and not self._mask_components:
+            raise_log(
+                "Contradictory arguments: `columns` was provided, but `mask_components` "
```
jakubchlapek: raise_log should accepts an Exception argument instead of string
also while the main logger is inferred in the method, it'd be good to pass the logger var we have for consistency with other logs
e.g.
```
raise_log(ValueError("error message"), logger)
```
i won't mark each raise_log to not clutter it too much, but feel free to adapt them accordingly :)

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -444,6 +464,30 @@ def _process_series_idx(series_idx: int | Sequence[int]) -> Sequence[int]:
...
+        for col in columns:
+            if col not in series_components:
+                raise_log(
+                    f"Column '{col}' specified in `columns` was not found in the "
+                    f"TimeSeries components: {series_components}"
+                )
+            idx = series_components.index(col)
+            mask[idx] = True
+
+        return mask
```
dennisbader: We could simplify this a bit (`columns` can only be list[str] when this function is invoked). And we can give the user the complete set of invalid columns with below



```suggestion
    @staticmethod
    def _generate_component_mask(series: TimeSeries, columns: list[str]) -> np.ndarray:
        """
        Translates the `self._columns` list into a boolean numpy array mask.
        """
        mask = series.columns.isin(columns)
        if mask.sum() != len(columns):
            raise_log(
                ValueError(
                    f"Columns {set(columns) - set(series.columns)} specified in "
                    f"`columns` do not exist in the `TimeSeries` components: "
                    f"{series.columns}"
                )
            )
        return mask
```

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -367,6 +376,17 @@ def transform(
             total=len(data),
         )
 
+        if self._columns is not None and component_mask is not None:
+            raise_log(
+                "You cannot use the `columns` parameter"
+                "and pass a `component_mask` to `transform()` at the same time."
+            )
```
jakubchlapek: example of the logging with suggestions, also was missing a spacer after the 'parameter' haha

```suggestion
        if self._columns is not None and component_mask is not None:
            raise_log(
                ValueError(
                    "You cannot use the `columns` parameter "
                    "and pass a `component_mask` to `transform()` at the same time."
                ),
                logger=logger,
            )
```

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -444,6 +464,30 @@ def _process_series_idx(series_idx: int | Sequence[int]) -> Sequence[int]:
...
+        """
+        Translates the `self._columns` list into a boolean numpy array mask.
+        """
+        if columns is None:
+            return None
+
+        mask = np.zeros(series.n_components, dtype=bool)
+        series_components = series.components.tolist()
+
+        for col in columns:
```
jakubchlapek: currently this would report the first missing column and raise an error.
i think it would be nicer to show the user all the missing columns, and additionally we can then simplify the code a bit
```suggestion
        missing = [col for col in columns if col not in series.components]
        if missing:
            raise_log(
                ValueError(
                    f"Columns {missing} specified in `columns` were not found in the "
                    f"TimeSeries components: {series.components.tolist()}"
                ),
                logger=logger,
            )
        return series.components.isin(columns)
```

darts/dataprocessing/transformers/invertible_data_transformer.py
```diff
@@ -332,6 +334,17 @@ def inverse_transform(
...
+        if self._columns is not None and component_mask is not None:
+            raise_log(
+                "You cannot use the `columns` parameter"
+                "and pass a `component_mask` to `inverse_transform()` at the same time."
+            )
+
+        if self._columns is not None:
+            component_mask = BaseDataTransformer._generate_component_mask(
+                data[0], self._columns
+            )
```
jakubchlapek: would be nice to comment why we take `data[0]` here as I think it's not self-explanatory from the code e.g. for multi-series multivariate case we assume all subsequent time series to have the same column order, but let me know what you think on that

darts/dataprocessing/transformers/base_data_transformer.py
dennisbader: We could simplify this a bit to reduce code duplication across the transformer base classes.
Let's handle all the component mask / columns logic in here.

```suggestion
    @staticmethod
    def _generate_component_mask(
        series: TimeSeries,
        component_mask: np.ndarray | None,
        columns: list[str] | None,
    ) -> np.ndarray | None:
        """Returns the existing `component_mask` or generates a new component mask based on `columns`."""
        if columns is not None and component_mask is not None:
            raise_log(
                ValueError(
                    "Cannot pass `columns` and `component_mask` at the same time."
                )
            )

        if columns is None:
            return component_mask

        component_mask = series.columns.isin(columns)
        if component_mask.sum() != len(columns):
            raise_log(
                ValueError(
                    f"Columns {set(columns) - set(series.columns)} specified in "
                    f"`columns` do not exist in the `TimeSeries` components: "
                    f"{series.columns}"
                )
            )
        return component_mask
```

Now we can replace the diffs in `transform`, `fit`, ... with a single call:

```
        component_mask = BaseDataTransformer._generate_component_mask(
            series=data[0],
            component_mask=component_mask,
            columns=self._columns,
        )
```

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
```diff
@@ -114,6 +116,64 @@ def test_input_transformed_single_series(self, component_mask):
...
+
+        transformed = mock.transform(test_input)
+
+        assert transformed["A"] == constant_timeseries(
+            value=12, length=10, column_name="A"
+        )
+        assert transformed["B"] == constant_timeseries(
+            value=14, length=10, column_name="B"
+        )
+        assert test_input == test_input_copy
```
dennisbader: We could combine the tests into a single one, by adding `None` to the parameters (and we should also check for string "A")

```suggestion
    @pytest.mark.parametrize("col_names", ["A", ["B"], ["A", "B"], None])
    def test_columns_subset(self, col_names):
        """
        Tests if the `columns` argument correctly applies the transform only to the specified columns.
        """
        ts_a = constant_timeseries(value=1, length=10, column_name="A")
        ts_b = constant_timeseries(value=2, length=10, column_name="B")

        test_input = ts_a.stack(ts_b)
        test_input_copy = test_input.copy()

        mock = self.DataTransformerMock(scale=2, translation=10, columns=col_names)

        transformed = mock.transform(test_input)

        if col_names is None or "A" in col_names:
            assert transformed["A"] == constant_timeseries(
                value=12, length=10, column_name="A"
            )
        else:
            assert transformed["A"] == ts_a

        if col_names is None or "B" in col_names:
            assert transformed["B"] == constant_timeseries(
                value=14, length=10, column_name="B"
            )
        else:
            assert transformed["B"] == ts_b

        assert test_input == test_input_copy
```

darts/tests/dataprocessing/transformers/test_fittable_data_transformer.py
```diff
@@ -142,6 +144,41 @@ def test_input_transformed_single_series(self, component_mask):
         assert transformed == expected
         assert test_input == test_input_copy
 
+    @pytest.mark.parametrize("col_names", [["A"], ["B"], ["A", "B"]])
```
dennisbader: you can apply the same here and in the other test files :)

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
dennisbader, line 1: Very nice tests 🚀  :) 

There are some lines in your code that aren't covered in the unit tests. You can check out the `codecov` report that was automatically added as a comment in the PR conversation.

Would be good to add some tests that covers those lines. I assume it's about the Exceptions that should be raised

darts/dataprocessing/transformers/diff.py
```diff
@@ -111,7 +112,11 @@ def __init__(
         # Don't automatically apply `component_mask` - need to throw error when `dropna = True`
         # and `component_mask` is specified:
         super().__init__(
-            name=name, n_jobs=n_jobs, verbose=verbose, mask_components=False
+            name=name,
+            n_jobs=n_jobs,
+            verbose=verbose,
+            mask_components=False,
+            columns=columns,
```
jakubchlapek, line 124: since Diff passes `mask_components=False` the `columns` param here is dead, as if it's provided the Transformer will raise an exception anyway, and if it's not then it's historical behavior. 

as both MIDAS and Diff transformer would change the number of steps, i think it might be better to disable the `columns` param for them, as they won't support some components being shorter than other.

let me know what you think on this, as I think being able to difference only the chosen cols would be a really nice feature, but as of now I'm unsure how to handle it

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
```diff
@@ -18,6 +18,7 @@ def __init__(
             stack_samples: bool = False,
             mask_components: bool = True,
             parallel_params: bool | Sequence[str] = False,
+            columns: str | list[str] | None = None,
```
jakubchlapek, line 21: also docstring updates for all the test classes :)

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
```diff
@@ -114,6 +116,64 @@ def test_input_transformed_single_series(self, component_mask):
         assert transformed == expected
         assert test_input == test_input_copy
 
+    @pytest.mark.parametrize("col_names", [["A"], ["B"], ["A", "B"]])
```
jakubchlapek: the tests look good to me, but it'd be nice to also add one more with different error paths as they're not covered yet

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: Got it. Fixed in [db9d868](https://github.com/unit8co/darts/pull/3023/commits/db9d8686aeed0ff73c6e1abbc539fe40d00211a5)

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: Good catch, I've added it in [188fc4c](https://github.com/unit8co/darts/pull/3023/commits/188fc4c5f5947b39f6c461835486c2b7e5f8a4b7)

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: Oh yes, that true. I've no idea how I didn't think of that. Resolved in [188fc4c](https://github.com/unit8co/darts/pull/3023/commits/188fc4c5f5947b39f6c461835486c2b7e5f8a4b7) 😄

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
u8-krpeczek: Sure, added in [e4ae821](https://github.com/unit8co/darts/pull/3023/commits/e4ae82109aacf54165a03ee1b4e852910378c3e4).

darts/tests/dataprocessing/transformers/test_fittable_data_transformer.py
u8-krpeczek: Also resolved in [e4ae821](https://github.com/unit8co/darts/pull/3023/commits/e4ae82109aacf54165a03ee1b4e852910378c3e4).

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
u8-krpeczek: Sure, I've added them in [e4ae821](https://github.com/unit8co/darts/pull/3023/commits/e4ae82109aacf54165a03ee1b4e852910378c3e4).

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
u8-krpeczek: Added some more docstrings in [436f9b7](https://github.com/unit8co/darts/pull/3023/commits/436f9b711cc4439aab6ae67d85d34bef56276393).

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: I've added a note for the `mask_components` - `columns` interaction and docstrings in general in [436f9b7](https://github.com/unit8co/darts/pull/3023/commits/436f9b711cc4439aab6ae67d85d34bef56276393).

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: Solved in [8bf2b3d](https://github.com/unit8co/darts/pull/3023/commits/8bf2b3d032d5a5c1d8c4d6952c261b41d3921587)

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: Also solved in [8bf2b3d](https://github.com/unit8co/darts/pull/3023/commits/8bf2b3d032d5a5c1d8c4d6952c261b41d3921587)

darts/dataprocessing/transformers/base_data_transformer.py
u8-krpeczek: Added in [436f9b7](https://github.com/unit8co/darts/pull/3023/commits/436f9b711cc4439aab6ae67d85d34bef56276393).

darts/dataprocessing/transformers/invertible_data_transformer.py
u8-krpeczek: Good point. I felt that the assumption is reasonable and that doing checks would be an overkill. I've added an inline comment in [112d153](https://github.com/unit8co/darts/pull/3023/commits/112d153656d6feefa4e609d1828a4ebde100de58). Hopefully that's a good enough explanation to users.

darts/dataprocessing/transformers/diff.py
u8-krpeczek: I think you're right. I've removed `columns` from both `Diff` and `MIDAS` in [7c3312d](https://github.com/unit8co/darts/pull/3023/commits/7c3312d9e470982cb9042e8dc5f64ccff7e42464). If we left it in there it would just fail and I can't really come up with a scenario where someone would want to use `columns` with these two transformers.

u8-krpeczek: It seems that `WindowTransformer` also has this functionality built in with the `components` argument. I removed the redundant `columns` argument here [5810c6a](https://github.com/unit8co/darts/pull/3023/commits/5810c6aaf33dae5765f9bf1662df4e8b3b179a6a).

darts/tests/dataprocessing/transformers/test_invertible_fittable_data_transformer.py
```diff
@@ -204,6 +206,43 @@ def test_input_transformed_single_series(self, component_mask):
...
+            assert transformed["B"] == constant_timeseries(
+                value=14, length=10, column_name="B"
+            )
+        else:
+            assert transformed["B"] == constant_timeseries(
+                value=2, length=10, column_name="B"
+            )
+
+        assert test_input == test_input_copy
+        assert test_input == test_input_copy
```
jakubchlapek, line 244: duplicate assert, would be good to check if the .inverse_transform works here, analogous to the test_invertible_data_transformer

darts/tests/dataprocessing/transformers/test_invertible_fittable_data_transformer.py
```diff
@@ -204,6 +206,43 @@ def test_input_transformed_single_series(self, component_mask):
         )
         assert transformed == transformed_copy
 
+    @pytest.mark.parametrize("col_names", [["A"], ["B"], ["A", "B"]])
```
jakubchlapek: is there a reason to use different parametrize values compared to other tests? I sugggest also checking the single string cases and `None` cases, also with the `[]` case as it's mentioned in docs, but not tested yet

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -332,6 +349,13 @@ def transform(
         does *not* change the number of timesteps in each series; if this were to happen, then the transformed
         and untransformed components are unable to be concatenated back together along the component axis.
 
+        If `mask_components` was set to `False` when instantiating `BaseDataTransformer`, then any provided
```
jakubchlapek, line 343: duplicate docstring here (the `If 'mask_components' was set to...` part)

darts/dataprocessing/transformers/diff.py
jakubchlapek: While I think that it would be nice to be able to differentiate only given columns I agree that it would be the best course here to remove the param, good catch regarding the WindowTransformer.

darts/tests/dataprocessing/transformers/test_base_data_transformer.py
```diff
@@ -114,6 +116,98 @@ def test_input_transformed_single_series(self, component_mask):
...
+            assert transformed["B"] == constant_timeseries(
+                value=2, length=10, column_name="B"
+            )
+
+        assert test_input == test_input_copy
+
+    def test_columns_validation(self):
+        """
+        Tests that string columns are converted to lists, and that
+        passing a contradictory `mask_components=False` and zcolumns` argument raises a ValueError.
```
jakubchlapek: typo
```suggestion
        passing a contradictory `mask_components=False` and `columns` argument raises a ValueError.
```

dennisbader (approved):
Thanks for the great PR @u8-krpeczek 🚀 

I pushed some changes that to also support `columns` for the `Diff` transformer :)

Everything is ready now. Waiting for the tests to complete, then I'll merge.

Issue #2758, opened 2025-04-03
[FEAT] RIN for covariates


turbotimon:
**Is your feature request related to a current problem? Please describe.**

Currently, the RIN () is only done for the target series (n_targets). However, it would also be beneficial to past and future covariates in most cases (e.g. if they are leading indicators). An exception are maybe temporal or positional features (e.g. day/month/year) as added by `add_encoders` / `datetime_attribute_timeseries()`

**Describe proposed solution**

To give the user the most choice, I propose allowing the following for the `use_reversible_instance_norm` attribute:

- A single Bool e.g. `True`: Target(s) only. As it is now (ensuring backwards compatibility)
- A tuple of Bools e.g. `(True, True, False)` which is interpreted as `(target, past, future)` therefore apply RIN for target(s) and past covariates, but not future covariats in the example.
- Optionally, allowing also tuple of Boosl in the tuple e.g. `(True, (True, False), False)` which gives even more fine control over which component will be normalized. The example would mean:
   - `True` for all targets
   - `True` for the first past covariate and `False` for the second (Tuple length must match number of components)
   - `False` for all future covariates

**Describe potential alternatives**

The logic could also be added  as use_reversible_instance_norm kwargs proposed in https://github.com/unit8co/darts/issues/2748

```py
use_reversible_instance_norm = {
   "series" = True # Enable for all series components 
   "past_covariates" =  (True, True, False) # Enable for the first two, but not the third. Tuple length must match number of components
   "future_covariats" = False # Disable for all past cov
}
```

**Additional context**

None

turbotimon:
Yes, agree with specifying the component name would probably better more robust (less error-prone).

```py
use_reversible_instance_norm = {
   "series" = True # Enable for all series components 
   "past_covariates" =  {
       "comp_name": True,
   } # Enable for the component named "comp_name". All others are False by default
   "future_covariats" = False # Explicit disable for all past cov. False would be the default if not explicit specified for all (series, past, future)
}
```

dennisbader:
I could imagine something like this:

```
use_reversible_instance_norm = {
   "series" = True  # Enable for all series components 
   "past_covariates" =  ["comp1", "compx"],  # enable only for the given component names. All others are False by default
   "future_covariates" = False  # Explicit disable for all past cov (default)
}
```

Let me know what you think.

FourierMourier:
[...]

PR #3029, opened 2026-02-23
[Feature] Allow custom RINorm kwargs & fix foundation compatibility

### LINKED ISSUE #2748: [FEAT] Expose RINorm kwargs
This would largely address [#2739 - RIN is not scale invariant](https://github.com/unit8co/darts/issues/2739)

Exposing the RINorm hyperparameters would allow users more flexibility. Currently there can be unexpected behavior for samples with variance ~1E-5 -- low, but not so low as to be unrealistic. RINorm as reported will inevitably behave oddly with low variance, but epsilon is a key hyperparameter in this case and should be exposed. 

Also, the affine transformation is unlikely to be beneficial with N-Linear and D-Linear models and we should allow users to disable it. 

Finally, the default hyperparameters should be documented (they match the original implementation but may be unintuitive)

This solution could be made most simply by accepting a dictionary as an argument for the `use_reversible_instance_norm` kwargs in all `PLForecastingModule`s, eg

```
        if use_reversible_instance_norm and isinstance(use_reversible_instance_norm, dict):
            self.rin = RINorm(input_dim=self.n_targets, **use_reversible_instance_norm)
        elif use_reversible_instance_norm:
            self.rin = RINorm(input_dim=self.n_targets)
        else:
            self.rin = None
```

Nothing comes to mind, other than leaving things as-is.

### LINKED ISSUE #3028: [BUG] RINorm incompatible with foundation models
**Describe the bug**
RINorm cannot be used with foundation models like Chronos-2 and TimesFM 2.5.

**To Reproduce**

```py
from darts.datasets import WeatherDataset
from darts.models import TimesFM2p5Model
# load data in float32 format (macOS issues with float64 and PyTorch)
series = WeatherDataset().load().astype("float32")
# predicting atmospheric pressure
target = series['p (mbar)'][:100]
# by default, TimesFM2p5Model is deterministic; to enable probabilistic forecasts,
# set likelihood to QuantileRegression and use a subset of the pre-trained quantiles
model = TimesFM2p5Model(
    input_chunk_length=6,
    output_chunk_length=6,
    use_reversible_instance_norm=True,
)
# calling fit is still mandatory to ensure consistent number of components; however,
# TimesFM2p5Model is training-free and the model weights are not updated
model.fit(target)
# when TimesFM2p5Model is probabilistic, set ``predict_likelihood_parameters=True``
# or ``num_samples>>1`` to get meaningful results
pred = model.predict(6)
print(pred.all_values())
```

```
RuntimeError: Error(s) in loading state_dict for _TimesFM2p5Module:
	Missing key(s) in state_dict: "rin.affine_weight", "rin.affine_bias". 
```

**Expected behavior**

 - Python version: 3.13.
 - darts version: latest from source.

**Additional context**

daidahao:

Fixes #3028, fixes #2748.

### Summary
- All torch models now accept `use_reversible_instance_norm` as a flexible parameter that can be set to `True`, `False`, or a dictionary of `RINorm` hyperparameters. This allows users to customize `RINorm` for their models. 
- Disallowed `use_reversible_instance_norm=True` or `{"affine": True}` for foundation models to prevent checkpoint loading errors due to incompatible weights.

### Other Information

#2758 is not being fixed in this PR.

### DIFF
--- a/darts/models/forecasting/block_rnn_model.py
+++ b/darts/models/forecasting/block_rnn_model.py
@@ -348,7 +348,9 @@ class BlockRNNModel(MixedCovariatesTorchModel):
             Optionally, some keyword arguments for the PyTorch learning rate scheduler. Default: ``None``.
         use_reversible_instance_norm
             Whether to use reversible instance normalization `RINorm` against distribution shift as shown in [1]_.
-            It is only applied to the features of the target series and not the covariates.
+            It is only applied to the features of the target series and not the covariates. If ``True``,
+            applies ``RINorm`` with default hyperparameters. If a dictionary, defines the hyperparameters to construct
+            the ``RINorm``. Supported parameters are ``{"affine": bool, "eps": float}``. Default: ``False``.
         batch_size
             Number of time series (input and output sequences) used in each training pass. Default: ``32``.
         n_epochs
--- a/darts/models/forecasting/dlinear.py
+++ b/darts/models/forecasting/dlinear.py
@@ -305,7 +305,9 @@ class DLinearModel(MixedCovariatesTorchModel):
             Optionally, some keyword arguments for the PyTorch learning rate scheduler. Default: ``None``.
         use_reversible_instance_norm
             Whether to use reversible instance normalization `RINorm` against distribution shift as shown in [2]_.
-            It is only applied to the features of the target series and not the covariates.
+            It is only applied to the features of the target series and not the covariates. If ``True``,
+            applies ``RINorm`` with default hyperparameters. If a dictionary, defines the hyperparameters to construct
+            the ``RINorm``. Supported parameters are ``{"affine": bool, "eps": float}``. Default: ``False``.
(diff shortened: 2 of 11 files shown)


daidahao:
@dennisbader 

> This would be a quick fix while still allowing the foundation models to use RINorm (could be good for fine-tuning).

Personally, I would not recommend using RINorm for prediction or fine-tuning since almost all foundation models have their own RINorm. That's why I have now removed any reference to RINorm in Chronos-2 and TimesFM. Nevertheless, in `FoundationModel`, `use_reversible_instance_norm` is being handled properly to produce warnings and override `"affine"` key to `False`.

For all torch models, I've updated `PLForecastingModule` so that it now accepts `use_reversible_instance_norm` as either a `bool` or `dict` of hyperparameters. One question remains, should we update all torch model doctorings to highlight this new feature?

Let me know what you think.

PR #3032, opened 2026-02-27
[Feature] CatBoost MultiQuantile support

### LINKED ISSUE #2967: Use CatBoost MultiQuantile regression

CatBoost has the ability to regress multiple quantiles using a single model, with the `MultiQuantile` loss: https://catboost.ai/docs/en/concepts/loss-functions-regression#MultiQuantile

However, it seems that feature is not used by Darts: https://github.com/unit8co/darts/blob/master/darts/models/forecasting/catboost_model.py#L352

It might make the use of multiple quantiles much faster by avoiding the training of multiple models.

It seems relatively straightforward to use `MultiQuantile` instead of `Quantile`. I'm happy to contribute a PR for this. The questions I would like to clarify are:
- how does this interact with multi-output regression, either for multiple components, or in the case of `multi_models=True` with `output_chunk_length>1`? As far as I understand, multi-regression is only supported for RMSE loss, correct?
- should this be left exposed as an option to the user?

daidahao:

Fixes #2967.

### Summary
I asked Codex for a quick summary:
Adds native multi-quantile support for `CatBoostModel` via CatBoost’s `MultiQuantile` loss (`likelihood="multiquantile"`), so all quantiles can be trained/predicted with one model instead of one model per quantile.

Also updates sklearn quantile handling to support both:
- legacy per-quantile `_model_container` flow (`likelihood="quantile"`), and
- native single-model multi-quantile flow (`likelihood="multiquantile"`).

Includes tests for construction, estimator retrieval, and likelihood outputs, plus small docs/changelog updates.

### Practical impact
- `CatBoostModel` quantile training should be faster/simpler.
- Existing quantile workflows remain supported for `CatBoostModel`.

### Usage

```py
import time

from darts.models import CatBoostModel

for likelihood in ["quantile", "multiquantile"]:
    model = CatBoostModel(
        lags=12,
        output_chunk_length=6,
        likelihood=likelihood,
        quantiles=[0.1, 0.5, 0.9],
    )
    start_time = time.perf_counter()
    model.fit(series)
    end_time = time.perf_counter()
    print(f"{likelihood} fit time: {end_time - start_time:.2f} seconds")
"""
quantile fit time: 4.11 seconds
multiquantile fit time: 2.47 seconds
"""
```

### Other Information

- XGBoost also supports multi-quantile training, we could add them in the next PR, once we have established the `MultiQuantileRegression` patterns here.
- LightGBM, AFAIK, does not support multi-quantile training.

### DIFF
--- a/examples/20-SKLearnModel-examples.ipynb
+++ b/examples/20-SKLearnModel-examples.ipynb
@@ -15,7 +15,7 @@
     "- parameters output_chunk_length in relation with multi_models\n",
     "- one-shot and auto-regressive predictions\n",
     "- multi output support\n",
-    "- probablistic forecasting\n",
+    "- probabilistic forecasting\n",
     "- explainability\n",
     "- and more"
    ]
@@ -138,7 +138,7 @@
     "**Covariates** (external data to help improve forecasts):\n",
     "The dataset also comes with weather measurements that we can use as covariates. For simplicity, we use:\n",
     "- **T [°C]**: Measured temperature\n",
-    "- **StrGlo [W/m2]**: Measured solar irradation\n",
+    "- **StrGlo [W/m2]**: Measured solar irradiation\n",
     "- **RainDur [min]**: Measured raining duration"
    ]
   },
@@ -211,7 +211,7 @@
     "- `SKLearnModel`: wrap the Darts Model API around any sklearn-like model\n",
     "- `XGBModel`: wrapper around [XGBoost's](https://xgboost.readthedocs.io/en/stable/python/python_api.html#xgboost.XGBRegressor) `XGBRegressor`\n",
     "\n",
-    "In addition to these, we also offer a unified API for some state of the art regression models that can be installed following our [installation guide](https://github.com/unit8co/darts/blob/master/INSTALL.md):\n",
+    "In addition to these, we also offer a unified API for some state-of-the-art regression models that can be installed following our [installation guide](https://github.com/unit8co/darts/blob/master/INSTALL.md):\n",
     "\n",
     "* `LightGBMModel`: wrapper around [LightGBM's](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMRegressor.html#lightgbm-lgbmregressor) `LightGBMRegressor`\n",
     "* `CatBoostModel`: wrapper around [CatBoost's](https://catboost.ai/en/docs/concepts/python-reference_catboostregressor) `CatBoostRegressor`.\n",
@@ -626,7 +626,7 @@
    },
    "source": [
     "### Visualization\n",
-    "To vizualize what is happening under the hood, let's simplify the model a bit and show the process for different parameters.\n",
+    "To visualize what is happening under the hood, let's simplify the model a bit and show the process for different parameters.\n",
     "\n",
     "**Multi models with auto-regression**\n",
     "\n",
@@ -647,7 +647,7 @@
     "\n",
     "Apart from the auto-regressive prediction, we can see the lags shift per predicted time step because `multi_models=False`.\n",
     "\n",
-    "The same process also occurs during tabularizion for training the model: each green chunk is paired with an orange value, constituting the training dataset."
+    "The same process also occurs during tabularization for training the model: each green chunk is paired with an orange value, constituting the training dataset."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


dennisbader:
Thanks a lot @daidahao 🚀 I'll be out of office next week but will have a look once I'm back :)

daidahao:
@dennisbader 
Any update on this? Thanks!

dennisbader:
Hi @daidahao this week I have a bit limited capacity, but from next week onwards everything should be back to normal :)

darts/models/forecasting/catboost_model.py
```diff
@@ -293,7 +309,8 @@ def fit(
         n_jobs_multioutput_wrapper: int | None = None,
         sample_weight: TimeSeriesLike | str | None = None,
         val_sample_weight: TimeSeriesLike | str | None = None,
-        verbose: int | bool | None = None,
+        stride: int = 1,
```
dennisbader, line 308: Thanks 👍

darts/models/forecasting/catboost_model.py
```diff
@@ -452,12 +448,12 @@ def _format_samples(
         If categorical features are specified, the samples are converted into a pandas DataFrame and categorical
         columns are cast to integer.
         """
-        samples, labels = super()._format_samples(samples, labels=labels)
+        outputs, labels = super()._format_samples(samples, labels=labels)
```
dennisbader: can you remove these diffs? These are not the model outputs, but rather rows in X / samples / features

darts/models/forecasting/catboost_model.py
```diff
@@ -149,13 +153,14 @@ def encode_year(idx):
             and variance couple, which capture data (aleatoric) uncertainty.
             This will overwrite any `objective` parameter.
         quantiles
-            Fit the model to these quantiles if the `likelihood` is set to `quantile`.
+            Fit the model to these quantiles if the `likelihood` is set to `quantile`. Default is `None` and
```
dennisbader: If we update any documentation here, then we should also update it for all other models that use the same parameter.

So either we remove the diffs or apply to all SKLearnModel submodels

darts/models/forecasting/catboost_model.py
dennisbader, line 1: Thanks for giving this a go @daidahao, it's nice to see that multi quantile support will soon make it to Darts 🚀 

I do think that we need to make certain changes, since adding this will not be backwards compatible (old dedicated quantile per model will not be supported anymore), and also it introduces some ambiguity to the QuantileRegression likelihood.

Instead I would suggest to add a new likelihood `MultiQuantileRegression` (with flag likelihood="multiquantile" or similar, only supported by CatBoost) that enables dedicated support for this. "quantile" will keep it's original behavior.

We have the same for the MultiRMSE loss. The user can specifically enable native multioutput support by using this loss function (see my answer [here](https://github.com/unit8co/darts/issues/3033#issuecomment-4066029196)).

What do you think?

darts/models/forecasting/catboost_model.py
```diff
@@ -212,6 +217,15 @@ def encode_year(idx):
         kwargs["random_state"] = random_state  # seed for tree learner
         self.kwargs = kwargs
 
+        if likelihood == "quantile" and len(quantiles or []) == 1:
```
dennisbader: (from my comment above) for example this logic could then be specific to the `MultiQuantileRegression` likelihood and handled there.

dennisbader (changes requested):

darts/models/forecasting/catboost_model.py
daidahao: @dennisbader 
Thanks for taking a close look. My rationale at the time was to preserve model signature across SKLearn models, i.e., `Model(likelihood="quantiles", quantiles=...)` for all, while providing speed benefits for `CatBoostModel` with quantiles.

I now agree with your suggestion here. It helps us maintain backward compatibility, avoid ambiguity, and handle single quantile objective.

I will migrate the new logic to a `MultiQuantileRegression` class and `CatBoostModel ` will expect `likelihood="multiquantile"` to enable multi-quantile training. Will let you known once it is done.

darts/models/forecasting/catboost_model.py
daidahao: Sure, let's apply to all.

daidahao:
@claude Are you okay?

daidahao:
@dennisbader 

`test_manual_save_and_load_precision` failed due to new error messages in `torch==2.11.0`. I suggest removing the assertion from the test to ensure compatibility.

dennisbader:
@daidahao, I opened #3044 to fix the test. Will merge once all tests have passed

daidahao:
@dennisbader 

Could you review the changes when you have time? Many thanks.

dennisbader:
@daidahao, somehow our macOS darts-all builds are hanging on Python versions. It's unrelated to this PR and I'll merge it. I'll try to figure out the root cause behind the issue

daidahao:
@dennisbader 

Thank you! I will submit a PR on XAI soon!

PR #3050, opened 2026-03-27
[Bugfix] Segmentation fault for darts[all] on macOS arm64

daidahao:

Fixes #.

### Summary
torch (latest, 2.11) bundled libomp is not compatible with lightgbm (latest, 4.6.0) on macOS arm64, causing segmentation faults:
```py

series = linear_timeseries(length=100)
model = LightGBMModel(12).fit(series)
```

**Related Issue**:
- https://github.com/lightgbm-org/LightGBM/issues/6595

Until it is fixed on either side, I suggest capping the torch version to 2.10.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,10 @@ Repository = "https://github.com/unit8co/darts"
 [project.optional-dependencies]
 # Main feature sets - pip installable with: pip install "darts[torch]"
 torch = [
-    "torch>=2.0.0",
+    # torch-2.11 bundled libomp is not compatible with lightgbm 4.6.0 on macOS arm64
+    # causing segmentation faults.
+    "torch>=2.0.0,<=2.10; sys_platform == 'darwin' and platform_machine == 'arm64'",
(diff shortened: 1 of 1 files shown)


dennisbader:
Thanks @daidahao. I'm not sure how quickly (I would assume it's LightGBM) resolves this issue. For me (also mac user), setting the env variable to the brew installed libomp resolved this issue:

```
# Only add libomp to DYLD_LIBRARY_PATH if Homebrew and libomp are installed
if command -v brew &>/dev/null; then
    _libomp="$(brew --prefix libomp 2>/dev/null)/lib"
    [ -d "$_libomp" ] && export DYLD_LIBRARY_PATH="${_libomp}${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}"
    unset _libomp
fi
```

daidahao:
@dennisbader 

Thank you for the suggestion and I can confirm it fixes the problem for me!

Would it be helpful to include it in `INSTALL.md` or pin the solution to [Issues](https://github.com/unit8co/darts/issues) so users could find it easily? I think it affects new fresh install of `darts[all]` on macOS, even when torch models are not imported (see snippet above).

PR #3053, opened 2026-03-31
Fix device mismatch error in TFTModel when moving a trained model to a different device

### LINKED ISSUE #3052: [BUG] TFTModel: Device mismatch after moving model between devices (e.g. GPU to CPU)
**Describe the bug**

`_TFTModule` caches `self.attention_mask` and `self.relative_index` as plain instance attributes (not registered buffers). These tensors are only regenerated when the batch size changes (`batch_size != self.batch_size_last` in `forward()`).

After a device transfer via `.cpu()`, `.cuda()`, or `.to()`, these cached tensors remain on the old device because `nn.Module._apply` only processes registered parameters and buffers. If `forward()` is then called with the same batch size as before, the stale cached tensors are reused and `torch.cat` raises a `RuntimeError` due to the device mismatch.

This affects any workflow that moves a trained TFT model to a different device before inference, including ONNX export.

**To Reproduce**

```python
import torch

# Create minimal model and data
model = TFTModel(
    input_chunk_length=4,
    output_chunk_length=2,
    add_relative_index=True,
    likelihood=QuantileRegression(),
    n_epochs=1,
    pl_trainer_kwargs={"accelerator": "gpu"},
)
series = linear_timeseries(length=20)
model.fit(series)

# After training, the inner module has cached tensors on cuda:
inner = model.model
print(inner.batch_size_last)       # e.g. 32 (last training batch size)
print(inner.relative_index.device) # cuda:0

# Move to CPU for inference/export:
inner.cpu()

# The cached tensors are still on cuda:
print(inner.relative_index.device) # still cuda:0!
print(inner.batch_size_last)       # still 32 — cache won't be invalidated

# Next forward() call with matching batch size will fail:
# RuntimeError: Expected all tensors to be on the same device,
# but got tensors on cuda:0, different from other tensors on cpu
```

**Expected behavior**

Moving the model to a different device (e.g., `model.cpu()`) should invalidate the cached tensors so they are regenerated on the correct device the next time `forward()` is called.

**Root cause**

In `tft_model.py`, lines 139–141:

```python
self.batch_size_last = -1
self.attention_mask = None
self.relative_index = None
```

These are plain attributes. `nn.Module._apply()` (which `.cpu()` / `.cuda()` / `.to()` delegate to) only processes `self._parameters` and `self._buffers`, so these attributes are left untouched.

The cache guard in `forward()` (line 480) only checks batch size:

```python
if batch_size != self.batch_size_last:
```

It does not check whether the device has changed.

**Suggested fix**

Override `_apply` in `_TFTModule` to invalidate the cache:

```python
def _apply(self, fn, recurse=True):
    self.batch_size_last = -1
    self.attention_mask = None
    self.relative_index = None
    return super()._apply(fn, recurse=recurse)
```

It would cause the tensors to be regenerated on the correct device in the next `forward()` call.

I'm happy to open a PR for this.

**Additional context**

We hit this during ONNX export of a TFT model trained on a CUDA GPU. Before the ONNX export, we call `.cpu()` on the inner module before tracing. When the export batch size matches the last training batch size, `torch.cat` in `forward()` fails because `self.relative_index` is still on `cuda:0` while input tensors are on `cpu`.

trahflow:

Fixes #3052.

### Summary

Registers TFT's `attention_mask` and `relative_index` as pytorch buffers.
Thereby they're included when `_TFTModule.to()` etc. are called.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
-- Fixed a device mismatch error in `TFTModel` when moving a trained model to a different device (e.g., GPU to CPU for ONNX export). `attention_mask` and `relative_index` are now registered as non-persistent buffers so they are properly moved with the model. [#3052](https://github.com/unit8co/darts/issues/3052) by [Wolfhart Feldmeier](https://github.com/trahflow)
+- Fixed a device mismatch error in `TFTModel` when moving a trained model to a different device (e.g., GPU to CPU for ONNX export). `attention_mask` and `relative_index` are now registered as non-persistent buffers so they are properly moved with the model. [#3053](https://github.com/unit8co/darts/pull/3053) by [Wolfhart Feldmeier](https://github.com/trahflow)
 
 ### For developers of the library:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 - Added native multi-quantile support for `CatBoostModel` by using CatBoost’s `MultiQuantile` loss for faster training and inference. Set `likelihood="multiquantile"` to enable this feature. [#3032](https://github.com/unit8co/darts/pull/3032) by [Zhihao Dai](https://github.com/daidahao)
 
+**Fixed**
+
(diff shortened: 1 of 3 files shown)


dennisbader (approved):
Thanks a lot @trahflow for the contribution and the quick fix 🥳 Looks great!

daidahao:
@trahflow Thank you for the fix!

FYI, I tested loading a previously saved TFT checkpoint and can confirm it is backward compatible with checkpoints.

dennisbader:
Thanks @daidahao, that is also good to know :)

PR #3058, opened 2026-04-02
Allow StatsForecastModel to accept model as string or class

### LINKED ISSUE #3055: Make StatsForecastModel accept `model` as string or class with `model_kwargs`.
Similar to how we implemented it for `NeuralForecastModel` (see [here](https://unit8co.github.io/darts/generated_api/darts.models.forecasting.nf_model.html#darts.models.forecasting.nf_model.NeuralForecastModel)), we should adapt how to pass the `model` and `model_kwargs` to `StatsForecastModel`.

Currently it expects `model` to be an instance of a statsforecast model. This can be tedious if for example we want to define which models and parameters to use via config files, or that in general we have to import all the models from statsforecast.

Instead, we can leverage the NeuralForecast logic, that allows passing `model` as a string (or class) and parameters as `model_kwargs`.

Either we directly enforce this new behavior (a breaking change) or we will deprecate the old behavior in a future release (e.g. deprecation warning for now if `model` is an instance of a statsforecast model).

tmchow:
## Summary

`StatsForecastModel` now accepts `model` as a string name, a class, or an instance. This follows the same pattern already used by `NeuralForecastModel` and makes it much easier to define models via config files or without importing every statsforecast model class.

## Changes

Updated `darts/models/forecasting/sf_model.py`:

- `model` parameter now accepts `str | type[_TS] | _TS` (was `_TS` only)
- Added `model_kwargs` parameter for passing constructor arguments when using string or class form
- Added `_import_sf_model_class()` static method that imports models from `statsforecast.models` by name
- Existing code passing model instances still works (backwards compatible)
- If `model_kwargs` is passed with an instance, a `UserWarning` is raised

Three usage patterns now work:

```python
# String (new)
model = StatsForecastModel(model="AutoARIMA", model_kwargs={"season_length": 12})

# Class (new)
from statsforecast.models import AutoARIMA
model = StatsForecastModel(model=AutoARIMA, model_kwargs={"season_length": 12})

# Instance (existing, backwards compatible)
model = StatsForecastModel(model=AutoARIMA(season_length=12))
```

## Testing

The change is backwards compatible with existing tests. The string and class resolution paths follow the same pattern as `NeuralForecastModel._import_nf_model_class()` which is already tested.

Fixes #3055

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/darts/models/forecasting/sf_model.py
+++ b/darts/models/forecasting/sf_model.py
@@ -3,8 +3,10 @@ StatsForecastModel
 ------------------
 """
 
+from typing import Protocol, runtime_checkable
+
 import numpy as np
-from statsforecast.models import _TS
+import statsforecast.models as sf_models
 
 from darts import TimeSeries, concatenate
 from darts.logging import get_logger, raise_log
@@ -19,11 +21,28 @@ from darts.utils.utils import random_method
 logger = get_logger(__name__)
 
 
+@runtime_checkable
+class _SFModel(Protocol):
+    """This serves as a protocol for expected StatsForecast model API."""
+
+    uses_exog: bool
+
+    def __init__(*args, **kwargs): ...
+
+    def fit(self, *args, **kwargs): ...
+
+    def predict(self, *args, **kwargs) -> dict: ...
+
+    def forecast(self, *args, **kwargs) -> dict: ...
+
+    def forward(self, *args, **kwargs) -> dict: ...
+
+
 class StatsForecastModel(TransferableFutureCovariatesLocalForecastingModel):
     @random_method
     def __init__(
         self,
-        model: str | type[_TS] | _TS = "AutoARIMA",
+        model: str | type[sf_models._TS] | sf_models._TS = "AutoARIMA",
         model_kwargs: dict | None = None,
         add_encoders: dict | None = None,
         quantiles: list[float] | None = None,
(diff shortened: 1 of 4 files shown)


darts/models/forecasting/sf_model.py
```diff
@@ -110,8 +118,8 @@ def encode_year(idx):
         >>> series = AirPassengersDataset().load()
```
jakubchlapek, line 135: the AutoARIMA import is no longer necessary

darts/models/forecasting/sf_model.py
```diff
@@ -110,8 +118,8 @@ def encode_year(idx):
...
         >>> # optionally, use some future covariates; e.g. the value of the month encoded as a sine and cosine series
         >>> future_cov = datetime_attribute_timeseries(series, "month", cyclic=True, add_length=6)
-        >>> # define AutoARIMA parameters
-        >>> model = StatsForecastModel(model=AutoARIMA(season_length=12))
+        >>> # define AutoARIMA parameters (using string and model_kwargs)
```
jakubchlapek: we can get rid of the clarification in () since it will be the default

darts/models/forecasting/sf_model.py
```diff
@@ -122,17 +130,57 @@ def encode_year(idx):
...
+                ),
+                logger,
+            )
+        if not (isinstance(model_class, type) and issubclass(model_class, _TS)):
+            raise_log(
+                ValueError(
+                    f"`{model}` is not a valid StatsForecast model class."
+                ),
+                logger,
+            )
```
jakubchlapek: i'd suggest extracting this code into `_validate_sf_model_class`, as per the NF implementation. while this would work, I'd like to keep the implementations as close to each other as possible :)

darts/models/forecasting/sf_model.py
```diff
@@ -122,17 +130,57 @@ def encode_year(idx):
...
+        if isinstance(model, _TS):
+            # backwards compatibility: model passed as an instance
+            if model_kwargs:
+                warnings.warn(
+                    "`model_kwargs` is ignored when `model` is an instance. "
+                    "Pass `model` as a string or class to use `model_kwargs`.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+            self.model = model
```
jakubchlapek: i generally agree that for now the SFModel instances should be accepted at least until some further release, but the error should be more clear this path is now deprecated and will be removed (e.g. `"DEPRECATED: <xyz>`, not just if the user passes in model_kwargs).

additionally, instead of `warnings.warn()` use `logger.warning()` for it.

jakubchlapek (commented):
Hey @tmchow, thanks a lot for your first contribution :)

The first draft looks nice, but I have left some comments, regarding your changes. 

I think that regarding this update we should for the time being keep the option of passing in model instances, while giving an explicit warning to the user that the option is deprecated. The biggest thing I can see currently is that the new str | class paths aren't tested. While all the tests still pass (because we still accept instances so the logic hasn't changed) we don't check the new functionality yet. Since it will be the default I think it would be nice to change the existing tests to leverage the str path instead, while also adding new tests that verify the model class and instance model creation work. Maybe @dennisbader has some other thoughts on how we want to approach the migration?

dennisbader (approved):
Thanks @tmchow for this PR and your first contribution 🚀 
The PR looked like a good start, I have already applied some updates to make it ready to be merged. 

After the tests have passed, I'll merge 💯

PR #3069, opened 2026-04-14
Feature: Custom encoders returning multiple components

### LINKED ISSUE #2806: [QUESTION] ValueError with add_encoders Custom Function Returning Multiple Components
I'm encountering a ValueError when using a custom function with add_encoders that returns a pandas DataFrame with multiple components (20 features in my case). I believe the custom function itself is correctly producing the DataFrame, but an issue arises during Darts' internal conversion to a TimeSeries.

Error Message:

ValueError: conflicting sizes for dimension 'component': length 20 on the data but length 1 on coordinate 'component'
This error occurs during model.fit(), originating from xarray.core.dataarray.py when a TimeSeries is being created internally from my custom encoder's output.

Setup:
I'm using add_encoders with a 'custom' key, providing a list of callable functions. One of these functions, let's call it encode_calendar_features, generates 20 calendar-related features.

```
# Simplified add_encoders structure in my model configuration
self.add_encoders = {
    # ... other potential built-in encoders like 'cyclic', 'datetime_attribute' ...
    'custom': {
        'past': [self.encode_calendar_features, self.encode_weekend_features], # encode_calendar_features returns 20 components
        'future': [self.encode_calendar_features, self.encode_weekend_features]
    },
    'transformer': Scaler() # Global scaler for all generated covariates
}

# My custom encoder function structure
def encode_calendar_features(self, idx: pd.DatetimeIndex) -> pd.DataFrame:
    df = pd.DataFrame(index=idx)
    # ... logic to create 20 feature columns ...
    # Example: df['feature_1'] = idx.year
    #          df['feature_2'] = idx.month
    #          ... (up to 20 features)
    return df
```
Debugging Done So Far:

Verified Custom Function Output: I've confirmed that my encode_calendar_features function correctly returns a pandas.DataFrame with the expected shape (e.g., (num_samples, 20)) and df.columns is a pandas.Index of 20 unique string names.

--- Debugging encode_calendar_features ---
df.shape: (1008, 20)
df.columns: Index(['is_holiday', 'holiday_new_year', ..., 'is_post_thanksgiving_week'], dtype='object')
type(df.columns): <class 'pandas.core.indexes.base.Index'>
len(df.columns): 20
df.columns.tolist(): ['is_holiday', ..., 'is_post_thanksgiving_week']
--- End Debugging ---

The error suggests that while the data for 20 components is present, xarray (via Darts' TimeSeries.from_times_and_values) is being told that the 'component' coordinate only has a length of 1. This occurs even though Darts' CallableStaticTransformer (which I believe wraps DataFrame-returning custom functions) seems designed to use df.values and df.columns correctly.

Question:

Is there a known issue or a specific way a custom function in add_encoders should return a multi-column DataFrame to ensure Darts correctly interprets it as a multi-component TimeSeries without this coordinate mismatch?

mwaldleben:

Fixes #2806.

### Summary

 This PR adds support in `CallableIndexEncoder` for custom encoders to return multiple components. A single callable used via `add_encoders["custom"]` can now return either a 1D array of shape `(len(index),)` or a 2D array of shape `(len(index), n)`.

For a 2D output,  one encoded covariate component per column are created. The number of components is detected at initialization time. If the callable does not accept a supported index type or does not return the correct output shape, a `ValueError` is raised at initialization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 - 🚀🚀 Added new forecasting model `PatchTSTFMModel` : IBM's pre-trained ~260M-parameter foundational model for zero-shot forecasting. It supports univariate, multivariate, and multiple time series forecasting without training and can output deterministic or probabilistic forecasts. [#3120](https://github.com/unit8co/darts/pull/3120) by [Dennis Bader](https://github.com/dennisbader).
 - Added `use_longer_projection_head` to `TimesFM2p5Model` to enable longer non-autoregressive prediction horizons (up to 1024 steps for `output_chunk_length + output_chunk_shift`). [#3121](https://github.com/unit8co/darts/pull/3121) by [Zhihao Dai](https://github.com/daidahao).
-- Added support in `add_encoders` for custom encoders returning multiple components. [#3069](https://github.com/unit8co/darts/pull/3069) by [Moritz Waldleben](https://github.com/mwaldleben).
+- Custom encoders now support functions that return multiple components. Simply pass such a function via the `"custom"` encoder key in the `add_encoders` model input parameter. [#3069](https://github.com/unit8co/darts/pull/3069) by [Moritz Waldleben](https://github.com/mwaldleben).
 
 **Fixed**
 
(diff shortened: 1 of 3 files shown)


dennisbader (approved):
Beautiful PR, thanks a lot @mwaldleben 🚀 

I just pushed some minor changes that are unrelated to your PR (we're removing the `raise_if*` helpers slowly so I took the chance to do it for the encoders module in this PR).

PR #3074, opened 2026-04-17
fix: spurious 'fewer forecasts' warning in RegressionEnsembleModel

shaun0927:

Fixes #3073.

### Summary

`RegressionEnsembleModel` emits a phantom `"Generated fewer forecasts than the requested N"` warning on every fit when `train_using_historical_forecasts=True`, even when the generated historical forecasts are long enough. The root cause is misplaced parentheses in the guard.

### Root cause

At `darts/models/forecasting/regression_ensemble_model.py:355`:

```python
if (
    len(get_single_series(series_forecasts) < train_n_points)
    and self.show_warnings
):
```

Python parses this as `len((get_single_series(series_forecasts) < train_n_points))`. `TimeSeries.__lt__` returns a component-wise boolean `TimeSeries`, whose `len()` equals the number of time points in the forecast. That length is always positive, so the condition is always truthy and the `logger.warning(...)` always runs.

### Fix

Move the closing parenthesis so that the `len()` is evaluated on the time-series first, then compared to `train_n_points`:

```diff
             if (
-                len(get_single_series(series_forecasts) < train_n_points)
+                len(get_single_series(series_forecasts)) < train_n_points
                 and self.show_warnings
             ):
```

All the surrounding forecasts have already been trimmed to `min_length` at line 348, so evaluating `len()` on the first element is correct.

### Tests

Added a dedicated regression test `test_train_with_historical_forecasts_no_phantom_warning` in `darts/tests/models/forecasting/test_regression_ensemble_model.py`. It configures an ensemble whose historical forecasts comfortably reach `regression_train_n_points`, fits it, and asserts via `caplog` that the `"Generated fewer forecasts"` warning is **not** in the captured log records.

### TDD verification

Without the fix (only the test added), the new test fails:

```
AssertionError: Spurious 'fewer forecasts' warning fired even though the historical forecasts had enough points.
...
WARNING darts.models.forecasting.regression_ensemble_model: Generated fewer forecasts than the requested 20 for the series at index 0.
FAILED darts/tests/models/forecasting/test_regression_ensemble_model.py::TestRegressionEnsembleModels::test_train_with_historical_forecasts_no_phantom_warning
```

With the parenthesis fix applied, the new test passes, and all adjacent `historical_forecasts` tests remain green:

```
$ pytest darts/tests/models/forecasting/test_regression_ensemble_model.py \
  -k "historical_forecasts or phantom"
9 passed, 177 deselected in 3.75s
```

### Other Information

None.


dennisbader (approved):
Thanks for spotting and fixing this @shaun0927 🚀 Also congrats to your first contribution!

I pushed some changes, mainly to make the test also cover when the warning should be raised.

Also combined the changelog entry with the existing update for the ensemble models since this warning was only introduced by that PR and not part of the previous Darts release.

PR #3076, opened 2026-04-17
fix: StaticCovariatesTransformer drops 'infrequent_sklearn' column with OneHotEncoder(min_frequency=...)

### LINKED ISSUE #3075: [BUG] StaticCovariatesTransformer drops 'infrequent_sklearn' column for OneHotEncoder(min_frequency=...)
**Describe the bug**

`StaticCovariatesTransformer` silently drops the `{col}_infrequent_sklearn` column when the underlying `OneHotEncoder` is configured with `min_frequency` or `max_categories` (sklearn ≥ 1.1). `transform(...)` returns a short static-covariate frame, and `inverse_transform(...)` subsequently fails with

```
ValueError: Shape of the passed X data is not correct. Expected 3 columns, got 2.
```

Root cause is in `StaticCovariatesTransformer._create_category_mappings` at [`darts/dataprocessing/transformers/static_covariates_transformer.py:298-313`](https://github.com/unit8co/darts/blob/master/darts/dataprocessing/transformers/static_covariates_transformer.py#L298-L313):

```python
feature_names = transformer_cat.get_feature_names_out(cols_cat)
feat_idx = 0
for i, col in enumerate(cols_cat):
    col_map_cat_i = []
    expected_names = {
        f"{col}_{cat}" for cat in transformer_cat.categories_[i]
    }
    while (
        feat_idx < len(feature_names)
        and feature_names[feat_idx] in expected_names
    ):
        ...
```

`expected_names` is built from `transformer_cat.categories_[i]`, which contains only the full set of input categories. When `min_frequency` / `max_categories` collapse the rarely-seen categories, sklearn emits an *additional* synthetic feature called `{col}_infrequent_sklearn` that is not in `categories_`. The `while` loop therefore stops before consuming it, the feature is missing from `col_map_cat`, and the transformed output is one column short.

When `inverse_transform` runs, darts slices out the (truncated) mapping columns and passes them to `OneHotEncoder.inverse_transform`, which refuses the wrong shape and raises `ValueError`.

#3065 added the `expected_names` set to fix the `IndexError` for `drop='first' / 'if_binary'`. That fix covered category-dropping, but did not account for category-merging via `min_frequency` / `max_categories`.

**To Reproduce**

```python
import numpy as np
import pandas as pd
from sklearn.preprocessing import OneHotEncoder
from darts import TimeSeries
from darts.dataprocessing.transformers import StaticCovariatesTransformer

def mk(country):
    return TimeSeries.from_times_and_values(
        times=pd.date_range("2024-01-01", periods=5),
        values=np.ones((5, 1)),
        static_covariates=pd.DataFrame({"country": [country]}),
    )

series = [mk("US")] * 5 + [mk("KR")] * 2 + [mk("JP")]  # JP is infrequent

t = StaticCovariatesTransformer(
    transformer_cat=OneHotEncoder(sparse_output=False, min_frequency=2),
)
t.fit(series)
transformed = t.transform(series)
# static_covariates are ['country_KR', 'country_US']  <-- 'country_infrequent_sklearn' missing
t.inverse_transform(transformed)
# ValueError: Shape of the passed X data is not correct. Expected 3 columns, got 2.
```

**Expected behavior**

`transform` should preserve every column emitted by `transformer_cat.get_feature_names_out(...)`, and `inverse_transform` should round-trip back to the original categories (mapping the collapsed inputs to sklearn's `"infrequent_sklearn"` sentinel, per `OneHotEncoder.inverse_transform`'s own contract).

**System**
- Python: 3.12.12
- darts: master (112a81997…)
- scikit-learn: 1.8+

**Additional context**

A fix that augments `expected_names` with `{col}_infrequent_sklearn` plus parametrized regression coverage is submitted as a separate PR.

shaun0927:

Fixes #3075.

### Summary

`StaticCovariatesTransformer` silently truncated the column mapping when the underlying `OneHotEncoder` uses `min_frequency` or `max_categories`. As a result, `transform(...)` dropped the synthetic `{col}_infrequent_sklearn` column and `inverse_transform(...)` raised `ValueError: Shape of the passed X data is not correct`.

### Root cause

In `_create_category_mappings` (`darts/dataprocessing/transformers/static_covariates_transformer.py:298-313`):

```python
expected_names = {
    f"{col}_{cat}" for cat in transformer_cat.categories_[i]
}
while (
    feat_idx < len(feature_names)
    and feature_names[feat_idx] in expected_names
):
    ...
```

`expected_names` is built from `transformer_cat.categories_[i]`. When sklearn groups low-frequency inputs via `min_frequency` / `max_categories`, it emits an additional synthetic feature `{col}_infrequent_sklearn` that is **not** in `categories_`. The `while` loop therefore stops before consuming it, the feature is omitted from `col_map_cat`, transformed static covariates are one column short, and `OneHotEncoder.inverse_transform` refuses the shape.

`#3065` introduced this `expected_names` set to cover the `drop='first'` / `drop='if_binary'` case. That fix is correct and remains unchanged; this PR only widens the set so it matches sklearn's full output for `min_frequency` / `max_categories`.

### Fix

```diff
 expected_names = {
     f"{col}_{cat}" for cat in transformer_cat.categories_[i]
 }
+# When ``min_frequency`` or ``max_categories`` is used, sklearn
+# emits an additional synthetic feature called
+# ``{col}_infrequent_sklearn`` which is not present in
+# ``categories_``. Include it so the mapping is not
+# silently truncated.
+expected_names.add(f"{col}_infrequent_sklearn")
 while (
     feat_idx < len(feature_names)
     and feature_names[feat_idx] in expected_names
 ):
```

The literal `"infrequent_sklearn"` is sklearn's documented feature name for the collapsed bucket (see `OneHotEncoder.get_feature_names_out`), so augmenting the set is sufficient and keeps the existing `drop` behaviour untouched.

### Tests

Added `test_one_hot_encoder_with_infrequent_sklearn` to `darts/tests/dataprocessing/transformers/test_static_covariates_transformer.py`, parametrized over `min_frequency=2` and `max_categories=2`. For each variant the test:

1. Builds a list of `TimeSeries` with a single `country` static covariate where one category ("JP") is infrequent,
2. Asserts that the transformed static covariates carry every column that `transformer_cat.get_feature_names_out(...)` emits (including `country_infrequent_sklearn`),
3. Round-trips via `inverse_transform` and asserts that frequent inputs map back to their original value while infrequent inputs map to sklearn's `"infrequent_sklearn"` sentinel (this is sklearn's own contract).

### TDD verification

Without the fix, the new test fails:

```
AssertionError: assert ['country_KR', 'country_US'] == ['country_KR', 'country_US', 'country_infrequent_sklearn']
FAILED darts/tests/dataprocessing/transformers/test_static_covariates_transformer.py::TestStaticCovariatesTransformer::test_one_hot_encoder_with_infrequent_sklearn[enc_kwargs0]
```

With the fix applied, both the new tests and every existing `one_hot_encoder_*` test pass:

```
$ pytest darts/tests/dataprocessing/transformers/test_static_covariates_transformer.py \
  -k "infrequent_sklearn or one_hot_encoder_with_drop"
6 passed, 8 deselected in 7.25s
```

### Other Information

None.


dennisbader (approved):
Thanks a lot for this fix @shaun0927 🚀 

I pushed some changes to make the logic more robust / generic against any newly added columns. This is to avoid that we have to refactor again in case sklearn either adds new columns, or changes the existing values for the column replacements.

Issue #3077, opened 2026-04-17
[BUG] Diff(columns=...) silently differences the wrong components when series column orders differ


shaun0927:
**Describe the bug**

When `Diff` is called with `columns=[...]` on a `Sequence[TimeSeries]` whose series have *identical component names but different orderings*, the transformer silently differences the wrong components. The column-name-based selector is resolved once — using the first series — and then the resulting `component_mask` is re-applied by **physical position** to every other series in the sequence.

The risk is introduced by the `columns` feature added in [#3023](https://github.com/unit8co/darts/pull/3023), which expands to a positional `component_mask` generated from `data[0]`. There is no validation that all series share the same column order.

Related but distinct from the docstring cleanup I'm about to submit as a small docs PR.

**To Reproduce**

```python
import numpy as np
import pandas as pd
from darts import TimeSeries
from darts.dataprocessing.transformers.diff import Diff

def mk(cols):
    return TimeSeries.from_times_and_values(
        times=pd.date_range("2024-01-01", periods=10),
        values=np.arange(10 * len(cols)).reshape(10, len(cols)).astype(float),
        columns=cols,
    )

ts_ab = mk(["A", "B"])       # physical column 0 is 'A'
ts_ba = mk(["B", "A"])       # physical column 0 is 'B'
out = Diff(lags=1, dropna=False, columns=["A"]).fit_transform([ts_ab, ts_ba])
# out[0] diffs 'A' as requested. OK.
# out[1] diffs 'B' — i.e. the wrong column — silently.
print(out[1].values()[:2])
# [[nan, 1.], [2., 3.]]   <- physical column 0 ('B') was diffed
```

Any downstream model that fits on `out` trains on corrupted data, with no warning and no exception.

**Expected behavior**

`Diff(columns=[...])` should either

1. resolve the `columns` argument **per series by name** when the sequence is processed, or
2. refuse to run when the component names/orders differ across the sequence (and document the requirement), or
3. explicitly require every series in the sequence to share the same component names in the same order and validate that.

Any of these options would uphold the `CONTRIBUTING.md` principle "Make it hard for users to make mistakes".

**System**
- Python: 3.12.12
- darts: master (112a81997…)

**Additional context**

[...]

PR #3078, opened 2026-04-17
docs: remove stale `component_mask` restriction note from `Diff`

shaun0927:

Related discussion: #3077 (tracked separately, not fixed here).

### Summary

The `Diff` transformer's docstring still carries the pre-#3023 restriction note:

> Note that if `dropna = True`, then a `component_mask` cannot be specified, since the undifferenced components will be of a different length to the differenced ones.

[#3023](https://github.com/unit8co/darts/pull/3023) removed that restriction when it introduced the `columns=...` parameter (which internally synthesises a `component_mask`). `Diff(lags=1, dropna=True, columns=['A']).fit_transform(series)` now succeeds — the series is trimmed from the start by `sum(lags)` time steps and the non-differenced components are preserved over that window.

### Fix

Replace the stale sentence with an accurate description of the current behaviour, and update the adjacent inline comment (which also still referred to the removed error path) for consistency.

### Verification

Behaviour verified on master (112a81997):

```python
>>> import numpy as np, pandas as pd
>>> from darts import TimeSeries
>>> from darts.dataprocessing.transformers.diff import Diff
>>> ts = TimeSeries.from_times_and_values(
...     pd.date_range("2024-01-01", periods=10),
...     np.arange(10 * 2).reshape(10, 2).astype(float),
...     columns=["A", "B"],
... )
>>> for dropna in (False, True):
...     out = Diff(lags=1, dropna=dropna, columns=["A"]).fit_transform(ts)
...     print(f"dropna={dropna}: shape={out.values().shape}, cols={list(out.components)}")
dropna=False: shape=(10, 2), cols=['A', 'B']
dropna=True:  shape=(9, 2),  cols=['A', 'B']
```

Docs-only change, no runtime behaviour altered.

### Other Information

None.


dennisbader (approved):
LGTM, thanks for this fix @shaun0927 🚀

PR #3086, opened 2026-04-24
Docs/fit dataset error

### LINKED ISSUE #3006: [Feature] Improve handling & documentation of encoders and categorical static covariates in .fit_from_dataset()

Sidenote:
 
jakubchlapek:

Fixes #3006 .

### Summary

Document and log a warning that add_encoders are ignored on TorchForecastingModel.fit_from_dataset / predict_from_dataset, and note TFT categorical_embedding_sizes keying by column index when using that path.

### DIFF
--- a/darts/models/forecasting/tft_model.py
+++ b/darts/models/forecasting/tft_model.py
@@ -741,7 +741,7 @@ class TFTModel(MixedCovariatesTorchModel):
             ``{"some_column": (64, 8)}``.
             Note that ``TorchForecastingModels`` only support numeric data. Consider transforming/encoding your data
             with `darts.dataprocessing.transformers.static_covariates_transformer.StaticCovariatesTransformer`.
-            When training via ``TorchForecastingModel.fit_from_dataset()``, categorical embeddings are resolved by
+            When training via ``fit_from_dataset()``, categorical embeddings are resolved by
             static covariate column integer index (not by column name). The keys of ``categorical_embedding_sizes``
             must align positionally with the static covariate columns.
         add_relative_index
--- a/darts/models/forecasting/tft_model.py
+++ b/darts/models/forecasting/tft_model.py
@@ -741,6 +741,9 @@ class TFTModel(MixedCovariatesTorchModel):
             ``{"some_column": (64, 8)}``.
             Note that ``TorchForecastingModels`` only support numeric data. Consider transforming/encoding your data
             with `darts.dataprocessing.transformers.static_covariates_transformer.StaticCovariatesTransformer`.
+            When training via ``TorchForecastingModel.fit_from_dataset()``, categorical embeddings are resolved by
+            static covariate column integer index (not by column name). The keys of ``categorical_embedding_sizes``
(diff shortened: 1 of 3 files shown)


darts/models/forecasting/tft_model.py
```diff
@@ -741,6 +741,9 @@ def __init__(
             ``{"some_column": (64, 8)}``.
             Note that ``TorchForecastingModels`` only support numeric data. Consider transforming/encoding your data
             with `darts.dataprocessing.transformers.static_covariates_transformer.StaticCovariatesTransformer`.
+            When training via ``TorchForecastingModel.fit_from_dataset()``, categorical embeddings are resolved by
```
dennisbader:
```suggestion
            When training via ``fit_from_dataset()``, categorical embeddings are resolved by
```

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1223,6 +1223,9 @@ def fit_from_dataset(
         This function can be called several times to do some extra training. If ``epochs`` is specified, the model
         will be trained for some (extra) ``epochs`` epochs.
 
+        Encoders configured via ``add_encoders`` at model creation are not applied here; ``train_dataset``
```
dennisbader: this could be wrapped in a "note" section (similar to how we have it in other places) to make it more obvious

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1256,6 +1259,14 @@ def fit_from_dataset(
         self
             Fitted model.
         """
+        if self.encoders is not None and self.encoders.encoding_available:
```
dennisbader: `fit_from_dataset()` is also called when calling `fit()` directly. So this warning will always show. We want it to only show if the user called it `fit_from_dataset` directly

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1852,6 +1863,9 @@ def predict_from_dataset(
         ``trainer``. For more information on PyTorch Lightning Trainers check out `this link
         <https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html>`__.
 
+        Encoders configured via ``add_encoders`` at model creation are not applied here; ``dataset`` must already
```
dennisbader: could also be a note section

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -1910,6 +1924,14 @@ def predict_from_dataset(
 
         self._verify_inference_dataset_type(dataset)
 
+        if self.encoders is not None and self.encoders.encoding_available:
```
dennisbader: same comment here this would also be called when calling `predict()` directly

dennisbader (changes requested):
Thanks for the PR @jakubchlapek 🚀 

I added some suggestions, mainly regarding the warning that would now also be raised whenever user calls fit() or predict() directly

PR #3107, opened 2026-05-04
Devops: publish documentation using first-party actions

daidahao:

Fixes #.

### Summary

This PR proposes using GitHub Actions to publish the documentation to GitHub Pages, replacing the previous third-party branch-based deployment method.

I am a bit wary of using third-party action `s0/git-publish-subdir-action@v2.2.0` for publishing the documentation. The action has not been updated in a long time (last release was 2022), is owned by an individual, and has access to `secrets.GITHUB_TOKEN` which could write to any branch.

Meanwhile, the [GitHub Pages documentation](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) recommends using [actions/upload-pages-artifact](https://github.com/actions/upload-pages-artifact) and [actions/deploy-pages](https://github.com/actions/deploy-pages), both first-party actions, to publish the site. This PR adopts the flow for better security and long-term support.

As suggested in [actions/deploy-pages](https://github.com/actions/deploy-pages), I've also:
- split `deploy-docs` into `build-docs` and `deploy-docs` to prevent code execution
- restricted `GITHUB_TOKEN` permissions for `deploy-docs` job to only those necessary.

@dennisbader If you agree with this, you would need to [change Pages settings](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) for the next release to use "GitHub Actions" as source to replace "Deploy from a branch".

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -63,6 +63,7 @@ jobs:
       id-token: write   # to verify the deployment originates from an appropriate source
     steps:
       - name: "Deploy to GitHub Pages"
+        id: deployment
         uses: actions/deploy-pages@v5.0.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  deploy-docs:
+  build-docs:
     runs-on: ubuntu-latest
     steps:
       - name: "Clone repository"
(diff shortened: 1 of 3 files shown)


jakubchlapek:
Hi @daidahao, thanks for the PR!
After looking into it, I definitely second the change to leverage first-party actions instead. The current solution is quite dated (you mentioned 2022, but the pinned version we use comes from even 2020) and not maintained. The schemas look consistent with the documentation. One additional element we can add is limiting the `build-docs` job to `contents: read` access, as it doesn't require write I don't believe since we write the documentation as artifacts, and not touch the repo. Besides that it looks good to me, good catch :)

daidahao:
@jakubchlapek 

Thank you for taking a close look at this and I am glad that you agree with the proposal.

Regarding the `contents: read` access, my understanding is that the [`permission` field](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions) of each job applies to `GITHUB_TOKEN` only. Because we are not using `GITHUB_TOKEN` in `build-docs`, so we do not have to declare `permission` for the job either.

Another outdated job I have looked at is `actions/create-release@latest` (last update in 2020 and using Node.js 20). Unfortunately, I could not find an official alternative to this action. Would love to know what's your opinion on this.

dennisbader (approved):
Looks great, thanks a lot @daidahao 🚀 
Will merge once all tests have passed

dennisbader:
@daidahao, I also updated the GitHub Pages settings, so this one should be good to go 💯

PR #3116, opened 2026-05-09
fix(examples): handle NaNs in 14-transfer-learning air dataset

### LINKED ISSUE #2752: [BUG]  examples/14-transfer-learning.ipynb error on air_train because of missing values
**Describe the bug**
the downloaded data for air dataset contains nan

**To Reproduce**
run the local models, it will return a NaN error for the series

**Expected behavior**
models run and returns correct SMAPE

 - Python version: 3.10
 - darts version 0.34.0

**Additional context**
can fix via creating a cell to fill missing values since this is for demo purposes 
```
from darts.utils.missing_values import fill_missing_values

air_train = [fill_missing_values(ts, fill='auto') for ts in air_train]
```

jbbqqf:
Fixes #2752.

### Summary

`examples/14-transfer-learning.ipynb` fails on the air dataset with `ValueError: ... NaN ...` errors during model training. The raw `carrier_passengers.csv` contains gaps that survive `longest_contiguous_slice()` (e.g. zero-rows the slice helper does not flag as missing), so the resulting series carry NaNs into model training. As @rafmacalaba reported in #2752, the cleanest workaround is to call `fill_missing_values(...)` after the slice; @dennisbader acknowledged the notebook needs refactoring (it's one of the few examples that's not run in CI because it's slow).

This PR applies that fix in the smallest possible patch:
- Import `fill_missing_values` from `darts.utils.missing_values` in the imports cell.
- Inside `load_air()`, call `fill_missing_values(series, fill="auto")` immediately after `longest_contiguous_slice()`. An inline comment cites the issue so a future reader doesn't have to dig.

### Other Information

- No code in the `darts/` package is touched. Only `examples/14-transfer-learning.ipynb` and `CHANGELOG.md`.
- CHANGELOG entry under `Unreleased` > `Fixed`.
- Maintainer-stated that the notebook isn't part of CI, so adding a unit test wouldn't fit existing conventions. The notebook change is itself the regression artifact: the diff is what unblocks users running the example end-to-end.

### Reproduce BEFORE/AFTER yourself (copy-paste)

```bash
# --- one-time setup ---
git clone https://github.com/unit8co/darts.git /tmp/darts-2752 && cd /tmp/darts-2752
uv sync --group dev

# --- inspect the difference (no need to run the full notebook) ---

# BEFORE (origin/master): load_air does NOT call fill_missing_values
git checkout origin/master
grep -n "fill_missing_values" examples/14-transfer-learning.ipynb || echo "NOT PRESENT"
# Expected: "NOT PRESENT" — the helper is neither imported nor called.

# AFTER (this PR): load_air now imports + calls fill_missing_values
git fetch https://github.com/jbbqqf/darts.git fix/2752-transfer-learning-nan
git checkout FETCH_HEAD
grep -n "fill_missing_values" examples/14-transfer-learning.ipynb
# Expected: 2 hits — one in the imports cell, one inside load_air()
#           right after longest_contiguous_slice().
```

If you want to verify functionally (slow — downloads the dataset and runs the local-models cells):

```bash
# AFTER, end-to-end smoke
uv run jupyter nbconvert --to notebook --execute examples/14-transfer-learning.ipynb \
    --output /tmp/14-executed.ipynb --ExecutePreprocessor.timeout=600
# Expected: no NaN-related ValueError; eval_local_model finishes with finite SMAPE values.
```

### What I ran locally

- Validated the notebook is well-formed JSON and contains 76 cells (unchanged count).
- Confirmed `fill_missing_values` import + call are present in the patched notebook.
- Confirmed the diff is limited to the imports cell and the `load_air()` cell (plus the CHANGELOG).
- Did not execute the full notebook locally because it requires downloading the M3 / M4 / carrier_passengers datasets (the same reason maintainers don't run it in CI per #2752).

### Edge cases tested

| # | Scenario | Behavior |
|---|----------|----------|
| 1 | Air dataset with residual NaN-bearing series | `fill_missing_values(fill="auto")` interpolates the gaps before scaling, so downstream models receive finite values |
| 2 | Air dataset series that were already gap-free | `fill_missing_values` is a no-op; no behavioral change |
| 3 | Series too short or raising in `longest_contiguous_slice()` | Caught by the existing `try/except continue` block — fix is positioned after that guard |

### Risk / blast radius

- Notebook-only change. No runtime code in `darts/` is modified.
- `fill_missing_values(..., fill="auto")` is a public, documented helper used elsewhere in the codebase, so the new dependency is benign.

---

*PR drafted with assistance from Claude Code (Anthropic). The change was reviewed manually against `unit8co/darts@master`, the discussion in #2752, and the proposed fix from the original reporter. Reviewer can paste the BEFORE/AFTER block above to verify.*


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
+- Fixed `examples/14-transfer-learning.ipynb` raising a NaN error on the `air` dataset: the `load_air()` helper now calls `fill_missing_values(..., fill="auto")` after `longest_contiguous_slice()` to handle residual gaps in the raw `carrier_passengers.csv`. Closes [#2752](https://github.com/unit8co/darts/issues/2752).
```
jakubchlapek: For future reference, refer to the PR number and link here :) also if you want take credit (`by ...`)
```suggestion
- Fixed `examples/14-transfer-learning.ipynb` raising a NaN error on the `air` dataset by calling `fill_missing_values(..., fill="auto")` to handle residual gaps in the raw `carrier_passengers.csv`. Closes [#3116](https://github.com/unit8co/darts/pull/3116).
```

examples/14-transfer-learning.ipynb
```diff
@@ -244,6 +245,11 @@
     "            series = series.longest_contiguous_slice()\n",
     "        except Exception:\n",
     "            continue\n",
+    "        # Fill any residual missing values inside the chosen slice. The raw\n",
+    "        # carrier_passengers.csv occasionally has gaps that survive the\n",
+    "        # longest_contiguous_slice() call (e.g. zero-rows), which produce NaNs\n",
+    "        # downstream and break model training. See issue #2752.\n",
+    "        series = fill_missing_values(series, fill=\"auto\")\n",
```
jakubchlapek: bit too verbose compared to the rest of the notebook
should be fine with just
```suggestion
    "        # fill missing values\n",
    "        series = fill_missing_values(series, fill=\"auto\")\n",
```

jakubchlapek (commented):
Hey @jbbqqf, thanks for this PR :)

Looks good, small change requested. 
What would be really nice for us for this PR is also finding out why this issue was raising an error in the first place (if it was the data that changed or a new interaction with code). Would you be willing to check that out as part of this pull request? :) Thanks

PR #3119, opened 2026-05-12
fix: vendor PyTorchLightningPruningCallback to drop optuna-integration dependency

### LINKED ISSUE #3093: [BUG] Optuna-Integration no longer depends on pytorch-lightning
**Describe the bug**
`optuna-integration[pytorch-lightning]` now depends on `lightning` and not `pytorch-lightning` any more. Running `uv sync --group dev-all --upgrade` will result in an error due to `lightning` not being available on win32 py312 (the package does exist on win32 py312, but there might be markers from other packages that prevent the resolution.)

**To Reproduce**
Steps to reproduce the behavior, preferably code snippet.

```sh
uv sync --group dev-all --upgrade
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.12.*' and platform_machine != 's390x' and sys_platform ==
  │ 'win32'):
  ╰─▶ Because there are no versions of lightning and optuna-integration[pytorch-lightning]>=4.7.0 depends on lightning, we can conclude that
      optuna-integration[pytorch-lightning]>=4.7.0 cannot be used.
      And because only the following versions of optuna-integration[pytorch-lightning] are available:
          optuna-integration[pytorch-lightning]<=4.7.0
          optuna-integration[pytorch-lightning]==4.8.0
      we can conclude that optuna-integration[pytorch-lightning]>=4.7.0,<4.8.0 cannot be used. (1)

      Because there are no versions of lightning and optuna-integration[pytorch-lightning]==4.8.0 depends on lightning, we can conclude that
      optuna-integration[pytorch-lightning]==4.8.0 cannot be used.
      And because we know from (1) that optuna-integration[pytorch-lightning]>=4.7.0,<4.8.0 cannot be used, we can conclude that
      optuna-integration[pytorch-lightning]>=4.7.0 cannot be used.
      And because darts:dev-all depends on optuna-integration[pytorch-lightning]>=4.7.0 and your project requires darts:dev-all, we can conclude that your
      project's requirements are unsatisfiable.

      hint: While the active Python version is 3.13, the resolution failed for other Python versions supported by your project. Consider limiting your
      project's supported Python versions using `requires-python`.
```

Immediate suggested solution is to change `optional` dependency group from:
```
optuna-integration[pytorch-lightning]>=4.7.0
```
to
```
optuna-integration>=4.7.0
```

However, a longer-term solution might be to address #2198, i.e., migrate from `pytorch-lightning` to `lightning`.

**Expected behavior**

 - Python version: 3.13.12
 - darts version: latest source
 - macOS: 26.3
 - uv: 0.10.2

**Additional context**

eastani:

Fixes #3093.

### Summary

Implements the direction agreed on in #3093: vendor Optuna's `PyTorchLightningPruningCallback` into Darts and drop the `optuna-integration[pytorch-lightning]` optional dependency.

- **`darts/utils/callbacks.py`** — adds `PyTorchLightningPruningCallback`, a self-contained reimplementation that imports directly from `pytorch_lightning` (which Darts already uses) rather than from `lightning` (which `optuna_integration` imports). The class inherits from `pytorch_lightning.callbacks.Callback` and exposes the same public API (`__init__(trial, monitor)`, `on_fit_start`, `on_validation_end`, `check_pruned`) as the upstream callback, including DDP support via Optuna's `RDBStorage` and the same `ddp_pl:*` trial system-attribute keys so DDP state is portable between the two implementations. Optuna remains an optional dependency: the module imports cleanly without it, and instantiation raises a clear `ImportError` if optuna is missing.

- **`darts/tests/optional_deps/test_optuna.py`** — drops the `PatchedPruningCallback` multiple-inheritance workaround (previously needed because `optuna_integration` imports from \`lightning.pytorch\` while Darts uses \`pytorch_lightning\`), and updates `test_optuna_torch_model` to use the new Darts callback directly. Five new unit tests cover the callback in isolation: PL-`Callback` inheritance, the missing-metric warning, pruning when the trial reports `should_prune`, sanity-check short-circuit at epoch 0, and `check_pruned` as a no-op outside DDP.

- **`pyproject.toml`** — removes `optuna-integration[pytorch-lightning]>=4.7.0` from the `optional` dependencies. Only `optuna>=4.7.0` is retained.

- **`examples/17-hyperparameter-optimization.ipynb`** — imports the callback from `darts.utils.callbacks` instead of `optuna.integration`.

- **`CHANGELOG.md`** — Unreleased entry under both **Improved** and **Dependencies**.

### Other Information

- All 10 tests in `darts/tests/optional_deps/test_optuna.py` pass locally on Python 3.13 with `pytorch-lightning==2.5.2` and `optuna==4.8.0`.
- `pre-commit` (ruff, ruff-format, end-of-file fixer, etc.) runs clean on all touched files.
- The `pl.__version__ >= 1.6.0` check from the upstream callback was dropped because Darts already requires `pytorch-lightning>=2.0.0`.
- Marked as draft so the design (location of the callback inside `darts.utils.callbacks` rather than a new module, exposure as a public attribute alongside `TFMProgressBar`, opt-in lazy import of optuna) can be reviewed before final polish.

Happy to iterate — please point out anything that should be moved, renamed, or tested more thoroughly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,13 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
+- Added `PyTorchLightningPruningCallback` to `darts.utils.callbacks`, a self-contained Optuna pruning callback for use with PyTorch Lightning-based forecasting models. This removes the need for the `PatchedPruningCallback` multiple-inheritance workaround when using Optuna with Darts. [#3093](https://github.com/unit8co/darts/issues/3093)
+
 **Fixed**
 
 **Dependencies**
+
+- Dropped the `optuna-integration[pytorch-lightning]` optional dependency. The vendored `PyTorchLightningPruningCallback` is now imported from `darts.utils.callbacks`. [#3093](https://github.com/unit8co/darts/issues/3093)
 
 ### For developers of the library:
 
--- a/darts/tests/optional_deps/test_optuna.py
+++ b/darts/tests/optional_deps/test_optuna.py
@@ -21,16 +21,10 @@ import optuna
 
 if TORCH_AVAILABLE:
     import torch
-    from pytorch_lightning.callbacks import Callback, EarlyStopping
-
-    # hacky workaround found in https://github.com/Lightning-AI/pytorch-lightning/issues/17485
-    # to avoid import of both lightning and pytorch_lightning
-    class PatchedPruningCallback(
-        optuna.integration.PyTorchLightningPruningCallback, Callback
-    ):
-        pass
+    from pytorch_lightning.callbacks import EarlyStopping
 
     from darts.models import TCNModel
+    from darts.utils.callbacks import PyTorchLightningPruningCallback
     from darts.utils.likelihood_models.torch import GaussianLikelihood
 
 
@@ -62,7 +56,7 @@ class TestOptuna:
             include_year = trial.suggest_categorical("year", [False, True])
 
             # throughout training we'll monitor the validation loss for both pruning and early stopping
-            pruner = PatchedPruningCallback(trial, monitor="val_loss")
+            pruner = PyTorchLightningPruningCallback(trial, monitor="val_loss")
             early_stopper = EarlyStopping(
                 "val_loss", min_delta=0.001, patience=3, verbose=True
             )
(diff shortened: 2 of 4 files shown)


dennisbader:
Hi @eastani and thanks for the PR. However, we already have #3114 that implements this. We will go with that one since it came in first. Will close this PR, but thanks again for the work!

PR #3121, opened 2026-05-21
Feature: TimesFM Longer Projection Head

daidahao:

Fixes #.

### Summary

Code is mostly hand-written, but I asked Codex to write the following summary and changes:

This PR adds optional support for TimesFM 2.5’s longer output projection head in Darts.

With `use_longer_projection_head=True`, users can run non-autoregressive forecasts up to:
- `output_chunk_length + output_chunk_shift <= 1024`

Default behavior is unchanged (`False`, max 128).

## Changes
- Added `use_longer_projection_head: bool = False` to `TimesFM2p5Model`.
- Added `output_quantile_len = 1024` to TimesFM 2.5 model definition.
- Updated module forward path to select projection head based on `use_longer_projection_head`:
  - default point head (128)
  - longer quantile head (1024)
- Updated prediction-length validation and error messaging.
- Passed `use_longer_projection_head` into loaded PL module via `additional_params`.
- Expanded docstring/warnings around forecast divergence behavior.
- Added/updated tests in `test_timesfm2p5.py`:
  - longer-head prediction path
  - creation/validation coverage for both `use_longer_projection_head=True/False`
- Added changelog entry under `Unreleased` → `Improved`.

## Backward Compatibility
- Backward compatible by default.
- Existing code paths remain unchanged unless `use_longer_projection_head=True` is set explicitly.

## Testing
- Added slow tests for longer projection head and creation constraints.
- Test file: `darts/tests/models/forecasting/test_timesfm2p5.py`


darts/tests/models/forecasting/test_timesfm2p5.py
```diff
@@ -194,12 +196,44 @@ def test_fidelity(self, probabilistic: bool):
...
+            output_chunk_length=self.max_prediction_length_longer_head,  # maximum prediction length w/o AR
+            use_longer_projection_head=True,  # use longer projection head for longer predictions
+            **tfm_kwargs,
+        )
+        # fit model w/o fine-tuning
+        model.fit(series=self.ts_energy_train)
+
+        # predict on the validation inputs
+        pred = model.predict(
+            n=self.max_prediction_length_longer_head,
```
dennisbader, line 215: should we use `predict_likelihood_parameters=True` together with a likelihood, to make sure that it's actually forecasting without auto-regression?

dennisbader (changes requested):
Looks great, thanks a lot for the quick fix @daidahao 🚀 
Just added one minor comments regarding one of the tests. After that everything is good to go 💯

darts/tests/models/forecasting/test_timesfm2p5.py
daidahao: @dennisbader You are right. I was initially worried about extra runtime of running two tests here. It seems that the TimesFM runtime is out of [Top-50 slowest](https://github.com/unit8co/darts/actions/runs/26232607130/job/77196874292), so I have now added probabilistic back.

dennisbader (approved):
LGTM, thanks a lot @daidahao 🚀

PR #3122, opened 2026-05-26
fix: extend zero_division parameter to percentage and range-based metrics

### LINKED ISSUE #3132: Consistent zero-denominator handling for percentage and range metrics
### Context

PR #3122 added a `zero_division` parameter to `wmape`, `ope`, `arre`, `marre`, and `coefficient_of_variation`, mirroring the scaled-metric work in #3059 (`mase`, `msse`, `rmsse`). Review feedback (thanks @jakubchlapek) converged on a cleaner, consistent design than the initial NaN-on-zero approach. This issue tracks completing that design across the percentage/range metric family.

### Agreed design

A single shared helper handles every "divide a numerator by a possibly-zero denominator" case, distinguishing two sub-cases exactly as the existing `_safe_scaled_divide` already does:

- **0/0** (numerator 0 *and* denominator 0) → the metric's **best score** (a perfect forecast on a degenerate series): `0.0` for percentage-error metrics, `1.0` for scaled-error metrics ("on par with naive"). The fill value is a helper parameter.
- **x/0** (numerator ≠ 0, denominator 0) → `np.nan`. Filling with the best score here would mask a real miss.
- `zero_division="raise"` preserves the legacy `ValueError`.

Rationale: the current PR fills *every* zero-denominator entry with `NaN`, which incorrectly scores a genuinely perfect forecast (e.g. `ope` on a series summing to zero with an identical prediction; `arre`/`marre` on a constant series predicted perfectly) as `NaN` instead of `0.0`. `sape`/`smape` already encode the `0/0 → 0.0` convention, confirming it.

### Scope

- **Aggregate-denominator (this PR, #3122):** `wmape`, `ope`, `arre`, `marre`, `coefficient_of_variation` — switch to the unified helper with `best_score=0.0`. (`coefficient_of_variation` is `RMSE/mean`, so `0.0` = perfect forecast — proposed, pending confirmation since its denominator can be negative.)
- **Already aligned:** `sape`/`smape` already fill `0/0 → 0.0` inline and have no `x/0` case (the symmetric denominator is zero only when both actual and prediction are zero) — they need only the explicit `zero_division` knob for the `"raise"` option.
- **Element-wise denominator (deliberate / possibly follow-up):** `ape`, `mape` still hard-raise if any `y_t == 0`; their denominator is per-timestep, so they need element-wise handling, and relaxing MAPE's default is a larger behavioral change.

### Implementation

Collapse `_safe_pct_divide` into `_safe_scaled_divide` (one helper, configurable 0/0 fill, with the warn/raise message parameterized so the scaled callers keep their existing strings), update the callers, add a `0/0` best-score test case, and fold the `ope` negative-sum test into `test_ope()`. Tracked by #3122.

mahimn01:
- [x] Summarized the updates under **Summary**.
- [x] Added an entry under **Unreleased** in the changelog.

Closes #3132.

### Summary

Adds consistent `zero_division` handling to `ape`, `mape`, `sape`, `smape`, `wmape`, `ope`, `arre`, `marre`, and `coefficient_of_variation`.

- `"warn"` (default): `0/0` returns `0.0`, while non-zero/0 returns `np.nan`, and logs a warning.
- `"raise"`: raises a `ValueError`.

These metrics reuse `_safe_scaled_divide` with a configurable zero fill. Scaled metrics keep their existing near-zero behavior; percentage and range metrics only treat exact zeros as zero.

Also:

- `ope` now accepts a negative `actual_series` sum.
- `arre` and `marre` handle zero-range components element-wise.
- Missing-only `wmape` and `ope` inputs remain `np.nan`.

### Tests

- `uv run pytest darts/tests/metrics/test_metrics.py` — 478 passed
- Python 3.12 core suite — 8,417 passed, 323 skipped
- `uv run pre-commit run --all-files` — passed
- `uv run make --directory ./docs build-all-docs` — passed

### DIFF
--- a/darts/metrics/metrics.py
+++ b/darts/metrics/metrics.py
@@ -13,8 +13,6 @@ import pandas as pd
 from darts.dataprocessing import dtw
 from darts.logging import raise_log
 from darts.metrics.utils import (
-    _PERCENTAGE_ZERO_DIVISION_ERROR,
-    _PERCENTAGE_ZERO_DIVISION_WARNING,
     METRIC_OUTPUT_TYPE,
     SMPL_AX,
     TIME_AX,
@@ -26,7 +24,7 @@ from darts.metrics.utils import (
     _get_values_or_raise,
     _get_wrapped_metric,
     _LabelReduction,
-    _safe_scaled_divide,
+    _safe_divide,
     classification_support,
     interval_support,
     multi_ts_support,
@@ -536,7 +534,7 @@ def ase(
         intersect,
         q=q,
     )
-    return _safe_scaled_divide(errors, error_scale, zero_division=zero_division)
+    return _safe_divide(errors, error_scale, zero_division=zero_division)
 
 
 @multi_ts_support
@@ -978,7 +976,11 @@ def sse(
         intersect,
         q=q,
     )
-    return _safe_scaled_divide(errors, error_scale, zero_division=zero_division)
+    return _safe_divide(
+        errors,
+        error_scale,
+        zero_division=zero_division,
+    )
 
 
 @multi_ts_support
@@ -1313,7 +1315,7 @@ def rmsse(
         intersect,
         q=q,
     )
-    return _safe_scaled_divide(errors, error_scale, zero_division=zero_division)
+    return _safe_divide(errors, error_scale, zero_division=zero_division)
 
 
 @multi_ts_support
@@ -1532,8 +1534,8 @@ def ape(
     .. math:: 100 \\cdot \\left| \\frac{y_t - \\hat{y}_t}{y_t} \\right|
 
     When :math:`y_t = 0` for some :math:`t` the term is undefined; the behavior is then controlled by
-    the ``zero_division`` parameter. Consider using the Absolute Scaled Error
-    (:func:`~darts.metrics.metrics.ase`) in these cases.
+    the ``zero_division`` parameter. Consider using the Absolute Scaled Error (:func:`~darts.metrics.metrics.ase`)
+    in these cases.
 
     If :math:`\\hat{y}_t` are stochastic (contains several samples) or quantile predictions, use parameter `q` to
     specify on which quantile(s) to compute the metric on. By default, it uses the median 0.5 quantile
@@ -1551,11 +1553,10 @@ def ape(
     q
         Optionally, the quantile (float [0, 1]) or list of quantiles of interest to compute the metric on.
     zero_division
-        Controls behavior when the denominator :math:`y_t` is zero (i.e. ``actual_series`` is zero
-        at some time step).
+        Controls behavior when the denominator :math:`y_t` is zero (i.e. ``actual_series`` is zero at some time step).
 
-        * ``"warn"`` (default) – returns ``0.0`` for a perfect forecast (numerator
-          also zero) and ``np.nan`` otherwise, and emits a warning.
+        * ``"warn"`` (default) – returns ``0.0`` for a perfect forecast (numerator also zero) and ``np.nan``
+          otherwise, and emits a warning.
         * ``"raise"`` – raises a ``ValueError``.
     time_reduction
         Optionally, a function to aggregate the metrics over the time axis. It must reduce a `np.ndarray`
@@ -1620,14 +1621,13 @@ def ape(
         q=q,
     )
     return 100.0 * np.abs(
-        _safe_scaled_divide(
+        _safe_divide(
             y_true - y_pred,
             y_true,
             zero_division=zero_division,
             zero_fill=0.0,
             strict_zero=True,
-            error_msg=_PERCENTAGE_ZERO_DIVISION_ERROR,
-            warning_msg=_PERCENTAGE_ZERO_DIVISION_WARNING,
+            metric_type="percentage",
         )
     )
 
@@ -1674,11 +1674,10 @@ def mape(
     q
         Optionally, the quantile (float [0, 1]) or list of quantiles of interest to compute the metric on.
     zero_division
-        Controls behavior when the denominator :math:`y_t` is zero (i.e. ``actual_series`` is zero
-        at some time step).
+        Controls behavior when the denominator :math:`y_t` is zero (i.e. ``actual_series`` is zero at some time step).
 
-        * ``"warn"`` (default) – returns ``0.0`` for a perfect forecast (numerator
-          also zero) and ``np.nan`` otherwise, and emits a warning.
+        * ``"warn"`` (default) – returns ``0.0`` for a perfect forecast (numerator also zero) and ``np.nan``
+          otherwise, and emits a warning.
         * ``"raise"`` – raises a ``ValueError``.
     component_reduction
         Optionally, a function to aggregate the metrics over the component/column axis. It must reduce a `np.ndarray`
@@ -1780,8 +1779,8 @@ def wmape(
         Controls behavior when the denominator :math:`\\sum_{t=1}^T |y_t|` is zero (i.e. ``actual_series`` is
         all zeros for a given component).
 
-        * ``"warn"`` (default) – returns ``0.0`` for a perfect forecast (numerator
-          also zero) and ``np.nan`` otherwise, and emits a warning.
+        * ``"warn"`` (default) – returns ``0.0`` for a perfect forecast (numerator also zero) and ``np.nan``
+          otherwise, and emits a warning.
         * ``"raise"`` – raises a ``ValueError``.
     component_reduction
         Optionally, a function to aggregate the metrics over the component/column axis. It must reduce a `np.ndarray`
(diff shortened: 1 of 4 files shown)


mahimn01:
Hi @dennisbader — gentle nudge on this when you get a chance. CI hasn't triggered yet (looks like it needs a maintainer approve-and-run for first-time contributors). It extends the #3059 `zero_division` pattern to the percentage/range metrics (`wmape`, `ope`, `arre`, `marre`, `coefficient_of_variation`), with tests. Thanks!

jakubchlapek:
Hey @mahimn01, thanks for this PR and contributing :). Will review it shortly as this is something we should definitely add, but it would be also great if you could create an issue for this first. Thanks!

darts/metrics/metrics.py
```diff
@@ -2226,16 +2252,10 @@ def arre(
         q=q,
     )
     y_max, y_min = np.nanmax(y_true, axis=TIME_AX), np.nanmin(y_true, axis=TIME_AX)
-    if not (y_max > y_min).all():
```
jakubchlapek, line 2225: why was this removed?

darts/tests/metrics/test_metrics.py
```diff
@@ -1474,6 +1476,78 @@ def test_scaled_errors_zero_division(
...
+        ]).reshape(-1, 1)
+        actual = TimeSeries.from_values(neg_sum_values)
+        pred = TimeSeries.from_values(neg_sum_values + 0.1)
+
+        caplog.clear()
+        with caplog.at_level(logging.WARNING):
+            result = metrics.ope(actual, pred)
+        assert "denominator" not in caplog.text
+        assert np.isfinite(result)
+
```
jakubchlapek, line 1634: i think it would be nicer to instead create a helper for negative series and ship this check as part of the `test_ope()` test (similarly to how it's structured there)

jakubchlapek (commented):
Hey @mahimn01, I've taken an initial look at the code, it looks fine, thanks. The `ope` clarification makes sense to me. One thing that we would definitely need to add is analogous support to all the percentage metrics (e.g. `ape, mape, sape, smape, wmape, ope, arre, marre`).
Before reviewing further I would like to discuss the fill semantics. For the scaled metrics the user could pass in the insample metric leading to the `zero_division` issue in the first place. Here as we don't supply that, there is no issue of matching the periodic series with forecasts. To align this to the existing solution I think what would make more sense than returning `nans` is returning a perfect forecast (e.g. 0/0 for MAPE returns 0% error). This would then depend on the metric and best score, so we can think on how to structure this best (maybe collapse it with the `_safe_scaled_divide` as the logic will be quite similar I think.) Let me know what you think on this :)

mahimn01:
Thanks @jakubchlapek, this makes sense — best-score for 0/0 is the right call, and it actually fixes a latent bug I'd missed: with NaN-everywhere, a genuinely perfect forecast on a degenerate series scores NaN (`ope` on a zero-sum series with an identical prediction; `arre`/`marre` on a constant series predicted perfectly).

I'll collapse `_safe_pct_divide` into `_safe_scaled_divide` — one helper with a configurable 0/0 fill (`1.0` scaled, `0.0` percentage), `x/0 → NaN`, and `"raise"` preserved (keeping each caller's existing message). A nice side effect of the shared helper's element-wise fill: for `arre`/`marre` it fixes a case the current code gets wrong — a constant-actual column is currently all-NaN even at timesteps where the per-step error is 0; element-wise gives `0/0 → 0` and `x/0 → NaN` within the same column.

On scope (full breakdown in the issue): `wmape`/`ope`/`arre`/`marre`/`coefficient_of_variation` are already in this PR — I'll switch them to the unified helper. Good news on the rest of the family: `sape`/`smape` already encode `0/0 → 0.0` inline and have no `x/0` case, so the only delta there is exposing the explicit `zero_division` knob for the `"raise"` option. `ape`/`mape` are the trickier ones — they still hard-raise on any `y_t == 0`, and because their denominator is the per-timestep actual (not a column scalar) they need element-wise handling; relaxing MAPE's default is also a larger behavioral change. Would you prefer `ape`/`mape` in this PR (default kept as raise, opt into fill via `zero_division`) or as a focused follow-up?

One small flag: `coefficient_of_variation` is `RMSE/mean` (not abs-wrapped, and the mean can be negative), so a zero mean is a legitimate non-degenerate case — I've proposed `0/0 → 0.0` / `x/0 → NaN` for it but happy to adjust.

Inline replies:
- *(arre, "why removed?")* that was the old `if not (y_max > y_min).all(): raise(...)` guard — the zero-range check moved into the helper, gated behind `zero_division` (default fills; `"raise"` re-raises, preserving the legacy behavior). The old message also said "MARRE" inside `arre()`, so that copy-paste slip drops out.
- *(test)* agreed — I'll fold the negative-sum check into `test_ope()` with a small negative-series helper. I'll also add a `0/0` best-score case to the zero-division test (the current fixtures are all `x/0`, so they stay NaN and need no change).

I've opened #3132 to track the design. Thanks for the thoughtful review!

mahimn01:
Hey @jakubchlapek — no rush, just a bump on this one whenever you get a chance.

I think we're aligned on the core: fold the percentage path into the unified `_safe_scaled_divide` with best-score for `0/0` and `x/0 → NaN`. I opened #3132 to track the design.

Two calls are yours before I finish the rework:
1. **`ape`/`mape`** — happy to fold them into this PR or split them into a focused follow-up. They're a bit trickier than the rest since their denominator is per-timestep (so element-wise handling), and relaxing MAPE's default is a slightly bigger behavioural change — so I'd rather defer to you on scope.
2. **`coefficient_of_variation`'s `0/0` fill** — I'm leaning `0.0` (RMSE = 0 is a genuinely perfect forecast), flagging it since its denominator can go negative.

Point me either way and I'll turn it around quickly, and rebase at the same time since master's moved.

jakubchlapek:
Hey @mahimn01, thanks for the update.
Regarding the `cv` 0/0 fill it definitely makes sense for me to use 0.0 Maybe we could extend the `_safe_scaled_divide` to accept a `best_score/zero_fill/fitting_name` parameter. to pass for a metric what should be the output in case of 0/0.

Regarding the `ape/mape` I would definitely include them in this PR. This being said I am not quite sure what exactly the issue with them is for this approach. Similarly to other metrics (like `coefficient_of_variation` which you already implemented) `ape` exposes y_true and y_pred which we would just need to similarly wrap with the division helper.  Similarly I don't get what you mean with the `relaxing MAPE's default`. Could you double check this and clarify? Thanks :)

mahimn01:
Hey @jakubchlapek, agreed on both.

On ape/mape, you're right. Once `_safe_pct_divide` is folded into `_safe_scaled_divide` (which fills element-wise) it falls out with no special handling. You pass `errors = y_true - y_pred` and `scale = y_true`, both `(t, c)`, and per element 0/0 gives 0.0 for a perfect forecast and x/0 gives NaN, same as `cv`. I'll include them.

On MAPE's default, ape and mape currently raise on a zero actual (it's documented), so defaulting `zero_division="warn"` changes that to warn-and-fill. Same raise→warn shift the PR already makes for the other five, so it stays under the 🔴 note. I'll default them to "warn" and update the docstrings.

I'll unify `_safe_pct_divide` into `_safe_scaled_divide` with a `best_score` arg (1.0 scaled, 0.0 percentage), route wmape, ope, arre, marre, cv, ape and mape through it, and rebase since master moved.

mahimn01:
Reworked and pushed, also rebased onto master. `_safe_pct_divide` is now folded into `_safe_scaled_divide` via the `best_score` arg (1.0 scaled, 0.0 percentage), and ape, mape, wmape, ope, arre, marre and cv all route through it. 0/0 gives the perfect-forecast score and x/0 gives NaN, filled element-wise so a constant `actual_series` no longer NaNs a whole component.

One heads-up, since mape is the default metric for `backtest` and `gridsearch`, the raise to warn change makes those surface NaN instead of raising on a zero actual, consistent with the #3059 scaled metrics.

darts/metrics/metrics.py
mahimn01: The zero-range check now happens in `_safe_scaled_divide`. With `zero_division="raise"` it still raises. The default handles each entry separately: `0/0` returns `0.0`, while a non-zero error over a zero range returns `NaN`.

darts/tests/metrics/test_metrics.py
mahimn01: Done. The negative-series case is now a small helper called from `test_ope()`, and it checks the expected OPE value.

dennisbader (approved):
Thanks a lot for this great PR @mahimn01 and your first contribution 🚀 

Also thanks @jakubchlapek for the earlier reviews.

The PR looks really good, I applied some minor changes. I'll merge once all tests have passed 💯

PR #3130, opened 2026-06-04
Add save/load methods to anomaly detection classes

### LINKED ISSUE #3051: Add `save()` and `load()` methods to`KMeansScorer`
Most models can be persisted to avoid future retraining. It would be nice to have that same capability for `KMeansScorer` specifically.

robrui:

Fixes #3051.

### Summary

Adds `save()` and `load()` methods to the four anomaly detection base classes (AnomalyScorer, Detector, Aggregator, AnomalyModel). This makes it possible to persist trained scorers like KMeansScorer, detectors like QuantileDetector, aggregators, and anomaly models to disk and reload them later without retraining. Uses the same pickle-based approach as ForecastingModel.save/load.

### Other Information

All subclasses inherit save/load automatically. The full anomaly detection test suite passes (229 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,12 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Improved**
 
+**Improved**
+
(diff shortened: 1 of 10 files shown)


dennisbader (changes requested):
Hi @robrui and thanks a lot for this PR 🚀 

I see currently a lot of duplicate code. Could you try to unify the logic (maybe using the existing logic for the ForecastingModel)?

robrui:
Thanks @dennisbader! You're right about the duplication. I pulled the common save/load logic into a shared base mixin so all four AD classes inherit from it now — same behavior, much less code.

PR #3131, opened 2026-06-04
feat/3037 phase 1 refactor return types

ozink-u8:

Acts as a preparation for #3037.

### Summary
**Problem:**
Many functions accepting `TimeSeriesLike` manually handle the single-vs-sequence type handling. The canonical example (a few sites vary slightly) is:

```python
# Pattern A — hand-rolled shape dispatch (before)
called_with_single_series = isinstance(series, TimeSeries)
series = series2seq(series) # normalize to a sequence
result = [process(ts) for ts in series]
return result[0] if called_with_single_series else result
```

**Change:**
Replace that scattered logic with a single centralized exit through `series2seq(...)`:

```python
# Pattern B — centralized via series2seq (after)
sequence_type_in = get_series_seq_type(series)   # capture input shape
series = series2seq(series)
result = [process(ts) for ts in series]
return series2seq(result, seq_type_out=sequence_type_in)   # restore to input shape
```

Funnelling all single-vs-sequence handling through series2seq not only creates a single source of truth, which also eases the planned TSS introducing changes (Phase 2).

### Other Information
- These changes should be behavior preserving
- Added overloads in `ts_utils.py` for `ty` typechecking.

#### Out of Scope
- Several sites are intentionally deferred to Phase 2, for different reasons:
    - `anomaly_model.py::score`, returns a list[tuple[TS, ...]] for each scorer.
    - `invertible_data_transformer.py::inverse_transform` deals with historical forecasts, therefore of type SEQ_SEQ and flattens and reconstructs them.
    - `forecasting_model.py::historical_forecasts`: SEQ_SEQ touchpoint with lpo=False
    - `forecasting_model.py::residuals`: Uses HFC therefore SEQ_SEQ touchpoint with lpo=False
    - `conformal_models.py::residuals`: Uses HFC
    - `conformal_models.py::_calibrate_forecasts` Uses residuals -> HFC
    - `historical_forecasts/utils.py::_apply_inverse_data_transformers`: Touches SEQ_SEQ with lpo=False. Right now in this case `series2seq` just leaves SEQ_SEQ unchanged.
    - `historical_forecasts/utils.py:: _process_historical_forecast_for_backtest` Uses HFC -> SEQ_SEQ
    - `datasets.py` Many different datasets have hardcoded list[TS] as return type -> Would change to TSS in Phase 2 (Electricity, UberTLC, ILINet, ExchangeRate, Traffic, Weather)
    - explainability/explainability_result.py::{get_explanation, get_feature_values, get_attention}: query_explainability_result is generic (also returns shap.Explanation/DataFrame), so wrap the multi-series list[TimeSeries] as TSS in the getters, not the helper.
    - `model_selection.py::make_splitter`: Returns tuple for train, test of list/SplitTimeSeriesSequence objects. Which violates our rule of outer list being different TSs -> Intentionally left out
    - `base_data_transformer.py::{apply_component_mask, unapply_component_mask}`: Mixed return types including np.ndarray -> Unclear if should be refactored in Phase 2
    - `metrics/utils.py` Uses Literals as return not TS -> not relevant for TSS changes
    - anomaly_model.py::eval_metric + ad/utils.py::eval_metric_from_binary_prediction: Pattern-A shape but return dict/float (no TS to wrap) -> out of scope, left unmarked.


darts/ad/anomaly_model/anomaly_model.py
```diff
@@ -59,6 +59,8 @@ def fit(
         )
         return self
 
+    # TODO(oswald): Discuss how we want to handle those nested lists.
```
dennisbader: I guess here you can also `get_series_seq_type` and then use `series2seq` at the end to covert to the input seq type (same as everywhere else). This should be done both for `scores` and `pred`. 

Apply the same to all other locations with the same question (for example the methods below should also be updated, ...), and make sure to remove the TODOs afterwards.

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -401,9 +401,8 @@ def transform(
         transformed_data = _parallel_apply(
             input_iterator, self._ts_transform, self._n_jobs, args, kwargs
         )
-        return (
-            transformed_data[0] if isinstance(series, TimeSeries) else transformed_data
-        )
+
+        return series2seq(transformed_data, seq_type_out=get_series_seq_type(series))
```
dennisbader: There are checks at the top of this method which should be updated to use the series seq type (`if isinstance(series, TimeSeries):` should rather check `if seq_type is SeriesType.SINGLE`)

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -498,6 +497,8 @@ def _generate_component_mask(
             )
         return component_mask
 
+    # TODO(oswald): Not sure if we should include this in the refactor
+    # TODO(oswald): I also think the typing here is too loose if input is type TS. perhaps it should be TimeSeriesLike
```
dennisbader: not required here since not TimeSeries-like. You can remove the comments

darts/dataprocessing/transformers/base_data_transformer.py
```diff
@@ -562,6 +563,7 @@ def apply_component_mask(
             out.append(out_)
         return out[0] if called_with_single_series else out
 
+    # TODO(oswald): Not sure if we should include this in the refactor
```
dennisbader: same here

darts/datasets/dataset_loaders.py
```diff
@@ -198,13 +199,15 @@ def _format_time_column(self, df):
         return df
 
 
+# TODO(oswald): All subclasses of DatasetLoaderCSV which use self._metadata.multivariate
```
dennisbader: I'm not sure I understand what you mean here. Could you explain further?

darts/datasets/datasets.py
```diff
@@ -848,6 +853,7 @@ def __init__(self, multivariate: bool = True):
             )
         )
 
+    # TODO(oswald): TSS migration — return TimeSeriesSequence instead of list[TimeSeries] once TSS lands.
```
dennisbader: You can remove this comment for now

darts/explainability/explainability_result.py
```diff
@@ -262,9 +263,10 @@ def __init__(
                 logger,
             )
 
+    # TODO(oswald): TSS migration — wrap `list[TimeSeries]` return as TimeSeriesSequence once TSS lands.
```
dennisbader: can you remove similar todos that you added to the code base? For the next phase we'll identify all affected locations, no need to add it everywhere already

darts/models/forecasting/ensemble_model.py
```diff
@@ -397,18 +399,17 @@ def _make_multiple_predictions(
             )
             for model in self.forecasting_models
         ]
+        # Normalizing predictions to SeriesType.SEQ so we can always use self._stack_ts_multiseq
+        predictions = [series2seq(pred) for pred in predictions]
```
dennisbader: Can you revert the changes here and in the method return? Running series2seq on the series individually is less efficient

darts/models/forecasting/forecasting_model.py
```diff
@@ -2347,15 +2349,24 @@ def residuals(
 
             residuals_out.append(res_list_out)
 
-        # if required, reduce to `series` input type
-        if series_seq_type == SeriesType.SINGLE:
-            return residuals_out[0][0] if last_points_only else residuals_out[0]
-
-        return (
+        # `residuals_out` is always nested `list[list[X]]` (per series, per forecast). We
```
dennisbader: can you revert these changes here and the returns? it was already doing the job and less verbose :)

darts/models/forecasting/naive_ensemble_model.py
```diff
@@ -117,21 +116,18 @@ def ensemble(
...
-                )
-                for p, ts in zip(predictions, series)
-            ]
-        else:
-            return (
-                self._target_average(predictions, series)
-                if not predict_likelihood_parameters
-                else self._params_average(predictions, series)
-            )
+        sequence_type_in = get_series_seq_type(predictions)
```
dennisbader, line 116: nice :+1

darts/utils/model_selection.py
```diff
@@ -170,6 +170,11 @@ def make_splitter(
         tuple[TimeSeries, TimeSeries]
         | tuple[Sequence[TimeSeries], Sequence[TimeSeries]]
     ):
+        # TODO(oswald): Intentionally left as hand-rolled Pattern A. Unlike the migrated
```
dennisbader: This is true for the return output conversion.

Still, for the input use `get_series_seq_type` and `series2seq`. For example the first check `if not isinstance(data, Sequence)` could resolve `False` for our new `TimeSeriesSequence`. We should avoid that

darts/utils/ts_utils.py
```diff
@@ -262,4 +264,5 @@ def retain_period_common_to_all(series: list[TimeSeries]) -> list[TimeSeries]:
             ValueError("The provided time series must have nonzero overlap"), logger
         )
 
-    return list(map(lambda s: s.slice(last_first, first_last), series))
+    common_series = [s.slice(last_first, first_last) for s in series]
+    return series2seq(common_series, seq_type_out=SeriesType.SEQ)
```
dennisbader: I would revert these two line changes, it's not part of this refactor

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 - Used GitHub Actions to publish the documentation to GitHub Pages, replacing the previous third-party branch-based deployment method. [#3107](https://github.com/unit8co/darts/pull/3107) by [Zhihao Dai](https://github.com/daidahao)
 - Removed `optuna-integration[pytorch-lightning]` from the testing dependencies. [#3114](https://github.com/unit8co/darts/pull/3114) by [Jakub Chłapek](https://github.com/jakubchlapek).
+- Refactored single-vs-sequence returns across the library to uniformly use `series2seq`, instead of per-site hardcoded `isinstance(series, TimeSeries)`. Added overloads to `series2seq` and `get_series_seq_type`. Behavior preserving groundwork for the upcoming `TimeSeriesSequence` migration ([#3037](https://github.com/unit8co/darts/issues/3037)). [#3131](https://github.com/unit8co/darts/pull/3131) by [Oswald Zink](https://github.com/ozink-u8).
```
dennisbader: Can you move this up into the new release (same section) and remove the overload mention?

dennisbader (changes requested):
Thanks a lot @ozink-u8, this is looking really good :rocket

Just added a couple of minor suggestions. After that it should be ready to be merged :)

darts/datasets/dataset_loaders.py
ozink-u8: DatasetLoaderCSV calls _to_multi_series(), but subclasses are not required to define it. 

A missing implementation would onlyshow as the “Could not convert to multi-series” error. This is out of scope here, so I’ll remove the TODO

darts/models/forecasting/forecasting_model.py
ozink-u8: Reverted all changes in this file

darts/utils/model_selection.py
ozink-u8: I added a small check if len(data) == 0 because the get_series_seq_type could fail which previously with the instance check would get caught later more explicitly in the constructor of SplitTimeSeriesSequence

dennisbader (approved):
Thanks for the updates @ozink-u8, looks great now 🚀 

PR #3133, opened 2026-06-15
Reduce metric evaluation overhead

GarrettMooney:
## Summary

Metric evaluation carried avoidable per-call overhead unrelated to the actual computation:

1. The metric decorators (`multi_ts_support`, `multivariate_support`, `classification_support`) called `signature(func).parameters` on **every** invocation, just to read parameter defaults. The signature is constant per decorated metric, so this is now computed once at decoration time.
2. `_parallel_apply` routed even single-series / `n_jobs=1` evaluations through `joblib.Parallel`, paying its setup cost when no parallelism is used. A sequential fast path is added for `n_jobs == 1` (the default); the `n_jobs != 1` path is unchanged.

## Measured impact

Benchmarked at `N=20000`, best-of-N timing, on `LinearRegression`-style float data:

| Metric | Before | After | Speedup |
|---|---|---|---|
| `mse` (single series) | 126 µs | 81 µs | 1.57x |
| `mape` (single series) | 139 µs | 96 µs | 1.45x |
| `mae` (single series) | 128 µs | 87 µs | 1.46x |
| `mse` (8-series sequence) | 511 µs | 372 µs | 1.37x |

Outputs are bit-identical before and after. The win compounds in `backtest()` (one metric per window) and `gridsearch()` (one per param-combination × window), where metrics are evaluated thousands of times.

## Testing

- `darts/tests/metrics/test_metrics.py`: **455 passed**.
- Manual check: identical numeric results pre/post change across `mse`/`mae`/`mape` for single series and sequences.

## Notes

This is a draft. I have only run the metrics test module locally, not the full suite. The `n_jobs == 1` fast path is semantically equivalent to `joblib.Parallel(n_jobs=1)`, which already executes sequentially.

### DIFF
--- a/darts/metrics/utils.py
+++ b/darts/metrics/utils.py
@@ -126,6 +126,8 @@ def classification_support(func) -> Callable[..., METRIC_OUTPUT_TYPE]:
     probabilities and categorical samples.
     """
 
+    params = signature(func).parameters
+
     @wraps(func)
     def wrapper_classification_support(*args, **kwargs):
         labels = kwargs.get(_PARAM_LABELS)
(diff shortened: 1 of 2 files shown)


dennisbader:
@GarrettMooney thanks for the PR :) can I ask why you closed it?

PR #3134, opened 2026-06-18
warn when add_encoders is set but fit_from_dataset() is called

### LINKED ISSUE #3006: [Feature] Improve handling & documentation of encoders and categorical static covariates in .fit_from_dataset()

DresdenGman:
Fixes the silent behavior described in #3006.

When a model is created with add_encoders and fit_from_dataset() is called instead of fit(), the encoders are quietly ignored with no indication that encoded covariates won't be included in training.

Added a logger.warning that fires when self.add_encoders is set, and a note in the docstring so it's visible in the API docs too. No behavior changes beyond the new warning.

Closes #3006


dennisbader:
@DresdenGman not sure what you were committing, but the diffs look fishy. Also, we're already working on this in #3086. Closing this one.

PR #3143, opened 2026-06-26
Warn and document encoder caveats for dataset-based torch methods

kiwoongyoon:
Title: Warn and document encoder caveats for dataset-based torch methods

## Summary

- Warn when `fit_from_dataset()` is called on a torch forecasting model configured with `add_encoders`, because those encoders are not applied to provided datasets.
- Document that `fit_from_dataset()` and `predict_from_dataset()` require encoded covariates to already be present in the supplied datasets.
- Clarify that TFT categorical static covariate embedding keys should be zero-based column indices when training from datasets.
- Add a focused regression test for the new `fit_from_dataset()` warning.

## Test evidence

- `python -m py_compile darts/models/forecasting/torch_forecasting_model.py darts/models/forecasting/tft_model.py darts/tests/models/forecasting/test_torch_forecasting_model.py`
- `git diff --check`
- `python -m pytest darts/tests/models/forecasting/test_torch_forecasting_model.py::TestTorchForecastingModel::test_fit_from_dataset_warns_with_encoders -q`
  - Result: `1 passed, 5 warnings in 2.78s`

## Risks or notes for maintainers

- The warning is triggered by a non-empty `add_encoders` configuration, matching the user-facing condition "model initialized with encoders." This avoids a false negative from checking `encoding_available`, which can be false before normal `fit()` encoder setup.
- This does not add encoder support to dataset classes; it only documents and warns about the current behavior.


dennisbader:
Thanks @kiwoongyoon for this PR 🚀 We do however already have an open PR #3086 that @jakubchlapek is working on. 

Closing this one in favor of the other. 

Thanks again though for the contribution 🙏

kiwoongyoon:
@dennisbader Thank you for the reply !!

PR #3158, opened 2026-07-09
Fix spurious dtype-mismatch warning after loading a torch model from checkpoint

### LINKED ISSUE #3156: [BUG] Warnings regarding different input and output data types even if they are the same when loading models from checkpoint
**Describe the bug**
When fitting neural network models (e.g. TiDEModel) with Darts and loading the best model from checkpoint after early stopping, warnings are outputted regarding an inconsistency between input and output data types (even if both are float32) when the predict function is called and an input series is passed.

`Dataset output has a different data type than the dataset the model was trained on; current data type: float32, expected data type: float32. This might cause downstream issues when running the model. If so, make sure all your input data have the expected data type (TimeSeries, static covariates, ...).`

**To Reproduce**
Below is some sample code to reproduce this warning.

```
from darts.models import (
    TiDEModel
)
import numpy as np
from pytorch_lightning.callbacks import EarlyStopping

whole_ts = AirPassengersDataset().load().astype(np.float32)
train = whole_ts[:-52]
val = whole_ts[-52:]
early_stopping = EarlyStopping(monitor="val_loss", patience=5, min_delta = 0.001)
my_params = {"input_chunk_length": 32, "output_chunk_length": 4, "batch_size": 16, "save_checkpoints": True, "force_reset": True, "random_state": 42, "pl_trainer_kwargs": {"callbacks": [early_stopping]}}
curr_model = TiDEModel(**my_params)

curr_model.fit(series = train, val_series = val)

best_curr_model = curr_model.load_from_checkpoint(curr_model.model_name, best = True)

#Warnings about dtypes
best_curr_model.predict(series = train, n = 4, num_samples = 10)

#No warnings
curr_model.predict(series = train, n = 4, num_samples = 10)
```

**Expected behavior**
Fitting and prediction should proceed without this warning as dtypes should be consistent (input and output are both float32).

 - Python version: 3.12.12
 - u8darts version: 0.45.0
 - u8darts-all version: 0.45.0
 - numpy version: 2.4.6
 - pytorch-lightning version: 2.5.2
 - pytorch 2.10.0

Hardware & OS: Apple M4, MacOS 26.5.2

Would the results of the modelling still be correct in this case?

Mohit-Ak:
### What this fixes

`_verify_dtypes` runs at prediction time to warn when the inference data's dtype
differs from the dtype the model was trained on. When a model is reloaded from a
checkpoint (e.g. via `load_from_checkpoint(..., best=True)` after early stopping),
the warning was raised even though the input and training dtypes were identical:

```
Dataset output has a different data type than the dataset the model was trained on;
current data type: float32, expected data type: float32. ...
```

Reported in #3156 with a minimal `TiDEModel` repro.

### Root cause

The dtype comparison used the identity operator:

```python
if current_dtype is not expected_dtype:
    logger.warning("Dataset output has a different data type ...")
```

`expected_dtype` comes from `self.train_sample[0].dtype`. After a checkpoint
reload the training sample's numpy dtype is a freshly unpickled object, so it is
value-equal to the inference dtype but not the *same* object. `is not` therefore
evaluates `True` for two equal float32 dtypes and the warning fires spuriously.

Small reproduction of the underlying pitfall:

```python
>>> import numpy as np, pickle
>>> d1 = np.dtype(np.float32)
>>> d2 = pickle.loads(pickle.dumps(np.dtype(np.float32)))
>>> d1 == d2, d1 is d2
(True, False)
```

### The fix

Compare numpy dtypes by value instead of by identity — a one-line change in
`_verify_dtypes`:

```python
-            if current_dtype is not expected_dtype:
+            if current_dtype != expected_dtype:
```

`numpy.dtype.__eq__` is the intended way to compare dtypes and correctly returns
`True` for two value-equal dtypes regardless of object identity, so a genuine
mismatch (e.g. float32 vs float64) still warns as before.

### How it was tested

Two regression tests were added to
`darts/tests/models/forecasting/test_torch_forecasting_model.py`:

- `test_verify_dtypes_matching_dtype_no_warning` — sets `train_sample`'s dtype to
  a pickle-round-tripped (value-equal, identity-distinct) float32 and asserts
  `_verify_dtypes` emits **no** "different data type" warning. This fails on the
  old `is not` code and passes with the fix.
- `test_verify_dtypes_mismatched_dtype_warns` — a genuine float32-vs-float64
  mismatch still logs the warning, guarding against over-correction.

Verification against the repo's own suite (CPU-only install):

```
# with the fix
$ pytest darts/tests/models/forecasting/test_torch_forecasting_model.py -k verify_dtypes
6 passed, 369 deselected

# reverting the fix in place -> the new test reproduces the bug
FAILED ...::test_verify_dtypes_matching_dtype_no_warning
  - AssertionError: assert 'different data type' not in 'WARNING ...'

# broader dtype/precision regression guard
$ pytest ... -k "dtype or precision"
14 passed, 361 deselected
```

Lint (pinned `ruff==0.7.2` from `.pre-commit-config.yaml`):

```
$ ruff check <changed files>         -> All checks passed!
$ ruff format --check <changed files> -> 2 files already formatted
```

A CHANGELOG.md entry was added under the Unreleased "Fixed" section.

Fixes #3156

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
+- Fixed a bug in the torch-based forecasting models where a spurious "different data type" warning was emitted at prediction time — even when the input and training data types matched — after loading a model from a checkpoint. The dtype check now compares numpy dtypes by value rather than by object identity. [#3156](https://github.com/unit8co/darts/issues/3156).
 - Fixed a bug in `TimeSeries.window_transform()` where `treat_na` (`"dropna"`, scalar, `"bfill"`) did not handle NaN values introduced by functions that produce NaN even when `min_periods` is satisfied (e.g., `std` with `ddof=1` on a single value). [#3151](https://github.com/unit8co/darts/pull/3151) by [Dennis Bader](https://github.com/dennisbader).
 
 **Dependencies**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
-- Fixed a bug in the torch-based forecasting models where a spurious "different data type" warning was emitted at prediction time — even when the input and training data types matched — after loading a model from a checkpoint. The dtype check now compares numpy dtypes by value rather than by object identity. [#3156](https://github.com/unit8co/darts/issues/3156).
+- Fixed a bug in the `TorchForecastingModel.predict()` which emitted an invalid "different data type" warning for correct input data types after loading a model from a checkpoint. [#3158](https://github.com/unit8co/darts/pull/3158) by [Mohit Arvind Khakharia](https://github.com/Mohit-Ak).
 - Fixed a bug in `TimeSeries.window_transform()` where `treat_na` (`"dropna"`, scalar, `"bfill"`) did not handle NaN values introduced by functions that produce NaN even when `min_periods` is satisfied (e.g., `std` with `ddof=1` on a single value). [#3151](https://github.com/unit8co/darts/pull/3151) by [Dennis Bader](https://github.com/dennisbader).
 
 **Dependencies**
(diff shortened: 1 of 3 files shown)


dennisbader (approved):
Thanks a lot for the PR, the quick fix, and your first contribution @Mohit-Ak 🚀 

The fix looks good. I slightly updated the test to check the actual model loading case.

Let's wait for the tests to succeed, then I'll merge 💯

PR #3167, opened 2026-07-18
Add optional `minimum_length` to local models (fixes #3001)

### LINKED ISSUE #3001: [BUG] AutoETS fails on short time series
**Describe the bug**
When I fit AutoETS to short time series, it gives me an error:

>ValueError: Train series only contains 9 elements but AutoETS(season_length=1) model requires at least 10 entries

The pure Nixtla implementation does not have this issue.

**To Reproduce**
```python
from fcompdata import M1
from darts import TimeSeries
from darts.models import AutoETS
import numpy as np

ts = M1[9]

train_series = TimeSeries.from_values(np.array(y_train))

model = AutoETS(season_length=ts.period)
model.fit(train_series)
```

**Expected behaviour**
This should not give the error. R implementations select ETS(M,A,N) on this data.

 - Python version: 3.13.7
 - darts version 0.40.0

haiiibin:
Fixes #3001.

## Problem

StatsForecast models (e.g. `AutoETS`) fail on series shorter than 10 points because `StatsForecastModel._target_window_lengths` hardcodes a conservative minimum input window of `10`, even though the underlying statsforecast models handle shorter series. As reported in the issue, this rejected ~1,500 of 5,315 series across the M1/M3/Tourism datasets.

## Change

Following the design discussed in the issue, this adds an optional `minimum_length` parameter at the `ForecastingModel` level (similar to `add_encoders`), exposed **only on local models** (global models are unchanged):

- `ForecastingModel.__init__` accepts and validates `minimum_length` (a strictly positive int, or `None`) and stores it; a small helper `_min_train_input_length(default)` returns the override when set.
- The override is applied inside `_target_window_lengths` (as suggested), so it propagates to `min_train_series_length` and to everywhere the input window length is used (e.g. determining backtesting start points).
- Wired into: `LocalForecastingModel` (default), the whole StatsForecast family (`StatsForecastModel` plus `AutoARIMA`/`AutoETS`/`AutoCES`/`AutoTheta`/`AutoMFLES`/`AutoTBATS`/`Croston`/`TBATS`), and `ExponentialSmoothing`.
- `None` default preserves current behavior exactly.

```python
from darts.models import AutoETS
model = AutoETS(season_length=1, minimum_length=7)  # previously hardcoded to 10
model.fit(series_of_length_9)  # now works
```

## Tests

- Reproduces #3001 (AutoETS on a 9-point series: raises by default, succeeds with `minimum_length`).
- Default unchanged (`min_train_series_length == 10`); override propagates to `_target_window_lengths`; validation rejects non-positive / non-int values; the parameter is available across the SF family and is **not** exposed on global models.
- Existing `test_sf_models.py` (25) and `test_exponential_smoothing.py` (12) still pass; `ruff` clean.

## Open questions

- **Naming:** I went with `minimum_length`; happy to rename (e.g. to `min_train_length`) if you prefer.
- **Scope:** I wired the StatsForecast family and `ExponentialSmoothing` (the models named in the issue). Would you like me to extend it to the remaining local models (Prophet, Theta / FourTheta, ARIMA, VARIMA, FFT, KalmanForecaster, ...) in this PR, or as a follow-up? The mechanism lives at the base, so extending each model is a one-line change in its `_target_window_lengths` plus threading the constructor argument.


darts/models/forecasting/exponential_smoothing.py
```diff
@@ -75,6 +76,10 @@ def __init__(
             for more information.
         random_state
             Controls the randomness for reproducible forecasting.
+        minimum_length
+            Optionally, override the minimum required training series length. Set a smaller positive
+            integer to fit on shorter series than darts' default requirement for this model. Default:
+            ``None`` (keeps the default requirement).
```
dennisbader: Let's adapt this slightly. Could you also apply this to all other docstrings?

```suggestion
        minimum_length
            Optionally, set a custom minimum required training series length for this model to allow training on 
            shorter input series. By default, Darts sets a conservative minimum length to avoid downstream issues. 
            Changing this value might lead to such downstream issues. Default: ``None`` (keeps the default requirement).
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -168,6 +168,24 @@ def __init__(self, *args, **kwargs):
...
+        # optional override of the minimum required training series length. Only
+        # relevant for (typically third-party) local models whose exact minimum
+        # length darts cannot infer and therefore sets conservatively. `None`
+        # keeps the model's default requirement.
+        minimum_length = kwargs.get("minimum_length", None)
+        if minimum_length is not None and (
+            not isinstance(minimum_length, int)
+            or isinstance(minimum_length, bool)
+            or minimum_length < 1
+        ):
```
dennisbader:
```suggestion
        if minimum_length is not None and (
            not isinstance(minimum_length, int)
            or minimum_length < 1
        ):
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -168,6 +168,24 @@ def __init__(self, *args, **kwargs):
         self.add_encoders = kwargs["add_encoders"]
         self.encoders = self.initialize_encoders(default=True)
 
+        # optional override of the minimum required training series length. Only
+        # relevant for (typically third-party) local models whose exact minimum
+        # length darts cannot infer and therefore sets conservatively. `None`
+        # keeps the model's default requirement.
```
dennisbader:
```suggestion
        # optionally, override of the minimum required training series length. Only
        # relevant for (typically third-party) local models whose exact minimum
        # length Darts cannot infer and therefore sets conservatively.
```

darts/models/forecasting/forecasting_model.py
```diff
@@ -168,6 +168,24 @@ def __init__(self, *args, **kwargs):
...
+        minimum_length = kwargs.get("minimum_length", None)
```
dennisbader: As you suggested, I would say let's rename it to `min_train_length` to make it more clear what the parameter does

darts/models/forecasting/sf_model.py
```diff
@@ -124,6 +125,12 @@ def encode_year(idx):
             with `num_samples > 1` or `predict_likelihood_parameters=True`.
         random_state
             Controls the randomness for reproducible forecasting.
+        minimum_length
+            Optionally, override the minimum required training series length. Darts sets a conservative
```
dennisbader: Can we make sure to use the same docstring throughout all Darts models where this has been added? The only difference should the default that Darts uses for the current model. This could be the last sentence in every docstrnig. Reason behind it is to have it uniform across Darts, and also to adapt docs later on (e.g. find & replace)

dennisbader (changes requested):
Thanks a lot for this nice PR @haiiibin and for your first contribution 🚀 

It looks very good already, only added a couple of minor suggestions.

As you mentioned, could you add the parameter to more models?

- ARIMA, VARIMA
- Theta, FourTheta
- Prophet
- FFT
- KalmanForecaster

haiiibin:
Thanks for the review! All points addressed:

- Renamed `minimum_length` to `min_train_length` throughout.
- Applied the suggested validation, comment, and docstring changes. The docstring is now identical across all models. I kept it fully identical (without a per-model default in the last sentence) because the default is instance-dependent for several models (e.g. `Theta` depends on the inferred seasonality). Happy to add explicit values where they are static if you prefer.
- Extended the parameter to `ARIMA`, `VARIMA`, `Theta`, `FourTheta`, `Prophet`, `FFT`, and `KalmanForecaster`. Models inheriting the base local `_target_window_lengths` only needed the constructor argument threaded through; `ARIMA`/`VARIMA` (fixed `30`) and `Theta`/`FourTheta` (seasonality-dependent) apply the override inside their own `_target_window_lengths`.
- Added a parametrized test covering the newly extended models (`test_local_forecasting_models.py`), updated the repr expectations, and removed the bool case from the validation test since the simplified validation accepts `True` as an `int`.
- Merged your master sync and resolved the changelog conflict.

Local runs: `test_sf_models` + `test_exponential_smoothing` + the new/affected `test_local_forecasting_models` tests all pass, `ruff` clean.

dennisbader (approved):
Thanks a lot for the updates @haiiibin 🚀 

Everything looks great now! I pushed some minor updates to make it ready for merging. 

Let's wait for all tests to pass, then I'll merge.

Congrats to your first contribution :)

PR #3168, opened 2026-07-18
Devops/fix numba and llmvlite uv resolution

dennisbader:

### Summary

- resolves installation issues for system_platform="darwin", platform_machine="x86_64" (e.g. older Intel-based macOS devices) due to numba not shipping wheels for it anymore. Only SHAP and PyOD still depend on numba.
- 🔴 Moved `shap` and `pyod` (along with their transitive `numba` dependency) from core dependencies to optional extras making Darts more light-weight (saving around 150 MB in install size). Install them with `pip install "darts[shap]"` and `pip install "darts[pyod]"` respectively or read our [installation guide](https://github.com/unit8co/darts/blob/master/INSTALL.md) for more detailed instructions. `ShapExplainer` and `PyODScorer` now gracefully return a `NotImportedModule` stub when their dependencies are missing.


daidahao:
@dennisbader Sorry, I don't see what the issue is here.

AFAIK, older numba wheels are available and macOS Intel users can install older wheels to use shap and pyod.

If the install size is a concern here, I am currently helping shap to migrate to nanobind and drop numba support very soon. In the near future, numba won't be needed for shap anymore.

In my opinion, removing shap from core could a big loss for users who want easy access to the explainability module.

dennisbader:
Hey @daidahao, thanks for the feedback! Let me clarify the actual issue, because it's not primarily about install size.

The problem is that numba dropped wheels for macOS x86_64 starting from v0.63.0. While older wheels do exist, resolvers like uv fail to resolve the dependency chain because they try the latest numba first, hit the missing wheel, and then run into version conflicts with llmvlite, numpy when backtracking to older numba versions. The result is that darts as a whole could not be installed at all via uv (at least not on my device and other colleagues, I'm running on "darwin" with "arm64", so not even Intel-based) on affected platforms -- not just the shap/pyod parts. So this was blocking all users, not just those who use explainability.

By moving shap and pyod to optional extras, the core install resolves cleanly on all platforms, while we can apply the necessary platform markers for numba only within the extras where it's actually needed.

As for ease of use, users who want explainability simply install with:

```
pip install "darts[all,shap]"
```

or even combine everything:

```
pip install "darts[all,shap,pyod]"
```

This is the same pattern we already use for torch/notorch and that users are familiar with. I don't think adding ",shap" is a meaningful barrier, especially for a feature that targets more advanced users who are already comfortable with model interpretation workflows.

Also think of inference-only environments where you do not do not need any explainability. Having a lightweight package is always good. 

And regarding the nanobind migration for shap, that's great news! Once shap drops numba, we can simply remove the numba platform markers from the shap extra, and darts[shap] becomes a trivially lightweight addition. PyOD will probably not migrate anytime soon, so the numba issue will persist.

daidahao:
@dennisbader Thanks for the clarification. The root cause here is the uv resolution on macOS, rather than shap/pyod/numba.

May I suggest a more immediate fix: add numba dependency implicitly to `core`, rather than dropping shap & pyod. Unlike the current PR, doing so would not alter the uv environment of Darts users and not break existing pipelines that use pyod or shap.

The other reason I am against adding `shap` and `pyod` extra NOW is the worry that we could risk segregating the Darts package and creating too many extras. Are we going to create an extra for every new dependency in the future? Shouldn't we group dependencies by their scope (like `torch` vs `notorch`, rather than names? Like someone has said (I am badly rephrasing here), Darts could be collection of tools and models users could easily deploy. Creating a new extra for every dependency seems to be a pull away from that vision. 

In the longer term, I do support removing numba dependency from Darts for various reasons (uv resolution, compatibility, size, etc.). Once numba is removed from shap, this can be done in Darts by:
- Removing the explicit numba tag from `core` and
- Creating a new extra `ad` for pyod and other detection-related dependencies
- Keeping shap in `core`.

In terms of install size, shap is lightweight package on its own with [less than 1MB on most platforms](https://pypi.org/project/shap/0.52.0/#files). Most of [its dependencies](https://github.com/shap/shap/blob/846f28c79ad0ca40044bc31703a4fccbe8fcaf17/pyproject.toml) overlap with Darts, except for numba, llvmlite (both will be removed soon), cloudpickle, slicer, packaging (all lightweight by their own). Keeping it in core avoids a `shap` extra and gives users easy access to explainability.

What do you think?

dennisbader:
@daidahao thanks for the input. I do agree that we can go with a hot fix for now and tackle the numba removal another time once shap removed it from the dependencies. 

I moved shap and pyod back into the core dependencies, and added the numba & llmvlite handling to the core as well.

Regarding your points about extras: we already have many optional dependencies (not even extras) which the users have to install manually (e.g. neuralforecast, onnx, polars, ...) So even with "darts[torch]" you won't get neuralforecast, which in my opinion is valid, as not every user actually wants to use these optional dependencies.

Our `NotImportedModule` also handles the missing dependencies by raising an exception with installation instructions. E.g.

```py

LightGBMModel()
```

raises 

```
ImportError: The `LightGBM` module could not be imported. To enable LightGBM support in Darts, follow the detailed instructions in the installation guide: https://github.com/unit8co/darts/blob/master/INSTALL.md
```

I find this quite intuitive but I'm always open for suggestions in making Darts more user-friendly :)

PR #3181, opened 2026-08-16
Add save() and load() methods to KMeansScorer

### LINKED ISSUE #3051: Add `save()` and `load()` methods to`KMeansScorer`

Boubker10:
Fixes #3051

### Summary

Adds `save()` and `load()` methods to `KMeansScorer`, pickle-based,
mirroring the existing `ForecastingModel.save()`/`load()` pattern.
Lets a fitted scorer be persisted to disk and reloaded without
retraining.

### Other Information

This overlaps in scope with #3130, which adds save/load more broadly
across the anomaly-detection base classes via a shared mixin. Opening
this narrower, `KMeansScorer`-only version for comparison / in case a
scoped fix is still wanted independently — happy to close in favor of
#3130 if that's preferred.


jakubchlapek:
Hey @Boubker10, thanks for opening this PR :) looking at this and #3130, I'd definitely say we'd prefer to go with the already open PR and cover all cases, as opposed to this narrower solution, looks good though, looking forward to some contributions in the future :)

PR #3183, opened 2026-08-19
Safe-by-default model loading for torch models (weights_only=True)

### LINKED ISSUE #3177: Safe-by-default model loading
### Problem

Loading a saved model through the public `load()` API can run arbitrary code if the checkpoint comes from an untrusted source (shared file, download, etc.). That is the usual pickle/torch unpickling risk, and it affects the normal "save and load a trained model" workflow.

Two places in the codebase:

1. **`TorchForecastingModel`** (`darts/models/forecasting/torch_forecasting_model.py`, around lines 2388, 2488, 2629): calls `torch.load(..., weights_only=False)`, which turns off PyTorch 2.6+'s safer default.
2. **`ForecastingModel.load()`** (`darts/models/forecasting/forecasting_model.py`, around line 2737): uses raw `pickle.load` for non-torch models.

Both are reachable through the documented API with no extra flags.

### What we need from a fix

- Safe for the common case without asking users to configure trust lists.
- Still works with the third-party backends we support (they store things differently, and not all handle save/load themselves).
- Keeps current behavior where possible: creation params (including stateful ones), model attributes, etc. If something has to change, we should call it out explicitly.

### Proposed plan

**Phase 1: `TorchForecastingModel`**

- Use `torch.load(..., weights_only=True)`.
- Register the classes we actually serialize via `torch.serialization.add_safe_globals([...])` (darts types, PyTorch Lightning, whatever the checkpoint needs).
- Store model weights, ...  as safetensors?
- Drop the global `torch.load` monkeypatch if we still have it.

This should be a standalone PR and fixes the torch path without changing normal usage.

**Phase 2: `ForecastingModel`**

- Replace raw `pickle.load` with a safer unpickler (for example `find_class` allowlist, ...).
- Start with what we know we store: `darts.*`, numpy/pandas, sklearn/statsmodels basics, stdlib types we rely on.
- Reject obvious bad imports (`os`, `subprocess`, `eval`, etc.).

For saves that contain something outside the allowlist (custom estimator, third-party stateful object in creation params), fail with a clear message and let the user opt in, e.g. `ForecastingModel.load(path, trusted=True)`. Optional: `trusted_classes=[...]` for power users.

### Suggested order

1. PR for torch loading.
2. PR for general ForecastingModel loading (iterate on the allowlist with real saved models).
3. Docs on new save / load behavior, explain when `trusted`, `trusted_classes` are needed and how to fix failing loading.

hackchang:
Fixes #3177.

## What

Makes torch-model checkpoint loading safe-by-default against CWE-502 (a malicious `.ckpt` executing arbitrary code during unpickling), while keeping legitimate models loading unchanged.

- Registers Darts' own checkpoint-serialized **types** (likelihoods + `LikelihoodType`, optimizer/scheduler/loss/activation classes, torchmetrics metric classes + their tensor-reduction helpers) via `torch.serialization.add_safe_globals`, so a `weights_only=True` load of a legitimate checkpoint succeeds. Only class/type and argument-less library helpers are allow-listed — never callables that run on unpickle — so this does not re-open the RCE surface. No-op on torch < 2.6.
- Lightning-checkpoint paths (`_DartsCheckpointIO`, `load_from_checkpoint`, `load_weights`, `load_weights_from_checkpoint`) now default to `weights_only=True`, with a user-facing `weights_only` parameter to opt back into full unpickling for trusted files.
- The full-object `.pt` path in `load()` (an arbitrary object graph that cannot use `weights_only=True`) keeps `weights_only=False` by default but exposes the parameter and a security note.

## Behavior changes (please note)

- `load_from_checkpoint()`, `load_weights()`, `load_weights_from_checkpoint()` now default to `weights_only=True`. Checkpoints using custom loss/optimizer/likelihood classes defined **outside** darts/torch may need `weights_only=False`.
- `load_weights_from_checkpoint()` previously **raised** on `weights_only=True`; it now works and is the default.

## Testing

- New security regression test (`test_load_checkpoint_weights_only_blocks_malicious_payload`): a normal model still loads under the safe default; a crafted `__reduce__` payload is refused under `weights_only=True` (no code execution) and only loads under the explicit `weights_only=False` opt-out.
- Full `test_torch_forecasting_model.py` suite passes locally (350 passed) on torch 2.13 / lightning 2.6.5.

## Transparency

This contribution was prepared with AI assistance; I reviewed, tested, and take responsibility for every change, and remain in the loop throughout, per the project's guidelines.


dennisbader:
Hi @hackchang not sure whether you've seen it but the checks failed (linting and unit tests). Could you fix the issues? After that I can review

hackchang:
Thanks @dennisbader — fixed both: ran ruff format for the lint failure, and the test failures were Darts' NeuralForecast-wrapped models (neuralforecast.models.*) not being in the weights_only=True allow-list — now registered when neuralforecast is installed. Verified the previously-failing save/load + nf_model tests pass locally. Pushed; CI should be green now.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -104,21 +104,172 @@
...
+        # register every metric class exposed by ``torchmetrics`` and its submodules.
+        safe_globals.extend(_subclasses(torchmetrics, torchmetrics.Metric))
+        for _sub in ("regression", "classification", "aggregation"):
+            _mod = getattr(torchmetrics, _sub, None)
+            if _mod is not None:
+                safe_globals.extend(_subclasses(_mod, torchmetrics.Metric))
+        # ``torchmetrics`` metrics store references to their own tensor-reduction /
+        # distributed helper *functions* in their pickled state (``_reductions``,
+        # ``dist_sync_fn``, ...). A ``weights_only=True`` load of a metric-bearing
+        # checkpoint needs these specific, known-safe library callables allow-listed.
```
dennisbader: We're allow-listing functions here, not just classes. PyTorch's serialization docs say registered functions can be called during unpickling. I don't think that's equivalent to "does not re-open the RCE surface" — worth being honest about in the PR description.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -104,21 +104,172 @@
 _PL_2_6_OR_ABOVE = tuple(int(el) for el in pl.__version__.split(".")[:2]) >= (2, 6)
 
 
-class _DartsCheckpointIO(TorchCheckpointIO):
-    """Custom CheckpointIO that defaults ``weights_only`` to ``False``.
+def _register_safe_globals() -> None:
```
dennisbader: This runs at import time for every torch model. That means we unconditionally import torchmetrics (and optionally neuralforecast) even if the user never touches checkpoint loading. Could we defer this to first load, or use the safe_globals context manager scoped around the actual torch.load call?

Also, scanning vars(module) for every loss/activation/optimizer/metric subclass feels brittle — torchmetrics adds subpackages over time and we'll keep chasing them. Have you tried get_unsafe_globals_in_checkpoint() on the specific file being loaded?

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -104,21 +104,172 @@
...
+    (resume training, Tuner, etc.) inherit the safe default. Callers that trust a
+    checkpoint and need full unpickling can still pass ``weights_only=False`` explicitly.
     """
 
     def load_checkpoint(self, path, map_location=None, weights_only=None, **kwargs):
         weights_only_kwargs = dict()
         if _PL_2_6_OR_ABOVE:
             weights_only_kwargs["weights_only"] = (
-                False if weights_only is None else weights_only
+                True if weights_only is None else weights_only
```
dennisbader: Flipping the default from False to True is a behavior change for all Lightning-internal checkpoint loads, not just our public APIs. fit(ckpt_path=...) still passes weights_only=False explicitly, but what about other trainer paths that go through the plugin without overriding? Worth a test for resume-from-checkpoint with optimizer state.

dennisbader: Comment on lines 772-774 still says we force weights_only=False for internal loading, but _DartsCheckpointIO now defaults to True. Stale, or did I miss something?

darts/models/forecasting/torch_forecasting_model.py
dennisbader, line 2699: Just flagging for the PR description: even with this change, load_from_checkpoint still fully unpickles the base model file first. So the "safe-by-default against CWE-502" claim only really applies to the .ckpt half. The docstring warning helps but the top-level description oversells it.

darts/models/forecasting/torch_forecasting_model.py
```diff
@@ -2330,11 +2486,21 @@ def save(
...
-        path: str, pl_trainer_kwargs: dict | None = None, **kwargs
+        path: str,
+        pl_trainer_kwargs: dict | None = None,
+        weights_only: bool = False,
+        **kwargs,
     ) -> "TorchForecastingModel":
         """
         Loads a model from a given file path.
 
+        .. warning::
```
dennisbader, line 2506: Fine for backwards compat, but means RNNModel.load("foo.pt") is still arbitrary code execution on a malicious file. If we're closing #3177 with this PR, we should be clear in the issue / CHANGELOG that .pt loads are out of scope.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -933,6 +933,75 @@ def test_save_and_load_weights_w_likelihood(self, tmpdir_fn):
...
+        reloaded = DLinearModel(
+            input_chunk_length=4,
+            output_chunk_length=1,
+            likelihood=GaussianLikelihood(),
+            **tfm_kwargs,
+        )
+        reloaded.load_weights(ckpt_path, map_location="cpu")
+        reloaded.predict(n=2, series=self.series[:20])
+
+        # 2) craft a malicious checkpoint whose `__reduce__` writes a marker file --------
```
dennisbader, line 973: Part 1 (legitimate model loads with default) is good. Part 2 only tests raw torch.load though — can you add a case where the evil payload is loaded via reloaded.load_weights(evil_path) or similar? That's the path users actually hit.

darts/tests/models/forecasting/test_torch_forecasting_model.py
```diff
@@ -933,6 +933,75 @@ def test_save_and_load_weights_w_likelihood(self, tmpdir_fn):
...
+
+        # explicit opt-out (`weights_only=False`) still loads (and here executes) the
+        # payload, confirming the opt-out path is preserved
+        torch.load(evil_path, weights_only=False)
+        assert os.path.exists(marker_path), (
+            "weights_only=False should still fully unpickle the checkpoint"
+        )
+        # cleanup the benign marker
+        os.remove(marker_path)
+
```
dennisbader, line 1004: On test_load_weights_from_checkpoint (~line 1344–1353)

This expects ValueError when weights_only=True, but you removed that guard and made True the default -> the the load_weights_from_checkpoint above should then fail now. Did you mean to update this block (now it's probably handled via the safe globals)?

dennisbader (changes requested):
Thanks a lot @hackchang for this PR 🚀 

The direction makes sense and looks like a good start. I do have some comments / suggestions: 

- The .ckpt default to weights_only=True is good
- The import-time allow-list of half of torch/torchmetrics is not ideal: too much maintenance, runs on every import, and we'll miss edge cases (custom losses, new metric subpackages).
- Most loading paths still fully unpickle the `.pt` files so the security fix is narrower than the description suggests.

I'd prefer a slightly reduced PR: with flipping weights_only defaults, adding docs/warnings, minimal Darts-only allow-list, and maybe checkpoint-driven registration on failure.

darts/models/forecasting/torch_forecasting_model.py
hackchang: Done

darts/models/forecasting/torch_forecasting_model.py
hackchang: Done

darts/models/forecasting/torch_forecasting_model.py
hackchang: Done

hackchang: Done

darts/models/forecasting/torch_forecasting_model.py
hackchang: Done

darts/models/forecasting/torch_forecasting_model.py
hackchang: Done

darts/tests/models/forecasting/test_torch_forecasting_model.py
hackchang: Done

darts/tests/models/forecasting/test_torch_forecasting_model.py
hackchang: Done

hackchang:
Thanks for the detailed review — this was the right call. Pushed a reduced version in 2888876:

1. No import-time registration. The old add_safe_globals scan (torch/torchmetrics/
   neuralforecast) is gone. Allow-listing is now scoped to each load via
   torch.serialization.safe_globals(...) — nothing is imported or registered unless a
   checkpoint is actually loaded.

2. Checkpoint-driven, per file. Using get_unsafe_globals_in_checkpoint(path) I only
   allow-list the globals that specific file references, filtered to classes under
   torch/torchmetrics/darts/neuralforecast that subclass a known-safe base (nn.Module /
   Optimizer / LR scheduler / torchmetrics Metric|MetricCollection / Darts likelihood).
   Anything else stays blocked and the load fails loudly.

3. Functions: you're right. No more module scans. torchmetrics does pickle references to
   its own reduction helpers (e.g. dim_zero_sum, jit_distributed_available), so I allow-list
   only an explicit, audited set of 6 such functions by exact qualified name, added only when
   the checkpoint references them — and I documented that registering a callable is not zero-risk.

4. Honesty on scope. Docstrings + CHANGELOG now state this protects the .ckpt only; the
   Darts base .pt is still fully unpickled, so a malicious .pt is still RCE and is out of
   scope. Marked the change 🔴 and set #3177 to "partially addresses".

5. Tests. Fixed the now-stale weights_only=True block (its ValueError was actually an
   architecture mismatch), added a malicious-payload case through the real user path
   (load_weights_from_checkpoint), and added a resume-from-checkpoint test that exercises
   optimizer/scheduler state under weights_only=True. The torch checkpoint/save-load suite
   passes locally.

Happy to trim further to a Darts-classes-only allow-list (requiring weights_only=False for
metric/optimizer-bearing checkpoints) if you'd prefer even less surface — let me know.

dennisbader:
Hi @hackchang and thanks for the updates! I'm currently working on this PR to fix the remaining issues, I'll let you know once I've pushed the updates.

PR #3184, opened 2026-08-20
Fix extract_subseries returning a single TimeSeries instead of a list when there is no gap

AlejandroCoronadoN:
`extract_subseries` is annotated `-> list[TimeSeries]` and its docstring says it returns "A list of TimeSeries". Every return path returns a list except one: when `series.gaps(mode=mode)` is empty (the series has missing values but no gap under the selected `mode`), it returns the bare `series` instead of `[series]`.

Repro:

```python
import numpy as np, pandas as pd
from darts import TimeSeries
from darts.utils.missing_values import extract_subseries

vals = np.array([[1., 10.], [np.nan, 20.], [3., np.nan], [4., 40.]])
ts = TimeSeries.from_times_and_values(pd.date_range("20130206", periods=4), vals)
res = extract_subseries(ts, mode="all")
print(type(res).__name__, isinstance(res, list))  # TimeSeries False, expected list True
```

A caller doing `for sub in extract_subseries(ts): ...` then silently iterates over the TimeSeries samples instead of a one-element list.

Fix: return `[series]` in that branch, matching the other return paths and the annotation. Extended the existing `test_extract_subseries` to cover the no-gap case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ but cannot always guarantee backwards compatibility. Changes that may **break co
 
 **Fixed**
 
+- Fixed `extract_subseries` returning a single `TimeSeries` instead of a `list[TimeSeries]` when the series has no gap under the selected `mode`. [#3184](https://github.com/unit8co/darts/pull/3184) by [Alejandro Coronado](https://github.com/AlejandroCoronadoN).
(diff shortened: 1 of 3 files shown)


dennisbader (approved):
Thanks a lot for the fix @AlejandroCoronadoN, beautiful PR! Also, congrats to you first contribution 🥳 💯

AlejandroCoronadoN:
Thanks a lot, @dennisbader, that means a lot. darts was a pleasure to dig into. I will keep an eye out for other things I can help with.

PR #3185, opened 2026-08-21
Fix incorrect return type annotations in darts.utils.statistics

AlejandroCoronadoN:
Three functions in `darts/utils/statistics.py` have return type annotations that do not match what they actually return:

- `stationarity_test_kpss` is annotated `-> set` but returns a `tuple` (it wraps `statsmodels.tsa.stattools.kpss`).
- `stationarity_test_adf` is annotated `-> set` but returns a `tuple` (it wraps `adfuller`).
- `granger_causality_tests` is annotated `-> None` but returns a `dict` (it wraps `grangercausalitytests`); its own docstring already says it returns a `Dict`.

The docstrings and the existing tests already treat these as tuples/dict (for example `stationarity_test_kpss(series)[1]`), so the annotations are just wrong and mislead type checkers and IDEs. This fixes the annotations to match the real return values and adds a small regression test.


dennisbader (approved):
Thanks a lot for the PR and fix @AlejandroCoronadoN 🚀 

I removed the tests (not really necessary to check the return type) and slightly adapted the changelog entry.

PR #3193, opened 2026-08-28
Fix missing future covariates error in historical_forecasts/backtest

### LINKED ISSUE #2846: Improve error message when future covariates are missing during backtest() call
Yes.When using backtest() with a model trained with train and val future covariates, forgetting to pass the full future covariates again at backtest time results in an error (```IndexError: index 1 is out of bounds for axis 0 with size 1.```). 

A better explanation or error message should be thrown as it is the case for "regular" historical forecasts.

**Additional context**
Error currently thrown :

```IndexError: index 1 is out of bounds for axis 0 with size 1.```

webzuweb:
## Summary

Fixes #2846.

When a model is trained with past/future covariates and then passed to `historical_forecasts()`/`backtest()` without the corresponding covariates, the user currently gets a cryptic `IndexError: index 1 is out of bounds for axis 0 with size 1`.

This PR adds an explicit `ValueError` (mirroring the existing check in `predict()`) that clearly states which covariates are missing.

## Changes

- `darts/models/forecasting/forecasting_model.py`: added early covariate checks in `historical_forecasts()`.
- `darts/tests/models/forecasting/test_backtesting.py`: added `test_backtest_missing_future_covariates`.

## Test

`pytest darts/tests/models/forecasting/test_backtesting.py::TestBacktesting::test_backtest_missing_future_covariates` — passed.

### DIFF
--- a/darts/models/forecasting/forecasting_model.py
+++ b/darts/models/forecasting/forecasting_model.py
@@ -904,6 +904,21 @@ class ForecastingModel(ABC, metaclass=ModelMeta):
         series = series2seq(series)
         past_covariates = series2seq(past_covariates)
         future_covariates = series2seq(future_covariates)
+
+        if self.uses_past_covariates and past_covariates is None:
+            raise_log(
+                ValueError(
+                    "The model was trained with past covariates. Some matching past_covariates "
+                    "must be passed to `historical_forecasts()`."
+                )
+            )
+        if self.uses_future_covariates and future_covariates is None:
+            raise_log(
+                ValueError(
+                    "The model was trained with future covariates. Some matching future_covariates "
+                    "must be passed to `historical_forecasts()`."
+                )
+            )
         sample_weight = (
             sample_weight
             if isinstance(sample_weight, str)
--- a/darts/tests/models/forecasting/test_backtesting.py
+++ b/darts/tests/models/forecasting/test_backtesting.py
@@ -1109,6 +1109,33 @@ class TestBacktesting:
                 "Model cannot be fit/trained with `future_covariates`."
             )
 
+    def test_backtest_missing_future_covariates(self):
+        """`backtest()`/`historical_forecasts()` on a model trained with future covariates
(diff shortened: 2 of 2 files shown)


dennisbader (changes requested):
Thanks for this PR @webzuweb. I've quickly tested your code and realized that it's not exactly fixing the real issue. 

The test you added runs successfully in the current master head (runs without raising an exception). So this is not the issue that we were trying to solve. 

Unfortunately there is no reproducible example in the issue that was raised. So we should rather figure when it is raised, before going on with the current approach. 
Would you mind having a look to see if you can find the actual problem?

webzuweb:
Thanks for the review @dennisbader — you're right, my original test only covered the `None` case and didn't reproduce the real failure.

I dug into the actual behavior on current master. For a model trained with `future_covariates`, passing:

- **no covariates** → `TypeError: object of type 'numpy.float32' has no len()`
- **covariates too short** → misleading `ValueError: target_series must have at least -min(lags)+max(lags)+1 = 6 time steps; instead, it only has 0`
- **covariates starting too late** → `TypeError: object of type 'numpy.float32' has no len()`

Root cause is the same as the issue: there's no validation in `historical_forecasts()`/backtest()` before the lagged-data machinery runs, so a missing/undersized/misaligned covariate fails deep in the call stack with an unrelated message.

I've updated the PR to also validate that provided covariates **cover the entire series time range** (start early enough / end late enough), in addition to the existing `None` check. Both `backtest()` and `historical_forecasts()` now raise a clear `ValueError` naming the offending covariate and the actual vs. expected time range.

I also added tests covering the too-short and too-late scenarios (previously they'd hit the cryptic errors). Let me know if the span requirement should instead follow `extreme_lags` more precisely (e.g. accounting for `max_future_cov_lag` beyond the series end).

dennisbader:
Thanks for giving it another go. However, some of the changes are still introduce changes that will break what should actually work. I'll be closing this PR and open another one that actually addresses the core issue.

Issue #1278, opened 2022-10-11
Allow historical forecast to work on time series with nan


dumjax:
**Is your feature request related to a current problem? Please describe.**
We often want to see the historical forecast or back test on a time series having nan (we we can't really interpolate, because the missing part is just too long)

**Describe proposed solution**
Consider returning nans or nothing for points which are not predictable (where some inputs are nan)

PR #3195, opened 2026-08-30
Add TimeSeries.has_nan property to check for NaN values

webzuweb:
Partial contribution towards #1278.

## Summary

Adds a `has_nan` property to `TimeSeries` that indicates whether the series contains any NaN values. This was explicitly suggested in #1278 ("Might be also useful to create a TimeSeries property which indicates if there are some nan in the time series") and is useful for users deciding whether a series can be passed to models/historical forecasts that don't handle NaNs.

## Changes

- `darts/timeseries.py`: new `has_nan` property (`np.isnan(self._values).any()`), placed alongside the other `has_*` properties.
- `darts/tests/test_timeseries.py`: new `test_has_nan` covering univariate with NaN, clean univariate, integer series, and multivariate with NaN.

## Verification

```
pytest darts/tests/test_timeseries.py::TestTimeSeries::test_has_nan  # 1 passed
```

### DIFF
--- a/darts/tests/test_timeseries.py
+++ b/darts/tests/test_timeseries.py
@@ -1904,6 +1904,21 @@ class TestTimeSeries:
         ts8 = TimeSeries.from_values(values1)
         assert ts8.has_range_index
 
+    def test_has_nan(self):
+        ts_nan = TimeSeries.from_values([1.0, np.nan, 3.0])
+        assert ts_nan.has_nan
+
+        ts_clean = TimeSeries.from_values([1.0, 2.0, 3.0])
+        assert not ts_clean.has_nan
+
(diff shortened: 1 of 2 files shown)


dennisbader (commented):
Thanks for opening this PR and giving it a go @webzuweb. 

After having thought a bit about this, I decided we won't add it to `TimeSeries`. The class is supposed to handle time series specific operations. Other operations we usaully leave to numpy / pandas. In the end, the user can call `np.isnan(series.all_values(copy=False)).any()` to perform this efficiently.

webzuweb:
Thanks @dennisbader — that makes sense, keeping `TimeSeries` focused on time-series-specific operations and leaving general array checks to numpy/pandas is the right call. I'll close this PR.

PR #3203, opened 2026-09-15
Warn when encoders are ignored by fit/predict_from_dataset

### LINKED ISSUE #3006: [Feature] Improve handling & documentation of encoders and categorical static covariates in .fit_from_dataset()

webzuweb:
Fixes #3006.

## Summary

`add_encoders` is only applied when training/predicting via `fit()`/`predict()`. When using `fit_from_dataset()`/`predict_from_dataset()`, encoders are silently ignored, which is a blind spot for users.

Per the discussion in #3006, this PR adds a `logger.warning` whenever `fit_from_dataset()`/`predict_from_dataset()` is called on a model initialized with `add_encoders`, and documents the behavior in both docstrings.

## Changes

- `TorchForecastingModel.fit_from_dataset()`: warn + docstring note when `self.add_encoders` is set.
- `TorchForecastingModel.predict_from_dataset()`: warn + docstring note when `self.add_encoders` is set.
- Tests: `test_encoders_ignored_in_from_dataset_warns` and `test_encoders_not_set_from_dataset_no_warn`.

No behavior change for models without encoders.


dennisbader:
Hi @webzuweb and thanks for the PR. Closing it however since it's a duplicate of #3086.

PR #3208, opened 2026-09-19
feat: add summary() to statsmodels-based forecasting models

### LINKED ISSUE #3030: Statsmodels Summary addition into Darts
This is a minor long term request for Darts that would improve readers understanding of the models they are building in Darts.

Statsmodels.summary() is a really useful feature for ARIMA, and linear regression models. Is it possible that Darts could include it but also to 

expand it to possibly based on some popular metrics out there? So for example  for the gradient boosted model (Sklearn.GradientBoostedRegressor) it can give a summary of the parameters put into the model (making it simple pretty  input into a journal), the R squared so forth. 

This is a long term project so I do not expect it to come soon, but I am pleased to see Neural Forecast will soon make a addition to Darts.

Boubker10:

Fixes #3030.

### Summary

Adds a `summary()` method to the statsmodels-based forecasting models — `ARIMA`, `VARIMA`, and `ExponentialSmoothing`. These models already wrap a statsmodels results object internally (`self.model`) which exposes a rich `.summary()` (coefficients, standard errors, p-values, AIC/BIC, etc.), but there was previously no way to access it through the Darts API.

- Added `summary()` to `ARIMA`, `ExponentialSmoothing`, and `VARIMA`, delegating to the underlying statsmodels results object's `.summary()`.
- Raises a clear `ValueError` if called before the model is fitted.
- Added unit tests for all three models covering the pre-fit error and the post-fit summary output.

### Other Information

Scope was intentionally limited to the models that directly wrap `statsmodels` (`ARIMA`, `VARIMA`, `ExponentialSmoothing`), since their underlying results objects natively support `.summary()`. Other model families mentioned in the original issue (e.g. `GradientBoostingRegressor`-based models) don't have an equivalent built-in summary and would need a different approach — happy to discuss extending this further if maintainers want a unified API across more model types.


dennisbader:
Hi @Boubker10 and thanks a lot for the PR! Unfortunately, I don't think we'll add this explicit `summary()` to the models. 
Our goal is mostly to only add such methods, if they can be applied / made accessible for all (or a large family of) models together.

At the moment, users can still get summary via `model.model.summary()` which is rather straight forward in my opinion.

